PR #20, opened 2017-03-24
Some refactoring and a bug fix

shaunwheelhouse:
I've included a (temporary) bug fix for an access violation caused by passing an empty keychain to `lib/netpgp.c:get_first_ring`. Exactly why that happens sometimes is the next issue to follow up. Also included is a newer version of `build_macos.sh`.


ronaldtse:
Thank you @shaunwheelhouse ! Much cleaner than what we had before.

PR #23, opened 2017-03-25
Exception cases for empty keyrings; remove bzlib and zlib

shaunwheelhouse:
I tried to be a bit clever before and accidentally nuked a couple of commits and my PR. 😅 

* Fixes the empty keyring bug; adds a warning; correctly aborts.
* Removes static bzlib and zlib - note that these _may_ need to be declared as dependencies in the Homebrew formula.


ronaldtse:
Thank you @shaunwheelhouse !

PR #32, opened 2017-03-27
Use .rnp instead of the key ID when selecting a --homedir subdirectory

shaunwheelhouse:
As advertised. The code in and around it needs some work but this should be enough in the meantime.


shaunwheelhouse:
Oops, wait up, left some debugging information in there.

shaunwheelhouse:
All good. 👍

ronaldtse:
Thanks @shaunwheelhouse ! Please feel free to merge it.

PR #38, opened 2017-03-28
Updated the build shell scripts

shaunwheelhouse:
I rolled the common parts of `build_macos.sh` into `build_common.inc.mk` to be used by `build.sh` as well and expanded on macOS-specific script. Following that pattern it should be easy to fork off new scripts for other targets.


ronaldtse:
Very nice @shaunwheelhouse !

Issue #15, opened 2017-03-23
Replace libbn (for BIGNUM) with external library


ronaldtse:
Currently `libbn` is basically a wrapper/copy of the [libtommath library](https://github.com/libtom/libtommath).

It is better to directly link with a well-maintained external multi-precision integer library, instead of incorporating its source code directly.

Options are:
* [MPIR](http://mpir.org)
* [GMP](https://gmplib.org)
* [libtommath](https://github.com/libtom/libtommath)

ronaldtse:
Thanks @randombit !

Issue #13, opened 2017-03-23
Replace libdigest and librsa with Botan


ronaldtse:
`libdigest` and `librsa` in NetPGP implements a number of cryptographic algorithms, but it would be much better to use a fully tested and maintained external library like Botan.

PR #42, opened 2017-03-29
Modify netpgpverify to use libnetpgp for hashes and signature verification

randombit:
netpgpverify has its own digest_t API distinct from netpgp's pgp_hash_t, for now I've just implemented digest_t in terms of pgp_hash_t. Converting everything to pgp_hash_t and removing `digest.{c,h}` would be an eventual goal.

Forward declares BIGNUM and DSA_SIG in crypto.h instead of including the OpenSSL headers, because netpgpverify's headers conflict with OpenSSL (or at least will until I finish ripping out netpgpverify's custom RSA/DSA/etc code). Adds necessary includes in lib, that previously were made available through the openssl include (`stdlib.h` etc)

This change removes support for Tiger completely, since it is lacking in OpenSSL and not supported by the existing `pgp_hash_t` API. I don't think Tiger was used or tested; the `prefix_tiger` previously set in the source (in digest.c) is not in any way the Tiger OID (as would be required for a valid Tiger PKCSv1.5 signature) but instead just the initial chaining value of the Tiger hash which is completely unrelated. Tiger support could be easily added back later once a different crypto lib supporting Tiger is in use.

I've tested this change using `rnpv -k pubring.gpg NetBSD-6.0_RC1_hashes.asc` and `rnpv -S sshtest-20140202.pub data.sig`


ronaldtse:
Thank you @randombit !

Tiger in any case is no longer relevant in OpenPGP as described on this [SO post](http://crypto.stackexchange.com/questions/11284/how-do-you-use-the-tiger-hash-function-with-gpg).

randombit:
Updated in a085e7d to also call RSA and DSA verification functions in librnp, removes BIGNUM wrapper and LibTomMath from rnpv.

I think this is ok to merge.

ronaldtse:
Thank you again @randombit , this is a huge improvement from before.

ronaldtse:
Hi @randombit , for this change, could you also help add instructions on how to build with the latest Botan 2.1? Since 2.1 isn't released on your site yet...

The specific files are:
* `README.md`
* `build*.sh`

Thanks!

randombit:
@ronaldtse This PR does not add any Botan dependency, it just changes rnpv to call the crypto functions in lib/crypto.c - so currently still OpenSSL, until #13 and #15.

I ended up having to write several additions for the C89 API for Botan 2.1 to offer necessary operations not previously offered by the C API such as extracting p,q values from RSA keys. These are all merged to master, but this has held up the release of Botan 2.1 by a few days. I expect to release 2.1 today, then will start sending PRs to rnp for first adding Botan 2.1 to the build/docs, then incrementally removing OpenSSL usage (digests, ciphers, RSA, ...), until eventually all OpenSSL calls are gone and ax_check_openssl.m4 is removed.

I already have digests and bn fully implemented (botan-conv on my fork). RSA is tricky because of rnp's handrolled PKCS padding, so further changes to rnp on my part are required before RSA can be completely switched over.

ronaldtse:
Thank you @randombit for the exciting work 👍

Issue #43, opened 2017-04-02
rnp_init homedir check


shaunwheelhouse:
It's not very idiomatic that netpgp_init/rnp_init performs an emptiness test on the homedir setting and raises an error if it isn't set. I propose that we change the interface from:

```
if (rnp_set_homedir(&context, new_home, key_directory, quiet))
    rnp_init(&context);
```

to either assuming a sane default (getenv("HOME"), ".gnupg") or accepting the homedir directory in the initialiser:

```
rnp_init(&context, getenv("HOME"), ".gnupg");
```

What are your thoughts?

shaunwheelhouse:
It also seems that by zeroing the context structure during initialisation any previously set home directory variable is lost... causing apps to invariably fail. In the interest of getting it working for testing tonight I'm going to go with a sane default for the time being and shuffle set_homedir calls below the init call.

ronaldtse:
@shaunwheelhouse agree. This direction is much more proper that what is there. Let's do this.

shaunwheelhouse:
@ronaldtse There's currently a bug breaking key loading that has me holding off for a pull tonight but you can review today's work at my fork under the 'rnpkeys-init-fix' branch.

I haven't forgotten about the improved command line interface you'd like to see either. I'll move on to that once this and perhaps the asprintf dependencies are knocked out. 👍

ronaldtse:
Thank you @shaunwheelhouse , this is giving us lots of hope 👍

PR #45, opened 2017-04-04
Misc. improvements

shaunwheelhouse:
This one deviated from its topic a little bit.

* I changed the way `rnp_init` behaves as discussed in issue #43 and added a new public library function, `rnp_load_keys`.
* `rnp` and `rnpkeys` no longer break around the home directory and function more reliably than before.
* Consolidated use of ".gnupg" and ".ssh" in the utilities to a common definitions header.
* Many misc. style and readability changes here and there.

Consequences:

* If @zgyarmati has any uncommitted changes renaming netpgp to rnp in hkpd and libpaa this may conflict - I changed them too. Sorry!

* Other accompanying projects that make use of `rnp_init`, like the language bindings @dewyatt, will have to change the way they initialise librnp - also sorry! However, these should be mostly painless and I think are a strong improvement. Essentially:

  ```
  memset(&rnp, '\0', sizeof(rnp));
  rnp_setvar(... etc.
  rnp_init(&rnp);
  ```

  Now becomes:

  ```
  rnp_init(&rnp);
  rnp_setvar(... etc.
  if (need_to_load_keys)
      rnp_load_keys(&rnp);
  ```

This will make it to master when the parent branch name_refactor_wip is merged.


ronaldtse:
Sorry for the delay @shaunwheelhouse , these are welcome changes 👍

PR #47, opened 2017-04-06
pgp_sprint_keydata refactor

shaunwheelhouse:
Refactoring of the pgp_sprint_keydata function. It still needs more work but for the time being it eliminates the pgp_asprintf call there for equivalent functionality.


ronaldtse:
Thank you @shaunwheelhouse !

Issue #8, opened 2017-03-23
Replace OpenSSL dependency with the Botan cryptographic library


ronaldtse:
OpenSSL while used ubiquitously, it comes with the burden that it is version locked due to its status of a system library.

We'd like to replace it with the [Botan](http://botan.randombit.net) library, which supports a number of algorithms we wish to work with.

ronaldtse:
@randombit thank you for the work! The OpenSSL dependency is indeed completely gone but there are a few remaining textual references to OpenSSL (below), and the usage of `USE_BN_INTERFACE` (looks temporary) remains.

Could you help clean them up so we'll close this issue? Thanks!

``` bash
src/cmocka/rnp_tests.c:46:  #include <openssl/bn.h>
src/lib/bn.c:37:/* OpenSSL BIGNUM emulation layer */
src/lib/crypto.h:280:int openssl_read_pem_seckey(const char *, pgp_key_t *, const char *, int);
src/lib/librnp.3:167:The library uses functions from the openssl library for multi-precision
src/lib/pem.c:89:openssl_read_pem_seckey(const char *f, pgp_key_t *key, const char *type, int verbose)
src/lib/ssh2pgp.c:378:  if (!openssl_read_pem_seckey(f, key, "ssh-rsa", 0)) {
src/lib/ssh2pgp.c:392:          /* openssh and openssl have p and q swapped */
src/cmocka/rnp_tests.c:#if defined(USE_BN_INTERFACE)
src/lib/bn.h:#define USE_BN_INTERFACE
src/lib/bn.h:#ifdef USE_BN_INTERFACE
src/lib/bn.h:#endif /* USE_BN_INTERFACE */
```

randombit:
@ronaldtse No problem PR coming shortly

PR #48, opened 2017-04-06
Convert from OpenSSL to Botan for crypto operations

randombit:
Fixing #8 

This replaces all OpenSSL usage with the Botan crypto library, using Botan's C API. The very latest version (2.1.0) is required. This took me a while and my commit history was a mess, so I restaged everything and recommitted in a somewhat logical order, which is why all of the commits are from this afternoon.

This PR needs more testing, but I think it is ready for a first review.

Some possible further improvements, which I am not going to try addressing in this PR:

- rnp rolls its own PKCSv1.5 encoding code and then uses OpenSSL's RSA_NO_PADDING (or now Botan's equivalent "Raw"). Instead rnp should use the library version of the padding operations. This will probably require some significant changes because rnp also wants to compute the signature digests itself, and it seems like in some cases the packet parsing code and digest update code are somewhat intertwined. rnp's requirement for message recovery on RSA signatures (RSA_public_decrypt in OpenSSL land) is not available through the Botan signature API. The message encryption case is probably easier to tackle.
- rnp implements ElGamal using OpenSSL's BN API, since it is not supported by OpenSSL. Botan already includes it.


ronaldtse:
Thank you @randombit , let me move the pending TODOs to new issues, and let's get this tested first -- if we had tests it would have been much easier!

randombit:
I'm working on crypto unit tests (using cmocka - https://cmocka.org/) in a new branch, which allows me to run the same tests against both openssl and botan versions of the lib. That has already found one issue in CFB cipher handling that I'm debugging now, so this PR should certainly not be merged yet.

shaunwheelhouse:
@ronaldtse, @randombit - I'm so sorry that I didn't get to this earlier! I didn't receive a notification!

Testing master is going to be very difficult because the --homedir functionality is broken and the holdover fixes are in the name_refactor_wip branch. We should get that branch completed and merged ASAP so that good work like this can be assessed on a stable base.

ronaldtse:
@randombit @shaunwheelhouse I've updated dependency building to include packages required by `botan`/`cmocka` and also updated the README to reflect latest build procedures.

Perhaps we will want to make `automake` handle these dependencies directly later.

ronaldtse:
@shaunwheelhouse I've merged the latest changes from master.

randombit:
@shaunwheelhouse  Thanks I'll take a look at the problem you mention sometime today

randombit:
@shaunwheelhouse Hi Shaun both bugs should be fixed now in f04e8dd let me know what you think. Thanks!

ronaldtse:
@shaunwheelhouse feel free to merge once it is good. Thanks!

shaunwheelhouse:
@randombit - Your changes are great, thanks for fixing those issues. :+1:

@randombit, @ronaldtse - Sorry again for the late attention! Now that the easter holidays are over I'll be more responsive. 🔧

ronaldtse:
Thank you @shaunwheelhouse and @randombit !!

PR #52, opened 2017-04-07
Add some unit tests using cmocka

randombit:
Current tests digests, ciphers, RSA keygen and encrypt/decrypt, and ElGamal encrypt/decrypt. Uses the cmocka library https://cmocka.org/

I'm not sure if there is a better way to integrate this into the build. Right now if cmocka is not available the build will fail, probably instead the cmocka subdir should just be skipped in that case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@
 *.loT
 *.o
 
+core.*
+vgcore.*
+
 # These artifacts can be removed by ./remove_artifacts.sh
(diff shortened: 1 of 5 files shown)


ronaldtse:
Thanks @randombit for the tests. Let's merge them up for now and we'll find a better way to build, but it's probably a good idea for the build system to be able to run the cmocka tests.

PR #58, opened 2017-04-18
Elgamal on botan conv

flowher:
* Reimplements `pgp_elgamal_private_decrypt' and `pgp_elgamal_public_encrypt'
      - Botan is used instead of OpenSSL BN API
    * Updates description of both function
    * This patch depends on newer version of Botan (see changes to README.md)

### DIFF
--- a/src/cmocka/rnp_tests.c
+++ b/src/cmocka/rnp_tests.c
@@ -212,66 +212,77 @@ static void raw_rsa_test_success(void **state)
 
 static void raw_elg_test_success(void **state)
 {
-   // largest prime under 512 bits
-   const uint8_t p512[64] = { 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
-                              0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
-                              0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
-                              0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
-                              0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
-                              0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
-                              0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
-                              0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFD, 0xC7,
-   };
+  // largest prime under 512 bits
+  const uint8_t p512[64] = {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+                            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+                            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+                            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+                            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+                            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+                            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+                            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFD, 0xC7,
+  };
 
-   pgp_elgamal_pubkey_t pub_elg;
-   pgp_elgamal_seckey_t sec_elg;
-   uint8_t   encm[1024];
-   uint8_t   g_to_k[1024];
-   uint8_t decr[1024];
-   const uint8_t plaintext[5] = { 1, 2, 3, 4, 23 };
-   BN_CTX* ctx = BN_CTX_new();
+  pgp_elgamal_pubkey_t pub_elg;
+  pgp_elgamal_seckey_t sec_elg;
+  uint8_t              encm[64];
+  uint8_t              g_to_k[64];
+  uint8_t              decryption_result[1024];
+  const uint8_t plaintext[] = { 0x01, 0x02, 0x03, 0x04, 0x17 };
+  BN_CTX* ctx = NULL;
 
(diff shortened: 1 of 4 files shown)


ronaldtse:
Thank you @flowher ! The filename `openssl_crypto.c` is now a misnomer since it is now all Botan. Since that file is really long, we could potentially split it according to function grouping, such as `elgamal.c`. What do you think?

randombit:
It should not happen in practice that any of the `destroy` calls fail (because under the covers it is just a C++ `delete` along with invoking a `nothrow` destructor), but chaining them as you do means some resources could leak if the first or second call results in an error.

```
if (botan_pk_op_decrypt_destroy(op_ctx) ||
      botan_privkey_destroy(key) ||
       botan_rng_destroy(rng)) {
```

Other than that LGTM

flowher:
Thanks guys for review. I'll fix it today

src/lib/openssl_crypto.c
```diff
@@ -95,6 +95,12 @@
 
 #include <botan/ffi.h>
 
+#define FAIL(str)  \
```
shaunwheelhouse: Could you please explain the the reason for wrapping these statements in `do {} while (false)` and perhaps comment it?

src/lib/openssl_crypto.c
```diff
@@ -861,165 +867,138 @@ openssl_read_pem_seckey(const char *f, pgp_key_t *key, const char *type, int ver
...
-	return ret;
+  botan_rng_t            rng          = NULL;
+  botan_privkey_t        key          = NULL;
+  botan_pk_op_decrypt_t  op_ctx       = NULL;
+  int                    ret          = -1;
+  size_t                 out_len      = 0;
+  size_t                 p_len        = 0;
+  uint8_t*               bt_plaintext = NULL;
+
+  if (botan_rng_init(&rng, NULL)) {
```
shaunwheelhouse: If I'm understanding the Botan documentation correctly this branch will never be taken as `botan_rng_init` returns POSIX-style codes (0 on success or -1 on failure). If that isn't the case please send me a message, I could be wrong!

src/lib/openssl_crypto.c
```diff
@@ -861,165 +867,138 @@ openssl_read_pem_seckey(const char *f, pgp_key_t *key, const char *type, int ver
...
+   * Allocate all buffers needed for encryption and post encryption processing */
+  out_len = p_len*2;
+
+  bt_plaintext = calloc(out_len, 1);
+  if (!bt_plaintext) {
+
+      FAIL("Memory allocation failure");
+  }
+
+  if (botan_privkey_load_elgamal(&key, pubkey->p->mp, pubkey->g->mp, seckey->x->mp) ||
```
shaunwheelhouse: I have similar concerns over `botan_privkey_load_elgamal` but I can't find the documentation for this one. Again, correct me if I'm wrong.

src/lib/openssl_crypto.c
```diff
@@ -861,165 +867,138 @@ openssl_read_pem_seckey(const char *f, pgp_key_t *key, const char *type, int ver
...
+  if (botan_pk_op_decrypt_create(&op_ctx, key, "Raw", 0) ||
+      botan_pk_op_decrypt(op_ctx, out, &out_len, bt_plaintext, p_len*2)) {
+
+      FAIL("Decryption fails");
+  }
+
+  ret = (int)out_len;
+
+end:
+  if (botan_pk_op_decrypt_destroy(op_ctx) ||
```
shaunwheelhouse: It would be best if these were performed sequentially so that a short-circuit doesn't create a (tiny) memory leak.

src/lib/openssl_crypto.c
```diff
@@ -861,165 +867,138 @@ openssl_read_pem_seckey(const char *f, pgp_key_t *key, const char *type, int ver
...
+  if (botan_privkey_load_elgamal(&key, pubkey->p->mp, pubkey->g->mp, seckey->x->mp) ||
+      botan_privkey_check_key(key, rng, 1)) {
+
+      FAIL("Wrong private key");
+  }
+
+  memcpy(bt_plaintext, g2k, p_len);
+  memcpy(bt_plaintext + p_len, in, p_len);
+
+  if (botan_pk_op_decrypt_create(&op_ctx, key, "Raw", 0) ||
```
shaunwheelhouse: As the last couple.

src/lib/openssl_crypto.c
flowher: Agreed

src/lib/openssl_crypto.c
flowher: This is new function, I've just comitted it to botan. But you are right

src/lib/openssl_crypto.c
flowher: Agreed

src/lib/openssl_crypto.c
flowher: Yes, I agree. I'll change it

src/cmocka/rnp_tests.c
```diff
@@ -212,66 +212,77 @@ static void raw_rsa_test_success(void **state)
...
+  ctext_size /= 2;
 
-   BN_mod_exp(pub_elg.y, pub_elg.g, sec_elg.x, pub_elg.p, ctx);
-
-   BN_CTX_free(ctx);
+#if defined(DEBUG_PRINT)
+  BIGNUM *tmp = BN_new();
 
-   pgp_elgamal_public_encrypt(g_to_k, encm, plaintext, sizeof(plaintext), &pub_elg);
+  printf("\tP\t= 0x%s\n",   BN_bn2hex(pub_elg.p));
```
flowher: Ok

src/lib/openssl_crypto.c
flowher: It's multi (2) statement macro. Please find more here:
https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Multi-statement_Macro

shaunwheelhouse: Makes sense! Thank you @flowher.

flowher:
All comments implemented.
Actually, almost all - @ronaldtse I think it is very good idea to split openssl_crypto.c into smaller files. Do you mind if I create separated PR for this change. I don't want to pollute this PR with unrelated changes. I've started to work on this change here https://github.com/flowher/rnp/commits/crypto_split

ronaldtse:
@flowher I absolutely agree with the approach and let's proceed with that as well. Thank you!

src/lib/openssl_crypto.c
ronaldtse: Thank you @flowher ! Always something to learn every day...

src/lib/openssl_crypto.c
shaunwheelhouse: You are completely correct @flowher. I find that counter-intuitive and thought that I knew the correct behaviour. In the future I'll double-check the standards myself before pointing things out, thank you for the citation.

ronaldtse:
Thank you @flowher and @shaunwheelhouse !

flowher:
Thanks guys for review !

PR #62, opened 2017-04-28
Shuffled around key cleanup in lib/pem.c:openssl_read_pem_seckey.

shaunwheelhouse:
Tiny change that allows builds to succeed on macOS; clang is a bit more aware and detects that `privkey` could potentially be destroyed uninitialised.


flowher:
Good stuff. LGTM.

ronaldtse:
Thanks @shaunwheelhouse !

PR #63, opened 2017-04-28
Remove remaining references to OpenSSL in the source

randombit:
Silly to leave these hanging around. 

@ronaldtse One thing you mentioned in the comment in #8 is `USE_BN_INTERFACE`, which controls if the BN function wrappers are exposed as `PGPV_BN_foo` or else `BN_foo`. What do you want to do here? Change all callers to `PGPV_BN_foo` [or some other prefix like `RNP_BN_foo`, easy to change this at the same time], or just remove the controlling macro and always have the `BN_foo` compatibility defines?

### DIFF
--- a/ax_check_botan.m4
+++ b/ax_check_botan.m4
@@ -5,7 +5,7 @@
 # DESCRIPTION
 #
 #   Look for Botan in a number of default spots, or in a user-selected
-#   spot (via --with-openssl).  Sets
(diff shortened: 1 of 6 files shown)


ronaldtse:
@randombit perhaps we just remove the controlling macro and always have the defines to keep things clean? Thanks!

PR #66, opened 2017-05-03
Adding test cases for the command --generate-key

MohitKumarAgniotri:
Note: Currently the test case execution has a manual step as key generation process waits for the user-input.


ronaldtse:
Thank you @MohitKumarAgniotri ! @shaunwheelhouse could you please help guide merging these tests in?

MohitKumarAgniotri:
@ronaldtse , @shaunwheelhouse : The current test case requires user input during the key-generation process. Is there any way to avoid it ? I have been looking at it, but wasn't able to find any yet.

shaunwheelhouse:
Speak of the devil! @MohitKumarAgniotri, it may be possible using the `pass-fd` option. I will take a closer look at that for you now.

shaunwheelhouse:
@MohitKumarAgniotri sorry about the wait. Try setting up a pipe (2), writing the password to it, and setting the application variable "pass-fd" to the file descriptor with `rnp_setvar`. See the relevant entry in the project wiki under "Settings" for a summary. I'll give that a try myself in the morning.

That flow is somewhat clumsy for intra-application development, this issue highlights a good argument for implementing a more direct alternative.

MohitKumarAgniotri:
@shaunwheelhouse : Thanks for looking into it. I will give it a try. Also, do you have any more comments on the test cases itself?

ronaldtse:
@MohitKumarAgniotri , could you help migrate the PDF file into Markdown format? It's much easier to handle using a terminal, and we don't need to manage the DOCX file too ;-) Thanks! 👍

MohitKumarAgniotri:
Sure! I will do that. This was the initial draft as I wanted to get the ball rolling. Thanks for the quick feedback.

MohitKumarAgniotri:
@shaunwheelhouse : I was able to make the test cases use the `pass-fd`, but apparently for some reason the call to read the password (i.e. `fgets()` ) waits for a keystroke to continue. 
Any ideas why this would be the case?

ronaldtse:
Merged with 6926b5b50aef6be6d408ab414c721a77cf0c1fca.

ronaldtse:
Hmm... the new tests are currently failing on master. My bad I shouldn't have merged them so early. @MohitKumarAgniotri could you please help with them? Thanks!

ronaldtse:
(`fgets` bug posted in #75)

Issue #50, opened 2017-04-07
Replace PKCS v1.5 code with Botan equivalent


ronaldtse:
From @randombit :

rnp rolls its own PKCSv1.5 encoding code and then uses OpenSSL's RSA_NO_PADDING (or now Botan's equivalent "Raw"). Instead rnp should use the library version of the padding operations.

This will probably require some significant changes because rnp also wants to compute the signature digests itself, and it seems like in some cases the packet parsing code and digest update code are somewhat intertwined. 

rnp's requirement for message recovery on RSA signatures (RSA_public_decrypt in OpenSSL land) is not available through the Botan signature API. The message encryption case is probably easier to tackle.

ronaldtse:
Thank you @randombit for handling this!

randombit:
Done! Closing

ronaldtse:
Thanks guys!

PR #74, opened 2017-05-06
Change to use Botan's PKCS padding code for encryption

randombit:
First part of GH #50. This PR changes the encryption/decryption padding, but not the signature padding. The signature padding changes will be more complicated and will come in a later PR (probably next week or week after).

Tested by encrypting a file with master rnp and decrypting it with this branch, and also the reverse direction. Everything worked. But only RSA is tested so I may have flubbed something in the ElGamal code.


ronaldtse:
Thanks @randombit !

Issue #44, opened 2017-04-04
Internal subdirectory handling


shaunwheelhouse:
The following piece of boilerplate is/was present in all of the utilities:

```
rnp_set_homedir(&rnp, getenv("HOME"),
    rnp_getvar(&rnp, "ssh keys") ? SUBDIRECTORY_SSH : SUBDIRECTORY_GNUPG, 1);
```

I get the strong impression that switching for gnupg and ssh keys was a featured added much later and patched onto the existing `homedir` context registry setting. I suggest that we solve this by implementing the following behaviour:

* Three new registry settings: `homesubdir`, `homesubdir_gnupg`, `homesubdir_ssh`.
* Move the key type check to a getter for `homedir`.

When the home directory is read:

* Check whether or not the subdirectory for the current key type is not null (`homesubdir_gnupg` or `homesubdir_ssh`). If non-null then use that.
* Check whether `homesubdir` is non-null. If non-null then use that.
* Otherwise fall back to `SUBDIRECTORY_GNUPG` or `SUBDIRECTORY_SSH`.

Finally:

* Alter `rnp_set_homedir` to separately set the `homedir` and `homesubdir` settings. We can now pass `NULL` to the subdirectory parameter and get correct behaviour.

* Add an `rnp_get_homedir` function that performs the behaviour set out above.

ronaldtse:
@shaunwheelhouse we will still allow setting `keydir`, `sshdir` separately from (to override) `homedir`/`$HOME`, right?

shaunwheelhouse:
@ronaldtse for the subdirectories, yes. If we need keydir and sshdir to be able to be set to totally different paths, e.g.:

```
keydir: /home/myuser
sshdir: /home/myotheruser
```

a different design to what I suggest here would be necessary.

ronaldtse:
@shaunwheelhouse suppose I could set the following either via arguments or environment variables:
```
homedir:/home/myuser/
keydir: /home/myotheruser/.rnp
```

Then the subdirectories will be set to:
```
sshdir: /home/myuser/.ssh
keydir: /home/myotheruser/.rnp
```

Correct?

shaunwheelhouse:
@ronaldtse I hadn't yet seen the `sshkeydir` variable. The whole subsystem there is very spread out and inconsistent. I'll get back to you on this one.

ronaldtse:
[...]

PR #81, opened 2017-05-12
Home directory fragmentation

shaunwheelhouse:
My apologies for being out of action; in reference to #44.

These changes have been stagnating for some time. I'm submitting them for your review and for a speedy merge to master so that you don't have to wait any longer, but some minor changes will need to be made in coming days to actually make use of these variables and update the documentation.

Besides some miscellaneous fixes this augments the internal "homedir" variable to instead work as follows:

* "homedir" just points to the user's home directory rather than to the gnupg or ssh key directory. The "subdir" parameter to rnp_set_homedir is now ignored.

* If "homedir_gpg" is set "homedir" will be overridden with it when using gnupg-style keys.

* If "homedir_ssh" is set "homedir" will be overridden with it when using ssh-style keys.

* "subdir_gpg" indicates the subdirectory to use for gnupg keys (by default `.rnp`).

* "subdir_ssh" indicates the subdirectory to use for ssh keys (by default `.ssh`).

Within lib/rnp.c a new module private interface should be used for interacting with these variables:

* The `keydir` function will write the path to the directory for the current type of key to a user supplied buffer. This supplants using `homedir` directly.

* The `keydir_gnupg` and `keydir_ssh` functions fulfil this role for gnupg and ssh key types respectively.

* The `get_homedir_gnupg` and `get_homedir_ssh` functions should be used to get home directories for each respective key type.

I will add some refinements to this tomorrow. Please feel welcome to raise any concerns or comments you may have about these changes.


ronaldtse:
@shaunwheelhouse isn't able to help us handle this due to scheduling. @dewyatt do you have time to help merge this PR? Right now some tests are failing due to `rnp: bad homedir` presumably due to the changes to `homedir` handling.

@shaunwheelhouse you mentioned:
> some minor changes will need to be made in coming days to actually make use of these variables and update the documentation

Have those changes been made, if not, could you let us know what exactly those variables/docs are? Thanks!

shaunwheelhouse:
@ronaldtse - the new variables need to be used by the client applications `rnp` and `rnpkeys` when the appropriate option switches are used. There were also some places, I forget exactly, where the application assembles key paths manually. These could benefit from using the new `keydir` functions. Besides that, the manpages should be updated to describe the new arrangement.

Depending on how you change the application's switches you might want to layer `keypath_` variables on top of the ones I have defined above that allow paths to keyrings to directly override the more friendly homedir/subdir variables, e.g. as with --sshkeyfile.

ronaldtse:
@riboseinc/rnp : Anyone willing to take over this work from @shaunwheelhouse ? Thanks!

dewyatt:
@ronaldtse I looked at it a bit earlier today, hoping to get it merged and fixed up in the next few days.

ronaldtse:
Thank you @dewyatt ! 👍

dewyatt:
Closing this since we have absorbed it into #99.

PR #82, opened 2017-05-13
Fix RSA signatures

dewyatt:
Hopefully closes https://github.com/riboseinc/rnp/issues/79

I think the main piece I'm not sure on here is the removal of the legacy code, but it was necessary to get `rnp --verify` to agree with the other tools.
rnp/rnpv/gpg now all agree that the signatures are valid (and invalid when tampered with).

I think the crux of the issue here is a mismatch in parameter names in botan_privkey_load_rsa vs RSA_PrivateKey, so I filed a PR there: https://github.com/randombit/botan/pull/1045

### DIFF
--- a/src/lib/crypto.h
+++ b/src/lib/crypto.h
@@ -147,7 +147,7 @@ int pgp_rsa_decrypt_pkcs1(uint8_t* out, size_t out_len,
 int pgp_rsa_public_decrypt(uint8_t *, const uint8_t *, size_t,
 			const pgp_rsa_pubkey_t *);
 
-int pgp_rsa_private_encrypt(uint8_t *, const uint8_t *, size_t,
+int pgp_rsa_private_encrypt(uint8_t *, size_t, const uint8_t *, size_t,
 			const pgp_rsa_seckey_t *, const pgp_rsa_pubkey_t *);
 
 /*
--- a/src/lib/rsa.c
+++ b/src/lib/rsa.c
@@ -80,6 +80,8 @@
 #include "readerwriter.h"
 #include "rnpdefs.h"
 
+#include <string.h>
+
 /**
    \ingroup Core_Crypto
    \brief Recovers message digest from the signature
@@ -112,6 +114,7 @@ pgp_rsa_public_decrypt(uint8_t *out,
         if(n_bytes < out_bytes)
            return 0;
 
+        memset(out, 0, n_bytes);
         botan_mp_to_bin(output, out + (n_bytes - out_bytes));
 
         return n_bytes;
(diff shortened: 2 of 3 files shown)


ronaldtse:
@randombit could you help check if the fix is appropriate? Thanks @dewyatt !

randombit:
Re the check you had to remove: PKCS signature formatting is 0001FFFF.... but how Botan implements unpadded RSA signatures the leading 00 byte is removed. I think the correct check here is changing the test to `if(hashbuf_from_sig[0] != 1)` and modifying the following loop to start looking at index 1 instead of 2.

dewyatt:
@randombit Thanks! Ok I see that makes sense from the RFC (`EM = 0x00 || 0x01 || PS || 0x00 || T.`) so I added that check back in.

Anyways, it took me a while to finally realize it, but it looks like the issue was something different.
I think the main culprit here was that we were relying on uninitialized memory in the verification code.
```
$ src/rnp/rnp --verify inputfile.gpg 
sigbuf: 
9B F3 19 9C 4E 8C CE 4C AD 01 62 82 E0 F1 46 B8 A1 0F 0C 9D 6E 5E BD 24 6E EC 47 EE FB 6D 
E3 26 BE 94 CE 74 39 AC 58 FD 8B F0 31 EC 75 A4 52 A9 B7 31 33 90 0C D2 06 2B 72 85 9A 9A 
35 B4 E6 A0 74 AE AD B9 FB 4C DE 33 64 BD CB 9A EB 03 56 EF 66 A1 B8 FB 17 0F 70 36 D5 C0 
EE 63 EE 35 7B 44 1C 5E 2D 3E 72 4D 7F EB 3D 34 D8 C5 A2 1C EA 98 33 68 D1 8C C1 66 62 AB 
55 F1 0A FF 50 42 D7 36 

hashbuf_from_sig: 
20 01 FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 00 30 31 30 0D 06 09 60 86 48 01 65 03 04 
02 01 05 00 04 20 62 9E 85 6F 0A DD 1F 4B 06 CE 4F 5E 86 38 51 B7 75 25 C6 0A F4 77 1E B5 
CE 23 DE 1C A4 C8 56 C3 
```
That first byte in hashbuf_from_sig is uninitialized memory (so it tended to fail that check I removed).
I added a commit to zero the buffer in `pgp_rsa_public_decrypt` so that we're not returning any random garbage in the buffer.

Can you give it another quick look and let me know if it looks good to you?

randombit:
Argh yes uninitialized memory would do it, sorry about that - what you have here looks good to me.

PR #85, opened 2017-05-16
Fix DSA signature verification

### LINKED ISSUE #84: DSA signatures do not validate with rnp utils
`rnp --verify` and `rnpv` report signatures made with DSA as bad.

The signatures are valid (after #83) according to gpg.

dewyatt:
It looks like there was a TODO in there for this particular part.
I'm not familiar with DSA at all, so would love confirmation that q/r/s are always equal in size.
Closes #84.


ronaldtse:
@randombit could you help @dewyatt on this? Thanks!

randombit:
Will look at this today

randombit:
@dewyatt LGTM and yes r/s will always be encoded as exactly q-bytes (padded with 0s as necessary)

PR #89, opened 2017-05-17
Don't try to execute regex from NULL memory

catap:
Small bugfix for https://github.com/riboseinc/rnp/issues/88

### DIFF
--- a/src/lib/rnp.c
+++ b/src/lib/rnp.c
@@ -543,7 +543,7 @@ isarmoured(pgp_io_t *io, const char *f, const void *memory, const char *text)
 		}
 		(void) fclose(fp);
 	} else {
-		if (regexec(&r, memory, 10, matches, 0) == 0) {
+		if (memory && regexec(&r, memory, 10, matches, 0) == 0) {
 			armoured = 1;
 		}
 	}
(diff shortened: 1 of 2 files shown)


dewyatt:
Can you also add a check to rnpkeys.c to print an error and exit early if `f==NULL`:
https://github.com/riboseinc/rnp/blob/c40152e4c6554ea54158f20f4e1d04aadae409b1/src/rnpkeys/rnpkeys.c#L230

catap:
@dewyatt, it prints the error:
```
➜  rnp git:(not_crash_at_import_key_without_option) ✗ ./src/rnpkeys/rnpkeys --import-key                     
rnp: warning - cannot turn off core dumps
rnp: warning: core dumps enabled, sensitive data may be leaked to disk
Bad address
cannot import key from file (null)
➜  rnp git:(not_crash_at_import_key_without_option) ✗ 
```

catap:
But, I think you're right. The safe way is check that it is NULL, because printf's behaviour on NULL string isn't specified.

dewyatt:
I went ahead and merged this since it definitely looks good to me, and I think @ronaldtse is probably busy!

ronaldtse:
Thank you @catap for the changes and thanks to @dewyatt for the merge!

Issue #79, opened 2017-05-11
RSA signatures are invalid


dewyatt:
Current master doesn't seem to be generating valid signatures:
```
[root@3df60558e2be rnp]# git rev-parse HEAD
e1e2e5a7037a0545ef5df7e034f9c321bd36e839

[root@3df60558e2be rnp]# ./build.sh

[root@3df60558e2be rnp]# src/rnpkeys/rnpkeys --generate-key
rnp: warning - cannot turn off core dumps
rnp: warning: core dumps enabled, sensitive data may be leaked to disk
signature  2048/RSA (Encrypt or Sign) 65acfb9c064af9bd 2017-05-11 
Key fingerprint: d78f 0e87 6824 d737 7c28 3a24 65ac fb9c 064a f9bd 
uid              RSA 2048-bit key <root@localhost>
rnp: generated keys in directory /root/.rnp
Enter passphrase for 65acfb9c064af9bd: 
Repeat passphrase for 65acfb9c064af9bd: 

[root@3df60558e2be rnp]# echo -n test > inputfile

[root@3df60558e2be rnp]# src/rnp/rnp --sign inputfile 
rnp: warning - cannot turn off core dumps
rnp: warning: core dumps enabled, sensitive data may be leaked to disk
signature  2048/RSA (Encrypt or Sign) 65acfb9c064af9bd 2017-05-11 
Key fingerprint: d78f 0e87 6824 d737 7c28 3a24 65ac fb9c 064a f9bd 
uid              RSA 2048-bit key <root@localhost>
rnp passphrase: 
```

All tools are showing invalid signatures:
```
[root@3df60558e2be rnp]# src/rnp/rnp --verify inputfile.gpg 
rnp: warning - cannot turn off core dumps
rnp: warning: core dumps enabled, sensitive data may be leaked to disk
"inputfile.gpg": verification failure: 1 invalid signatures, 0 unknown signatures

[root@3df60558e2be rnp]# src/rnpv/rnpv -k ~/.rnp/pubring.gpg inputfile.gpg 
Signature did not match contents -- Signature does not match 

[root@3df60558e2be rnp]# gpg --homedir=~/.rnp --verify inputfile.gpg 
gpg: Signature made Thu May 11 00:50:18 2017 UTC using RSA key ID 064AF9BD
gpg: BAD signature from "RSA 2048-bit key <root@localhost>" [uncertain]
```

The self-sig on the key seems invalid as well:
```
[...]

PR #90, opened 2017-05-17
Add a test for RSA signature generation and verification

randombit:
This would have caught #79

Fixes two bugs I found along the way: in `pgp_getpassphrase` the password read from the file descriptor had its last character truncated. The library uses different codepaths to read the passphrase during keygen vs all other operations, so the result was that after generating a key with passphrase "abc" the decryption would fail because it would treat the entered passphrase as "ab" even though the pipe contained "abc". I think this doesn't come up in manual testing because `rnp` command line program doesn't use pass-fd. 

Also fix a bug in rsa_sign - if the requested hash is not SHA-1 or SHA-256 then it would just default to SHA-256, then be confused when the hash output a different length. Found because initially I was testing with MD5 also, which failed.

### DIFF
--- a/src/cmocka/rnp_tests.c
+++ b/src/cmocka/rnp_tests.c
@@ -311,7 +311,6 @@ static void raw_elg_test_success(void **state)
     BN_clear_free(pub_elg.y);
 }
 
-
 char *convert(char *buff, const int buffsize, unsigned int num, int base) {
     char *ptr;    
     ptr = &buff[buffsize - 1];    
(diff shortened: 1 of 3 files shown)


MohitKumarAgniotri:
@dewyatt I have been trying to add some test cases, but so far verification of the generated key has been a problem for me. I hope with this test case can act a way through. :+1:

dewyatt:
@randombit Thanks! And thank you for wrapping that call to write(), my compiler always complains about that unused result.
@MohitKumarAgniotri No problem, it looks like this will be good. I confirmed that this catches the issue mentioned. The tests are failing for a different reason, I'll create a separate issue for that.

PR #92, opened 2017-05-17
- Modified hash function length from 2048 to 1024

MohitKumarAgniotri:
- Updated pass-fd function to use base 10


dewyatt:
Great, thanks!

Closes https://github.com/riboseinc/rnp/issues/91

PR #94, opened 2017-05-18
Use Botan for PKCS signature formatting

randombit:
Completing #50. This probably shouldn't be merged quite yet - it relies on an unmerged Botan PR (https://github.com/randombit/botan/pull/1051), still waiting on CI there, but it should be merged to master today or tomorrow.

But, otherwise ready for review.

### DIFF
--- a/src/lib/rsa.c
+++ b/src/lib/rsa.c
@@ -184,58 +184,76 @@ done:
    \param pubkey RSA public key
    \return number of bytes decrypted
 */
-int
-pgp_rsa_private_encrypt(uint8_t *out,
-			size_t out_length,
-			const uint8_t *in,
-			size_t in_length,
-			const pgp_rsa_seckey_t *seckey,
-			const pgp_rsa_pubkey_t * pubkey)
+int pgp_rsa_pkcs1_sign_hash(uint8_t * sig_buf, size_t sig_buf_size,
+                            const char* hash_name, const uint8_t *hash_buf, size_t hash_len,
+                            const pgp_rsa_seckey_t *seckey,
+                            const pgp_rsa_pubkey_t *pubkey)
 {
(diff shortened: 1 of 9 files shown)


randombit:
Necessary PR to Botan has been merged to master and CI is green now - ready for review.

ronaldtse:
Thanks @randombit ! Would it be possible to help add some CMocka tests around these so we don't run into the issue last time?

@dewyatt would you have time to review this? Thanks!

randombit:
@ronaldtse I think the test from #90 already covers this change pretty well, but I will extend it to test all the available hashes.

src/lib/rsa.c
```diff
@@ -174,6 +136,77 @@ pgp_rsa_encrypt_pkcs1(uint8_t *out,
    return retval;
 }
 
+static void rnp_hash_to_botan_pkcs1_padding(char padding_name[], size_t len_padding_name,
+                                            const char* hash_name)
+{
+        // rnp uses SHAx Botan uses SHA-x
+        if(strcmp(hash_name, "SHA1") == 0)
+        {
+                strcpy(padding_name, "EMSA-PKCS1-v1_5(Raw,SHA-1)");
```
dewyatt: We should use strncpy on all these with `len_padding_name - 1` as the size:
```
strncpy(padding_name, "EMSA-PKCS1-v1_5(Raw,SHA-1)", len_padding_name - 1);
```

randombit: Good call, fixed in 9c9aa564

randombit:
I extended the test a bit to cover more hashes, also fixed a crashing bug (segfault) which occurred while trying to validate a signature with an unknown hash id - the return value of `pgp_hash_any` wasn't being checked.

Issue #55, opened 2017-04-13
Support SM3 hash function using Botan implementation


ronaldtse:
Thanks to @dewyatt (and @randombit ) we have a SM3 hash implementation merged to Botan. This task is to implement support of it in rnp within the OpenPGP framework.

ronaldtse:
Thanks @randombit for handling this. I have to let you know that SM3 isn't currently in the OpenPGP spec (RFC4880 or RFC4880bis01). We will need to create an extension to the spec 👍

ronaldtse:
Since we're going down this track we should also support the other Chinese requirements, SM2 (Public Key ECC), SM4 (Symmetric Cipher), by adding them into Botan and then rnp.

randombit:
@ronaldtse For SM3 we just need to pick a value in the OpenPGP "Private/Experimental algorithm" range to identify the hash within the OpenPGP message format (range is 100 to 110 - I'm not sure if any other implementation uses any of these values, it would be nice to avoid conflicts if we can), and also create figure out what OID is used for PKCSv1.5+SM3 signatures (Botan will need to include those). I found this doc http://gmssl.org/docs/oid.html that specifies OIDs for SM3 hash as well as RSA+SM3 signatures under the OID arc reserved for the "Chinese Cryptography Standardization Technology Committee" so I guess those are probably the values to use.

There is a description in English of SM4 here http://eprint.iacr.org/2008/329.pdf
And of the SM2 algorithms here - https://tools.ietf.org/html/draft-shen-sm2-ecdsa-02

ronaldtse:
Indeed @randombit ! GMSSL is a fork of OpenSSL that contains implementations of the SM series, and both SM2 and SM3 have been published as drafts on IETF. Therefore we could do SM2/3 now as extensions on OpenPGP given support in Botan. For the SM4 cipher we could refer to the paper and also the GMSSL implementation.

I do not know of any experimental signature/encryption schemes currently being publicly used with OpenPGP except with the now standardized extensions. I assume we have to work with the IETF PGP group to get an assignment. The GMSSL OIDs are official AFAIK.

randombit:
Done!

PR #96, opened 2017-05-22
Add support for SM3 hash function

randombit:
GH #55. Currently uses a hash id in the private range (105).

### DIFF
--- a/src/cmocka/rnp_tests.c
+++ b/src/cmocka/rnp_tests.c
@@ -24,7 +24,6 @@
  * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-
 #include <stdarg.h>
 #include <stddef.h>
 #include <setjmp.h>
@@ -460,6 +459,7 @@ static void rnpkeys_generatekey_testSignature(void **state)
 {
     const char* hashAlg[] = {
         "SHA1",
+        "SHA224",
         "SHA256",
         "SHA384",
         "SHA512",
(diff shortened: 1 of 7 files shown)


ronaldtse:
@randombit do we need to also update the `pgp_is_hash_alg_supported` function? I did some initial tracing it seems that this function is only used in `process_dash_escaped`, but `process_dash_escaped` doesn't even seem used anywhere. If that's the case could we remove both unused functions too? Thanks!

dewyatt:
`process_dash_escaped` is used in reader.c for reading clearsigned messages, so `pgp_is_hash_alg_supported` probably does need some updates.

ronaldtse:
My bad about thinking `process_dash_escaped` isn't used. I need better eyes. Look forward to the updates.

ronaldtse:
Speaking of support within `pgp_is_hash_alg_supported`, it only returns `1` on:

``` c
PGP_HASH_MD5
PGP_HASH_SHA1
PGP_HASH_SHA256
```

But not for these:
``` c
PGP_HASH_SHA512
PGP_HASH_SHA384
```

We should add those in as well. Maybe there is a better way indicating supported hash functions instead of having these constants littered in multiple places?

randombit:
@ronaldtse Suggested changes all sound good. On another project tomorrow, but should have updates to this PR later in the week.

The current sign+verify test doesn't test clearsigned messages because while `rnp_sign_memory` can create one, I could not figure out how to verify a clearsigned message using `rnp_verify_memory` - hints welcome.

dewyatt:
> I could not figure out how to verify a clearsigned message using rnp_verify_memory - hints welcome.

It looks like clearsign verification with `rnp_verify_memory` doesn't work right now, so I'll create an issue. I recall the clearsign functionality was originally incomplete upstream so I did some work on that for signing (435f9fbd2fcba73e95eab884f150729efde28694) but not verifying.

`rnp_verify_file` can, however, verify clearsign messages.

dewyatt:
@randombit Wait...I think I'm wrong on that. I think you should currently be able to verify clearsigned messages with rnp_verify_memory. I'll get back to you on that.

dewyatt:
@randombit Ok `rnp_verify_memory` will indeed verify clearsign messages. I suspect you may have had trouble because you were including the null terminator in the signature payload. Something like this would work for testing clearsign:
https://gist.github.com/dewyatt/b9dd6cb8127056ec8bfddd830de76689

(and `pgp_is_hash_alg_supported` would indeed need to be expanded to support the other hashes in the current test.)

I'll file an issue to discuss whether there is a bug here (don't want to derail this PR).

dewyatt:
@randombit Hmm, actually maybe it's just my lack of sleep. It should verify fine with the NULL terminator as well. Just make sure `armored=1` and that you stick to hashes that `pgp_is_hash_alg_supported` currently likes. I do wonder if `pgp_sign_buf` should explicitly fail when `cleartext=1` and a null byte is in the data, but that's a separate discussion.

randombit:
@dewyatt Am I right in thinking that if clearsigned=1 and armored=0, nothing useful happens? `rnp_sign_memory` succeeds, but I can't get the output to verify and it doesn't seem like the combination makes sense since clearsigned is for plaintext transport/display. Should `rnp_sign_memory` (or some lower-level function down the call stack) fail when clearsigned=1,armored=0 combination is used?

dewyatt:
@randombit It looks like `pgp_sign_buf` completely ignores `armored` when `cleartext=1`. So it would probably produce valid armored output, but then you try to verify it as non-armored.

So I think we should have `pgp_sign_buf` fail with that combination of arguments as it's pretty misleading. Same thing for `pgp_sign_file`.

dewyatt:
Also, I noticed GnuPG does allow NULL bytes in clearsign messages, and they verify fine, etc. Seems weird, but I guess it's OK.

I think the remaining task here is updating `pgp_str_to_hash_alg`.
I think we need to have the tests check the return value of rnp_setvar for "hash" since it does fail: https://github.com/riboseinc/rnp/blob/2d21450e8f5601da0f203507966c3a7308469bd3/src/lib/rnp.c#L2181

randombit:
@dewyatt SM3 was already in `pgp_str_to_hash_alg` but I enabled SHA-224 while in there. RIPEMD-160 is not supported by hash.c at all right now so I disabled the attempt to test it in `rnpkeys_generatekey_verifySupportedHashAlg`

dewyatt:
@randombit That makes sense. Ready for merge?

randombit:
Yes I think all ready here

PR #117, opened 2017-05-25
Set feature macros for glibc making nftw and mktdtemp visible

randombit:
Both are not visible by default with glibc 2.25 (on Arch), I see the following errors

```
rnp_tests.c:144:78: error: 'struct FTW' declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
 int remove_cb(const char *fpath, const struct stat *sb, int typeflag, struct FTW *ftwbuf)
                                                                              ^~~
rnp_tests.c: In function 'delete_recursively':
rnp_tests.c:162:5: error: implicit declaration of function 'nftw' [-Werror=implicit-function-declaration]
     nftw(path, remove_cb, 64, FTW_DEPTH | FTW_PHYS);
     ^~~~
rnp_tests.c:162:31: error: 'FTW_DEPTH' undeclared (first use in this function)
     nftw(path, remove_cb, 64, FTW_DEPTH | FTW_PHYS);
                               ^~~~~~~~~
rnp_tests.c:162:31: note: each undeclared identifier is reported only once for each function it appears in
rnp_tests.c:162:43: error: 'FTW_PHYS' undeclared (first use in this function)
     nftw(path, remove_cb, 64, FTW_DEPTH | FTW_PHYS);
```

### DIFF
--- a/src/cmocka/rnp_tests.c
+++ b/src/cmocka/rnp_tests.c
@@ -25,6 +25,9 @@
  */
 
 
+#define _XOPEN_SOURCE 500
+#define _POSIX_C_SOURCE 200809
+
 #include <stdarg.h>
 #include <stddef.h>
 #include <setjmp.h>
@@ -34,6 +37,7 @@
 #include <stdio.h>
 #include <unistd.h>
 #include <limits.h>
+
 #include <ftw.h>
 
 #include <cmocka.h>


dewyatt:
Right, I knew this would be an issue when I wrote it, and then promptly forgot!
I think we should have a larger discussion about this though.

In CI, we actually define `_GNU_SOURCE` and use `--std=c11`.
Locally, we do neither.

C11 and `_GNU_SOURCE` are both required to build rnp with clang (even before my use of `nftw`/`mkdtemp`). I suspect most of us aren't building locally with clang so we typically don't see an issue. GCC seems to be much more lenient.

So I think short-term, we ought to update the local build scripts to use C11 and `_GNU_SOURCE` (or more specific feature macros, like you did here).

randombit:
Using C11 across the board seems best to me, as long as the slightly reduced portability (eg compiling rnp on old Solaris or the like) is not an issue.

dewyatt:
I think it's best to close this and look at it in more detail in https://github.com/riboseinc/rnp/issues/122.
I think this PR would only address the compilation issues with GCC, so we should investigate further.

Issue #111, opened 2017-05-25
Overflow in write_seckey_body


dewyatt:
I was playing with clang's sanitizers and confirmed an issue here:
https://github.com/riboseinc/rnp/blob/f20e993e7b7e9e11e70f9b71f1db2b005746454c/src/lib/create.c#L434
We're calculating a SHA1 hash but we're using the hash size of `key->hash_alg` to determine the size of the buffer we allocate.
So this overflows in the tests when `key->hash_alg` is something that would produce a digest smaller than SHA1, like MD5.

It's not immediately clear to me what the best fix is here so I'm just documenting it.

ronaldtse:
It seems that this function is pretty much hardcoded SHA1 and CAST, we should probably support different digest and symmetric algorithms properly here.

randombit:
I believe this was fixed by #140

dewyatt:
As mentioned, this should be fixed now, closing.

PR #119, opened 2017-05-25
Fix a 4-byte memory overwrite when encrypting private key with hash=MD5

randombit:
Found by running valgrind over the tests.

The buffer to hold the hash output was sized to pgp_hash_size(key->hash_alg) but the actual hash used for the S2K is always SHA-1, so with an MD5 hash this caused a 4-byte overflow.

Also, free the buffer :)

### DIFF
--- a/src/lib/create.c
+++ b/src/lib/create.c
@@ -435,11 +435,13 @@ write_seckey_body(const pgp_seckey_t *key,
 			hashsize = pgp_hash_size(key->hash_alg);
 			needed = PGP_CAST_KEY_LENGTH - done;
 			size = MIN(needed, hashsize);
-			if ((hashed = calloc(1, hashsize)) == NULL) {
+
+			if ((hashed = calloc(1, PGP_SHA1_HASH_SIZE)) == NULL) {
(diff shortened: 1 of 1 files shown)


dewyatt:
See also: https://github.com/riboseinc/rnp/issues/111

We may want to scrutinize this some more, as we're still using hashsize in here, which may be incorrect.

randombit:
@dewyatt Thanks had missed #111 

Looking at it more I think my fix here is bogus, `hashsize` should be equal to the output len of the hash being used to generate the key, and the key's hash_alg (which I think is just what key is used for signatures?) is irrelevant. (Or, that is my understanding currently.)

It does actually still work, because the function is hardcoded to use CAST5 (128 bit key) so the loop always completes after a single iteration. However it would be wrong if `write_seckey_body` supported another cipher (like AES-256) and was used with a hash with output len < cipher key len. Then the function would write to an incorrect offset in the session key, and produce an undecryptable private key.

Anyway this shouldn't be merged until we are sure about what's going on.

randombit:
Obsoleted by changes in #140

PR #120, opened 2017-05-25
Initialize memory from realloc in pgp_memory_pad

randombit:
I'm not so certain about this change, it might just be papering over some other problem. But I see the following in valgrind

```
Conditional jump or move depends on uninitialised value(s)
    at 0x4C2E128: strlen (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
    by 0x4E4CB6D: rnp_strdup (misc.c:1344)
    by 0x4E4AA38: pgp_export_key (keyring.c:1067)
    by 0x401996: rnpkeys_exportkey_verifyUserId (rnp_tests.c:928)
  Uninitialised value was created by a heap allocation
    at 0x4C2D13F: realloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
    by 0x4E4B6DB: pgp_memory_pad (misc.c:888)
    by 0x4E4B794: pgp_memory_add (misc.c:910)
    by 0x4E64DAF: memory_writer (writer.c:1271)
```

Zeroing the return value of `realloc` at least prevents using uninitialized memory.

Maybe the problem really is that `pgp_write_xfer_pubkey` and `pgp_write_xfer_seckey`, if emitting armored data, need to emit a trailing null at the end?

Consider this more of an issue for discussion than a patch, though maybe zeroing this memory really is the right thing to do.

### DIFF
--- a/src/lib/misc.c
+++ b/src/lib/misc.c
@@ -884,12 +884,15 @@ pgp_memory_pad(pgp_memory_t *mem, size_t length)
 		return;
 	}
 	if (mem->allocated < mem->length + length) {
-		mem->allocated = mem->allocated * 2 + length;
-		temp = realloc(mem->buf, mem->allocated);
+                size_t newsize = mem->allocated * 2 + length;
+		temp = realloc(mem->buf, newsize);
 		if (temp == NULL) {
 			(void) fprintf(stderr, "pgp_memory_pad: bad alloc\n");
+                        // TODO return an error here
 		} else {
 			mem->buf = temp;
+                        memset(mem->buf + mem->allocated, 0, newsize - mem->allocated);
+                        mem->allocated = newsize;
 		}
 	}
 	if (mem->allocated < mem->length + length) {


dewyatt:
I've been seeing the same in the sanitize build runs:
https://travis-ci.org/riboseinc/rnp/jobs/236404120

I'll see if I can find any deeper issue. Your point about a missing NULL terminator certainly would make some sense.

dewyatt:
I think you're right that this would be covering up other problems if merged.
This wouldn't cover cases where `mem->length==mem->allocated`, for example:

```c
pgp_memory_t mem = {0};
pgp_memory_init(&mem, 10);
pgp_memory_add(&mem, "0123456789", 10); //no need to realloc, it fits
strlen(pgp_mem_data(&mem)); //overrun
```

I would say the real issue is treating a `pgp_memory_t` buffer as a string.
I think they are generally never NULL-terminated, which is why they have a `length` member.

I think in this case, the `rnp_strdup` in `pgp_export_key` should have been something like this instead:
```c
cp = calloc(1, pgp_mem_len(mem) + 1);
memcpy(cp, pgp_mem_data(mem), pgp_mem_len(mem));
```

Generally, I think most functions in the library that write armored text to a buffer *do not* add NULL terminators, and only return a length. `rnp_export_key`/`pgp_export_key` are a bit different since they directly return `char*`.

If we look at `rnp_sign_memory`, for example, it returns the length and does not add a NULL terminator (whether armored or not). I think this is the correct behavior because the signature should not be considered to be a NULL-terminated string, as it could in fact contain embedded NULLs.

Specifically, rnp and gpg both allow a cleartext signature on data with NULLs. So NULL-termination wouldn't make any sense there.

```
$ echo -ne "test1\x00test2" | gpg --clearsign | xxd
gpg: selecting openpgp failed: No such device
00000000: 2d2d 2d2d 2d42 4547 494e 2050 4750 2053  -----BEGIN PGP S
00000010: 4947 4e45 4420 4d45 5353 4147 452d 2d2d  IGNED MESSAGE---
00000020: 2d2d 0a48 6173 683a 2053 4841 310a 0a74  --.Hash: SHA1..t
00000030: 6573 7431 0074 6573 7432 0a2d 2d2d 2d2d  est1.test2.-----
00000040: 4245 4749 4e20 5047 5020 5349 474e 4154  BEGIN PGP SIGNAT
00000050: 5552 452d 2d2d 2d2d 0a0a 6946 3045 4152  URE-----..iF0EAR
00000060: 4543 4142 3057 4951 524f 3843 2f6c 664e  ECAB0WIQRO8C/lfN
00000070: 3234 6b4e 5558 4177 5347 674c 6f34 3836  24kNUXAwSGgLo486
00000080: 4d71 4941 5543 5753 6837 4f77 414b 4352  MqIAUCWSh7OwAKCR
00000090: 4347 674c 6f34 3836 4d71 0a49 4d35 5a41  CGgLo486Mq.IM5ZA
000000a0: 4a34 7a58 4a74 6139 5064 702f 6361 5872  J4zXJta9Pdp/caXr
000000b0: 7549 546d 4741 3767 7044 454f 4143 6767  uITmGA7gpDEOACgg
000000c0: 4669 375a 5665 4252 306b 7751 3771 6f39  Fi7ZVeBR0kwQ7qo9
000000d0: 6f79 4a6f 6933 2b48 6a77 3d0a 3d63 6b2b  oyJoi3+Hjw=.=ck+
000000e0: 2f0a 2d2d 2d2d 2d45 4e44 2050 4750 2053  /.-----END PGP S
000000f0: 4947 4e41 5455 5245 2d2d 2d2d 2d0a       IGNATURE-----.
```

Anyways, as far as `pgp_write_xfer_seckey`/`pgp_write_xfer_pubkey`, I think the behavior of not adding a NULL terminator is probably correct and consistent as well.

dewyatt:
Closing this old one as this has been fixed for a while now. :)

PR #124, opened 2017-05-25
Add -std=c11 -D_GNU_SOURCE as global compiler flags.

dewyatt:
This would mostly be a quick solution to the issues in https://github.com/riboseinc/rnp/pull/117/files.

Related: https://github.com/riboseinc/rnp/issues/122


ronaldtse:
@dewyatt we'll first merge this but there are portability concerns esp. with BSD. It's certainly more important now to make rnp consistent first though. Could we document in a new issue what exact functions we are using through `GNU_SOURCE` definitions?

PR #131, opened 2017-05-26
Rework the logic surrounding coredumps a bit.

dewyatt:
I think this is right, though it's a bit confusing.
rnp_init can now determine the following and act accordingly:

* Core dumps are possibly enabled
* Core dumps are definitely disabled
* Core dumps are possibly enabled because we failed to disable them
  * because we were compiled without sys/resource.h
  * because calls to setrlimit/getrlimit failed

Previously the error messages "cannot turn off core dumps" and "core dumps enabled" were always printed in tandem.
Now, they are distinct conditions.

Normal case:
```
$ rnpkeys --generate-key
[...] no messages about coredumps
```

Explicitly enabled coredumps:
```
$ rnpkeys --generate-key --coredumps
rnp: warning: core dumps enabled, sensitive data may be leaked to disk
```

Compiled without sys/resource.h, or setrlimit/getrlimit fail:
```
$ rnpkeys --generate-key
rnp: warning - cannot turn off core dumps
rnp: warning: core dumps enabled, sensitive data may be leaked to disk
```

### DIFF
--- a/src/lib/rnp.c
+++ b/src/lib/rnp.c
@@ -908,7 +908,7 @@ disable_core_dumps(void)
            error = getrlimit(RLIMIT_CORE, &limit);
            if(error)
               {
-              return error;
+              return -1;
               }
            else if(limit.rlim_cur == 0)
               {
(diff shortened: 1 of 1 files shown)


randombit:
Looks ok to me. These messages were bugging me, especially seeing them printing over and over again in the tests :P

dewyatt:
@randombit Right! And they made it hard to see errors in the output.

Ok, merging.

Issue #107, opened 2017-05-23
Implement Blowfish/Twofish (symmetric algorithms 4, 10)


ronaldtse:
We should support Blowfish and Twofish symmetric ciphers. Usage as described in RFC 4880.

* Symmetric key algorithm IDs: 4 (Blowfish), 10 (Twofish)

ronaldtse:
Thanks for tackling this @randombit !

ronaldtse:
Done thanks to @randombit #132 .

Issue #101, opened 2017-05-23
Split `hash.c` hash functions and `symmetric.c` encryption functions into separate directories/files


ronaldtse:
For consistency with public/private key functions which have been split in #59.

Perhaps we can separate them into the following directories:

* asymmetric
* symmetric
* hash

?

ronaldtse:
Thanks @randombit for handling this!

PR #132, opened 2017-05-26
Clean up symmetric cipher code

randombit:
I started by trying to make the structure completely opaque but kind of ran into a quagmire with that, so for now I limited the change to removing the function pointers from `pgp_crypt_t`. This alone greatly simplifies `symmetric.c`

While there implemented support for Blowfish/Twofish (#107), and also Camellia-192 which was missing.

At this point the only algorithm-specific code in `symmetric.c` are a handful of switch statements, so I'm not sure if breaking up the file (as suggested in #101) makes sense anymore - @ronaldtse your thoughts?

 Changes to `src/lib` are done and ready for review, and there is even a test now.

### DIFF
--- a/src/cmocka/rnp_tests.c
+++ b/src/cmocka/rnp_tests.c
@@ -273,32 +273,32 @@ static void cipher_test_success(void **state)
 
     memset(iv, 0x42, sizeof(iv));
 
-    crypt.set_crypt_key(&crypt, key);
-    crypt.block_encrypt(&crypt, block, block);
+    pgp_cipher_set_key(&crypt, key);
+    pgp_cipher_block_encrypt(&crypt, block, block);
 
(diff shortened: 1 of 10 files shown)


ronaldtse:
@randombit agree that it's no longer necessary to split `symmetric.c`, and this is already so much cleaner. Apparently the previous merge in #135 introduced a conflict, would you be able to fix it?

On the other hand, `crypto.c` does contain RSA/ElGamal specific things that could be better structured...

randombit:
@ronaldtse Rebased and fixed the merge conflict. I agree `crypto.c` is still a total grab bag and could use some help, I'll look at doing some cleanups there in the spirit of #101

ronaldtse:
Thank you @randombit !

Issue #114, opened 2017-05-25
Implement RFC4880 3.7.1.3 Iterated and Salted S2K


ronaldtse:
As title. Location: https://github.com/riboseinc/rnp/blob/f20e993e7b7e9e11e70f9b71f1db2b005746454c/src/lib/create.c#L481

randombit:
Maybe useful reference, Botan used to include this (was dropped some years ago because it was not particularly useful outside the context of a full PGP impl, and hey here we are). The last commit before it was removed is

https://github.com/randombit/botan/tree/75db296a459a9e25b112207707cc5e26a6f2b872/src/pbkdf/pgps2k

ronaldtse:
Things do come circle! @randombit any chance of re-instating this in botan?

randombit:
@ronaldtse Yes I think re-adding this to Botan (as well as the non-iterated S2K currently implemented in rnp) then changing rnp to use the PBKDF ffi interface is the way to go.

randombit:
Weirdly packet-parse.c already implements iterated S2K, but create.c doesn't, and it appears there is a third implementation of it in `ssh2pgp.c`. Yuk. Anyway I'll clean all that up as part of this issue.

ronaldtse:
You must have had a good idea in mind when you found all these 3 implementations. Thanks @randombit 😄

ronaldtse:
Closed with #140 .

Issue #106, opened 2017-05-23
See if we can make hash handling a bit more consistent and improve related tests


dewyatt:
Related discussion in https://github.com/riboseinc/rnp/pull/96

It looks like there are a few different places that deal with hashes that are a bit scattered around.
I wonder if we can improve this:

* `pgp_is_hash_alg_supported`
* `pgp_hash_any`
* `pgp_str_to_hash_alg`

Additionally, with the current tests, there are a few errors that are printed out because these functions need updates, but the tests pass (probably by defaulting to a different algorithm):

* `pgp_hash_any: bad algorithm` (`pgp_hash_any`)
* `Ignoring unknown hash algo 'RIPEMD160'` (`pgp_str_to_hash_alg`)
* `Ignoring unknown hash algo 'SHA224'` (`pgp_str_to_hash_alg`)

ronaldtse:
Fully agree with @dewyatt . I'm hoping that we can consolidate all these in a separate file per hash algorithm.

ronaldtse:
@randombit will be working on this. Would be great if we can make it much simpler to add hash algorithms with a single touch point. Thanks!

randombit:
Think this is done with #135 - adding a new hash now requires just modifying `hash.h` to add the codepoint and `hash.c` to modify a couple of lookup tables.

ronaldtse:
Thanks @randombit !

PR #135, opened 2017-05-26
Cleanup digest code

randombit:
Addressing #106 

Lot of changes in here but nothing too scary. Did not make `pgp_hash_t` completely opaque because some places create `pgp_hash_t` structs on the stack and I didn't want to change them all. However it removes the function pointer members (init, add, finish) and forces all operations to go through functions, instead of accessing any `pgp_hash_t` struct members directly. So opaque in practice at least, and a stepping stone for making the struct fully (actually) opaque.

Another change is removing the (IMO) weird system of calling `pgp_hash_any` then having to call `hash->init` afterwards - now `pgp_hash_create` does both operations. Also removed the algorithm specific constructors (like `pgp_hash_md5`) in favor of `pgp_hash_create(&hash, PGP_HASH_MD5)`

I also removed the `digest_t` wrapper from `rnpv` that was just wrapping `pgp_hash_t`.

Again wrt #101 it doesn't seem like there is any need to split up `hash.c` into multiple files anymore, everything is just handled by a couple of switches. Adding a hash now just means modifying `hash.h` (to add the PGP codepoint to `pgp_hash_alg_t` - I moved this enum from `packet.h`) and then at the top of `hash.c` to add the name.

Oh, git in its infinite wisdom seems to have somehow deduced that I moved `rnpv/digest.h` to `src/lib/hash.h`. In fact `hash.h` was added in the first commit, then `digest.h` deleted later, but I guess they have enough text in common that git decided it was a move, even though the file was added in a commit before the commit where to the 'source' file was removed. So the diff looks weird there. (wtf git)


randombit:
@dewyatt Yeah that will go away when I fix #114

ronaldtse:
The removal of `hash->init` is extremely welcome and just makes the flow so much easier to understand. Thanks @randombit !

ronaldtse:
@randombit in the changes there are some places where it accounts for undefined SHA-1 and MD5 (via `BOTAN_HAS_x`), but both are "required" in RFC 4880. Maybe we should change it to a switch that if those aren't defined, then rnp should fail outright?

randombit:
@ronaldtse Sounds good I will make a change so configure tests for the essential feature macros (RSA, PKCS1v15 padding, SHA-1, etc) and fails if any are not available in the Botan build.

ronaldtse:
Great plan, thanks @randombit !

PR #138, opened 2017-05-28
Check for required features in Botan library at configure time

randombit:
See discussion in #135


ronaldtse:
Thank you @randombit !

PR #139, opened 2017-05-28
Jsonc port

zgyarmati:
Fixing #7, doesn't really touch any other parts of the code just the rnpkeys' sources and the corresponding json creating/dumping functions, unlikely to conflict with any ongoing changes, maybe only in the auto{make|conf} files, just let me know if this happen


src/rnpkeys/rnpkeys.c
```diff
@@ -182,8 +181,9 @@ print_usage(const char *usagemsg)
 static int
 match_keys(rnp_t *rnp, FILE *fp, char *f, const int psigs)
 {
-	char	*json;
+	char	*json = NULL;
 	int	 idc;
+//    rnp_list_keys_json(rnp, &json, psigs);
```
ronaldtse: This comment should be removed?

src/lib/rnp.c
```diff
@@ -637,10 +639,9 @@ p(FILE *fp, const char *s, ...)
 
 /* print a JSON object to the FILE stream */
 static void
-pobj(FILE *fp, mj_t *obj, int depth)
+pobj(FILE *fp,  json_object *obj, int depth)
```
ronaldtse: Extra space in front of `json_object`?

src/lib/rnp.c
```diff
@@ -705,77 +702,96 @@ ptimestr(char *dest, size_t size, time_t t)
...
-	mj_t	*sub;
-	int	 i;
 
 	if (pgp_get_debug_level(__FILE__)) {
-		mj_asprint(&s, obj, MJ_HUMAN);
-		(void) fprintf(stderr, "formatobj: json is '%s'\n", s);
-		free(s);
+		(void) fprintf(stderr, "formatobj: json is '%s'\n", json_object_to_json_string(obj));
 	}
+#if 0 //?
```
ronaldtse: Could we handle the "revoked" case here which was commented out as well?

ronaldtse:
@zgyarmati thank you for this and to @dewyatt on the CI! Just some minor changes requested.

src/lib/rnp.c
zgyarmati: Good point, i missed to clean this up, and there is a potential issue here:
we handle the scenario when an userid is revoked (see line 767), but this code-part supposed to check whether the whole cert is revoked, however so far i can't find any evidence that this information can get into the generated .json (see the pgp_sprint_mj() function in the lib/packet-print.c file, which might be a flaw already in the original json generation, but i might missed something here, what do you think?

ronaldtse:
@zgyarmati the commit 71a1247 seems to have changed a lot of formatting between whitespaces and tabs. I understand that our current unclear policy on whitespaces vs tabs is confusing. In any case, can we revert 71a1247 first, and only change the two minor issues to prevent whitespaces clobbering up proper code history? Sorry for the trouble 👍

zgyarmati:
OK, gonna do that. Basically what happened is that my editor was set to use space indent, so the code i added was space indented while the rest of those files used tabs, so i thought better to set it back. Indeed, a policy would come handy on this :)

dewyatt:
I'm actively working on a whitespace/style policy as part of https://github.com/riboseinc/rnp/issues/61, which will likely include an auto-formatting tool (`uncrustify` or `clang-format`). Sorry it's not quite ready for review yet.

zgyarmati:
OK, sounds perfect, then IMHO till that we can ignore any whitespace formatting and get it right in one step.

ronaldtse:
Thank you @zgyarmati ! Merged!

dewyatt:
Actually, I hate to do this, but can we undo this and revert https://github.com/riboseinc/rnp/commit/71a124702a7bf65b6f705014d540f0110471bdb9 ?
It looks like the whitespace changes also occurred inside literal strings, which is probably not something we should let through right now.

zgyarmati:
@dewyatt it was reverted (2019ccb056d9921d0e7c510e9b72f98a2c44210a) before merge (per @ronaldtse's request), so we are set, sorry for the noise!

dewyatt:
@zgyarmati I see now, thanks!

ronaldtse:
@dewyatt I probably should have been more diligent with clearing off the revert entirely with a rebase. 😿

PR #140, opened 2017-05-29
Use Botan for S2K operations

randombit:
Now the 3 (!) different S2K implementations are removed, salted+iterated S2K mode is supported (#114), and the S2K hash is no longer hardcoded to use SHA-1 (instead whatever hash is specified for the key is used). Requires latest Botan master to work since PGP-S2K was only just added back.

I'm on another project for the rest of this week, so I may be slow about addressing any review comments.


randombit:
The first build failed due to the cached libbotan not having PGP-S2K - I cleared the caches on Travis and restarted.

dewyatt:
It looks like `testSignature` failed with the clang (normal) build, but passed with the gcc (coverage) build:

https://github.com/randombit/rnp/blob/b55aa38b09d652ed14ccfe2e2c436f1a68ef3055/src/cmocka/rnp_tests.c#L536

I even purged the cache again and gave it another shot, but no change (nice that it's consistent).

It sort of reminds me of the issue we saw in https://github.com/randombit/botan/pull/966 but that's a bit of a long shot.

Anyways, I guess we can follow up next week.

ronaldtse:
@randombit I have to say that this cleaned up so much of the legacy code we have, and I'm blown away. Hope Travis' cache gets cleared properly soon and shows this as a go!

ronaldtse:
@randombit BTW seems that this obsoletes #119 too?

flowher:
LGTM, good stuff

ronaldtse:
It's currently failing because of this, probably can be fixed if the PR is rebased on master since the JSON stuff was merged?
```
./ci/install.sh: line 26: JSON_C_INSTALL: unbound variable
```

randombit:
I was not able to replicate the test failure locally, but valgrind pointed to an uninitialized variable usage in the tests, around a call to strlen in the s2k code. Likely I'm just getting 'lucky' and that uninitialized stack memory happened to be zeroed out on my machine. Zeroing out the passphrase buffer in 6b1cbb0 resolves the valgrind warning and also seems to please CI.

ronaldtse:
Thank you @randombit for getting the checks working!

PR #142, opened 2017-05-29

flowher:
* Improves `set_core_dumps` readability


flowher:
argh....

PR #143, opened 2017-05-29

flowher:
* Makes set_core_dumps more readable


flowher:
Argh... let me re-do this commit

PR #145, opened 2017-05-30

flowher:
* Makes set_core_dumps more readable


ronaldtse:
Looks great. Thanks @flowher for the fix!

ronaldtse:
It looks like places like write_seckey_body still have hard coded key lengths for CAST. We should make a new issue to clean them up all at once.

randombit:
@ronaldtse All of those cases are resolved in #140

ronaldtse:
Got it, thanks @randombit !

PR #146, opened 2017-05-30
This code can't be build without json-c, fix it

catap:
The beset way is made it optional and don't try to build the code with json-c dependencies but stop at configure is better that pass it and crash at building.


zgyarmati:
Indeed, thx for spotting and patching

Issue #61, opened 2017-04-27
Make indentation whitespace consistent throughout code


ronaldtse:
Right now the code uses a mix of tabs and spaces for indentation. We should make it consistent and document the choice in a "Contribution guidelines" section of README.

dewyatt:
Some conventions we should probably nail down:

- Indentation (tabs vs spaces, count, etc.)
- General style (K&R etc.)
- General return values (int vs unsigned). Some functions return 0/1, some return -1, etc.

We may want to pull some of the C-specific portions of [Google's C/C++ style guide](https://google.github.io/styleguide/cppguide.html).

ronaldtse:
@dewyatt thanks for this. Would you mind being the coordinator for this and authoritatively set the rules in a "Contribution Guidelines" section within `README.md`?

dewyatt:
Closed in 48d36df5a9ae2aa51a00c72f0ebce16b4aa26f59

dewyatt:
I think I better re-open this because:

1. I missed `include/rnp.h` in the re-format (oops).  I think I must have been in the `src/` subdir instead of top-level like I should have been.
2. Clang 5.0.0 apparently fixes block comment alignment, so I'd like to evaluate whether it's feasible to get everyone on that version.
3. There are still some stubborn tabs in the source that I should purge.

PR #148, opened 2017-05-31
Developers guide and clang-format styling

### LINKED ISSUE #128: Write a "Developer's guide"
Topics:

* Getting started with developing rnp
* Coverity instructions
* How to use fuzzer (#97)
* Code conventions

dewyatt:
Feedback would be much appreciated.

There's mostly 2 things here:

#### Developer's Guide

This is just a simple introduction to some of the things we use (travis, codecov, coverity, ...), and some general guidelines, etc.
It can probably be expanded later on, and I guess we would want to add a link from the README.md.

#### clang-format

The idea here is to introduce clang-format for keeping a consistent code base.
I basically pinned down the settings for the current code base, as closely as possible.
The only major change I added in here was using spaces, rather than tabs.

I think all of us have had issues with tabs and accidentally committed spaces.
Every editor seems to handle tabs differently, and I think we may be better off avoiding them.

I created a git pre-commit hook that works well for me. It looks like this:
```
$ git commit -m 'do some commit with incorrect formatting'
--- src/lib/crypto.c	2017-05-30 20:58:48.279764787 -0400
+++ -	2017-05-30 20:58:48.309445645 -0400
@@ -495,8 +495,7 @@
 	pgp_memory_add(inmem, input, insize);
 
 	/* set up to read from memory */
-pgp_setup_memory_read(io,
- &parse, inmem, NULL, write_parsed_cb, 0);
+	pgp_setup_memory_read(io, &parse, inmem, NULL, write_parsed_cb, 0);
 
 	/* setup for writing decrypted contents to given output file */
 	pgp_setup_memory_write(&parse->cbinfo.output, &outmem, insize);

Formatting changes requested.
See /tmp/git-clang-format.lTv6U4.patch

Apply now (y/n)? y
[master b9c3b10] do some commit with incorrect formatting
```
It just shows a colorized diff of the formatting violations and asks if you would like to apply that before committing (else the commit is aborted).

It defaults to using a docker container (which has a slight delay), but if you have clang-format v4 (Arch Linux does) locally, then you can set `USE_DOCKER="no"` in `pre-commit.sh`.

Before we would use the hook, we would have to run clang-format on the existing code base.

The only issue I'm aware of with the styling that clang-format does is with the asterisk alignment in function definitions (IIRC):
https://github.com/llvm-mirror/clang/blob/6878143d92cd036e9916ff853b8bc522e6e118b7/lib/Format/WhitespaceManager.cpp#L406

It's a nuisance but hopefully will get fixed.

Closes #128 and #61

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,67 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Mozilla
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: true
+AlignEscapedNewlinesLeft: true
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: All
+AlwaysBreakAfterReturnType: AllDefinitions
+AlwaysBreakBeforeMultilineStrings: false
+BinPackArguments: false
+BinPackParameters: false
+BraceWrapping:
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   true
+  AfterStruct:     false
+  AfterUnion:      false
+  BeforeElse:      false
+  IndentBraces:    false
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Custom
+BreakBeforeTernaryOperators: false
+BreakStringLiterals: true
+ColumnLimit:     95
+ContinuationIndentWidth: 2
+DerivePointerAlignment: false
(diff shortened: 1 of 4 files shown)


ronaldtse:
@dewyatt this looks really good and very timely!

dewyatt:
@ronaldtse At some point we should do the initial clang-format run on the code base (before anyone can use the pre-commit hook). It would be a very noisy commit that would require merges for any code people are working on locally. How should we handle this?

ronaldtse:
The best way to do something drastic is to do it as early as possible. I think only @catap and @flowher are working on stuff right now. Are you guys OK for a drastic ("formatting") change? If they're OK @de

dewyatt:
@catap @flowher @randombit @zgyarmati Just FYI I did the code base reformat run.
You'll probably want to pull and merge any local stuff you have (the tabs to spaces conversion touched almost every line). Sorry for the inconvenience, but we had to do it at some point!

Let's all try to use the pre-commit hook from now on to stay consistent (Developer's Guide.md has some info).
If you have any issues using it, please let me know.
Also, if anyone uses a GUI for commits, please let me know as I can possibly make some updates for that.

PR #150, opened 2017-05-31
Fixes memory corruptions

flowher:
Some fixes found during EC development. 

Removes some memory corruptions and memory leaks in tests, so that test run can be finished successfully. This commit also uncovers memory leaks that are reported by leak sanitizer.


flowher:
argh.... my editor is automatically removing white spaces on save. I'll close this and prepare clean PR

ronaldtse:
Thanks @flowher !

Issue #33, opened 2017-03-27
Abstract out keyring read/write functionality in preparation for keybox read/write


ronaldtse:
GnuPG 2.1 now uses the `.kbx` format for key storage instead of the previous `.gpg` keyring.

In any case, the `.gpg` keyring format is not defined in RFC 4880, but is a GnuPG-proprietary format that uses an OpenPGP message to store keys together with "Trust Packet"s. Normal OpenPGP messages do not have the "Trust Packet" (Tag 12), and the "Trust Packet"s are not exported. The contents of the "Trust Packet" is also implementation-specific as defined in RFC 4880.

Given that GnuPG now uses a different `.kbx` format, it would be beneficial to have a general interface to manipulate stored keys.

We should abstract out keyring manipulation code into generic primitives such as:

* `keystore_save_key(keystore, key)`
* `keystore_load_key_by_fpr`
* `keystore_load_key_by_key_id`
* `keystore_load_key_by_user_id`
* `keystore_trust_key`

etc.

And bridge them to the specific "GnuPG keyring" or "GnuPG kbx" implementations.

ronaldtse:
@catap this issue is for the extraction of the key storage interface.

Preferably it should be FFI-friendly/accessible (#64).

ni4:
@ronaldtse @dewyatt  I think we abstracted out keyring saving/loading functionality enough?

ni4:
Closing this as completed as we have support for both gpg/kbx for a while.

PR #151, opened 2017-05-31
Abstract keyring

catap:
Made keyring as abstract (right now is SSH or GPG backend but easy to add KBX format) for #33 .


ronaldtse:
Thanks @catap , great first step for keybox. It's quite a huge change -- @dewyatt and @randombit would you have time to review this?

catap:
@ronaldtse I've tried split it to many commits to make review as easy as I can. Meanwhile a lot of code just moved from one files to another.

ronaldtse:
@catap one thing I noticed is that the name "Keyring" is OpenPGP 4880 specific.

* the term in SSH is called "Keychain" (keys in the `~/.ssh` directory)
* later on we will have "Keybox"

It would be better if we start using a prefix "`key_container_`" or "`key_store_`" or "`key_store_`" to differentiate between all these terms.

catap:
@ronaldtse, agre and I think that's using `rnp_key_store` is good.

Meanwhile, I'm waiting feedback from @randombit and @dewyatt, because one more rename will made review much more terrible ;)

ronaldtse:
Thanks @catap 👍

dewyatt:
I don't think `pgp_export_key` should live in `packet-print.c` but that's a minor thing, so let's merge.

PR #159, opened 2017-05-31
Fixes memory corruptions

flowher:
Some fixes found during EC development.

Removes some memory corruptions and memory leaks in tests, so that test run can be finished successfully. This commit also uncovers memory leaks that are reported by leak sanitizer.


ronaldtse:
I noticed the conversion of `rnp_strdup` into `malloc` + `memcpy`. Is this a change that should be propagated? There are maybe 10 other calls to `rnp_strdup`.

dewyatt:
@ronaldtse It's basically my fix from https://github.com/riboseinc/rnp/pull/120, not something that should necessarily be propagated. Generally the rule is don't treat pgp_memory_t's buffer as a NULL-terminated string.

ronaldtse:
@dewyatt I see, in this case we should probably review those calls to make sure they are acting on NULL-terminated strings instead of a `pgp_memory_t` buffer? I'll put that in a new issue.

dewyatt:
@flowher Thanks for fixing that conflict, I forgot about this PR or I would have take care of it before the re-formatting.
This fixes multiple things that ASan has been complaining about for a while, definitely eager to get this in.

flowher:
@dewyatt No worries. All good now. Thanks for committing re-formatting.

PR #161, opened 2017-05-31
Fixes memory corruption in pgp_export_key (Lost in a merge)

flowher:


flowher:
@dewyatt  thanks for merging previous commit. Nevertheless, there was a problem with merge after reformatting. So again - this fixes heap overflow caused by strlen() on not NULL terminated string

ronaldtse:
Thanks @flowher for following up!

PR #170, opened 2017-06-01
Abstract keyring

catap:
Small fix that change prefix `keyring` to `rnp_key_store`


dewyatt:
@catap Can you show the output of `clang-format --version`? It seems like yours is doing slightly different  formatting.

catap:
@dewyatt,
```
clang-format version 5.0.0 (tags/google/stable/2017-03-17)
```

It is standard clang from macOS 10.12.5

dewyatt:
@catap Ok I guess clang-format 5.0.0 is aligning block comments a bit differently from 4.0.0. Unless we all use the same version, it looks like we'll be encountering issues. I'll see if it's possible to get us all on 5.0.0, but for now, we're better off sticking to 4.0.0. Would you be able to use 4.0.0 (maybe via docker) for now?

catap:
@dewyatt, ok as temporrary solution I'm using docker-based clang-format and remove my formating commit from pull request.

dewyatt:
@catap Thanks! This is a pretty simple rename, so let's merge.

PR #174, opened 2017-06-02
Add list of officially supported platforms

ronaldtse:
Adds a list of officially supported platforms to README.

This is for #120.


ronaldtse:
@zgyarmati agree and made the separation between currently vs. upcoming supported. Thanks!

PR #175, opened 2017-06-03
Fixing remove_artifacts.sh 

MohitKumarAgniotri:
Currently, remove_artifacts.sh deletes the entire m4 folder hosting the configuration files. After modification from @dewyatt, it is required that we preserve these files and hence the script has to be adapted.


MohitKumarAgniotri:
My apologies. Need to close this pull request as this has too many unrelated changes. I will create a new one to fix the remove_artifact.sh

PR #177, opened 2017-06-03
Cleanup in RSA key generation

flowher:
Few things I've found during EC implementation, but not really related to EC
 
   * Remove DSA and Elgamal key serialization from RSA key generation
    * User doesn't need to free memory after rsa_generate_keypair fails
    * Fixes memory leaks that occure when RSA key generation fails in the middle

### DIFF
--- a/src/cmocka/rnp_tests.c
+++ b/src/cmocka/rnp_tests.c
@@ -319,6 +319,7 @@ pkcs1_rsa_test_success(void **state)
     const pgp_rsa_seckey_t *sec_rsa;
 
     pgp_key = pgp_rsa_new_key(1024, 65537, "userid", "AES-128");
+    assert_true(pgp_key != NULL);
     sec_key = pgp_get_seckey(pgp_key);
     pub_key = pgp_get_pubkey(pgp_key);
     pub_rsa = &pub_key->key.rsa;
--- a/src/common/utils.h
+++ b/src/common/utils.h
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2017, [Ribose Inc](https://www.ribose.com).
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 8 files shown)


ronaldtse:
@flowher very interesting that `rsa_key_generation` contains references to non-RSA related keys. It looks like it was trying to handle generation of different key types such as described in [GnuPG](https://www.gnupg.org/gph/en/manual/c14.html#AEN33) but was not completed:
> ```
> Please select what kind of key you want:
>   (1) DSA and ElGamal (default)
>   (2) DSA (sign only)
>   (4) ElGamal (sign and encrypt)
>  ```
> GnuPG is able to create several different types of keypairs, but a primary key must be capable of making signatures. There are therefore only three options. Option 1 actually creates two keypairs. A DSA keypair is the primary keypair usable only for making signatures. An ElGamal subordinate keypair is also created for encryption. Option 2 is similar but creates only a DSA keypair. Option 4 (Option 3 is to generate an ElGamal keypair that is not usable for making signatures) creates a single ElGamal keypair usable for both making signatures and performing encryption. In all cases it is possible to later add additional subkeys for encryption and signing. For most users the default option is fine.
 
Any idea why this is so and the best way to handle this?

flowher:
Yes. My idea is to first add --expert option (same option exist in gpg2). When option specified user will be asked for kind of key he wants to get/modulus bit length for RSA/curve type for EC/...
I have started implement it on my 'ecdh' branch already (still very much beta). This change comes with EC implementation next week.
Also I would prefer we have different function for generating different type of key. So next thing on my list is to implement proper DSA/elgamal/... key generation.
It may still sounds a bit cryptic, but please give my some time and I'll show you exactly what's on my mind in next commits

ronaldtse:
I agree the interactive mode does help some users on what they wish to generate but personally I am not a fan of interactive (or `expect`) prompts.

Also agree to separate key generation functions for different key types. Look forward to the upcoming commits!

PR #178, opened 2017-06-04
Re-factored Cmocka test case file

MohitKumarAgniotri:
@ronaldtse:  I have split the large test case file into multiple files. Also, remove_artifacts.sh is updated for the new requirements.


ronaldtse:
Obsoleted by #179

PR #185, opened 2017-06-06
use vsnprintf/vsprintf instead of vasprintf, fixes #125

### LINKED ISSUE #125: Portability: replace `vasprintf` in `obuf_printf` of `libverify.c`
As mentioned in #9, there is an instance of `vasprintf` within `libverify.c` that is of portability concerns. This issue is to replace it with standard alternatives such as `snprintf` and static allocation.

Related to #122.

zgyarmati:
This also removes src/rnpv/platform.h as it's not needed, so one less file to carry around


ronaldtse:
Looks good to me! Thanks @zgyarmati !

PR #187, opened 2017-06-07
Specify clang-format version required as v4.0.0.

dewyatt:
It looks like clang-format 4.0.0 is the newest official stable version.
At some point we may want to look at google's stable branch of clang for a newer clang-format (I think - may have to check on that).
But for now it's probably safest for us all to stick to 4.0.0.

This just clarifies things in the Developer's Guide, and adds a version check to the pre-commit hook.


ronaldtse:
Happy with this @dewyatt .

PR #188, opened 2017-06-07
ci: Add sanitize-leaks build to matrix (failure allowed for now).

dewyatt:
This just separates "sanitize" and "sanitize-leaks".
The "sanitize" run is no longer permitted to fail.
This way, we can still catch other things in the sanitize run, like heap overflows, but allow memory leaks for now.


ronaldtse:
Great move indeed for code quality. Thanks @dewyatt !

Issue #171, opened 2017-06-01
pass-fd usage is broken, tests are not communicating passphrases correctly


dewyatt:
It looks like 4 or so tests are succeeding but probably should *not* be.
Essentially, we're probably generating keys without a passphrase (and signing/decrypting, etc), even when we *think* we're using a passphrase.

While coming up with a fix, I realized there is actually a relevant fix upstream (and a few others). I'll create a separate issue to evaluate and pull in upstream changes.

Anyhow, I suspect there are multiple issues here:
* Some of the tests are calling `setupPassphrasefd` once, which writes the passphrase once, and then they loop and call multiple things that require a passphrase. So that's one issue - the passphrase has to be written to the pipe before each call that would require a passphrase.
* Secondly, calling `setupPassphrasefd`+`rnp_setvar` is only a valid thing to do *before* calling `rnp_init`. The `fdopen` call is only made in `rnp_init`, IIRC. After that, `pass-fd` doesn't matter (`rnp->passfp` does).
* I think 6b1cbb0925e5962188ebc1facfb4f963a7c2c961 and https://github.com/riboseinc/rnp/commit/594dd010f02233a4dc0a65518016f074915446d9#diff-675d74573e8c0ca275d41dd41e2a0a72 actually ended up covering up the underlying issue. (`fgets` reads up to `EOF` or `\n`, including `\n`. So sometimes we do need to truncate a character, sometimes we don't. `getpass` has code to do this internally.)
* At first glance I would also think that when `fgets()` fails in `find_passphrase`, we should be indicating an error to the caller (rather than saying "we read a passphrase of length 0"). I haven't looked at this much, it should be revisited after pulling in the upstream changes.

Related: #39

MohitKumarAgniotri:
@dewyatt I will have a look at this one.

ni4:
I fixed some testing/pass-fd issues in 2daa2c4 and ded5a07, however they are not merged to master yet.

ni4:
I'm closing this since it was fixed in #258 or some of the previous PR.

dewyatt:
This wasn't fixed, unfortunately, so reopening.

dewyatt:
Nevermind, the issues I'm seeing are of a slightly different form from this, so closing again.
I think it's just the `ask_expert_details` tests, since they don't write any passphrases.
But since I started a bit of work on https://github.com/riboseinc/rnp/issues/39 again, it may be better to handle things there.

PR #199, opened 2017-06-09
Coverity Fixes 

MohitKumarAgniotri:
Fixed a lot of coverity resource leaks.

### DIFF
--- a/src/lib/bn.c
+++ b/src/lib/bn.c
@@ -181,7 +181,7 @@ PGPV_BN_div(PGPV_BIGNUM *      dv,
             const PGPV_BIGNUM *d,
             PGPV_BN_CTX *      ctx)
 {
-    if ((dv == NULL && rem == NULL) || a == NULL || d == NULL) {
+    if ( (dv == NULL) || (rem == NULL) || (a == NULL) || (d == NULL)) {
         return 0;
     }
     USE_ARG(ctx);
--- a/src/lib/create.c
+++ b/src/lib/create.c
@@ -301,7 +301,7 @@ hash_bn(pgp_hash_t *hash, BIGNUM *bignum)
 static int
 hash_key_material(const pgp_seckey_t *key, uint8_t *result)
 {
-    pgp_hash_t hash;
+    pgp_hash_t hash = {0};
     pgp_hash_create(&hash, PGP_HASH_SHA1);
 
     switch (key->pubkey.alg) {
@@ -1147,6 +1147,7 @@ pgp_fileread_litdata(const char *filename, const pgp_litdata_enum type, pgp_outp
     mem = pgp_memory_new();
     if (!pgp_mem_readfile(mem, filename)) {
         (void) fprintf(stderr, "pgp_mem_readfile of '%s' failed\n", filename);
+        pgp_memory_free(mem);
         return 0;
     }
     len = (int) pgp_mem_len(mem);
@@ -1228,6 +1229,7 @@ pgp_write_symm_enc_data(const uint8_t *data, const int len, pgp_output_t *output
     done = (int) pgp_encrypt_se(&crypt_info, encrypted, data, (unsigned) len);
     if (done != len) {
         (void) fprintf(stderr, "pgp_write_symm_enc_data: done != len\n");
+        free(encrypted);
         return 0;
     }
 
--- a/src/lib/key_store_ssh.c
+++ b/src/lib/key_store_ssh.c
@@ -217,7 +217,15 @@ ssh2pubkey(pgp_io_t *io, const char *f, pgp_key_t *key, pgp_hash_alg_t hashtype)
         fprintf(stderr, "ssh2pubkey: can't open '%s'\n", f);
         return 0;
     }
-    stat(f, &st);
+
+    
+
+    /* check the pub file exists */
+    if (stat(f, &st) != 0) {
+        (void) fprintf(stderr, "ssh2pubkey: bad filename '%s'\n", f);
+        return 0;
+    }
+
     if ((buf = calloc(1, (size_t) st.st_size)) == NULL) {
         fprintf(stderr, "can't calloc %zu bytes for '%s'\n", (size_t) st.st_size, f);
         bufgap_close(&bg);
@@ -496,6 +504,7 @@ readsshkeys(rnp_t *rnp, char *homedir, const char *needseckey)
     } else {
         rnp_key_store_append_keyring(rnp->pubring, pubring);
     }
+
     if (needseckey) {
         rnp_setvar(rnp, "sshpubfile", filename);
         /* try to take the ".pub" off the end */
--- a/src/lib/key_store_ssh.c
+++ b/src/lib/key_store_ssh.c
@@ -218,11 +218,10 @@ ssh2pubkey(pgp_io_t *io, const char *f, pgp_key_t *key, pgp_hash_alg_t hashtype)
         return 0;
     }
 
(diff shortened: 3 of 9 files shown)


src/lib/create.c
```diff
@@ -301,7 +301,7 @@ hash_bn(pgp_hash_t *hash, BIGNUM *bignum)
 static int
 hash_key_material(const pgp_seckey_t *key, uint8_t *result)
 {
-    pgp_hash_t hash;
+    pgp_hash_t hash = {0};
```
catap: I think that `memset(&hash, 0, sizeof(pgp_hash_t)` better because this way doesn't depend on `pgp_hash_t` structure size/fields.

src/lib/misc.c
```diff
@@ -376,21 +376,18 @@ int
 pgp_fingerprint(pgp_fingerprint_t *fp, const pgp_pubkey_t *key, pgp_hash_alg_t hashtype)
 {
     pgp_memory_t *mem;
-    pgp_hash_t    hash;
+    pgp_hash_t    hash = {0};
```
catap: same idea with `memset`

src/lib/s2k.c
```diff
@@ -85,7 +85,7 @@ pgp_s2k_round_iterations(size_t iterations)
 uint8_t
 pgp_s2k_encode_iterations(size_t iterations)
 {
-    for (uint8_t c = 0; c <= 255; ++c) {
+    for (uint8_t c = 0; c < 255; ++c) {
```
catap: Are you sure? Because I found one more implementation with old logic: https://github.com/golang/crypto/blob/master/openpgp/s2k/s2k.go#L75

Meawhile, RFC 4880, Section 3.7.2.3:
> For compatibility, when an S2K specifier is used, the special value
> 254 or 255 is stored in the position where the hash algorithm octet
> would have been in the old data structure.  This is then followed
> immediately by a one-octet algorithm identifier, and then by the S2K
> specifier as encoded above.

Looks like both possible but I think that old way is compabilited with another implementation.

ronaldtse: The section @catap quoted is actually in 3.7.2.1 in RFC 4880.

The further description is:
> Therefore, preceding the secret data there will be one of these possibilities:
>        0:           secret data is unencrypted (no passphrase)
>        255 or 254:  followed by algorithm octet and S2K specifier
>        Cipher alg:  use Simple S2K algorithm using MD5 hash

More documentation here is needed anyway 👍

ronaldtse:
Thank you @MohitKumarAgniotri !

src/lib/create.c
dewyatt: @catap Can you elaborate? AFAIK this is a perfectly acceptable way of initializing a structure to zeroes in C99.

catap: @dewyatt you're right and I've checked supported distributives and all of them has compiler that works. I expect only one case when `{0}` doesn't work — some tolchain for embedded development but I think it is out of scope.

src/lib/bn.c
```diff
@@ -181,7 +181,7 @@ PGPV_BN_div(PGPV_BIGNUM *      dv,
             const PGPV_BIGNUM *d,
             PGPV_BN_CTX *      ctx)
 {
-    if ((dv == NULL && rem == NULL) || a == NULL || d == NULL) {
+    if ( (dv == NULL) || (rem == NULL) || (a == NULL) || (d == NULL)) {
```
catap: @flowher it is a very good changes because few lines below I see
`botan_mp_div(dv->mp, rem->mp, a->mp, d->mp) == 0;`

So, if dv or rem is NULL it's crashed with SEGFAULT but I haven't see any code at whole rnp where this function calls

src/lib/key_store_ssh.c
```diff
@@ -217,7 +217,15 @@ ssh2pubkey(pgp_io_t *io, const char *f, pgp_key_t *key, pgp_hash_alg_t hashtype)
         fprintf(stderr, "ssh2pubkey: can't open '%s'\n", f);
         return 0;
     }
-    stat(f, &st);
+
+    
+
+    /* check the pub file exists */
+    if (stat(f, &st) != 0) {
+        (void) fprintf(stderr, "ssh2pubkey: bad filename '%s'\n", f);
```
MohitKumarAgniotri: @flowher Missed that one. Will fix it :+1:

src/lib/s2k.c
MohitKumarAgniotri: @catap , @ronaldtse : The better fix would change the datatype of the iterator from uint8_t to uint16_t and change the boundary value to 256. 
`for (uint16_t c = 0; c < 256; ++c)`

src/lib/s2k.c
```diff
@@ -85,7 +85,7 @@ pgp_s2k_round_iterations(size_t iterations)
 uint8_t
 pgp_s2k_encode_iterations(size_t iterations)
 {
-    for (uint8_t c = 0; c <= 255; ++c) {
+    for (uint16_t c = 0; c < 256; ++c) {
```
ronaldtse: @MohitKumarAgniotri could we add some documentation referencing the RFC here?

e.g,. some part of the following
> For compatibility, when an S2K specifier is used, the special value
> 254 or 255 is stored in the position where the hash algorithm octet
> would have been in the old data structure.  This is then followed
> immediately by a one-octet algorithm identifier, and then by the S2K
> specifier as encoded above.
> 
> Therefore, preceding the secret data there will be one of these
> possibilities:
> 
>     0:           secret data is unencrypted (no passphrase)
>     255 or 254:  followed by algorithm octet and S2K specifier
>     Cipher alg:  use Simple S2K algorithm using MD5 hash
> 
> This last possibility, the cipher algorithm number with an implicit
> use of MD5 and IDEA, is provided for backward compatibility; it MAY
> be understood, but SHOULD NOT be generated, and is deprecated.

src/lib/packet-print.c
```diff
@@ -536,7 +536,7 @@ format_uid_notice(char *                 buffer,
     for (i = 0; i < key->subsigc; i++) {
         pgp_subsig_t *   subsig = &key->subsigs[i];
         const pgp_key_t *trustkey;
-        unsigned         from;
+        unsigned         from = 0u; //\todo: is this the right initial value ?
```
ronaldtse: Let's clear off the comment and just use `0`?

src/lib/s2k.c
MohitKumarAgniotri: @ronaldtse: Will do it.

MohitKumarAgniotri:
@ronaldtse: Review comments have been fixed. Thanks :)

ronaldtse:
Thank you @MohitKumarAgniotri !

ronaldtse:
Somehow the test failed here `rnp_tests_generatekey.c`.

``` c
retVal = rnp_verify_memory(
  &rnp, signatureBuf, sigLen, recoveredSig, sizeof(recoveredSig), armored);
assert_int_equal(retVal,
                 0); // Ensure signature verification fails for invalid sig
```

I've just un-done it here to keep master sane. Sorry for the noise.

dewyatt:
I would guess it's #171, `testSignature` tends to fail sporadically because we haven't really fixed up the passphrase handling in there.
Just a guess though, haven't checked it myself.

ronaldtse:
Seems like it was a broken merge from master in `src/lib/bn.c` where the `NULL` conditions were updated. Hope #209 fixes it.

Issue #109, opened 2017-05-23
RFC 4880 bis-01: Support EdDSA


ronaldtse:
We should support EdDSA, originally defined in [draft-koch-eddsa-for-openpgp-04](https://tools.ietf.org/html/draft-koch-eddsa-for-openpgp-04) and now included in [RFC 4880 bis-01](https://tools.ietf.org/html/draft-ietf-openpgp-rfc4880bis-01).

This work is to implement support for:
* Public key algorithm: 22 (EdDSA)
* Ed25519 curve

ronaldtse:
EdDSA support relies on enhancing Botan with Ed25519 support.

Existing Ed25519 implementations can be found:
  * https://github.com/floodyberry/ed25519-donna
  * https://github.com/orlp/ed25519

randombit:
Addition of Ed25519 to Botan would be great.

https://github.com/orlp/ed25519 has zlib license which is not compatible with BSD-2, no go

Floodyberry's Ed25519 is (IIRC) public domain so ok, but we wouldn't want to take all of his versions. Just use the 64-bit impl (curve25519-donna-64bit.h, modm-donna-64bit.h) and use either a native uint128_t or the `donna128` 128-bit integer type (this is how we do X25519). It leaves some performance on the table but simplifies implementation/testing/maintenence.

Another implementation (one I planned to use) was the public domain code in SUPERCOP (http://bench.cr.yp.to/supercop.html) either `crypto_sign/ed25519/ref` or `crypto_sign/ed25519/ref10`.

Any 3 of these would require some substantial modifications to be acceptable to upstream, `ref` is probably the easiest (though `ref` is much slower than `ref10` - http://bench.cr.yp.to/impl-sign/ed25519.html)

ronaldtse:
@randombit Thank you for the detailed info. SUPERCOP's `ref10` does seem more complex than `ref`...

Let's see who's free to work on this!

flowher:
Hi guys, I'm definitely planning to implement this after coming back from vacation.

ronaldtse:
This enhancement is apparently of interest to many. @randombit probably will be adding EdDSA to botan. @flowher perhaps you can work with @randombit on this?

flowher:
ok

randombit:
[...]

Issue #27, opened 2017-03-26
Support RFC 6637 (ECC for OpenPGP)


ronaldtse:
[RFC 6637](https://tools.ietf.org/html/rfc6637) defines ECC usage in OpenPGP.

Specifically, it defines:

* ECC Curves: "Curve P-256", "Curve P-384", and "Curve P-521".
* Public Key Algorithms with ID 18 (ECDH public key algorithm) and 19 (ECDSA public key algorithm)

Support of Ed25519 (EdDSA) is in a separate issue #109

flowher:
#217

flowher:
Last part  - ECDH is coming soon

flowher:
Waiting for https://github.com/randombit/botan/pull/1135 to be reviewed and merged

PR #200, opened 2017-06-09
Add Ed25519 signatures

randombit:
#109 

Currently keygen works by requesting a 255 bit key. It would be nice to clean this up by adding an algorithm identifier to `rnp_generate_key`. The underlying API `pgp_generate_keypair` does take an alg id now.

EdDSA keys are actually stored the same as ECC keys, so this helps towards #27 also.

Splits apart `crypto.h` a bit - adding `rsa.h`, `dsa.h`, `elgamal.h`, `eddsa.h`

Also extracts the function `rnp_strhexdump` out of `packet-print.c` because it's needed in `rnp.c` also now. Previously this function printed the key info then extracted the keyid at the expected offset in the string - this broke as soon as something other than RSA was used.

Tested by generating a key with `rnpkeys`, singing a file, exporting the public key to `gpg`, verifying the signature.

### DIFF
--- a/src/cmocka/rnp_tests_cipher.c
+++ b/src/cmocka/rnp_tests_cipher.c
@@ -216,6 +216,7 @@ void rnp_test_eddsa(void** state)
 
    BN_free(r);
    BN_free(s);
+   pgp_keydata_free(pgp_key);
    }
 
 void
--- a/src/cmocka/rnp_tests_cipher.c
+++ b/src/cmocka/rnp_tests_cipher.c
@@ -213,6 +213,9 @@ void rnp_test_eddsa(void** state)
 
    // cut one byte off hash -> invalid sig
    assert_int_equal(pgp_eddsa_verify_hash(r, s, hash, sizeof(hash) - 1, &pgp_key->key.seckey.pubkey.key.ecc), 0);
+
+   BN_free(r);
+   BN_free(s);
    }
 
 void
--- a/src/cmocka/rnp_tests_cipher.c
+++ b/src/cmocka/rnp_tests_cipher.c
@@ -24,6 +24,10 @@
  * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <rsa.h>
+#include <dsa.h>
+#include <eddsa.h>
+#include <elgamal.h>
 #include <crypto.h>
 #include <key_store_pgp.h>
 #include <packet.h>
@@ -138,7 +142,7 @@ pkcs1_rsa_test_success(void **state)
     const pgp_rsa_pubkey_t *pub_rsa;
     const pgp_rsa_seckey_t *sec_rsa;
 
-    pgp_key = pgp_rsa_new_key(1024, 65537, "userid", "AES-128");
+    pgp_key = pgp_generate_keypair(PGP_PKA_RSA, 1024, NULL, "SHA-256", "AES-128");
     assert_true(pgp_key != NULL);
     sec_key = pgp_get_seckey(pgp_key);
     pub_key = pgp_get_pubkey(pgp_key);
@@ -188,6 +192,28 @@ pkcs1_rsa_test_success(void **state)
     assert_int_equal(decrypted_size, 3);
     pgp_keydata_free(pgp_key);
 }
+
+void rnp_test_eddsa(void** state)
+   {
+   pgp_key_t* pgp_key = pgp_generate_keypair(PGP_PKA_EDDSA, 255, NULL, "SHA-256", "AES-128");
+   assert_non_null(pgp_key);
+
+   const uint8_t hash[32] = { 0 };
+   BIGNUM* r = BN_new();
+   BIGNUM* s = BN_new();
+
+   assert_int_equal(pgp_eddsa_sign_hash(r, s, hash, sizeof(hash),
+                                        &pgp_key->key.seckey.key.ecc,
+                                        &pgp_key->key.seckey.pubkey.key.ecc), 0);
+
+   assert_int_equal(pgp_eddsa_verify_hash(r, s, hash, sizeof(hash), &pgp_key->key.seckey.pubkey.key.ecc), 1);
+
+   // swap r/s -> invalid sig
+   assert_int_equal(pgp_eddsa_verify_hash(s, r, hash, sizeof(hash), &pgp_key->key.seckey.pubkey.key.ecc), 0);
+
+   // cut one byte off hash -> invalid sig
+   assert_int_equal(pgp_eddsa_verify_hash(r, s, hash, sizeof(hash) - 1, &pgp_key->key.seckey.pubkey.key.ecc), 0);
+   }
 
 void
 raw_elg_test_success(void **state)
--- a/src/lib/eddsa.c
+++ b/src/lib/eddsa.c
@@ -63,8 +63,8 @@ int pgp_genkey_eddsa(pgp_seckey_t* seckey, size_t curve_len)
    seckey->pubkey.key.ecc.oid = calloc(1, sizeof(ed25519_oid));
    memcpy(seckey->pubkey.key.ecc.oid, ed25519_oid, sizeof(ed25519_oid));
 
-   // Hack to insert the required 0x04 prefix on the public key
-   key_bits[31] = 0x04;
+   // Hack to insert the required 0x40 prefix on the public key
+   key_bits[31] = 0x40;
    seckey->pubkey.key.ecc.point = BN_bin2bn(key_bits + 31, 33, NULL);
 
    retval = 1;
--- a/src/lib/eddsa.c
+++ b/src/lib/eddsa.c
@@ -83,7 +83,7 @@ int pgp_eddsa_verify_hash(const BIGNUM* r,
    {
    botan_pubkey_t eddsa = NULL;
    botan_pk_op_verify_t verify_op = NULL;
-   int result = -1;
+   int result = 0;
    uint8_t bn_buf[64];
 
    // Check curve OID matches 25519
@@ -97,7 +97,7 @@ int pgp_eddsa_verify_hash(const BIGNUM* r,
    BN_bn2bin(pubkey->point, bn_buf);
 
    // See draft-koch-eddsa-for-openpgp-04 section 3 "Point Format"
-   if(bn_buf[0] != 0x04)
+   if(bn_buf[0] != 0x40)
       goto done;
 
    if (botan_pubkey_load_ed25519(&eddsa, bn_buf + 1))
--- a/src/lib/eddsa.c
+++ b/src/lib/eddsa.c
@@ -29,15 +29,50 @@
  */
 
 #include "eddsa.h"
+#include "bn.h"
+#include <string.h>
+#include <botan/ffi.h>
+#include "../common/utils.h"
 
-pgp_key_t* pgp_eddsa_new_key(const char* hash_alg,
-                             const char* cipher_alg,
-                             size_t curve_len)
+static const uint8_t ed25519_oid[9] = { 0x2b, 0x06, 0x01 ,0x04, 0x01, 0xda, 0x47, 0x0f, 0x01 };
+
+int pgp_genkey_eddsa(pgp_seckey_t* seckey, size_t curve_len)
    {
    if(curve_len != 255)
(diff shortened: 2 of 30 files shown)


src/lib/rsa.c
```diff
@@ -294,180 +294,47 @@ pgp_rsa_decrypt_pkcs1(uint8_t *               out,
...
-    }
-    return keydata;
-}
+int pgp_genkey_rsa(pgp_seckey_t* seckey, size_t numbits)
+   {
+   botan_privkey_t rsa_key = NULL;
+   botan_rng_t     rng = NULL;
+   int ret = 0;
+
+   if (botan_rng_init(&rng, NULL) != 0)
```
flowher: CHECK_BOTAN() macro is not needed anymore

src/lib/rsa.c
```diff
@@ -294,180 +294,47 @@ pgp_rsa_decrypt_pkcs1(uint8_t *               out,
...
+
+   /* Calls below never fail as calls above were OK */
+   seckey->pubkey.key.rsa.n = BN_new();
+   seckey->pubkey.key.rsa.e = BN_new();
+   seckey->key.rsa.p = BN_new();
+   seckey->key.rsa.q = BN_new();
+   seckey->key.rsa.d = BN_new();
+   seckey->key.rsa.u = BN_new();
+
+   (void) botan_privkey_rsa_get_n(seckey->pubkey.key.rsa.n->mp, rsa_key);
```
flowher: Hmmm... I think it's better to check result. I know current implementation will always return with success. But Botan is external library- if behaviour (for whatever reason) will change in the future, then rnp code will be safer with those checks.
But maybe you have better argument why not to check.

ronaldtse:
Thank you @randombit ! It seems that this solves #203 as well. The structural changes are very welcome too. Look good to me.

flowher:
Good stuff.

src/lib/packet.h
```diff
@@ -582,6 +601,12 @@ typedef struct pgp_elgamal_sig_t {
     BIGNUM *s;
 } pgp_elgamal_sig_t;
 
+/** pgp_ecc_signature_t */
+typedef struct pgp_ecc_sig_t {
+    BIGNUM *r;
+    BIGNUM *s;
+} pgp_ecc_sig_t;
```
flowher: actually, we don't need new structure. This is the same as pgp_dsa_sig_t and can be used for DSA, EdDSA and ECDSA (it's just 2 bignums in the end)

src/lib/eddsa.c
```diff
@@ -0,0 +1,182 @@
...
+
+   // Unexpected size for Ed25519 signature
+   if (BN_num_bytes(r) > 32 || BN_num_bytes(s) > 32)
+      goto done;
+
+   memset(bn_buf, 0, sizeof(bn_buf));
+   BN_bn2bin(r, bn_buf);
+   BN_bn2bin(s, bn_buf + 32);
+
+   result = (botan_pk_op_verify_finish(verify_op, bn_buf, 64) == 0) ? 1 : 0;
```
flowher: nitpicking:

result = (botan_pk_op_verify_finish(verify_op, bn_buf, 64) == 0);

src/lib/eddsa.c
```diff
@@ -0,0 +1,182 @@
...
+   // Unexpected size for Ed25519 key
+   if (BN_num_bytes(seckey->x) > 32)
+      goto done;
+
+   if (botan_rng_init(&rng, NULL) != 0)
+      goto done;
+
+   BN_bn2bin(seckey->x, bn_buf);
+
+   if (botan_privkey_load_ed25519(&eddsa, bn_buf) != 0)
```
flowher: Does it make sanse to call this method "botan_privkey_load_ec()" ? I think always be a scalar and can be used for any ECC priv key (and I can reuse it for ECDSA)

src/lib/rsa.c
```diff
@@ -294,180 +294,47 @@ pgp_rsa_decrypt_pkcs1(uint8_t *               out,
...
+   (void) botan_privkey_rsa_get_e(seckey->pubkey.key.rsa.e->mp, rsa_key);
+   (void) botan_privkey_rsa_get_d(seckey->key.rsa.d->mp, rsa_key);
+   (void) botan_privkey_rsa_get_p(seckey->key.rsa.p->mp, rsa_key);
+   (void) botan_privkey_rsa_get_q(seckey->key.rsa.q->mp, rsa_key);
+
+   if (botan_mp_mod_inverse(seckey->key.rsa.u->mp,
```
flowher: Isn't this already done in RSA_PrivateKey::check_key() which (I think) is called by botan_privkey_check_key() ?

dewyatt: No I believe he's calculating `u` from `p` and `q` here, very different from the key check.
The key check is likely a read-only operation, this is not.

src/lib/rsa.c
```diff
@@ -294,180 +294,47 @@ pgp_rsa_decrypt_pkcs1(uint8_t *               out,
...
+   if (botan_rng_init(&rng, NULL) != 0)
+      goto end;
+
+   if (botan_privkey_create_rsa(&rsa_key, rng, numbits) != 0)
+      goto end;
+
+   if (botan_privkey_check_key(rsa_key, rng, 1) != 0)
+      goto end;
+
+   /* Calls below never fail as calls above were OK */
```
flowher: this doesn't make sense anymore

src/lib/rsa.c
```diff
@@ -294,180 +294,47 @@ pgp_rsa_decrypt_pkcs1(uint8_t *               out,
...
+                            seckey->key.rsa.q->mp) != 0)
+      {
+      RNP_LOG("Error computing RSA u param");
+      goto end;
+      }
+
+   ret = 1;
+
+   end:
+   botan_privkey_destroy(rsa_key);
```
flowher: would be good to free memory allocated to seckey if (!ret) == true

src/lib/rsa.c
dewyatt: I believe it's a sanity check to ensure the key was generated with valid values. It probably does slow things down though, esp. with the thorough flag here.

src/lib/rsa.c
flowher: ahh... yes, you're right. I misunderstood it

flowher:
LGTM. just some minor comments

src/cmocka/rnp_tests_cipher.c
```diff
@@ -189,6 +193,28 @@ pkcs1_rsa_test_success(void **state)
...
+                                        &pgp_key->key.seckey.pubkey.key.ecc), 0);
+
+   assert_int_equal(pgp_eddsa_verify_hash(r, s, hash, sizeof(hash), &pgp_key->key.seckey.pubkey.key.ecc), 1);
+
+   // swap r/s -> invalid sig
+   assert_int_equal(pgp_eddsa_verify_hash(s, r, hash, sizeof(hash), &pgp_key->key.seckey.pubkey.key.ecc), 0);
+
+   // cut one byte off hash -> invalid sig
+   assert_int_equal(pgp_eddsa_verify_hash(r, s, hash, sizeof(hash) - 1, &pgp_key->key.seckey.pubkey.key.ecc), 0);
+   }
```
flowher: Memleak - free pgp_key, r,s

src/lib/eddsa.c
```diff
@@ -0,0 +1,182 @@
+/*
+ * Copyright (c) 2017, [Ribose Inc](https://www.ribose.com).
+ * All rights reserved.
+ *
+ * This code is originally derived from software contributed to
+ * The NetBSD Foundation by Alistair Crooks (agc@netbsd.org), and
```
catap: I'm not sure that this code from NetBSD. I think we should remove this line from this file, shouldn't we?

ronaldtse: Yes, this is just ours. We should remove this. Good catch @catap !

src/lib/packet.h
```diff
@@ -355,6 +355,7 @@ typedef enum {
     PGP_PKA_RESERVED_DH = 21,             /* Reserved for Diffie-Hellman
                                            * (X9.42, as defined for
                                            * IETF-S/MIME) */
+    PGP_PKA_EDDSA = 22,                   /* EdDSA */
```
catap: Can we add link at comment to https://tools.ietf.org/html/draft-koch-eddsa-for-openpgp-04#section-8 because I don't see EDSA at RFC4880 9.1 and found here.

Logical I accept it but the constant may change when `EdDSA for OpenPGP` will be approved.

Or we can move to one of `Private/Experimental algorithm` constant here.

src/lib/rnp.c
```diff
@@ -1252,11 +1250,12 @@ rnp_generate_key(rnp_t *rnp, char *id, int numbits)
     pgp_key_t *    key;
     pgp_io_t *     io;
     uint8_t *      uid;
-    char           passphrase[128];
-    char           newid[1024];
-    char           filename[MAXPATHLEN];
-    char           dir[MAXPATHLEN];
-    char *         cp;
+    char           passphrase[128] = { 0 };
```
catap: We have a constant `MAX_PASSPHRASE_LENGTH` with value 128. May we use it?

src/lib/packet.h
ronaldtse: @catap the constant of `22` is also defined here: https://tools.ietf.org/html/draft-ietf-openpgp-rfc4880bis-01#section-9.1

I agree that we can add a comment about this.

src/lib/rnp.c
```diff
@@ -1252,11 +1250,12 @@ rnp_generate_key(rnp_t *rnp, char *id, int numbits)
...
+    char           newid[1024] = { 0 };
+    char           filename[MAXPATHLEN] = { 0 };
+    char           dir[MAXPATHLEN] = { 0 };
+    char           keyid[2*PGP_KEY_ID_SIZE + 1] = { 0 };
```
catap: We have a constant `MAX_ID_LENGTH`, I think better use it here, am I wrong?

src/lib/packet.h
catap: @ronaldtse yes, but it is also a draft ;)

ronaldtse: @catap indeed. The point is the IETF OpenPGP group is not going to publish https://tools.ietf.org/html/draft-koch-eddsa-for-openpgp-04 but this will be incorporated directly into https://tools.ietf.org/html/draft-ietf-openpgp-rfc4880bis-01, so it's better if we refer to the later one. 👍

src/lib/rnp.c
randombit: No that constant is (AIUI) for user ids - this var is holding an 8-byte long hex keyid

src/lib/packet.h
```diff
@@ -400,6 +401,17 @@ typedef struct {
...
 } pgp_elgamal_pubkey_t;
 
+/** Structure to hold an ECC public key params.
+ *
+ * \see RFC 6637
+ */
+typedef struct {
+    uint8_t oid_len;
+    uint8_t *oid;
+    BIGNUM *point; /* octet string encoded as MPI */
```
randombit: Different name might make sense but I don't think it's right to have x,y as distinct pair here - rfc4880bis sec 13.2 (and also RFC 6637) specify that ECC points are stored as a single MPI which contains both x and y coordinates so I think this structure is correct for both EdDSA and ECDSA/ECDH.

src/lib/eddsa.c
randombit: Oops yes will remove thanks

src/lib/packet.h
randombit: I'll update the references but EdDSA is basically set at 22 already by virtue of GPG already shipping EdDSA support using that codepoint.

src/lib/packet.h
flowher: ok, I see your point. yes, it's possible to reuse it

src/lib/rsa.c
randombit: Changed this to check the BN_new calls

randombit:
Review comments addressed should be good to go.

flowher:
Good to go by me. Can we please merge it? I would like to commit my ecdsa dev on top of that.

randombit:
OK merging

ronaldtse:
Thank you @randombit ! This is great progress 👍

PR #201, opened 2017-06-09
Remove several unused functions

randombit:
Noticed these will doing the EdDSA work, none of these are used and don't seem worth keeping.


flowher:
Yes, please, let them go away. LGTM

ronaldtse:
Thank you @randombit . So happy to remove obsolete code.

Issue #130, opened 2017-05-26
EC key generation


flowher:
Currently frontend for key generation ``rnpkeys`` supports only RSA key generation. Goal of this ticket is to implement support for generation of ECDSA and ECDH keys. 
It is first step towards implementation support for both algorithms.

ronaldtse:
Very nice first step indeed. Welcome back @flowher !

flowher:
Following comment from #200 should be taken into account:

"
Currently keygen works by requesting a 255 bit key. It would be nice to clean this up by adding an algorithm identifier to rnp_generate_key. The underlying API pgp_generate_keypair does take an alg id now.
"

ronaldtse:
Ah @flowher I found this from your comment in #207 . Are you planning to include this in #207 or do you plan to have a separate one for exposing key generation?

flowher:
Actually #207 and this one will be separated
#207 is strictly about supporting ECDSA and this one is about changes to rnpkeys (like see the comment above)

PR #207, opened 2017-06-13
ECDSA support

flowher:
Pull request opened in order to check against CI
Before merge I would like to reuse structures introduces in EDDSA dev.

Please don't merge it just yet.


dewyatt:
Can you try to use the git pre-commit hook, or manually re-format (see Developers Guide)?

ronaldtse:
@flowher adding this to check with CI is no problem, but FYI you can also check CI building locally using instructions in the [Developers Guide](https://github.com/riboseinc/rnp/blob/master/Developers%20Guide.md).

@dewyatt , maybe we should include clang-format in CI so the build fails if there are diffs?

dewyatt:
@ronaldtse I thought about it, but we would probably have to build clang+llvm manually and cache it alongside botan et al. It seems like it would be slow if we need to rebuild (for newer botan, for example), but maybe I ought to time it and see.

flowher:
Actuallyou I have precomit installed so this code should be ok

ronaldtse:
@flowher now I know why you're using the CI this way -- the Travis container is a whopping 8.64GB. Crazy.

flowher:
@ronaldtse @dewyatt yup, a bit heavy. also - after running "git-hooks/enable.sh" again I get correct code formatting

flowher:
@dewyatt hmmm... any idea which botan branch is used for travis build? I can't reproduce in my local docker the failure I get in travis (missing function that was added to botan)

ronaldtse:
@flowher it's just botan master. You can see it in ci/install.sh.

flowher:
@ronaldtse thanks. That's what I thought. But then, why travis complains about functions that are available in botan...

randombit:
@flowher You may need to clear the Travis build cache for master branch

ronaldtse:
It's passing now @flowher !

randombit:
Thought I had left a comment here - I zapped the Travis cache and restarted the build so the new symbols in botan were available.

flowher:
thanks guys!

ronaldtse:
@randombit yes you did leave the comment, it was pushed up due to the CodeCov stuff!

@flowher question: are you also planning to expose key generation to the command line tool which is locked-in with RSA only? I remember you mentioned it some time ago. Thanks!

flowher:
@ronaldtse this will be done in separate pull request. After this PR  is merged it will be possible to generate ECDSA by specifying ``-numbits`` option (256, 384, 521 will generate ECDSA key). This isn't good solution, but allows to test manually with gpg2 (which is what I"m doing now)
Then in the next step that will be implemented in #130, I'll change this behaviour - I'll add ``--expert`` option that will ask user which algorithm/bitlen/curve he wants to use for key generation. It is partially implemented on my ``poc`` branc, so it's just matter of merge/review/improve.
Then in the next step I'll add ECDH.

ronaldtse:
Got it @flowher !

flowher:
I'll clean up this PR and open new one

PR #209, opened 2017-06-13
Retrying #199 Coverity Fixes

ronaldtse:
Retrying #199 and reverting #208 . Fixed broken merge in `src/lib/bn.c`.


ronaldtse:
Finally it's passed 👍

PR #212, opened 2017-06-13
Update json-c header paths to proper ones

ronaldtse:
Change `#include <json.h>` to `#include <json-c/json.h>` to comply with yum package needs (e.g., `/usr/lib64/json-c/json.h`.

### DIFF
--- a/Developers Guide.md
+++ b/Developers Guide.md
@@ -14,21 +14,28 @@ Sometimes tests fail in Travis CI and you will want to reproduce them locally fo
 We can use a container for this, like so:
 
 ``` sh
-$ docker run -ti --rm travisci/ci-garnet:packer-1490989530 bash -l
+./travis.sh
+# or
+# docker run -ti --rm travisci/ci-garnet:packer-1490989530 bash -l
 ```
 
-(Refer to [here](https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image) and [here](https://hub.docker.com/r/travisci/ci-garnet/tags/))
+(Refer to
+[here](https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image)
+and [here](https://hub.docker.com/r/travisci/ci-garnet/tags/))
 
 Inside the container, you will need to perform steps like the following:
 
 ``` sh
-$ git clone https://github.com/riboseinc/rnp.git
-$ cd rnp
-$ export BOTAN_INSTALL="$HOME/builds/botan-install"
-$ export CMOCKA_INSTALL="$HOME/builds/cmocka-install"
-$ export JSON_C_INSTALL="$HOME/builds/json-c-install"
-$ ci/install.sh
-$ env BUILD_MODE=normal CC=clang ci/main.sh
+cd ~/
+git clone https://github.com/riboseinc/rnp
+# or if testing local copy
+# git clone /usr/local/rnp
+cd rnp
+export BOTAN_INSTALL="$HOME/builds/botan-install"
+export CMOCKA_INSTALL="$HOME/builds/cmocka-install"
+export JSON_C_INSTALL="$HOME/builds/json-c-install"
+ci/install.sh
+env BUILD_MODE=normal CC=clang ci/main.sh
 ```
 
 (The above uses clang as the compiler -- use `CC=gcc` for GCC)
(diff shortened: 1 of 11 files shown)


dewyatt:
I don't think this is right, I believe it should in fact be `#include <json.h>`.
See the output of `pkg-config --cflags json-c`.

For CentOS 7, I had no issues building with:

```sh
yum -y install epel-release
yum -y install git autoconf automake make gcc gcc-c++ libtool bzip2-devel zlib-devel json-c-devel libcmocka-devel
cd
git clone https://github.com/randombit/botan.git
cd botan
./configure.py
make -j4 install

cd
git clone https://github.com/riboseinc/rnp.git
cd rnp
autoreconf -vfi
./configure
make -j4
```

dewyatt:
I do agree, however, that removing JSON_LIBS and JSON_CFLAGS is a step in the right direction.

ronaldtse:
@dewyatt you're right -- I got confused without looking at the json-c documentation.

I've changed back behavior and fixed the CentOS/RHEL build scripts to work. They were not previously working with the build-install.sh script. One thing to note is that our json-c usage requires 0.12, which is incompatible with json-c 0.11 from EPEL.

dewyatt:
> One thing to note is that our json-c usage requires 0.12, which is incompatible with json-c 0.11 from EPEL.

Just a small note but master builds and tests OK with libjson-c-devel 0.11 from the official base repo (not EPEL). I'm not aware of anything we're *currently* doing that requires 0.12.

Regardless, LGTM.

ronaldtse:
@dewyatt I concur and share the same experience. I was told that by the Fedora json-c package maintainer the following:
> [T]here was a reason for json-c being retired in EPEL6 and there is no build for EPEL7:  json-c is shipped in the regular distribution repos for el6 (since 6.8) and el7. Updating these to 0.12 would imply an API-incompatible change, which is not desired for a stable enterprise OS.

There are some breaking API changes described [here](https://groups.google.com/forum/#!topic/json-c/FSzSyLOmmyQ), copied below.

Since we might run into incompatibility problems in the future using `json-c`, we should just run with the later version `json-c 0.12.1`, which we've packaged in https://github.com/riboseinc/yum.

>   * Address security issues: 
>     * CVE-2013-6371: hash collision denial of service 
>     * CVE-2013-6370: buffer overflow if size_t is larger than int 
> 
>   * Avoid potential overflow in json_object_get_double 
> 
>   * Eliminate the mc_abort() function and MC_ABORT macro. 
> 
>   * Make the json_tokener_errors array local.  It has been deprecated for 
>      a while, and json_tokener_error_desc() should be used instead. 
> 
>   * change the floating point output format to %.17g so values with 
>      more than 6 digits show up in the output. 
> 
>   * Remove the old libjson.so name compatibility support.  The library is 
>       only created as libjson-c.so now and headers are only installed 
>       into the ${prefix}/json-c directory. 
> 
>   * When supported by the linker, add the -Bsymbolic-functions flag. 
> 
>   * Various changes to fix the build on MSVC. 
> 
>   * Make strict mode more strict: 
>     * number must not start with 0 
>     * no single-quote strings 
>     * no comments 
>     * trailing char not allowed 
>     * only allow lowercase literals 
> 
>   * Added a json_object_new_double_s() convenience function to allow 
>     an exact string representation of a double to be specified when 
>     creating the object and use it in json_tokener_parse_ex() so 
>     a re-serialized object more exactly matches the input. 
> 
>   * Add support NaN and Infinity

zgyarmati:
Indeed, with the current usage of the json-c api we are not depending on 12.1, so i'm ok if we drop it.
Otherwise thepatch looks OK to me.

ronaldtse:
@zgyarmati I'd prefer using the newer version of 0.12.1 due to these two fixed vulnerabilities with CVEs on `json-c`:

* CVE-2013-6371: hash collision denial of service
* CVE-2013-6370: buffer overflow if size_t is larger than int

dewyatt:
Just FYI, the CVEs are already taken care of upstream ([RHSA-2014-0703](https://rhn.redhat.com/errata/RHSA-2014-0703.html)), you can verify this in the CentOS 7 SRPM for json-c.

ronaldtse:
Nice find @dewyatt . So the good news is we can build against both `json-c` 0.11 and 0.12.

ronaldtse:
The changes are mainly on the build scripts. @dewyatt if you're OK with the changes, can you merge this?

PR #216, opened 2017-06-13
Use all processor cores when building most things.

### LINKED ISSUE #213: CI and local build scripts should use all processor cores
Currently there's a few places where we have things like `make -j2`.
Many of us have 4 or more cores so when we build locally, this is a bit slower than what we expect.

I usually use something like `make -j$(grep -c '^$' /proc/cpuinfo)` which works well as long as procfs is mounted, etc.
I think we ought to default to 2 cores and then use /proc/cpuinfo to adjust if it's available.

dewyatt:
Closes #213.

### DIFF
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -1,12 +1,14 @@
 #!/bin/bash
 set -eu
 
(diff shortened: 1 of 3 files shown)


ronaldtse:
LTGM, thanks @dewyatt !

Issue #18, opened 2017-03-23
Support non-RSA public keys


ronaldtse:
While it is technically supposed to support ElGamal, etc non-RSA algorithms, the code is now hard-coded to only use RSA.

This task is to complete support of non-RSA schemes.

flowher:
@ronaldtse Should I use this record to implement ECDSA, ECDH and EDDSA ?
I've just started implementing support for ECDSA

ronaldtse:
@flowher yes please, please go ahead here. Thanks!

PR #217, opened 2017-06-14
ECDSA support

flowher:


flowher:
Please review and merge

ronaldtse:
Thanks @flowher ! @dewyatt and @randombit would you mind reviewing this?

flowher:
There is a small missing part in this implementation. In crypto.c I'm checking if (hash_size < 32). In fact hash size depends on curve that's used. I'll try to fix it today

src/lib/crypto.c
```diff
@@ -198,109 +199,115 @@ pgp_elgamal_encrypt_mpi(const uint8_t *          encoded_m_buf,
...
+          (alg_params == 256) ?
+            PGP_CURVE_NIST_P_256 :
+            (alg_params == 384) ? PGP_CURVE_NIST_P_384 : PGP_CURVE_NIST_P_521;
+
+        pgp_hash_t   tmp;
+        const size_t hash_size =
+          pgp_hash_create(&tmp, seckey->hash_alg) ? pgp_hash_output_length(&tmp) : 0;
+
+        if (hash_size < 32) {
+            RNP_LOG("Hash output length to small (256 required minimum)");
```
flowher: That's wrong. 
Also check which hash is used when signing

src/lib/rnp.c
```diff
@@ -1240,6 +1240,8 @@ rnp_import_key(rnp_t *rnp, char *f)
     return rnp_key_store_list(io, rnp->pubring, 0);
 }
 
+#define ID_OFFSET 38
```
randombit: Is this used?

flowher: Nop. I'll remove it

flowher:
#18

randombit:
LGTM.

I think some more of the code paths for EdDSA and ECDSA can be further merged (like in reading public keys in `packet-parse.c`, and EdDSA OID handling can change to use the `find_curve_by_OID` logic). All minor stuff, I can send a patch for those later.

ronaldtse:
Thanks @randombit for the review (and in advance for the upcoming patch)! I'm going to merge this now.

PR #222, opened 2017-06-16
Add support for SM4

randombit:
SM4 support is not quite landed in Botan master yet but will shortly. This allows it to be used for bulk encryption when available.

Also renumbers Camellia ciphers to use the values from RFC 5581. I looked and Camellia was assigned to the experimental/private range in the original NetPGP import I guess because the support in NetPGP predated the RFC.


ronaldtse:
Thanks @randombit , maybe we should wait until Botan master contains SM4 to allow tests here.

But it will be useful to merge the Camilla algorithm ID changes now -- should we still support the experimental IDs for backwards compatibility though?

ronaldtse:
Just going to merge this given that new commits are need to trigger tests once botan includes SM4.

PR #229, opened 2017-06-19
Fix empty filename/modification date in the literal data packet

ni4:
This pull request also fixes crash with --list-packet command for the signed file, and introduces new rnp_ctx_t which later will be used in other cases where you need to transfer complicated parameters to the lower-layer functions.


src/lib/writer.c
```diff
@@ -1017,21 +1017,23 @@ encrypt_se_ip_writer(const uint8_t *src,
...
+    rnp_ctx_t *      ctx;
     encrypt_se_ip_t *se_ip = pgp_writer_get_arg(writer);
     pgp_output_t *   litoutput;
     pgp_output_t *   zoutput;
     pgp_output_t *   output;
     pgp_memory_t *   litmem;
     pgp_memory_t *   zmem;
     pgp_memory_t *   localmem;
-    unsigned         ret = 1;
+    unsigned         ret = 1;
```
flowher: Please clang-format before commiting

src/rnp/rnp.c
```diff
@@ -260,12 +261,19 @@ show_output(char *out, int size, const char *header)
...
-    int       cc;
+    const int  cleartext = 1;
+    unsigned   maxsize;
+    char *     out;
+    char *     in;
+    int        ret;
+    int        cc;
+    rnp_ctx_t *ctx;
+
+    ctx = rnp_cur_ctx_new();
```
flowher: Is this memory ever freed?

ni4: It is freed if another set_cur_ctx() call is made. So at max only single object will be kept alive until the application termination.

flowher: I see. Would be better to free it at some point, so that valgrind doesn't complain.

src/lib/writer.c
ni4: Actually, I run clang-format before created the PR, and don't see the space there. So it's kinda strange that GitHub is showing it.

src/rnp/rnp.c
ni4: Can you please tell me how do you run valgrind so it complains here? I tried valgrind --leak-check=full rnp --encrypt , it tells about 40 possible leaks, but no-one is about rnp_ctx

src/lib/writer.c
flowher: Yup, seems you already fixed it in some other commit. But I've noticed some additional spaces where added here and there (like signature.c:954), so maybe try running clang-format again before commiting

src/rnp/rnp.c
flowher: Memory is allocated but not freed as far as I can see. You just need to call rnp_free_ctx()/rnp_set_ctx(NULL) before leaving this funciton

dewyatt:
I think we should have a separate PR for introducing `rnp_ctx_t`, before this PR. As is, it introduces two static globals, which is probably not ideal.

src/lib/rnp.c
```diff
@@ -999,6 +1000,65 @@ rnp_end(rnp_t *rnp)
...
+int
+rnp_init_ctx(rnp_ctx_t *ctx)
+{
+    memset(&ctx, '\0', sizeof(ctx));
+    return 0;
+}
+
+/* set operation context as current */
+static rnp_ctx_t *curctx = NULL;
+static int        ownctx = false;
```
flowher: should be bool

ni4:
@dewyatt The problem is that to fully fix this issue I need the ctx - otherwise will need to pass additional two parameters to some of pgp_encrypt() calls. Other solution would be to pass ctx to pgp_encrypt calls but this would require modification of many function headers. So I sticked to globals approach. It can be used as temporary solution until we fully implement the ctx as a 'main' variable passed to the functions.

src/lib/writer.c
ni4: Strange, I believe I run it after the latest commit. Now it definitely should be fine.

src/lib/rnp.c
```diff
@@ -999,6 +1000,65 @@ rnp_end(rnp_t *rnp)
...
+
+void
+rnp_set_ctx(rnp_ctx_t *ctx)
+{
+    if (curctx && (curctx != ctx) && ownctx) {
+        rnp_free_ctx(curctx);
+        free(curctx);
```
flowher: this is call to free(NULL)

src/lib/rnp.c
```diff
@@ -999,6 +1000,65 @@ rnp_end(rnp_t *rnp)
...
+}
+
+/* free operation context */
+void
+rnp_free_ctx(rnp_ctx_t *ctx)
+{
+    if (ctx->filename != NULL)
+        free(ctx->filename);
+    if (ctx == curctx)
+        curctx = NULL;
```
flowher: Hmmm...  rnp_free_ctx(curctx) calls :

rnp_free_ctx(curctx)
free(curctx) 

As this line assigns NULL to curctx, then call to free(curctx)  will never free anything. 

Am I wrong?

flowher:
If you need temporary solution, can you use stack instead of heap? This should be a way to get rid of global variables. 
Please be aware I'm also introducing context - for some other development. Both approaches will need merigng at some point (no worries for now I would say)

src/lib/rnp.c
ni4: You are right. Thank you for pointing at this, mislooked this weird thing. Updated the PR.

ni4:
Sure, I can allocate context in stack and just save pointer to it as global static variable, will this be okay? And, where do you plan to use context? My idea is that it will be used to transfer all the possible flags/variable which would be impossible to drag through the all function calls/sdk layers.

src/lib/rnp.c
flowher: Ok. In such case  I agree with dewyatt   - global variables make this code complicated.

flowher:
Exactly. I need this to pass params for key generation (I don't consider rnp_getvar as an option). 

For example that's my context for key generation
https://github.com/flowher/rnp/blob/poc/include/rnp.h#L52

I'm just merging one PR that will use it, so you will see in comming pull request

ni4:
The only reason I did the global variable is that I need information from it in the functions where rnp_t is not transferred. And actually rnp_t seems to be suitable for upper-level (where user interacts with application, passing parameters), but context is needed on lower needed. So what do you think about adding of first parameter of type rnp_ctx_t into pgp-related function? You keygen_ctx greatly fit into this scenario as part of rnp_ctx_t.

flowher:
I think at some point we will end up with adding some ctx param to all/many API calls anyway. I think it is needed, so IMHO - go ahead.

ronaldtse:
@ni4 I think we can merge this once the remaining comments from @flowher are resolved. Thanks!

MohitKumarAgniotri:
I will review it today. Sorry for the delay guys.

ni4:
Thanks! I will add a bunch of changes in a few hours, so please don't spend time on it until I do some commits.

ni4:
Please see the updated pull request. It adds a lot of changes, most of them - to add rnp_ctx_t to inner levels, i.e. to pgp_output_t and pgp_writer_t . First one logically should become later the part of rnp_ctx_t, I didn't want to change it now since it would touch almost every method.

ni4:
@ronaldtse @flowher @dewyatt 
While I'll be fixing travis build, I would like to hear some critic/opinion regarding my breaking changes in functions interface:
1.  I get out from rnp_getvar(rnp*) inside of lib/rnp.c for the encryption, so encryption algorithm is sent via rnp_ctx_t as pgp_symm_alg_t, not as char*  - are we stick to this? Then the same should be done for the hash algorithm and other parameters. Basically, my idea is to (later) get rid of everything string/system dependent  on lib level, so rnp_getvar would be used only on CLI level. However we can leave it as it is, then I'll reverse some changes.
2. I also moved into rnp_ctx_t 'armored' parameter. From one side it is one of the basic parameters for the encryption/signing, from the other side - it is also some internal thing which can be stored in the context.

Thanks for all the input!

ronaldtse:
1. I agree `rnp_getvar` and likes should only be at the CLI level, not at lib level.
2. To me `armored` or not, is just a reader/writer thing and is not technically related to encryption/signing. Core code should not need to know whether the input/output is armored.

flowher:
I'm agree with with what's above

include/rnp.h
```diff
@@ -65,6 +80,11 @@ typedef struct rnp_t {
 int rnp_init(rnp_t *);
 int rnp_end(rnp_t *);
 
+/* init, reset and free rnp operation context */
+int  rnp_ctx_init(rnp_ctx_t *);
+void rnp_ctx_reset(rnp_ctx_t *);
+void rnp_ctx_free(rnp_ctx_t *);
+
 /* debugging, reflection and information */
 int         rnp_set_debug(const char *);
```
MohitKumarAgniotri: Missing usage of the coding guideline.

src/lib/create.c
```diff
@@ -1055,16 +1053,27 @@ pgp_write_litdata(pgp_output_t *         output,
...
-     * We will not implement them.
-     */
-    /* \todo do we need to check text data for <cr><lf> line endings ? */
+    char *   filename = NULL;
+    uint64_t modtime = 0;
+    unsigned flen = 0;
+    /* \todo do we need to check text data for <cr><lf> line endings ? - Yes, we need.
+    For non-PGP_LDT_BINARY we should convert line endings to the canonical CRLF style. */
+
+    if (output->ctx) {
```
MohitKumarAgniotri: Do all the callers ensure that the output and data are ALWAYS valid pointers?

src/lib/create.c
```diff
@@ -1055,16 +1053,27 @@ pgp_write_litdata(pgp_output_t *         output,
...
+            (void) fprintf(stderr, "pgp_write_litdata : filename %s too long\n", filename);
+            return 0;
+        }
+    }
+
     return pgp_write_ptag(output, PGP_PTAG_CT_LITDATA) &&
-           pgp_write_length(output, (unsigned) (1 + 1 + 4 + maxlen)) &&
-           pgp_write_scalar(output, (unsigned) type, 1) && pgp_write_scalar(output, 0, 1) &&
-           pgp_write_scalar(output, 0, 4) && pgp_write(output, data, (unsigned) maxlen);
+           pgp_write_length(output, (unsigned) (1 + 1 + flen + 4 + maxlen)) &&
```
MohitKumarAgniotri: Would be nicer if we can make a more readable macro for computations like (1 + 1 + flen + 4 + maxlen)

src/lib/create.c
ni4: output is the the main object which is used to write on all the underlying levels, so output is a valid pointer. ctx is part of the rnp_t and also is destroyed only when rnp_t is destroyed.

include/rnp.h
ni4: Do you mean function names, or something else?

src/lib/create.c
ni4: Here it is written in this way to easier check whether it conforms to the RFC packet structure, seems pretty clear to me.

include/rnp.h
MohitKumarAgniotri: ahk! sorry, but I looked at non-modified lines. 
I was thinking of the spacing in the function prototype.
int         rnp_set_debug(const char *);

src/lib/create.c
MohitKumarAgniotri: No worries then. Still loads of RFC to read for me.

include/rnp.h
ni4: Ah, I see. I now struggling with clang-format - version 5.0 which is automatically installed on Mac by brew produces incompatible with 4.0 output, so travis tests fail.

ni4:
@flowher I was just lucky to run into sign/verification problem of ECDSA test. This problem is reproducible locally if run test 20-30 times in a row. Can you please take a look at this?

Other then this all seems to be fixed according to comments, so waiting for your and somebody else approval.

flowher:
Thanks. I've noticed. I think I have fix already

flowher:
LGTM

randombit:
Looks ok to me on a quick pass.

ni4:
Ok, merging then. Thank you all!

PR #232, opened 2017-06-19
Randomized iv right after the generation.

ni4:
Basically, just one-line fix. Or in such cases it is safe to commit directly to master?


dewyatt:
> Basically, just one-line fix. Or in such cases it is safe to commit directly to master?

I would say that a PR is still the best move, so that the entire test matrix gets a chance to complete.

ronaldtse:
Agree with @dewyatt . I think @dewyatt and @randombit are in the best positions to approve this.

ronaldtse:
Thanks @randombit !

PR #233, opened 2017-06-19
Changed pgp_add_time to use enum instead of char* parameter.

### LINKED ISSUE #219: pgp_add_time needs some refactoring
It's a bit weird - it has char* parameter type, which can be "birth" or "expiration", and then it is mapped to `tag = (strcmp(type, "birth") == 0) ? PGP_PTAG_SS_CREATION_TIME : PGP_PTAG_SS_EXPIRATION_TIME`

I think char* parameter should be replaced to some enum or directly to the `PGP_PTAG_SS_CREATION_TIME/PGP_PTAG_SS_EXPIRATION_TIME`

ni4:
Fixes #219 , mostly for better code readability.

### DIFF
--- a/src/lib/packet-key.c
+++ b/src/lib/packet-key.c
@@ -490,8 +490,9 @@ pgp_add_selfsigned_userid(pgp_key_t *key, const uint8_t *userid)
 
     /* create sig for this pkt */
     sig = pgp_create_sig_new();
-    pgp_sig_start_key_sig(sig, &key->key.seckey.pubkey, userid, PGP_CERT_POSITIVE, key->key.seckey.hash_alg);
-    pgp_add_time(sig, (int64_t) time(NULL), "birth");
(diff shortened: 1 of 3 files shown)


flowher:
LGTM

ni4:
What policy do we have? May I merge it now, or wait for @dewyatt approval as well?

dewyatt:
@ronaldtse Do you want to formalize a PR review process (see @ni4 comment above)? For example, require 2 approvals on each PR, or something like that (you can enforce a policy via github too, of course).

ni4:
Thanks! Yeah, possibly it would be easier/more organized to have some simple formal rules, like naming of the branches, number of approves needed for the merge and so on.

ronaldtse:
Yes @dewyatt let's formalize a policy to require two approvals to merge, and the other typical rules we should follow, could you help add them to the Dev Guide?

PR #241, opened 2017-06-20
Introduces --expert flag for rnpkeys

flowher:
Introduces --expert flag for rnpkeys. This flag can be used if user want's to generate key different than default (RSA/2048). In such case rnpkeys will ask user for kind of key/size he wants to generate. Currently only RSA, ECDSA and EDDSA is supported.

 I propose we introduce structure that can be used for passing complete set of parameters needed for key generation (currently this structure is used to pass number of bits or curve name).

Patch contains also some small improvements for EC key generation (moves some function from ecdsa.c to crypto.c)

#130

### DIFF
--- a/src/cmocka/rnp_tests_cipher.c
+++ b/src/cmocka/rnp_tests_cipher.c
@@ -143,11 +143,10 @@ pkcs1_rsa_test_success(void **state)
     const pgp_rsa_pubkey_t *pub_rsa;
     const pgp_rsa_seckey_t *sec_rsa;
 
-    rnp_keygen_desc_t key_desc = {0};
-    key_desc.key_alg = PGP_PKA_RSA;
-    key_desc.hash_alg = PGP_HASH_SHA256;
-    key_desc.sym_alg = PGP_SA_AES_128;
-    key_desc.rsa.modulus_bit_len = 1024;
+    const rnp_keygen_desc_t key_desc = {.key_alg = PGP_PKA_RSA,
+                                        .hash_alg = PGP_HASH_SHA256,
+                                        .sym_alg = PGP_SA_AES_128,
+                                        .rsa = {.modulus_bit_len = 1024}};
     pgp_key = pgp_generate_keypair(&key_desc, NULL);
 
     assert_true(pgp_key != NULL);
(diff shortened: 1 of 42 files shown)


src/rnpkeys/rnpkeys.c
```diff
@@ -234,10 +242,34 @@ rnp_cmd(rnp_t *rnp, prog_t *p, char *f)
...
+        if (PGP_HASH_UNKNOWN == ctx->hash_alg) {
+            // Use default hash
+            if (ctx->key_alg == PGP_PKA_ECDSA) {
+                ctx->hash_alg = (ctx->ecc.curve == PGP_CURVE_NIST_P_256) ?
+                                  PGP_HASH_SHA256 :
+                                  (ctx->ecc.curve == PGP_CURVE_NIST_P_384) ?
+                                  PGP_HASH_SHA384 :
+                                  /*PGP_CURVE_NIST_P_521*/ PGP_HASH_SHA512;
+            } else {
+                // TODO: Shouldn't it be PGP_DEFAULT_HASH_ALGORITHM ?
```
ronaldtse: `PGP_DEFAULT_HASH_ALGORITHM` is now specified as SHA256 in code, but in RFC 4880 (and bis01) "SHA-1" is still the only mandatory implementation? Can we confirm this?

ronaldtse:
Thank you @flowher for the very welcome improvement! We probably need a corresponding README update on how to use this.

flowher:
@ronaldtse good point! I'll do that

src/cmocka/rnp_tests_cipher.c
```diff
@@ -316,8 +327,13 @@ ECDSA_signverify_success(void **state)
     uint8_t       message[32];
     pgp_ecc_sig_t sig = {NULL, NULL};
 
-    pgp_key_t *pgp_key1 = pgp_generate_keypair(PGP_PKA_ECDSA, 256, NULL, "SHA256", "AES-128");
-    pgp_key_t *pgp_key2 = pgp_generate_keypair(PGP_PKA_ECDSA, 256, NULL, "SHA256", "AES-128");
+    rnp_keygen_desc_t key_desc = {0};
+    key_desc.key_alg = PGP_PKA_ECDSA;
+    key_desc.hash_alg = PGP_HASH_SHA256;
+    key_desc.sym_alg = PGP_SA_AES_128;
```
randombit: Shouldn't this be setting some kind of indicator of curve size?

randombit:
LGTM

src/rnpkeys/rnpkeys.c
ni4: Sure, they should be configurable as well. However users and software developers prefer that as most as possible things are solved for them, so good and secure defaults is a must have thing.

flowher:
I'll open new PR as this doesn't look good after merge

PR #243, opened 2017-06-21
Remove the unused BN_CTX type

randombit:
Was a stub to allow OpenSSL compat, but not really used anymore.

It's kind of telling that almost no code changes were required to handle the removal of the BN_CTX args - it looks like (now that ElGamal and co are not implemented via BN) the only BN functions actually in use are `BN_bn2bin`, `BN_bin2bn`, `BN_num_bytes`, `BN_num_bits`, and maybe one or two others. So maybe there is more pruning that can occur here.

### DIFF
--- a/src/cmocka/rnp_tests_cipher.c
+++ b/src/cmocka/rnp_tests_cipher.c
@@ -247,7 +247,6 @@ raw_elg_test_success(void **state)
     uint8_t              g_to_k[64];
     uint8_t              decryption_result[1024];
     const uint8_t        plaintext[] = {0x01, 0x02, 0x03, 0x04, 0x17};
-    BN_CTX               ctx;
 
     // Allocate needed memory
     pub_elg.p = BN_bin2bn(p512, sizeof(p512), NULL);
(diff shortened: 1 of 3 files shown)


randombit:
Huh, clang-format check fails in CI but running `clang-format 4.0.0` locally does not reproduce the requested style changes, not sure what's happening.

ni4:
Yesterday I had problems with clang-format as well. Latest travis docker container has only clang-format-3.9, and when I used 3.9 locally everything went fine.

dewyatt:
@randombit Are you using the pre-commit hook? If not, what was the full command line you used to invoke clang-format?

@ni4 If you use the pre-commit hook, an appropriate container will be created for you. Otherwise, you can easily create one that has clang-format 4.0.0 with:

```
docker run --name=clang-format alpine:latest apk --no-cache add clang
docker commit clang-format clang-format
docker rm clang-format
```

randombit:
@dewyatt This was running it just from the command line (just enabled the pre-commit hook). I forgot the `-style=file` option, adding that seems to do the right thing.

I was (am) confused because even without the command line option, I could see (by running `clang-format` under `strace`) that it was opening the `.clang-format` file, so I assumed everything was working. *shrug*

dewyatt:
@randombit I figured that might be the case, super easy mistake to make (kind of a bad design there). Weird indeed that it still opens the file. `¯\_(ツ)_/¯`

ronaldtse:
Looks good and always happy to get rid of unnecessary code 👍

PR #244, opened 2017-06-21
Introduces --expert flag for rnpkeys (ex #241) 

flowher:
Introduces --expert flag for rnpkeys. This flag can be used if user want's to generate key different than default (RSA/2048). In such case rnpkeys will ask user for kind of key/size he wants to generate. Currently only RSA, ECDSA and EDDSA is supported.

I propose we introduce structure that can be used for passing complete set of parameters needed for key generation (currently this structure is used to pass number of bits or curve name).

Patch contains also some small improvements for EC key generation (moves some function from ecdsa.c to crypto.c)

#130

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Upcoming supported platforms:
 
 ## Generating an RSA Private Key
 
-Only RSA key supported right now.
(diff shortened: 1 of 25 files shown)


flowher:
@ronaldtse I've update README.md. Let me know if OK with you.

PR #246, opened 2017-06-23
docs: Update dev guide with info on PR approvals and branch use

dewyatt:


ronaldtse:
The updates look good and descriptive. Thank you @dewyatt !

Issue #249, opened 2017-06-25
ECDSA and EDDSA signature verification fails 


flowher:
## Description

Wrong representation of 'r' and 's' components of ECDSA/EDDSA  as octet-string causes failures during signature verification. Verification may fail even if signature is completely correct.

## Steps to Reproduce

Assuming uniform distribution of numbers produced by RNG, the only way to 
1. Ensure RNG is initialized once and ecdsa/eddsa uses initialized RNG
2. Run either ECDSA_signverify_success or rnp_test_eddsa tests in a never ending loop

## Expected Behavior
Test must not fail

## Actual Behavior
Test will fail when high-order bytes of generated number are 0

ni4:
If I correctly understand, problem happens when high-order byte of one of the (r, s) is zero. This should happen at least in 1 in 256 runs, if think about (r, s) pair as a semi-random thing, isn't it?

flowher:
It happens if high-order byte is zero.
r and s are not uniformly random, but they depend on value chosen at random. but my explanation above about number of trials isn't actually correct (as there is many such numbers)

PR #250, opened 2017-06-25
Fixes a bug in ECDSA and EDDSA signature verification  

flowher:
I needed to reformat eddsa.c before commiting a fix, so in order to see what was really changed in EdDSA see commit, 9036fb0

#249


ronaldtse:
Good catch @flowher , a very tricky problem to debug 👍

PR #253, opened 2017-06-27
CID: 1447250 and CID 1447251

flowher:
Fixes coverity issues introduced in recent commits
CID 1447250
CID 1447251


ronaldtse:
Thanks @flowher !

PR #255, opened 2017-06-28
Memory fuzzying

catap:
First step to run rnp with retrace's memory fuzzing.

### DIFF
--- a/src/cmocka/rnp_tests.c
+++ b/src/cmocka/rnp_tests.c
@@ -27,23 +27,43 @@
 #include <setjmp.h>
 #include <stdio.h>
 #include <cmocka.h>
-#include <rnp_tests.h>
-#include <rnp_tests_support.h>
+
+#include "rnp_tests.h"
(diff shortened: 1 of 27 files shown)


ronaldtse:
@catap this is very nice!

catap:
@ronaldtse I feel that retrace should have a way to switch off retracing and enable only fuzzy, because it produce a lot of garbages.

and I think the right way is create a issue for this :)

ronaldtse:
Agree @catap !

ronaldtse:
@catap these are really good fixes, could we make the build pass so we can merge this soon? Thanks!

catap:
@ronaldtse strange, the 'style check' had failed, but I restarted a job... and it passed.

So, I think we can merge it :)

ni4:
Yeah, please merge so I can apply those changes and finally make PR with refactoring.

ronaldtse:
Thank you guys!!

PR #257, opened 2017-06-30
Fixes bug introduced in commit 8c3031d

flowher:
``rnpkeys`` uses wrong curve size for generating ECDSA keys

### DIFF
--- a/src/rnpkeys/tui.c
+++ b/src/rnpkeys/tui.c
@@ -94,7 +94,7 @@ ask_curve()
     } while (!ok);
 
     if (ok) {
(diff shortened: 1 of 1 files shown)


flowher:
I'm not really sure why codecov/project fails. This commit has a change in 1 line.

ronaldtse:
@flowher could you help add a test that guards against this case? Thanks!

flowher:
OK

ronaldtse:
Merged this first and please use #262 to add the test. Thanks @flowher !

Issue #153, opened 2017-05-31
Come up with consistent return values


dewyatt:
The code base seems to have a mixture of functions returning different values indicating success, failure, other conditions.
I think we should come up with a standard and stick to it.
I think we may benefit from some macros like `RNP_OK(call_some_func())`/`RNP_OK(retval)` as well.
I'd like to dig into this some more. Botan has a convention for the FFI interface that might be a good fit.

ronaldtse:
Nice one @dewyatt ! The macro approach does look good.

dewyatt:
@flowher Indeed, it's partially a dupe.
I think you mean `pgp_errcode_t`, which does provide a decent set of errors that are used in a lot of the modules.
However, a common convention is for errors to be <0 when using int return values.
Some functions return buffer sizes or have other positive return values that indicate different successful conditions, for example, so a positive error value can't be used with those.
So we may want to rework that a bit.

Those error codes are currently mostly used with the pgp_error_t linked list, so you can read multiple parse errors, etc.
I think this issue is more about the ultimate return values of functions, but it would probably make sense to utilize pgp_errcode_t (though I think negative error codes would make more sense here).

flowher:
hmm... I'm not sure why returning `<0` as an error code is better than returning `>0`. Buffer size is actually `size_t`, it's wise not to cast it to `int`, so I think it is good example why not to do it.
We can always return values by pointer, so that's not a problem. If we assume that all our functions alwyas return error code and results are always returned by pointer, then it makes quite unified API. Also caller is less likely to make a mistake - when returning int, caller must first check if returned value is <>0 and take different action depending on returned value.

dewyatt:
@flowher It's a very common convention, see the Linux kernel, botan FFI API, etc.
On size_t, you're right, and some of the existing functions do handle this incorrectly.

flowher:
I know posix style is common practice (but not necessarily good). Whatever type we return (signed or unsigned), we should be should be using `pgp_errcode_t` (you can always return `-pgp_errcode_t`).

dewyatt:
Agreed that we don't want to duplicate error codes.

ni4:
[...]

PR #258, opened 2017-06-30
Major rnp refactoring: separated CLI and library layers

ni4:
In this PR I did the following:
- removed rnp_getvar/rnp_setvar and corresponding fields from the rnp_t. Now rnp_t is responsible for keyrings, keyring pathes,d default key and some other options
- rnp_ctx_t now has pointer to rnp_t and is responsible for per-operation options
- new type rnp_cfg_t exists only on CLI level and responsible for storage of user-specified options and converting/passing them to the rnp_t structure
- new type rnp_params_t specifies rnp initialization options

So, library level now isolated from the user specified options.
Unfortunately this added some complications, like rnp initialization now has more parameters/requires more calls, and separate rnp_ctx_t now should be setup before each operation.

Also all tests were changed to work correctly with the new interfaces.

@dewyatt @catap @ronaldtse  @flowher Since these changes are major the discussion is welcome - probably, I missed something/something can be made better/etc. 

Also I did not update readme/mans yet - probably we'll need to change some interfaces after the discussion, so I'll update readme's right before merging this PR.


src/common/constants.h
```diff
@@ -36,6 +36,14 @@
  */
 
 #define SUBDIRECTORY_GNUPG ".rnp"
```
ronaldtse: Now that there is `SUBDIRECTORY_RNP` should this be ".gnupg" then?

ronaldtse:
@ni4 thank you for this! This is a very good attempt that is sorely needed. I imagine that we will need to further separate the CLI / library usage and provide good man pages for both. Next we will want to combine the executables or maybe move them to a separate subdirectory to fully separate from the lib. It's quite confusing to have both the CLI / library in the same directory.

ronaldtse:
@dewyatt @flowher any comments here?

src/common/constants.h
ronaldtse: @ni4 What do you think about this?

flowher:
@ronaldtse I need a bit more time here. I'll finish review this week

src/common/constants.h
ni4: @ronaldtse Agree on this, will update the PR.

ni4:
@ronaldtse Thanks. Yeah, after this PR is merged I'll start working on CLI, exactly on things you described - utility should be one, separated, and using the library in the way that any other application should use it.

ni4:
I merged with the latest master and added some more refactoring by replacing (where I found it/where it is applicable)  return 0/return 1 with RNP_FAIL/RNP_OK.

@ronaldtse @dewyatt What about updating the developers guide and then codebase with some rules regarding the function output types/constatns? I.e. whether it is success/error - then return rnp_err_t, if it is length - rnp_size_t, and so on?

@dewyatt @catap @flowher Would greatly appreciate one more review so I can merge this so nobody will spend a lot of time on merging further updates from the old master.
Thanks!

catap:
@ni4 I vote for `RNP_FAIL`/`RNP_OK` because I feel that few places have different logic.

About review — I will do it tomorrow at end of my day (UTC+0400)

flowher:
I've to say I don't see a point in using ``RNP_FAIL``/``RNP_OK``. It's  not different than using ``true``/ ``false`` (it even has same value). We should use values from  ``pgp_error_t`` enum.

ni4:
@flowher The reason of this was to put something easily distinguishable from 0/1 in functions where result is 'success'/'error'. So when later on we'll agree on some other return type it would be easy as search and replace.

Also pgp_errcode_t has some reverse logic - 0 for success/1 for error, so moving right now to this return type could break a lot of things, even if edit code extra carefully.

flowher:
@ni4 ah, so that's just to make it easier to grep the code. yep, I agree fixing the return code issue will be a bit of a mess. but definitely must be done at some point

src/rnp/rnpcfg.c
```diff
@@ -0,0 +1,502 @@
...
+    char **temp;
+
+    if (cfg->size == 0) {
+        /* only get here first time around */
+        cfg->keys = calloc(sizeof(char *), newsize);
+        cfg->vals = calloc(sizeof(char *), newsize);
+
+        if ((cfg->keys == NULL) || (cfg->keys == NULL)) {
+            (void) fprintf(stderr, "rnp_cfg_resize: bad alloc\n");
+            return RNP_FAIL;
```
flowher: 1. bug
you are checking cfg->keys (twice)

but let say you have fixed it
2. memory leak
if cfg->keys != NULL and cfg->vals == NULL
you should always free before returning - it cost nothing

src/rnp/rnpcfg.c
```diff
@@ -0,0 +1,502 @@
...
+    unsigned i;
+
+    for (i = 0; i < cfg->count && strcmp(cfg->keys[i], key) != 0; i++)
+        ;
+    return (i == cfg->count) ? -1 : (int) i;
+}
+
+/* resize keys/vals arrays to the new size. Only expanding is supported */
+static int
+rnp_cfg_resize(rnp_cfg_t *cfg, unsigned newsize)
```
flowher: again, Please provide proper comments - for example - what should I do if this function fails? is memory always freed() ? it seems sometimes it is sometimes not

src/rnp/rnpcfg.c
```diff
@@ -0,0 +1,502 @@
...
+            return RNP_FAIL;
+        }
+    }
+
+    if ((i = rnp_cfg_find(cfg, key)) < 0) {
+        /* add the element to the array */
+        if (rnp_cfg_resize(cfg, cfg->size + 15)) {
+            newkey = rnp_strdup(key);
+            if (newkey == NULL) {
+                (void) fprintf(stderr, "rnp_cfg_set: bad alloc\n");
```
flowher: memory leak: newval is not freed

src/rnp/rnpcfg.c
```diff
@@ -0,0 +1,502 @@
...
+        if (newval == NULL) {
+            (void) fprintf(stderr, "rnp_cfg_set: bad alloc\n");
+            return RNP_FAIL;
+        }
+    }
+
+    if ((i = rnp_cfg_find(cfg, key)) < 0) {
+        /* add the element to the array */
+        if (rnp_cfg_resize(cfg, cfg->size + 15)) {
+            newkey = rnp_strdup(key);
```
flowher: what happens when key is not null terminated?

flowher:
@ni4 I had few comments. I'm not sure if all of them apply to new code or code that was moved (diff is big and it's hard to check). I havn't found anything major but there are some places you should take a look.

src/rnp/rnpcfg.h
```diff
@@ -0,0 +1,91 @@
...
+/* rnp CLI config : contains all the system-dependent and specified by the user configuration
+ * options */
+typedef struct rnp_cfg_t {
+    unsigned count; /* number of elements used */
+    unsigned size;  /* allocated number of elements in the array */
+    char **  keys;  /* key names */
+    char **  vals;  /* values */
+} rnp_cfg_t;
+
+int rnp_cfg_init(rnp_cfg_t *cfg);
```
flowher: please use pgp_error_t for all those new functions

src/rnp/rnpcfg.c
```diff
@@ -0,0 +1,502 @@
...
+    rnp_cfg_setint(cfg, CFG_OVERWRITE, 1);
+    rnp_cfg_set(cfg, CFG_OUTFILE, NULL);
+    rnp_cfg_set(cfg, CFG_HASH, DEFAULT_HASH_ALG);
+    rnp_cfg_set(cfg, CFG_CIPHER, "cast5");
+    rnp_cfg_setint(cfg, CFG_MAXALLOC, 4194304);
+    rnp_cfg_set(cfg, CFG_SUBDIRGPG, SUBDIRECTORY_RNP);
+    rnp_cfg_set(cfg, CFG_SUBDIRSSH, SUBDIRECTORY_SSH);
+    rnp_cfg_setint(cfg, CFG_NUMTRIES, MAX_PASSPHRASE_ATTEMPTS);
+
+    return RNP_OK;
```
flowher: please remove

src/rnp/rnpcfg.c
```diff
@@ -0,0 +1,502 @@
...
+
+    /* detecting keystore pathes and format */
+    if (!rnp_cfg_get_ks_info(cfg, params))
+        return RNP_FAIL;
+
+    /* default key/userid */
+    if (!rnp_cfg_get_defkey(cfg, params))
+        return RNP_FAIL;
+
+    return RNP_OK;
```
flowher: return true/false or    PGP_E_OK + more descriptive error codes in case of error

src/rnp/rnpcfg.c
```diff
@@ -0,0 +1,502 @@
...
+
+/* find the value name in the rnp_cfg */
+static int
+rnp_cfg_find(rnp_cfg_t *cfg, const char *key)
+{
+    unsigned i;
+
+    for (i = 0; i < cfg->count && strcmp(cfg->keys[i], key) != 0; i++)
+        ;
+    return (i == cfg->count) ? -1 : (int) i;
```
flowher: return bool

src/rnp/rnpcfg.c
```diff
@@ -0,0 +1,502 @@
...
+        /* If the path doesn't exist then fail. */
+        fprintf(stderr, "rnp: warning homedir \"%s\" not found\n", homedir);
+        return RNP_FAIL;
+    } else if (ret != 0) {
+        /* If any other occurred then fail. */
+        fprintf(stderr, "rnp: an unspecified error occurred\n");
+        return RNP_FAIL;
+    }
+
+    return RNP_OK;
```
flowher: PGP_E_OK + more descriptive error codes

src/rnp/rnpcfg.c
```diff
@@ -0,0 +1,502 @@
...
+rnp_cfg_unset(rnp_cfg_t *cfg, const char *key)
+{
+    int i;
+
+    if ((i = rnp_cfg_find(cfg, key)) >= 0) {
+        free(cfg->vals[i]);
+        cfg->vals[i] = NULL;
+        return RNP_OK;
+    }
+    return RNP_FAIL;
```
flowher: do we have to return anything from this function?

src/rnp/rnpcfg.c
```diff
@@ -0,0 +1,502 @@
...
+  res should point to the allocated buffer.
+*/
+int
+rnp_path_compose(const char *dir, const char *subdir, const char *filename, char *res)
+{
+    int pos;
+
+    /* checking input parameters for conrrectness */
+    if (!dir || !filename || !res) {
+        return RNP_FAIL;
```
flowher: this is not needed. if any of those is NULL, program should crash and developer gets perfect core dump to analyse. function should not return anything.
additionally - no single call to this function checks return value.

src/rnp/rnpcfg.c
```diff
@@ -0,0 +1,502 @@
...
+            if (id[0] != 0x0) {
+                params->defkey = strdup(id);
+                rnp_cfg_set(cfg, CFG_USERID, id);
+            }
+        }
+    } else {
+        params->defkey = strdup(userid);
+    }
+
+    return RNP_OK;
```
flowher: please remove

include/rnp.h
```diff
@@ -33,6 +33,7 @@
 
 #include <stddef.h>
 #include "packet.h"
+#include "../common/constants.h"
```
ni4: Then it doesn't build.

include/rnp.h
```diff
@@ -48,85 +49,101 @@ __BEGIN_DECLS
...
-    enum key_store_format_t key_store_format; /* keyring format */
-    union {
-        rnp_keygen_desc_t generate_key_ctx;
-    } action;
-} rnp_t;
-
-/* begin and end */
-int rnp_init(rnp_t *);
+/* initialize rnp using the init structure  */
+int rnp_init(rnp_t *, rnp_params_t *);
```
ni4: 1. It has the same return type as before, the error code, as this is done for most function. Imho it is much better to change return types all at once then to have some functions returning int, and some - returning pgp_error_t, with different meaning for 0 and 1.

2. See 1.
3. Yeah, it should, thanks.

src/cmocka/rnp_tests_support.c
```diff
@@ -56,6 +56,14 @@ file_exists(const char *path)
     return stat(path, &st) == 0 && S_ISREG(st.st_mode);
 }
 
+/* Check if a directory exists */
+int
```
ni4: We don't use bool anywhere, that's why I'm using int here, while it returns only 'true'/'false'

src/cmocka/rnp_tests_support.c
```diff
@@ -241,50 +249,84 @@ setupPassphrasefd(int *pipefd)
...
+        return res;
     }
+    rnp_params_free(&params);
 
-    return rnp_init(rnp);
+    return RNP_OK;
+}
+
+void
+set_default_rsa_key_desc(rnp_keygen_desc_t *key_desc, pgp_hash_alg_t hashalg)
```
ni4: The same code was used in rnp_tests_exportkey and rnp_tests_generatekey, so I moved it here to reuse the code. Also 1024 bits were hardcoded, so I reused them.

src/lib/misc.c
```diff
@@ -793,7 +794,7 @@ pgp_mem_readfile(pgp_memory_t *mem, const char *f)
         mem->mmapped = 1;
     }
     (void) fclose(fp);
-    return (mem->allocated == mem->length);
+    return (mem->allocated == mem->length) ? RNP_OK : RNP_FAIL;
```
ni4: Logically this function returns error code (0 for failure, 1 for success, probably later other int for more specific error code), while (mem->allocated == mem->length) logically is bool, not int.

src/lib/misc.c
```diff
@@ -1139,4 +1140,76 @@ rnp_filename(const char *path)
     } else {
         return res + 1;
     }
-}
\ No newline at end of file
+}
+
+/* find the time - in a specific %Y-%m-%d format - using a regexp */
+int
+grabdate(const char *s, int64_t *t)
```
ni4: This function was just moved 'as it is' from the write.c or something similar.
I'm chasing 1 to RNP_OK, and 0 to RNP_FAIL, and logically those return values are ints. While here it is bool. But we don't use bool anywhere so it is int.

src/lib/misc.c
```diff
@@ -1139,4 +1140,76 @@ rnp_filename(const char *path)
...
+        tm.tm_mday = (int) strtol(&s[(int) matches[3].rm_so], NULL, 10);
+        *t = mktime(&tm);
+        return 1;
+    }
+    return 0;
+}
+
+/* get expiration in seconds */
+uint64_t
+get_duration(const char *s)
```
ni4: Don't know. As stated before - these 3 functions were just moved to the more suitable place.

src/lib/misc.c
```diff
@@ -1139,4 +1140,76 @@ rnp_filename(const char *path)
...
+{
+    uint64_t now;
+    int64_t  t;
+    char *   mult;
+
+    if (s == NULL) {
+        return 0;
+    }
+    now = (uint64_t) strtoull(s, NULL, 10);
+    if ((mult = strchr("hdwmy", s[strlen(s) - 1])) != NULL) {
```
ni4: Thanks, will add an issue to check these functions

src/rnp/rnpcfg.h
ni4: I didn't use pgp_error_t because it has reverse logic - 0 for success, 1 for failure...

src/rnp/rnpcfg.c
ni4: This cannot return bool since returns index of the key.

src/rnp/rnpcfg.c
ni4: Return result can be used as indication whether key existed, maybe somewhere we'll make use of it.

ni4:
@flowher Thanks for you comments. To sum up, mostly those are about 'what to return'. I tried to follow as much as possible logic of the similar functions, so I return int and RNP_OK/RNP_FAIL where (even theoretically) we need to return error code. And since we do not use bool anywhere (but probably we should?) I used int also for functions with the true/false result.

include/rnp.h
flowher: yes, of course. the thing is it must build. you need to change include. that's a main include file, you can't expect anybody it will have a ../common/constant.h file in it's directory tree

src/cmocka/rnp_tests_support.c
flowher: yes, we use. it's c11

src/lib/misc.c
flowher: it was static  before and local to single file. please move it back then

src/rnp/rnpcfg.c
flowher: Ok. good point

flowher:
@ni4 we use bool a lot. please include <stdbool.h> and go ahead with true / false

ni4:
Hm, I searched for true/false in all sources in rnp and found only in few places in libverify and crypto.c. Don't think that I have something against true/false, it's much better then 0/1, but I just sticked to other code in rnp :)

src/lib/misc.c
ni4: It is actually used only to parse command-line parameters, so logically it should be moved even further, to the rnp.c

include/rnp.h
ni4: I used this include as it was done previously. But I agree - we should consolidate usage of headers, but in some other PR - this one is too huge already.

ni4:
Despite the strange codecov report (maybe, because of changing return 0/return 1 to defines?) I will merge this PR since it's too heave already. It would be easier to rise issues/fix them on already merged code.

src/lib/misc.c
flowher: yep, this function must return bool. not int (not RNP_* )

src/rnp/rnpcfg.c
flowher: Actually, whole function must be removed as it's never used

src/rnp/rnpcfg.h
flowher: it doesn't matter as it's new function.

src/lib/compress.c
```diff
@@ -75,6 +75,7 @@ __RCSID("$NetBSD: compress.c,v 1.23 2012/03/05 02:20:18 christos Exp $");
 #include <string.h>
 #include <stdlib.h>
 
+#include "../common/constants.h"
```
catap: Looks like you should fix it on build-system level.

src/lib/key_store.c
```diff
@@ -142,26 +103,16 @@ rnp_key_store_load_keys(rnp_t *rnp, char *homedir)
...
     if (((rnp_key_store_t *) rnp->pubring)->keyc < 1) {
         fprintf(io->errs, "pub keyring is empty\n");
-        return 0;
-    }
-
-    /* If a userid has been given, we'll use it. */
-    if ((userid = rnp_getvar(rnp, "userid")) == NULL) {
-        /* also search in config file for default id */
-        memset(id, 0, sizeof(id));
-        conffile(rnp, homedir, id, sizeof(id));
```
catap: Why you remove this logic? I think if user has more than one userid inside keychain they may would like to switch default to second one for example.

Am I wrong?

src/common/constants.h
```diff
@@ -35,7 +35,21 @@
  *       developers' .gnupg installations.
  */
 
-#define SUBDIRECTORY_GNUPG ".rnp"
+#define SUBDIRECTORY_GNUPG ".gnupg"
```
catap: and you just switch all our users from `~/.rnp` to `~/.gnupg`.

src/lib/Makefile.am
```diff
@@ -20,7 +20,6 @@ librnp_la_SOURCES	= \
 	key_store.c \
 	key_store_pgp.c \
 	key_store_kbx.c \
-	key_store_ssh.c \
```
catap: Why you just remove it? Or I missed something?

catap:
@ni4 why have you merge it?

I think it's bad idea because I see at master right for example `include "../common/constants.h"` [here](https://github.com/riboseinc/rnp/blob/master/include/rnp.h#L36)

and this is huge PR and I think it should be reviewed more of one person, and review this huge changes request a lot of time.

I did small review but I missed a lot of things.

cc: @ronaldtse @dewyatt 

Also, may we extend our developers guidlines to prevent rush and forbiden merge PR by author?

flowher:
I thought we had a rule - at least 2 approvals  before merge.

catap:
@flowher yes, we had. But I think it rules should be ignored for huge PR.

flowher:
I would say for huge PRs we need at least 2 approvals. Here I see only 1

src/lib/compress.c
ronaldtse: Agree with @catap . @ni4 could we fix this as a follow up?

src/rnp/rnpcfg.c
ronaldtse: @flowher you mean we should guard this function with descriptive errors instead of just crashing, right?

src/rnp/rnpcfg.c
ronaldtse: Agree with @flowher to have descriptive errors here.

src/rnp/rnpcfg.c
ronaldtse: @ni4 can you help fix this? Thanks.

src/rnp/rnpcfg.c
ronaldtse: @ni4 can you help fix this? Thanks.

src/rnp/rnpcfg.c
ronaldtse: @ni4 can you help fix this? Thanks.

src/rnp/rnp.c
```diff
@@ -252,253 +237,278 @@ show_output(char *out, int size, const char *header)
...
-    int       ret;
-    int       cc;
+    unsigned    maxsize;
+    char *      out = NULL;
+    char *      in = NULL;
+    const char *outf;
+    const char *userid;
+    int         ret;
+    int         cc;
+    int         clearsign = (cmd == CMD_CLEARSIGN) ? 1 : 0;
```
ronaldtse: @ni4 can you help fix this? Thanks.

src/lib/rnpsdk.h
```diff
@@ -77,4 +78,8 @@ int64_t rnp_filemtime(const char *path);
 
 const char *rnp_filename(const char *path);
 
+int grabdate(const char *s, int64_t *t);
```
ronaldtse: @ni4 could you add some comments here? Thanks.

src/lib/Makefile.am
ronaldtse: I think @catap means, while we are not planning to generate SSH keys, we still need to access the SSH keychain to allow using PGP keys for SSH purposes, so we need the SSH keystore code?

src/lib/key_store.c
ronaldtse: I think @catap is correct here?

ronaldtse:
I agree there's a lack of approved reviews here and that next time we should have at least 2.

I'd like to propose the following actions given this merge:
* @catap @dewyatt feel free to suggest changes to the Developers Guide.
* agree on the default PGP key storage location (whether it is `.gnupg`, `.rnp` or a common, cross-tool `.openpgp`), and how to access them
* handle the remaining comments in this PR
* come up with a comprehensive solution regarding descriptive error codes and error output, such as using a struct to carry and output them.

src/rnp/rnpcfg.c
flowher: @ronaldtse I mean - if this is function that can be called by external library then we should return descriptive error, otherwise we should crash as it's implementation error.

ronaldtse: Agree. Instead of "crashing" though it might be better to exit gracefully still.

src/rnp/rnpcfg.c
ronaldtse: Maybe we can handle this as part of #153?

src/rnp/rnpcfg.c
flowher: @ronaldtse I agree, instead of crashing we may also handle error correctly. One approach or the other, just not "return error and not handle it".

include/rnp.h
ni4: Ok, moving constants.h to the include/

src/lib/key_store.h
```diff
@@ -114,9 +114,7 @@ typedef struct rnp_key_store_t {
     pgp_hash_alg_t hashtype;
 } rnp_key_store_t;
 
-int rnp_key_store_extension(rnp_t *rnp, char *buffer, size_t buffer_size);
-
-int rnp_key_store_load_keys(rnp_t *rnp, char *homedir);
+int rnp_key_store_load_keys(rnp_t *rnp, int loadsecret);
```
ni4: Ok

src/lib/misc.c
ni4: Agree, my fault.

src/lib/rnp.c
```diff
@@ -1198,16 +888,17 @@ rnp_match_pubkeys(rnp_t *rnp, char *name, void *vp)
...
     pgp_io_t *io;
 
     io = rnp->io;
     if (id == NULL) {
         (void) fprintf(io->errs, "NULL id to search for\n");
-        return 0;
+        return RNP_FAIL;
     }
-    return rnp_key_store_get_key_by_name(rnp->io, rnp->pubring, id) != NULL;
+    return rnp_key_store_get_key_by_name(rnp->io, rnp->pubring, id) != NULL ? RNP_OK :
```
ni4: It's again int vs bool. I will fix this method to return bool, since it is more suitable here.

src/lib/rnpsdk.h
ni4: Ok, commented in the code. Actually it was old function just moved from the wrong place. Now it is in the CLI.

src/rnp/rnp.c
```diff
@@ -252,253 +237,278 @@ show_output(char *out, int size, const char *header)
     }
     if (cc < size) {
         fputs("Short write\n", stderr);
-        return 0;
+        return RNP_FAIL;
     }
-    return cc == size;
+    return cc == size ? RNP_OK : RNP_FAIL;
```
ni4: ok

src/rnp/rnp.c
ni4: Done. Needed to fix some signing function declarations to bool as well.

src/rnp/rnpcfg.c
ni4: Thanks for noticing that, was just a typo.

src/rnp/rnpcfg.c
ni4: The key must be null terminated, I added this to the function description now.
Also it is used internally so our task to ensure that all calls are made according to this requirement.

src/rnp/rnpcfg.c
ni4: Thanks, fixed.

src/rnp/rnpcfg.h
ni4: Ok, changed to bool. Despite it doesn't return false right now probably we'll update init with more complicated logic later and will not need to change the function interface.

src/rnp/rnpcfg.c
ni4: ok

src/rnp/rnpcfg.c
ni4: Changed return value to bool - since it is actually a CLI level all errors can be printed out to the stderr.

src/rnp/rnpcfg.c
ni4: Changed to bool and left for future use - so in case it is needed somehow/somewhere we will not need to type it in again.

src/rnp/rnpcfg.c
ni4: Changed to bool

src/lib/compress.c
ni4: Yeah, I fixed this by moving constants.h and utils.h to the include. Since we'll need multiple include headers sooner or later anyway we need also to change build stuff (Makefiles?) so we'll put rnp/ folder with contents to include, not the single header. Don't know how to do that yet.

src/lib/key_store.c
ni4: I moved this logic to the CLI level, it is more suitable place then lib level. Please see the rnp_cfg_get_defkey.

src/common/constants.h
ni4: Nope, since now SUBDIRECTORY_RNP will be used.

src/lib/Makefile.am
ni4: Ok, I thought that we are at least disabling SSH key code for now. Actually nobody answered in duscussion regarding the SSK keys on Ribose.
Will fix this.

catap: Why you just dissable it? Have we got any discussion about it tat I missed?

Anyway I see your question at Ribose and it was created 10 minutes ago when this comment was created 40 minutes...

src/lib/compress.c
catap: @ni4 I still think that it is a bad design, because all files from includes will be installed to user's include folder.

Why we can't merge it one single file or make all files inside subfolder `rnp` for example?

src/lib/key_store.c
```diff
@@ -36,11 +36,14 @@
...
 #include "key_store_kbx.h"
-#include "key_store_ssh.h"
 #include "packet-print.h"
 #include "packet-key.h"
 #include "packet.h"
 
+#ifdef ENABLE_SSH
+#include "key_store_ssh.h"
+#endif
+
```
catap: I see your point with `enable ssh`, but I don't see any point when you add this configure flag and moved it inside `Makefile.am`

src/lib/key_store.c
catap: ok

src/lib/Makefile.am
ni4: @catap I disabled it since it needed some time to refactor it while we talked with Ronald that most likely SSH code is obsolete.

Sorry, but it's a bit insulting. The discussion was created at Jun, 28 and 10 minutes ago I edited it since found small typo.

src/lib/compress.c
ni4: That's what I tried to say - we should put includes into single rnp folder. Or merge into single one. All advices how do this better are welcome, actually this is the main goal of reviewing PRs, isn't it?

src/lib/Makefile.am
catap: @ni4 I see. Looks like I just missed it :)

Anyway, I agree that ssh keysotre looks like an orphan and it requests complete redisign because it has (had?) hardcoded 'id_rsa'.

Can you just make it build optional for now over confgure flag?

src/lib/compress.c
flowher: yes it is!
I would say we should have many different files in one folder

catap: I vote for forlder with a lot of files.

src/lib/Makefile.am
ni4: Ok, I will make SSH keystore usage configurable

src/rnp/rnpcfg.c
```diff
@@ -0,0 +1,502 @@
...
+{
+    int pos;
+
+    /* checking input parameters for conrrectness */
+    if (!dir || !filename || !res) {
+        return RNP_FAIL;
+    }
+
+    /* concatenating dir, subdir and filename */
+    strcpy(res, dir);
```
dewyatt: I'm pretty sure Coverity is going to (correctly) complain about a number of things in this function.

PR #261, opened 2017-07-01
Code review fixes

MohitKumarAgniotri:
- Fixed some minor resource leak.
- Fixed the initialization of the alloc and free routines.

### DIFF
--- a/src/lib/validate.c
+++ b/src/lib/validate.c
@@ -887,6 +887,7 @@ pgp_validate_file(pgp_io_t *             io,
             }
         }
     }
+    (void) close(outfd);
     pgp_memory_free(validation.mem);
     return ret;
 }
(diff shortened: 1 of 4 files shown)


ronaldtse:
@MohitKumarAgniotri somehow the build is failing. Could you help fix it? Thanks!

MohitKumarAgniotri:
@ronaldtse I am working on that. Somehow, I don't have the Clang 4.0 and I am finding it quite hard to install the same on Cento OS 7, but I am on it.

ronaldtse:
@MohitKumarAgniotri perhaps you can use the docker pre-commit script for clang-format?

src/lib/validate.c
```diff
@@ -887,6 +887,7 @@ pgp_validate_file(pgp_io_t *             io,
             }
         }
     }
+    (void) close(outfd);
```
flowher: I think it should be ``if (!strcmp(outifle, "-")) (void) close(outfd);`` otherwise you may endup closing STDOUT_FILENO (see above)

MohitKumarAgniotri: Yeah, you are right. Let me fix it. Thanks for the review,

flowher: Other than that - LGTM

Issue #263, opened 2017-07-02
find_passphrase is not guaranteed to clear sensitive data


flowher:
## Description

``find_passphrase`` function stores password provided by user in a ``buf`` buffer. It is not guaranteed that this buffer is cleared before function exits. Similarly, ``rnp_generate_key`` doesn't clear password stored in ``passphrase`` buffer.

See [here](https://www.securecoding.cert.org/confluence/display/CINT/MSC06-C.+Beware+of+compiler+optimizations) for more details.

``rnp`` needs utility (a function) similar to ``memset_s`` from C11 (Appendix K), which will guarantee to wipe out requested data buffer. Also please notice that ``memset_s`` is rarely implemented by compilers (C11 standard doesn't require it to be implemented, unfortunately).

There is number of implementations of such function available in open source software (especially crypto libraries).

## Steps to Reproduce

1.  Compile program with ``-O2`` (or higher) optimization level
2.  Enter debug session with gdb
3. Examine ``buf`` after function exists

Behaviour depends on compiler and it's settings. Reproducibility is not guaranteed

## Expected Behavior

It must be guaranteed that ``buf`` is cleared.

flowher:
Also we are using ``getpass``. man(3) says interesting things about ``getpass``:

```
DESCRIPTION
       This  function  is  obsolete.   Do not use it.  If you want to read input without terminal echoing enabled, see the description of the ECHO flag in  termios(3).
``

PR #259, opened 2017-06-30
Memory leaks

catap:
I found and fix few memory leaks, meanwhile my tests shown that it isn't all :(

PR #264, opened 2017-07-02
Ensure data provided in expert mode key generation are correctly parsed

flowher:
Number of changes here:
1. Code to be tested is in ``rnpkyes`` module. In order to test ``rnpkeys`` it needs to be build to static library that doesn't contain main() function. This patch splits rnpkeys.c in two files.
2. ``rnp_t`` structure contains ``passfp`` field which points to user input stream. Password is not the only input that user provides. Field name is now changed to  ``user_input_fp``. But, I think in the future we should add it to ``pgp_io_t`` and change code accordingly. Also function ``find_passphrase`` requires some changes (see 
#263). I believe both - changes to ``pgp_io_t`` and #263 should be addressed at the same time
3. tui.c: Must read from custom stream, not hardcoded ``stdin``.
4. Finally tests for expert mode were added.

Additionally, I've noticed that calling ``rnp_init`` ``rnp_end`` and again ``rnp_init`` causes segmentation fault as array ``rnp_t::names`` is incorrectly freed/initialized. I didn't investigated this further.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
   global:
     - BOTAN_INSTALL="$HOME/builds/botan-install"
     - CMOCKA_INSTALL="$HOME/builds/cmocka-install"
-    - JSON_C_INSTALL="$HOME/builds/json-c-install"
+    - JSONC_INSTALL="$HOME/builds/jsonc-install"
     - LOGNAME="Travis"
     - CLANG_FORMAT_DIFF="clang-format-diff-4.0"
 
@@ -44,7 +44,7 @@ cache:
   directories:
     - "$BOTAN_INSTALL"
     - "$CMOCKA_INSTALL"
-    - "$JSON_C_INSTALL"
+    - "$JSONC_INSTALL"
 
 install:
   - ./ci/install.sh
--- a/Developers Guide.md
+++ b/Developers Guide.md
@@ -93,7 +93,7 @@ git clone https://github.com/riboseinc/rnp
 cd rnp
 export BOTAN_INSTALL="$HOME/builds/botan-install"
 export CMOCKA_INSTALL="$HOME/builds/cmocka-install"
-export JSON_C_INSTALL="$HOME/builds/json-c-install"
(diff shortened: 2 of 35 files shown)


catap:
> Additionally, I've noticed that calling rnp_init rnp_end and again rnp_init causes segmentation fault as array rnp_t::names is incorrectly freed/initialized. I didn't investigated this further.

I think I fixed it at #259

flowher:
Yes, I've noticed it may be fixed. Havn't tried yet as it wasn't merged when I was working on this PR

catap:
Ok, about `pgp_io_t` and streams. I think we should change the API and provides the way to setup callbacks for interactive interaction with user.

By default we should use current logic with read from `stdio`, but we should provide a way to setup callbacks. And first user of new callback will be our tests :)

@flowher, @ronaldtse what do you think about it?

flowher:
+1. That's how I've imagined this working. Nevertheless, notice that in order to test this functionality, input must be provided from stdin and rnpkeys binary still uses stdin.

MohitKumarAgniotri:
@flowher I will review the changes overnight.

ni4:
Please note that soon we are going to get rid of rnpkeys as a separate utility - all functionality will be merged into one. I will start working on this as soon as my PR is approved and merged. I will merge it after this PR.

Also names/values in rnp_t will be removed, so it's not a problem if something works incorrectly.
Somewhere we agreed on the fact that rnp_io_t should be removed/merged into rnp_ctx_t as well.

ni4:
@catap Agree on the fact that pgp_io_t should be replaced by something callback-based, and be stored in rnp_ctx_t, having initializers for common situations - memory read/write/password request, file read/write/password callback, etc.

catap:
@ni4 and we have this issue: https://github.com/riboseinc/rnp/issues/157

flowher:
@ni4 yes,that's obvious. We don't have it yet, though.

flowher:
@MohitKumarAgniotri @ronaldtse @ni4 @catp Thanks for review and comments !

PR #265, opened 2017-07-02
use pkgconfig for getting json-c

zgyarmati:
With the introduction of AC_CHECK_HEADERS in 4cdd9b5 the configure script couldn't find out of the box
the headers of a standard json-c implementation. To save the user from the need of manually
tinkering with the CFLAGS and co. we (re-)introduce using pkg-config to get the proper
compiler- and linker flags, while keeping the CI build (with non-standard json-c install path)
functioning by passing a custom PKG_CONFIG_PATH envvar to ./configure.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -79,9 +79,8 @@ AC_CHECK_HEADERS([sys/cdefs.h sys/file.h sys/mman.h sys/param.h \
 AC_CHECK_HEADERS([bzlib.h],
                  [],
                  [AC_MSG_FAILURE([missing <bzlib.h>; is bzip2 installed?])])
-AC_CHECK_HEADERS([json.h],,
-  [AC_MSG_FAILURE([missing <json.h>; is json-c installed?])]
-)
+# Check JSON
+PKG_CHECK_MODULES(JSON, json-c,,,,[AC_MSG_ERR("json-c missing")])
 
 # Checks for typedefs, structures, and compiler characteristics.
 #
@@ -111,8 +110,6 @@ AX_CHECK_BOTAN_DEFINES([CFLAGS="$CFLAGS $BOTAN_INCLUDES"
 AC_SEARCH_LIBS([_cmocka_run_group_tests], [cmocka],,AC_MSG_ERROR(CMocka not found!))
 AC_SEARCH_LIBS([gzopen], [z],,AC_MSG_ERROR(libz not found!))
 AC_SEARCH_LIBS([BZ2_bzDecompress], [bz2],,AC_MSG_ERROR(Libbz2 not found!))
-AC_SEARCH_LIBS([json_object_new_object], [json-c],,AC_MSG_ERROR(json-c not found!))
-
 
 # Initialize the testsuite
 #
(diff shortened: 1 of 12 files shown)


zgyarmati:
And this also makes the packaging (on different distros) more reliable and future-proof, as pkg-config will always know the correct flags for the installed json-c. 
I'm considering doing the same for botan and cmocka after this one is set.

configure.ac
```diff
@@ -79,9 +79,8 @@ AC_CHECK_HEADERS([sys/cdefs.h sys/file.h sys/mman.h sys/param.h \
 AC_CHECK_HEADERS([bzlib.h],
                  [],
                  [AC_MSG_FAILURE([missing <bzlib.h>; is bzip2 installed?])])
-AC_CHECK_HEADERS([json.h],,
```
ronaldtse: I heard it is good practice when doing `PKG_CHECK_MODULES` to also include the actual check on the lib/headers.

For example, in this [link from Android](https://android.googlesource.com/platform/external/cryptsetup/+/7ae863e380dac362d5df269d9338d83238543b5d/configure.ac) they do the following:

```cmake
dnl Try to detect needed device-mapper static libraries, try pkg-config first.
LIBS="$saved_LIBS -static"
PKG_CHECK_MODULES([DEVMAPPER_STATIC], [devmapper >= 1.02.27],,[
	...
])
LIBS="$saved_LIBS $DEVMAPPER_STATIC_LIBS"
AC_CHECK_LIB(devmapper, dm_task_set_uuid,,
	AC_MSG_ERROR([Cannot link with static device-mapper library.]))
```

And at the end they do:
```cmake
AC_CHECK_HEADER(libdevmapper.h,,[AC_MSG_ERROR([You need the device-mapper library.])])
```

ronaldtse:
@zgyarmati thank you for this! As you mentioned we should really do the same for botan and cmocka!

configure.ac
zgyarmati: Thanks for the review. 
Well, i'm not sure. If we use AC_CHECK_HEADER, it means we have to manually tinker with CFLAGS if json-c is installed on a non-standard location, while the main purpose of this change is to avoid that. Anyway, i push a commit with this and we can review

ronaldtse:
Ping @wizzard you're our other autotools expert here, could you give some comments on best practice? Thanks!

dewyatt:
If Paul doesn't get to it, let me know and I'll take a look. There are a number of improvements I've been meaning to make in this area anyways.

ronaldtse:
@dewyatt it's all yours. Didn't know you were interested in this!

dewyatt:
I pushed a commit on how I think we ought to handle this.
I noticed Jack actually did the same thing for botan (`ax_check_botan.m4`).

This allows the user to provide an optional configure argument `--with-jsonc=DIR`, or to default to pkg-config, or fallback to trying some typical directories.

Let me know if this looks OK (feel free to back out the commit otherwise).

ronaldtse:
@dewyatt the `ax_check_*` approach is definitely the best -- it's just a PITA to do it 👍 

@zgyarmati if you're OK with this feel free to merge it!

zgyarmati:
It's a more comprehensive implementation, ACK

PR #270, opened 2017-07-05
Fixes #210: --output option doesn't work when input is stdin

### LINKED ISSUE #210: The output option doesn't seem to work.
As far as I can tell, `rnp` completely ignores the `--output` option, which makes tests more complicated because it's necessary to have multiple copies of input files in order to get different output files.

ni4:
For the rnp utility, during the signing/encryption/whatever else, --output option worked only for file input, i.e. when input filename is specified, and was ignored when input is stdin. Now it should work for stdin as well.

### DIFF
--- a/src/lib/rnp.c
+++ b/src/lib/rnp.c
@@ -1072,24 +1072,23 @@ rnp_encrypt_file(rnp_ctx_t *ctx, const char *userid, const char *f, const char *
 {
     const pgp_key_t *key;
     const char *     suffix;
-    pgp_io_t *       io;
     char             outname[MAXPATHLEN];
 
-    io = ctx->rnp->io;
     if (f == NULL) {
-        (void) fprintf(io->errs, "rnp_encrypt_file: no filename specified\n");
+        (void) fprintf(ctx->rnp->io->errs, "rnp_encrypt_file: no filename specified\n");
         return RNP_FAIL;
     }
-    suffix = (ctx->armour) ? ".asc" : ".gpg";
     /* get key with which to sign */
     if ((key = resolve_userid(ctx->rnp, ctx->rnp->pubring, userid)) == NULL) {
         return RNP_FAIL;
     }
+    /* generate output file name if needed */
     if (out == NULL) {
+        suffix = (ctx->armour) ? ".asc" : ".gpg";
         (void) snprintf(outname, sizeof(outname), "%s%s", f, suffix);
         out = outname;
     }
-    return (int) pgp_encrypt_file(ctx, io, f, out, key);
+    return (int) pgp_encrypt_file(ctx, ctx->rnp->io, f, out, key);
 }
 
 #define ARMOR_HEAD "-----BEGIN PGP MESSAGE-----"
--- a/src/rnp/rnp.c
+++ b/src/rnp/rnp.c
@@ -43,6 +43,8 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <fcntl.h>
+#include <stdbool.h>
 
 extern char *__progname;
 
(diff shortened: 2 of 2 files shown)


catap:
@ni4 why have you merge it?

I think it's bad idea because I see at master right for example `include "../common/constants.h"` [here](https://github.com/riboseinc/rnp/blob/master/include/rnp.h#L36)

and this is huge PR and I think it should be reviewed more of one person, and review this huge changes request a lot of time.

src/rnp/rnp.c
```diff
@@ -221,24 +223,55 @@ stdin_to_mem(rnp_cfg_t *cfg, char **temp, char **out, unsigned *maxsize)
...
+            flags |= O_TRUNC;
+        else
+            flags |= O_EXCL;
+
+        fd = open(outfile, flags, 0600);
+        if (fd < 0) {
+            if (overwrite && (errno == EEXIST)) {
+                fprintf(stderr, "Failed to write to the %s : file already exists.\n", outfile);
+            } else {
+                fprintf(
```
catap: I think here better to use `strerror` or `perror`

src/rnp/rnp.c
```diff
@@ -221,24 +223,55 @@ stdin_to_mem(rnp_cfg_t *cfg, char **temp, char **out, unsigned *maxsize)
...
+                  stderr, "Failed to open file %s for writing : error %d.\n", outfile, errno);
+            }
+            return RNP_FAIL;
+        }
+    }
+
     for (cc = 0; cc < size; cc += n) {
-        if ((n = write(STDOUT_FILENO, &out[cc], size - cc)) <= 0) {
+        if ((n = write(fd, &out[cc], size - cc)) <= 0) {
```
catap: maybe also show the error?

catap:
1) I think that better keep different changes as different commit. Here I see changes inside two different files and I have no idea why you change `src/lib/rnp.c`.

2) I have no idea that the commit `--output option doesn't work when input is stdin` means. This just broke me :(

src/lib/rnp.c
```diff
@@ -1072,24 +1072,23 @@ rnp_encrypt_file(rnp_ctx_t *ctx, const char *userid, const char *f, const char *
...
         return RNP_FAIL;
     }
-    suffix = (ctx->armour) ? ".asc" : ".gpg";
     /* get key with which to sign */
     if ((key = resolve_userid(ctx->rnp, ctx->rnp->pubring, userid)) == NULL) {
         return RNP_FAIL;
     }
+    /* generate output file name if needed */
     if (out == NULL) {
+        suffix = (ctx->armour) ? ".asc" : ".gpg";
```
ronaldtse: I think we need a better method than hard-coding file extensions? I know it's a moved line but maybe for the future.

src/rnp/rnp.c
ni4: Thanks, will fix this.

src/rnp/rnp.c
ni4: Ok, will improve this.

src/lib/rnp.c
ni4: Yeah, I don't like this one as well. Just didn't fix this since it went a bit far from the purpose of this commit.

ni4:
@catap , thanks for the review.
1.  It was just some cleanup in lib/rnp.c which I forgot to do in previous PR
2. I updated the description, also commit includes link to the issue #210 where it was discussed.

catap:
@ni4, ok, can you also update the commit log by `git rebase -i`? And make style-checker happy!

dewyatt:
FYI I haven't really tested the pre-commit hook with rebasing/amending much. And thinking about it now, it probably *will* do things you don't want (though you shouldn't lose data). Anyways, it may be better to run clang-format manually and then squash (and then enable the pre-commit hook).

ni4:
Sure, I'll update this commit as soon as I get to it. A bit busy during these days. Usually I run clang-format manually before each push to PR, but this time I forgot.

ni4:
@ronaldtse @catap Please see the updated PR. Also, since you mentioned rebase. Do we merge or rebase? This should be also described in developers guide, I think.

catap:
@ni4 I do rebase, but merge it's easy :) So, I think it is up to you in general.

But rebase helps commits log clear and smaller.

dewyatt:
Agree with @catap that it's something to evaluate for every PR, not something to set a solid rule on.
Ronald usually does a plain merge, I usually rebase, but sometimes squash when there are a lot of minor "fix it" commits.

ronaldtse:
@dewyatt personally I prefer a rebase, but a merge is good to preserve the commits (given they are clean) for an easier revert. Now that this is mentioned I'll stick to a rebase 👍

ni4:
Thanks for the comments. I would appreciate some approvals or requests for changes so I can rebase and merge this :)

dewyatt:
> Thanks for the comments. I would appreciate some approvals or requests for changes so I can rebase and merge this :)

LGTM (but note that the reviewer should merge, not the PR creator)

src/rnp/rnp.c
ronaldtse: @ni4 is this fixed? Thanks.

src/rnp/rnp.c
ronaldtse: @ni4 just following up.

src/rnp/rnp.c
```diff
@@ -315,10 +347,12 @@ rnp_cmd(rnp_cfg_t *cfg, rnp_t *rnp, int cmd, char *f)
         ctx.sigcreate = get_birthtime(rnp_cfg_get(cfg, CFG_BIRTHTIME));
         ctx.sigexpire = get_duration(rnp_cfg_get(cfg, CFG_DURATION));
 
+        clearsign = (cmd == CMD_CLEARSIGN) ? true : false;
```
flowher: I think you should also declare ``clearsign`` as ``bool``. Also 
``clearsign = (cmd == CMD_CLEARSIGN)``
means same thing with less code

ni4: @flowher I believe I changed it to bool (among with other things in rnp.c/rnpcfg.c) in some other branch which did not become pull request yet :) Sorry, was a bit out of rnp this week, will get back tomorrow.
@ronaldtse everything should be fixed here, just was waiting for approvals.

ronaldtse:
I'm just going to merge this first and leave the comments as separate issues.

ronaldtse:
Thanks @ni4 !

PR #281, opened 2017-07-07
Fix uninitialized variable read in rnp_cmd.

### LINKED ISSUE #280: Uninitialized variable read in rnp_cmd
## Description

My compiler keeps complaining about this, and it does look valid at a glance. Fix incoming.

## Steps to Reproduce

rnp_cmd uses userid when it may be uninitialized.

dewyatt:
Note: I have not looked in-depth at whether a larger problem exists here.
I have only noticed that my compiler does not like this and that it does appear there's a chance to read uninitialized data.

Fixes #280.

### DIFF
--- a/src/rnp/rnp.c
+++ b/src/rnp/rnp.c
@@ -250,7 +250,7 @@ rnp_cmd(rnp_cfg_t *cfg, rnp_t *rnp, int cmd, char *f)
     char *      out = NULL;
     char *      in = NULL;
     const char *outf;
-    const char *userid;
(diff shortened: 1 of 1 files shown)


ni4:
@dewyatt I was just working on this function, and re-checked it. It is just uninitialized variable, thanks for fixing.

ronaldtse:
@ni4 could you approve this and merge then? Thanks!

ni4:
@ronaldtse Sure, done.

Issue #35, opened 2017-03-27
Key flags are not really handled


dewyatt:
(RFC4880 5.2.3.21)
Key flags are *parsed* but not utilized beyond that. There is also currently no support to save key flags out (this is relatively easy to add, I played with it before).

I think this would also tie in with the current way of handling subkeys. Currently netpgp has the concept of an enckey and a sigkey, and that's about the extent of things. The primary key is sigkey, the last parsed subkey is enckey (if I recall correctly), which is pretty limited.

dewyatt:
TODO (before I forget): https://github.com/riboseinc/rnp/pull/285#issuecomment-313824599

PR #285, opened 2017-07-07
Add key flag loading/printing in listed keys.

dewyatt:
`rnpkeys --list-keys` (and some other commands) will now show something like "[SC]" (sign+certify), similar to gpg.

This also adds key flags to the pgp_key_t structure. These are not currently used beyond loading them from
signature packets and printing them out.

list-packets already handles key flags, so no update there.

The JSON produced is like `{"flags": 3, "usage": "SC"}`. So the raw flags value is available as well as our interpretation of them.

**Known Issues**:
* With subkeys, this will typically show the flags for the wrong key. This will be fixed after https://github.com/riboseinc/rnp/issues/277. Since this has no effect, other than printing some output, I think it's OK for the moment.
* Key flags are not set during key generation, are not persisted to disk, etc. (ongoing effort)

Part of #35

### DIFF
--- a/src/lib/packet-print.c
+++ b/src/lib/packet-print.c
@@ -100,6 +100,7 @@ __RCSID("$NetBSD: packet-print.c,v 1.42 2012/02/22 06:29:40 agc Exp $");
 #define SIGNATURE_PADDING "          "
 
 /* static functions */
+static bool format_key_usage(char *buffer, size_t size, uint8_t flags);
 extern ec_curve_desc_t ec_curves[PGP_CURVE_MAX];
 
 static void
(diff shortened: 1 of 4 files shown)


src/lib/packet-print.c
```diff
@@ -557,6 +561,29 @@ format_uid_notice(char *                 buffer,
     return n;
 }
 
+static bool
+format_key_usage(char *buffer, size_t size, uint8_t flags)
+{
+    const pgp_bit_map_t flags_map[] = {
```
flowher: map should be declared as static const

src/lib/packet-print.c
```diff
@@ -557,6 +561,29 @@ format_uid_notice(char *                 buffer,
...
+      {PGP_KF_ENCRYPT_COMMS | PGP_KF_ENCRYPT_STORAGE, "E"},
+      {PGP_KF_SIGN, "S"},
+      {PGP_KF_CERTIFY, "C"},
+      {PGP_KF_AUTH, "A"},
+    };
+
+    *buffer = '\0';
+    for (size_t i = 0; i < sizeof(flags_map) / sizeof(flags_map[0]); i++) {
```
flowher: let's use PGP_ARRAY_SIZE here

src/lib/packet-print.c
dewyatt: Agreed and reminded me I need to clarify a related part of the dev guide

src/lib/packet-print.c
dewyatt: Good call

flowher:
LGTM

ronaldtse:
@dewyatt Perhaps `usage: "SC"` could be changed to something like `usage: [ "sign", "certify" ]` or `usage: { sign: true, certify: true }` for easier parsing?

dewyatt:
@ronaldtse That's a good idea to make it more descriptive, I'll make an update.

PR #291, opened 2017-07-08
Update Developers Guide.md

dewyatt:
Proposed changes:
* Added a few links to riboseinc:guides.
* Completely flipped the guideline on single-line conditionals. I *thought* modern compilers were good at warning about things like:
  ```c
  if (something)
      do_thing();
      do_other_thing(); // sleepily added this line later, intending for it to be part of the if block
  ```
  But apparently not, as I did actually get bit by this recently in this code base. GCC does seem to warn in *some* cases, but not all. So I think it's probably safest to *always* use braces for if/while/etc.
* Added an exception for constant local static variable declarations. These are perfectly OK when all data is constant, and I neglected to mention this previously. Also added some code to demonstrate.
* Removed maheitsec.

### DIFF
--- a/Developers Guide.md
+++ b/Developers Guide.md
@@ -7,6 +7,8 @@ contributors.
 
 ## Pull Requests
 
+See also: [Guides - Code Review](https://github.com/riboseinc/guides/tree/master/code-review)
+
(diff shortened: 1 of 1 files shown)


ronaldtse:
Thanks @dewyatt !

PR #292, opened 2017-07-09
Botan autotools updates

dewyatt:
This does a few closely-related things:

* Moves `ax_check_botan.m4` into `m4/` with the others
* Fixes the issue where botan includes, etc are appended multiple times
* Adds a version check to make sure we're using Botan >= 2.1 (was a TODO item in configure.ac)


m4/ax_check_botan.m4
```diff
@@ -111,6 +111,23 @@ AC_DEFUN([AX_CHECK_BOTAN], [
...
+      #endif
+      ]])],
+      [
+          AC_MSG_RESULT([ok])
+          $1
+      ], [
+          AC_MSG_RESULT([failed])
+          $2
+      ]
+    )
```
zgyarmati: 2.2 is needed already. The current, unreleased git version of botan sets the version in build.h to 2.2.0, so we can use the same logic, just the value is different.

dewyatt: I had a feeling that might be the case, will update.

m4/ax_check_botan.m4
```diff
@@ -130,4 +149,5 @@ AC_DEFUN([AX_CHECK_BOTAN_DEFINES], [
             $2
         ],
         [#include <botan/build.h>])
+    CPPFLAGS="$save_CPPFLAGS"
```
catap: We can't build rnp without botan, can we? So, why we should keep `CPPFLAGS` without `BOTAN_INCLUDES`? Maybe just add it and don't touch all  `Makefile.am`?

dewyatt: 1. I would say it's bad behavior for an AC function to modify CPPFLAGS/etc.
2. I much prefer to keep things fine-grained and not set things like that globally. Not all of our makefile targets require botan headers.

ronaldtse:
LGTM, thanks @dewyatt !

m4/ax_check_botan.m4
catap: @dewyatt, got your point.

LGTM.

PR #293, opened 2017-07-11
Fixes key generation in expert mode

flowher:
I've found this during ECDH implementation.

Unfortunately recent refactoring completely breaks key generation when expert mode is requested.``rnpkeys`` will ask user for details regarding key to generate, ``rnp_generate_key_expert_mode`` returns with success and then default parameters (RSA/2048) for key to be generated are set as code is wrongly implemented.

Key generation in "expert mode" can be requested from command line and we don't have much tests for TUI. 

I haven't checked if any other command line option is broken.


ronaldtse:
@flowher good find -- could we add some tests so this routine is tested / guarded?

ni4:
Thanks, @flowher . Yeah, such massive changes could definitely break things. 
@ronaldtse I will create tests for CLI after joining rnp and rnpkeys together, to not do the same job twice.

flowher:
@ronaldtse yes, It's good idea to add tests for CLI at some point. I've opened #294 for that.

flowher:
Thanks for quick review!

PR #296, opened 2017-07-12
Fixes for digest algorithm to be used with ECDSA 

### LINKED ISSUE #295: Changes in behaviour of rnpkeys break key generation
## Description

Recently new behaviour in ``rnpkeys`` has been introduced  - rnpkeys is now initialized with some default values. This new functionality doesn't come with supporting tests (+ #294)

When generating a key user may or may not provide a digest algorithm - in case algorithm is provided it must be used, otherwise old ``rnp_getvar`` returns NULL and default algorithm is used. In case of ECDSA default depends on a curve that's used. Function ``adjust_key_params`` expected that NULL is returned in case no digest algorithm is provided. This behaviour has changed.

## Steps to Reproduce

1. ``rnpkeys --generate-key --expert``
2.  Choose ECDSA and NIST P-384
3. This message is displayed, because default HASH/256 is used
```
Generating a new key...
signature.c:228:ecdsa_sign Message hash to small
signature.c:829:pgp_write_sig ecdsa sign failure
signature  521/ECDSA 1c16c24487e33eff 2017-07-11 [] 
Key fingerprint: 42a5 5ad9 86cd f020 be41 30a8 1c16 c244 87e3 3eff 
```
Please notice that normally key generation in such case would fail, but we have another bug #206.

## Expected Behavior
Correct digest algorithm is used and generated key is signed correctly

*Fix is comming*

flowher:
Main goal of this PR is to fix #295 

Additionally I would like to introduce better testing of ECDSA key generation as this feature seems to be  broken every now and then. In order to test it correctly I need to test exact same code as it's called by ``main()`` function. Hence, I've moved implementation of main() into ``librnpkeys.so`` in order to make it testable.

Some additional test cases has been added + small cleanups.


ronaldtse:
Thanks @flowher !

src/cmocka/rnp_tests_generatekey.c
```diff
@@ -491,13 +491,26 @@ rnpkeys_generatekey_verifykeyHomeDirNoPermission(void **state)
...
-static void
-ask_expert_details(rnp_t *ctx, const char *rsp, size_t rsp_len)
+static bool
+ask_expert_details(rnp_t *ctx, rnp_cfg_t *ops, const char *rsp, size_t rsp_len)
 {
+    /* Run tests*/
+    bool      ret = true;
+    rnp_cfg_t cfg = {0};
+    if (setup_rnp_common(ctx, GPG_KEY_STORE, NULL, NULL) != RNP_OK) {
+        return false;
```
catap: I feel confused baceuse we have:
 - `RNP_OK`/`RNP_FAIL`;
 - `true`/`false`;
 - `0`/`1`/`-1`.

I didn't say that it is a error but I feel that we should find a standard way because right now it something strange.

src/lib/hash.c
```diff
@@ -279,3 +279,18 @@ pgp_is_hash_alg_supported(const pgp_hash_alg_t *hash_alg)
...
+
+    botan_hash_t handle = NULL;
+    if (botan_hash_init(&handle, pgp_hash_name_botan(alg), 0) ||
+        botan_hash_output_length(handle, output_length)) {
+        ret = false;
+    }
+
+    botan_hash_destroy(handle);
+    return ret;
+}
```
catap: Maybe add new line here?

src/cmocka/rnp_tests_generatekey.c
flowher: true/false when it makes sense - RNP_OK/RNP_FAIL 0/1/-1 goes away soon

src/rnpkeys/tui.c
```diff
@@ -137,16 +138,19 @@ ask_bitlen(FILE *input_fp)
...
+ * @param   cfg [in]  Requested configuration
  *
  * @returns PGP_E_OK on success
  *          PGP_E_ALG_UNSUPPORTED_PUBLIC_KEY_ALG algorithm not supported
+ *          PGP_E_FAIL indicates bug in the implementation
  *
 -------------------------------------------------------------------------------- */
 pgp_errcode_t
-rnp_generate_key_expert_mode(rnp_t *rnp)
+rnp_generate_key_expert_mode(rnp_t *rnp, const rnp_cfg_t *cfg)
```
catap: I don't see where have you use `cfg` at this function. Have you use it?

src/cmocka/rnp_tests_generatekey.c
catap: OK, I see.

src/lib/hash.c
flowher: done

flowher: done

src/lib/hash.h
```diff
@@ -85,4 +86,16 @@ pgp_hash_alg_t pgp_hash_alg_type(const pgp_hash_t *hash);
...
+ * @brief   Returns output size of an digest algorithm
+ *
+ * @param   [in] alg
+ * @param   [out] output_length: must be not NULL
+ *
+ * @return  true if provided algorithm is supported and it's size was
+ *          correctly retrieved, otherwise false
+ *
+-------------------------------------------------------------------------------- */
+bool pgp_hash_digest_length(pgp_hash_alg_t alg, size_t *output_length);
```
ni4: Aren't hash and digest are synonyms? imho pgp_hash_length or pop_digest_length would be enough.
Also from such function name it seems that it will return length, not put it in the variable. pop_get_hash_length seems to be more appropriate.

src/rnp/rnpcfg.c
```diff
@@ -500,3 +496,15 @@ rnp_cfg_get_defkey(rnp_cfg_t *cfg, rnp_params_t *params)
...
+rnp_cfg_copy(rnp_cfg_t *dst, const rnp_cfg_t *src)
+{
+    if (!src) {
+        return;
+    }
+
+    for (unsigned i = 0; i < src->count; i++) {
+        rnp_cfg_set(dst, src->keys[i], src->vals[i]);
+    }
+}
```
ni4: No newline at end of file?

flowher: yup, there is one

src/lib/hash.h
flowher: Yup, right function name is stupid - it must be called pgp_digest_length. And it should return false if invocation fails as stated in descripiton

src/rnp/rnpcfg.c
flowher: Done

ronaldtse:
@flowher indeed the newline changes are even highlighted by Github 😉

flowher:
I really think it is important. I tried to exaggerate with comment, because when you look at diff it's hard to see what was really changed.

flowher:
Anything more for anyone?

src/rnpkeys/tui.c
```diff
@@ -157,9 +161,40 @@ rnp_generate_key_expert_mode(rnp_t *rnp)
...
+
+        // Adjust hash to curve
+        switch (key_desc->ecc.curve) {
+        case PGP_CURVE_NIST_P_256:
+            if (digest_length < 32) {
+                key_desc->hash_alg = PGP_HASH_SHA256;
+            }
+            break;
+        case PGP_CURVE_NIST_P_384:
+            if (digest_length < 48) {
```
catap: I'm not sure but maybe add some comment or constant name for this magic 48?

Anyway it is very optional

dewyatt: I see what you mean but it seems OK since it's just 384/8 which is right above it.

flowher: It kind of comes from the fact how ecdsa works. But I agree, it's maybe not obvious and even if it is, additional comment doesn't hurt.
Done

flowher:
Thanks !

Issue #276, opened 2017-07-06
Segfault when user-id contains special character


MohitKumarAgniotri:
## Description

RNP library segfaults when trying to retrieve the key using user id which contains special characters like *

## Steps to Reproduce

1. Generate a key 
2. try to find/export the key 
    example: rnpkeys --find-key --userid=*
                   rnpkeys --export-key --userid=*max
3.  Observe segfault

## Expected Behavior

1. No segfault.
2. A Proper key is exported or listed.

## Actual Behavior

Segfault was observed when the commands were executed.

dewyatt:
Great find @MohitKumarAgniotri!
We can see a bit more info with `rnpkeys --debug=key_store.c --find-key --userid=*`.

`regcomp` return value needs to be checked:
https://github.com/riboseinc/rnp/blob/e98911dc3b00eed847c4b5e28e387a8e5b7bb7ab/src/lib/key_store.c#L703

On a side note, the `--find-key` option doesn't actually list keys. It only affects the exit code of `rnpkeys`.

catap:
@MohitKumarAgniotri can you confirm that it has gone?

MohitKumarAgniotri:
@catap Sorry for being away! (got stuck in a sticky place). 
I will test it tonight and confirm. 
Apologies again for the delayed response.

MohitKumarAgniotri:
Tested and couldn't reproduce the issue. LGTM.

ronaldtse:
Thanks @MohitKumarAgniotri ! Closing now.

ronaldtse:
Should we incorporate this in the CLI test suite?

@ni4 @dewyatt

ni4:
@ronaldtse Sure, this should be added to CLI tests. I'll create an issue to not forget about it.

PR #297, opened 2017-07-13
Check return value of `regcomp` for #276

catap:


src/lib/key_store.c
```diff
@@ -678,8 +682,10 @@ get_key_by_name(pgp_io_t *io, const rnp_key_store_t *keyring, const char *name,
     uint8_t          keyid[PGP_KEY_ID_SIZE + 1];
     size_t           len;
 
+    *key = NULL;
+
     if (!keyring || !name || !from) {
-        return NULL;
+        return RNP_OK;
```
dewyatt: Is this correct or should it fail?

catap: I think `RNP_FAIL` is better, yes. You're right.

MohitKumarAgniotri:
@catap: I will add my review by tomorrow evening.

catap:
@MohitKumarAgniotri np, no rush!

src/lib/misc.c
```diff
@@ -1153,8 +1153,12 @@ grabdate(const char *s, int64_t *t)
 
     if (!compiled) {
         compiled = 1;
-        (void) regcomp(
-          &r, "([0-9][0-9][0-9][0-9])[-/]([0-9][0-9])[-/]([0-9][0-9])", REG_EXTENDED);
+        if (regcomp(&r,
+                    "([0-9][0-9][0-9][0-9])[-/]([0-9][0-9])[-/]([0-9][0-9])",
+                    REG_EXTENDED) != 0) {
+            fprintf(stderr, "Can't compile regex\n");
```
flowher: Would be good to add error code at least.

src/lib/rnp.c
```diff
@@ -184,7 +184,10 @@ isarmoured(pgp_io_t *io, const char *f, const void *memory, const char *text)
     char       buf[BUFSIZ];
 
     armoured = 0;
-    (void) regcomp(&r, text, REG_EXTENDED);
+    if (regcomp(&r, text, REG_EXTENDED) != 0) {
+        fprintf(io->errs, "Can't compile regex\n");
```
flowher: and printing error code for this would be also useful

src/rnpkeys/rnpkeys.c
```diff
@@ -355,7 +355,10 @@ parse_option(rnp_cfg_t *cfg, int *cmd, const char *s)
 
     if (!compiled) {
         compiled = 1;
-        (void) regcomp(&opt, "([^=]{1,128})(=(.*))?", REG_EXTENDED);
+        if (regcomp(&opt, "([^=]{1,128})(=(.*))?", REG_EXTENDED) != 0) {
+            fprintf(stderr, "Can't compile regex\n");
```
flowher: again error code. also something saying that it is internal error because if ``regcomp`` fails here it means implementation is wrong

src/lib/rnp.c
```diff
@@ -167,14 +167,14 @@ resolve_userid(rnp_t *rnp, const rnp_key_store_t *keyring, const char *userid)
...
     }
     io = rnp->io;
-    if ((key = rnp_key_store_get_key_by_name(io, keyring, userid)) == NULL) {
+    if (rnp_key_store_get_key_by_name(io, keyring, userid, &key) != RNP_OK) {
         (void) fprintf(io->errs, "cannot find key '%s'\n", userid);
     }
     return key;
 }
 
 /* return 1 if the file contains ascii-armoured text */
```
flowher: so, 1 means armoured, 0 means not-armoured and -1 error. would be good to update comment

src/lib/rnp.c
```diff
@@ -184,7 +184,10 @@ isarmoured(pgp_io_t *io, const char *f, const void *memory, const char *text)
     char       buf[BUFSIZ];
 
     armoured = 0;
-    (void) regcomp(&r, text, REG_EXTENDED);
+    if (regcomp(&r, text, REG_EXTENDED) != 0) {
+        fprintf(io->errs, "Can't compile regex\n");
+        return -1;
```
flowher: you need to call ``regfree()`` before even if regcomp() fails.

src/rnpkeys/rnpkeys.c
dewyatt: Not necessarily, `regcomp` can fail for multiple reasons (incl. memory allocation failures, etc).

flowher: true. It must be freed if memory was allocated. Memory sometimes is allocated even if ``regcomp`` fails.

src/lib/misc.c
catap: I'm not sure because this may happened when rnp hasn't got enough memory or regex is wrong. This is hardcoded regex and it shouldn't be wrong (or we should get it by tests). So, the error means that we haven't enought memory or something like this.

Are you sure that it is a good idea use `pgp_error` for this?

src/lib/rnp.c
catap: why? `regcomp` can't compile the regex, so, I shoudn't try to free it. Am I wrong?

src/rnpkeys/rnpkeys.c
catap: and how can I check that regcomp allocated memory that I should free after fail?

src/lib/rnp.c
dewyatt: I would say @catap is right from looking at the implementation of `regcomp` on my system and the documenation.

src/rnpkeys/rnpkeys.c
dewyatt: @flowher 

> Memory sometimes is allocated even if regcomp fails.

`regfree` should only be called if `regcomp` returns 0, otherwise everything is just on the stack..

src/lib/misc.c
flowher: What I've meant is to add error code that's returned from regcomp. What you can also do is to use RNP_LOG(), so that if it happnes developer gets file name and line number (or add this info manually)

src/lib/rnp.c
flowher: It is easy to check, isn't it? Just compile this and use valgrind. 

```
#include <stdio.h>
#include <stdlib.h>
#include <sys/types.h>
#include <regex.h>

int main(){
	regex_t r;
	char t[100]={0};
	if(!regcomp(&r, t, 0)) {
//		regfree(&r);
		printf("Something bad happend\n");
		return 1;
	}
	regfree(&r);


    return 0;
}

```
```
> valgrind --leak-check=full ./a.out
==10028== Memcheck, a memory error detector
==10028== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==10028== Using Valgrind-3.12.0 and LibVEX; rerun with -h for copyright info
==10028== Command: ./a.out
==10028== 
Something bad happend
==10028== 
==10028== HEAP SUMMARY:
==10028==     in use at exit: 664 bytes in 12 blocks
==10028==   total heap usage: 17 allocs, 5 frees, 2,664 bytes allocated
==10028== 
==10028== 256 bytes in 1 blocks are definitely lost in loss record 11 of 12
==10028==    at 0x4C2AF1F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==10028==    by 0x4F09664: regcomp (in /usr/lib/libc-2.25.so)
==10028==    by 0x4005BC: main (tmp.c:9)
==10028== 
==10028== 408 (224 direct, 184 indirect) bytes in 1 blocks are definitely lost in loss record 12 of 12
==10028==    at 0x4C2AE4F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==10028==    by 0x4C2D1CF: realloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==10028==    by 0x4F08B99: re_compile_internal (in /usr/lib/libc-2.25.so)
==10028==    by 0x4F096C9: regcomp (in /usr/lib/libc-2.25.so)
==10028==    by 0x4005BC: main (tmp.c:9)
==10028== 
==10028== LEAK SUMMARY:
==10028==    definitely lost: 480 bytes in 2 blocks
```

src/rnpkeys/rnpkeys.c
flowher: @catap That's good question and I've no clue. Let's not free in error case as it may cause crash. 
In fact man says nothing about memory allocation. ``regfree()`` crashes (on my machine) if called on unallocated ``regex_t`` which is quite weird.

Issue #273, opened 2017-07-06
Handle remaining comments in #258


ronaldtse:
## Description

There are remaining comments #258 to be handled.

## Steps to Reproduce

See #258 .

ni4:
Fixed via PR #299

PR #299, opened 2017-07-14
Fix remaining comments from PR 258

ni4:
This pull request fixes non-adressed comments from PR #258, as described in issue #273.
They are mostly about:
- return type: int vs bool vs RNP_OK/RNP_FAIL. I changed return type to bool where logically operation succeeds or fails, and return result most likely will not be interesting for the user.
- documented newly added functions
- moved some functions to the right place
- enabled back and refactored SSH key storage. As SSH key storage usage is under discussion now I didn't polish it well. In case of any comments about the SSH key storage I believe they should be formalized as additional issue and be done together during the key storage redoing, so this PR do not hang around forever.

@flowher probably it will have merge conflicts with your ongoing PR #296 since we together changed the same code.


include/constants.h
```diff
@@ -46,6 +46,12 @@
 #define MAX_PASSPHRASE_ATTEMPTS 3
 #define INFINITE_ATTEMPTS -1
 
+/* SHA1 is now looking as though it should not be used.  Let's
+ * pre-empt this by specifying SHA256 - gpg interoperates just fine
+ * with SHA256 - agc, 20090522
+ */
+#define DEFAULT_HASH_ALG "SHA256"
```
flowher: fully agree

src/rnp/rnpcfg.c
```diff
@@ -497,6 +576,105 @@ rnp_cfg_get_defkey(rnp_cfg_t *cfg, rnp_params_t *params)
     } else {
         params->defkey = strdup(userid);
     }
+}
 
-    return RNP_OK;
+/**
```
flowher: I think comments should be in header files (as source files are not always delivered with library)

ni4:
@flowher Looks like I was lucky enough to get into some bug in eddsa (as far as I remember you were implementing it? or @catap?). Please see the #300 for the details.

flowher:
@ni4 - I've got to say, you're really lucky to find stuff like that.

include/constants.h
ni4: Updated, thanks.

src/lib/key_store_ssh.c
```diff
@@ -465,50 +465,27 @@ ssh2_readkeys(pgp_io_t *       io,
...
     rnp_key_store_t *secring;
     struct stat      st;
-    pgp_hash_alg_t   hashtype;
-    char *           hash;
-    char             f[MAXPATHLEN];
-    char *           filename;
-
-    if ((filename = rnp_getvar(rnp, "sshkeyfile")) == NULL) {
-        /* set reasonable default for RSA key */
-        (void) snprintf(f, sizeof(f), "%s/id_rsa.pub", homedir);
```
catap: Before behaviour was:
```sh
./src/rnpkeys/rnpkeys --keystore-format=SSH --list-keys
1 key found
signature  2048/RSA (Encrypt or Sign) 78f54270acf21f6a 1970-01-01
Key fingerprint: dacb0c54f989e60578f54270acf21f6a
uid              Kirills-MacBookPro.local (/Users/catap/.ssh/id_rsa.pub) <catap@satellite> 

```

are you sure that you keep it same?

src/rnp/rnp.c
```diff
@@ -666,7 +667,7 @@ main(int argc, char **argv)
 
     rnp_params_free(&rnp_params);
 
-    if (!rnp_key_store_load_keys(&rnp, rnp_cfg_getint(&cfg, CFG_NEEDSSECKEY))) {
+    if (!rnp_key_store_load_keys(&rnp, (bool) rnp_cfg_getint(&cfg, CFG_NEEDSSECKEY))) {
```
dewyatt: Would it make any sense to add `rnp_cfg_getbool` and avoid casting?

src/rnp/rnpcfg.c
```diff
@@ -192,33 +205,49 @@ rnp_cfg_set(rnp_cfg_t *cfg, const char *key, const char *val)
...
+ *  @param key must be null-terminated string
+ *  @param val value, which will be set
+ *
+ *  @return true if operation succeeds or false otherwise
+ **/
+bool
 rnp_cfg_setint(rnp_cfg_t *cfg, const char *key, int val)
 {
     char st[16] = {0};
     sprintf(st, "%d", val);
```
dewyatt: Would prefer to see `snprintf`, if only to satisfy coverity.

src/rnp/rnpcfg.c
```diff
@@ -497,6 +576,105 @@ rnp_cfg_get_defkey(rnp_cfg_t *cfg, rnp_params_t *params)
...
+grabdate(const char *s, int64_t *t)
+{
+    static regex_t r;
+    static int     compiled;
+    regmatch_t     matches[10];
+    struct tm      tm;
+
+    if (!compiled) {
+        compiled = 1;
+        (void) regcomp(
```
dewyatt: We should really be checking these return values, regcomp makes allocations and can fail. I think this was just a function move though, so feel free to dismiss.

catap: and it was fixed yesterday: https://github.com/riboseinc/rnp/pull/297 ;)

dewyatt: I knew it seemed familar! Ok we can ignore this comment. :)

catap: @dewyatt nope, because it introduced bug that was fixed at old place ;)

dewyatt: @catap Ah crap you're right, I forgot it was moved. I need more coffee.

src/cmocka/rnp_tests_exportkey.c
```diff
@@ -50,7 +50,7 @@ rnpkeys_exportkey_verifyUserId(void **state)
     rnp_assert_ok(rstate, rnp_key_store_load_keys(&rnp, 1));
```
flowher: Sorry this comment is about utils.h but I can't comment there.
Why do we move utils.h to "include" ?

My understanding was that this directory contains only files that are public API. Having public API in separated definitely is good idea. Now utils.h contains stuff which should be kept internal to rnp. We could have separated directory called "common" or "utils" which would contains stuff that's commonly used by different components but are not public API.
We should be careful about what we make public API.

src/rnp/rnpcfg.c
```diff
@@ -61,11 +55,18 @@ rnp_cfg_load_defaults(rnp_cfg_t *cfg)
...
 
-int
+static void rnp_cfg_get_defkey(rnp_cfg_t *cfg, rnp_params_t *params);
+static bool rnp_cfg_get_ks_info(rnp_cfg_t *cfg, rnp_params_t *params);
+
+/** @brief apply configuration from keys-vals storage to rnp_params_t structure
+ *  @param cfg [in] rnp config, must be allocated and initialized
+ *  @param params [out] this structure will be filled so can be further feed into rnp_init
+ *
+ *  @return true on success, false if something went wrong
```
flowher: If something went wrong, does user needs to release memory that potentially could be allocated to ``params`` before "something" occured ?

src/rnp/rnpcfg.c
```diff
@@ -116,8 +118,12 @@ rnp_cfg_find(rnp_cfg_t *cfg, const char *key)
     return (i == cfg->count) ? -1 : (int) i;
 }
 
-/* resize keys/vals arrays to the new size. Only expanding is supported */
-static int
+/** @brief resize keys/vals arrays to the new size. Only expanding is supported now.
```
flowher: Idea: So why not too call it ``rnp_cfg_expand``. To support shrinking you probably will need change function definition anyway, isn't?

src/rnp/rnpcfg.c
```diff
@@ -116,8 +118,12 @@ rnp_cfg_find(rnp_cfg_t *cfg, const char *key)
...
+ *  Pointers keys and vals will be freed in rnp_cfg_free
+ *
+ *  @return true on success, false if allocation fails.
```
flowher: does caller must call ``rnp_cfg_free`` if function fails?

src/rnp/rnpcfg.c
flowher: +1

src/rnp/rnpcfg.c
```diff
@@ -497,6 +576,105 @@ rnp_cfg_get_defkey(rnp_cfg_t *cfg, rnp_params_t *params)
...
+ * - 2017-07-12 : as the exact date when signature becomes invalid
+ * - 60000 : number of seconds
+ *
+ * @param s [in] NULL-terminated string with the date
+ * @param t [out] On successfull return result will be placed here
+ * @return duration time in seconds
+ */
+
+uint64_t
+get_duration(const char *s)
```
dewyatt: @flowher Maybe we ought to discuss this over in https://github.com/riboseinc/rnp/issues/192 instead?

flowher: @dewyatt I don't have strong opinion on that (yet). Let me think.

src/rnp/rnp.c
ni4: Yeah, was thinking about it. Now it seems to be much more applicable for a number of cfg variables.

src/rnp/rnpcfg.c
ni4: Sure, will change it. I'm quite new to coverity, so don't know all the things yet.

dewyatt: @ni4 Ah I see you haven't been added to Coverity yet, you may want to go ahead and request access: https://scan.coverity.com/projects/riboseinc-rnp

ni4: @dewyatt Sent request, thanks.

src/rnp/rnpcfg.c
ni4: Didn't get whether I should fix it or not so fixed.

src/rnp/rnpcfg.c
ni4: Updated the function description.

src/rnp/rnpcfg.c
ni4: It's just a static function now, and used only in one place. So let it be called resize, this will not make impact on anything.

src/rnp/rnpcfg.c
ni4: Actually rnp_cfg_free must be called on any cfg after it was created and initialized via rnp_cfg_init. Will update this function description.

src/cmocka/rnp_tests_exportkey.c
ni4: It was quick solution, agree that it is not a good place for utils.h. Also I think we should move some stuff to utils.c as well, but not in this PR. Moved it to the src/lib/utils.h, seems to be a good place.

ni4:
@dewyatt Please see the updated PR. I fixed and replied comments (while GitHub PR overview something is getting quite messy).
The only thing left - ssh key storage functionality, which will be addressed via the separate PR.

Issue #275, opened 2017-07-06
Need of `libverify`


ronaldtse:
## Description

`libverify` was previously used to support `rnpv` for a very light verification library, but since we're going to merge `rnp` and `rnpv` ( #69 ), `libverify` is no longer useful.

@ni4 suggests we should we remove it. What do you think?

Personally I believe just `librnp` is good enough, and signing/verification functionality should be placed together.

flowher:
Maybe I'm missing a point, but I don't see why too keep it as separated binary.

ronaldtse:
@flowher #69 describes several approaches for the command line util. Several individuals have mentioned that it is hard to use multiple CLIs, and for people who come from GnuPG expect everything provided in one binary.

We can always take an approach with aliasing, such as `alias rnpk=rnp --keys` or the other way round.

ronaldtse:
What do you think @catap @dewyatt ?

catap:
@ronaldtse I agree with you

ronaldtse:
@ni4 would you like to handle the culling of `libverify`? 😉

ni4:
@ronaldtse sure, will cut it out

PR #301, opened 2017-07-15
Remove rnpv and all the corresponding stuff.

ni4:
The goal of this PR is removal of libverify and rnpv according to issue #275.
Basically, I just removed directory rnpv and cleaned up all the build files and settings.

Given that rnp is not in release yet I didn't care about 'after install' cleanup, i.e. to delete old rnpv when installing/upgrading packages, and so on so on.


ronaldtse:
Thank you @ni4 !

dewyatt:
Merging since we have 2 approvals. :)

PR #306, opened 2017-07-16
Set hash algorithm while calculating fingerprint

flowher:
Found during ECDH development - hash algorithm was not set when fingerprint is calculated.

### DIFF
--- a/src/lib/misc.c
+++ b/src/lib/misc.c
@@ -397,7 +397,6 @@ pgp_fingerprint(pgp_fingerprint_t *fp, const pgp_pubkey_t *key)
         if (rnp_get_debug(__FILE__)) {
             hexdump(stderr, "v2/v3 fingerprint", fp->fingerprint, fp->length);
         }
-        fp->hashtype = PGP_HASH_MD5;
     } else if (key->version == 4) {
         mem = pgp_memory_new();
         if (mem == NULL) {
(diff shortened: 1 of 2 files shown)


dewyatt:
The field doesn't appear to be used anywhere, but this is obviously better than not populating it at all.
Eventually it'd be great to remove support for (deprecated) key versions < 4, but that's an issue for another day.

ronaldtse:
@dewyatt it'd be great if we can separate the functions for different key versions (since there's a proposal for v5) instead of an `if` statement here 😄

flowher:
Indeed the only usage was removed by one of the recent commits. Gone now.
BTW: 
```
typedef struct pgp_key_t {
...
    pgp_fingerprint_t sigfingerprint; /* pgp signature fingerprint */
   ....
    pgp_fingerprint_t encfingerprint; /* pgp encryption id fingerprint */
...
}
```

Any idea why do we have 2 fingerprints in the structure ? The encfingerprint is only used by keystore and comments says something that "it should be done better".

ni4:
@flowher As far as I understand those different fingerprints were added to store fingerprint of signing/encrypting key/subkey. Looks like an artifact - fingerprint must be single, and property of key or subkey itself.

flowher:
@ni4 Thanks, that would be also my understanding.

dewyatt:
@ni4 is correct on those fields and I've been looking at removing them as part of https://github.com/riboseinc/rnp/issues/277

flowher:
ok, thanks. 
I don't want to change code, but I've updated comment in the packet.h to reflect what was stated here.

flowher:
Guys, please don't hesitate add yourself to my reviews if you feel like you want to.

ni4:
Merging as it was second approval.

PR #307, opened 2017-07-17
Fixes #256: duplication of keys in keyring after the new key generation

### LINKED ISSUE #256: Keyring is duplicated when new key is generated
## Description

When new key is generated it should be appended to the keyring. However actually all existing keys are appended as well.

## Steps to Reproduce
```
MacBook-Pro:rnp Ribose$ rnpkeys --list-keys
pgp_mem_readfile: can't open "/Users/Ribose/.rnp/pubring.gpg"
cannot read pubring /Users/Ribose/.rnp/pubring.gpg
cannot read pub keyring
fatal: failed to load keys

MacBook-Pro:rnp Ribose$ rnpkeys --generate-key
pgp_mem_readfile: can't open "/Users/Ribose/.rnp/pubring.gpg"
cannot read pubring /Users/Ribose/.rnp/pubring.gpg
cannot read pub keyring
Generating a new key...
signature  2048/RSA (Encrypt or Sign) 574a1b37ac7ba267 2017-06-30 
Key fingerprint: ed9f 605e fff0 59bc ae0a 2aff 574a 1b37 ac7b a267 
uid              RSA (Encrypt or Sign) 2048-bit key <Ribose@localhost>
rnp: generated keys in directory /Users/Ribose/.rnp
Enter passphrase for 574a1b37ac7ba267: 
Repeat passphrase for 574a1b37ac7ba267: 

MacBook-Pro:rnp Ribose$ rnpkeys --list-keys
1 key found
signature  2048/RSA (Encrypt or Sign) 574a1b37ac7ba267 2017-06-30
Key fingerprint: ed9f605efff059bcae0a2aff574a1b37ac7ba267
uid              RSA (Encrypt or Sign) 2048-bit key <Ribose@localhost> 

MacBook-Pro:rnp Ribose$ rnpkeys --generate-key --expert
Please select what kind of key you want:
	(1)  RSA (Encrypt or Sign)
	(19) ECDSA
	(22) EDDSA
> 19
Please select which elliptic curve you want:
	(1) NIST P-256
	(2) NIST P-384
	(3) NIST P-521
> 1
Generating a new key...
signature  384/ECDSA 8fd534d81ac53199 2017-06-30 
Key fingerprint: 643c 3a8a 9c99 0e87 26d6 76a3 8fd5 34d8 1ac5 3199 
uid              ECDSA 384-bit key <Ribose@localhost>
rnp: generated keys in directory /Users/Ribose/.rnp
Enter passphrase for 8fd534d81ac53199: 
Repeat passphrase for 8fd534d81ac53199: 
MacBook-Pro:rnp Ribose$ 

MacBook-Pro:rnp Ribose$ rnpkeys --list-keys
3 keys found
signature  2048/RSA (Encrypt or Sign) 574a1b37ac7ba267 2017-06-30
Key fingerprint: ed9f605efff059bcae0a2aff574a1b37ac7ba267
uid              RSA (Encrypt or Sign) 2048-bit key <Ribose@localhost> 

signature  2048/RSA (Encrypt or Sign) 574a1b37ac7ba267 2017-06-30
Key fingerprint: ed9f605efff059bcae0a2aff574a1b37ac7ba267
uid              RSA (Encrypt or Sign) 2048-bit key <Ribose@localhost> 

signature  384/ECDSA 8fd534d81ac53199 2017-06-30
Key fingerprint: 643c3a8a9c990e8726d676a38fd534d81ac53199
uid              ECDSA 384-bit key <Ribose@localhost> 

```
## Expected Behavior
We have two keys, first RSA and second ECDSA 256-bits
## Actual Behavior
First key (RSA) is duplicated, also ECDSA key has 384 bits instead of 256.

ni4:
When key is loaded then all packets are saved into the sub packets array. Then, when keyring is re-saved all those packets were written together with public key packet + user id, causing duplication of key + user id packets on output. I changed this behavior that now for loaded key pub key and userid are saved from records only for non-loaded keys. This also have another meaning : already existing packets should written as they were to not break to the signature.

This solution is quick and dirty since within current implementation of key data structures it is hard to imagine correct working of multiple user ids, subkeys, signatures, direct key signatures and so on. 

@dewyatt I will write separate post about it somewhere, since it is closely related to what you are doing now.

### DIFF
--- a/src/lib/create.c
+++ b/src/lib/create.c
@@ -512,18 +512,23 @@ pgp_write_xfer_pubkey(pgp_output_t *         output,
         pgp_writer_push_armoured(output, PGP_PGP_PUBLIC_KEY_BLOCK);
     }
     /* public key */
-    if (!write_struct_pubkey(output, PGP_PTAG_CT_PUBLIC_KEY, &key->key.pubkey)) {
+    if (key->loaded) {
+        if ((key->packetc < 1) ||
+            (!pgp_write(output, key->packets[0].raw, (unsigned) key->packets[0].length))) {
+            return RNP_FAIL;
+        }
+    } else if (!write_struct_pubkey(output, PGP_PTAG_CT_PUBLIC_KEY, &key->key.pubkey)) {
         return RNP_FAIL;
     }
 
     /* TODO: revocation signatures go here */
-
+    /* TODO: the following is wrong, and works only for single uid */
     /* user ids and corresponding signatures */
     for (i = 0; i < key->uidc; i++) {
-        if (!pgp_write_struct_userid(output, key->uids[i])) {
+        if (!key->loaded && !pgp_write_struct_userid(output, key->uids[i])) {
             return RNP_FAIL;
         }
-        for (j = 0; j < key->packetc; j++) {
+        for (j = key->loaded ? 1 : 0; j < key->packetc; j++) {
             if (!pgp_write(output, key->packets[j].raw, (unsigned) key->packets[j].length)) {
                 return RNP_FAIL;
             }
--- a/src/lib/key_store.c
+++ b/src/lib/key_store.c
@@ -504,6 +504,7 @@ rnp_key_store_add_keydata(pgp_io_t *         io,
         pgp_fingerprint(&key->sigfingerprint, &keydata->pubkey);
         key->type = tag;
         key->key = *keydata;
+        key->loaded = 1;
     } else {
         // it's is a subkey, adding as enckey to master that was before the key
         // TODO: move to the right way — support multiple subkeys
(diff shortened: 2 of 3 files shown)


dewyatt:
Yeah this whole area is something I'm currently reworking, it's an area where we have a lot of things that *appear* to work but are incorrect, etc.

I think this hack is OK for now, assuming you've tested it.

src/lib/create.c
```diff
@@ -512,18 +512,23 @@ pgp_write_xfer_pubkey(pgp_output_t *         output,
...
-
+    /* TODO: the following is wrong, and works only for single uid */
     /* user ids and corresponding signatures */
     for (i = 0; i < key->uidc; i++) {
-        if (!pgp_write_struct_userid(output, key->uids[i])) {
+        if (!key->loaded && !pgp_write_struct_userid(output, key->uids[i])) {
             return RNP_FAIL;
         }
-        for (j = 0; j < key->packetc; j++) {
+        for (j = key->loaded ? 1 : 0; j < key->packetc; j++) {
```
flowher: "? 1 : 0" look weird.

ni4: It's kinda defensive coding, we need here exact integer, "just in case".

flowher: Ah. Got it. Good stuff!

PR #313, opened 2017-07-18
Define error codes: initial

flowher:
This is initial definition of error codes for #153.

* List is definitely not exhaustive, feel free to add whatever needed
* If you absolutely need those definitions, please merge and use. Otherwise, please keep the PR open until I update the code (+dev guide).
* RNP_SUCCESS in final version will be called RNP_OK


catap:
Good idea, I love it!

dewyatt:
Any thoughts on using an anonymous enum here?

```c
enum {
    RNP_SUCCESS                     = 0x00000000,
    RNP_ERROR_GENERIC               = 0x71000000,
    RNP_ERROR_BAD_FORMAT,
    RNP_ERROR_BAD_PARAMETERS,
    RNP_ERROR_NOT_IMPLEMENTED,
    RNP_ERROR_NOT_SUPPORTED,
    RNP_ERROR_OUT_OF_MEMORY,
    RNP_ERROR_SHORT_BUFFER,

    /* Storage */
    RNP_ERROR_STORAGE_NOT_AVAILABLE = 0x72000001,
    RNP_ERROR_READ,
    RNP_ERROR_WRITE,

    /* Crypto */
    RNP_ERROR_BAD_STATE             = 0x73000000,
    RNP_ERROR_MAC_INVALID,
    RNP_ERROR_SIGNATURE_INVALID,
    RNP_ERROR_KEY_GENERATION,

    /* Parsing */
    RNP_ERROR_NOT_ENOUGH_DATA       = 0x74000000,
    RNP_ERROR_UNKNOWN_TAG,
    RNP_ERROR_PACKET_NOT_CONSUMED,
    RNP_ERROR_NO_USERID,
};
```

That would help prevent accidental duplicates, for example.

flowher:
Cool. But why anonymous not typedef?

ni4:
Enum seems better for me, at least from the debugging point of view.
Also one concern regarding the uint32_t - why not just int? I guess we are not going to be compatible with less then 32 bit processor.

dewyatt:
@ni4 
> Also one concern regarding the uint32_t - why not just int?

At the moment, I can't think of any pros/cons on this really. I guess using int would make our headers *slightly* more compatible with pre-C99 standards. Anything else?

> I guess we are not going to be compatible with less then 32 bit processor.

It's up to the compiler to handle that (we also use `uint64_t`). Generally it doesn't matter if the processor's native width is less than what is needed for `uint32_t`/`uint64_t`, it will just end up generating slower code. So as long as there is a C99 compiler for the platform, it's fine.

dewyatt:
LGTM though I'm still leery of the bool returns because it sometimes makes it difficult to differentiate between an error and a result.

It's hard to imagine all possible use cases but say there was some web application that:

* Attempts to validate a signature from a user
* Disables a user's account if a signature is invalid

So assume we had a signature validation function that returns bool, and an internal error occurs (malloc fails or something), false will be returned and the user's account will be disabled as a result.

Also:

* bool returns tend to cause a cascading effect where a new function can't return a very specific error code because it calls some other function that only returns a bool.
* It's a little confusing sometimes if a function returns bool and it takes some result pointer parameter. If it returns false, there's a question of whether the data in the result is valid or if the function bailed early without setting valid data.

Anyways, I do at least like bool returns for internal functions that have no failure points.

Issue #304, opened 2017-07-15
Restructure src/cmocka


dewyatt:
## Description

I was thinking we can improve the file structure and naming in `src/cmocka`.
We have all source files prefixed with `rnp_tests_` which is a bit redundant. Also the directory name `cmocka` seems unnecessarily implementation-specific and could probably just be `tests` (or `tests/cmocka`)?

I was thinking something along these lines would be a minor improvement:

```
$ tree rnp/src/tests/
rnp/src/tests/
├── cipher.c
├── exportkey.c
├── generatekey.c
├── Makefile.am
├── rnp_tests.c
├── rnp_tests.h
├── support.c
└── support.h

0 directories, 8 files
```

In the future it may make sense to split the tests up more and maybe add some subdirs that test specific components, etc.

But I think the above is at least a small step towards something slightly more organized.

EDIT: typo

ronaldtse:
Can I give three thumbs up for this? :-)

ni4:
I can do this when will be updating cmocka with CLI tests.
Probably we should also remove current contents of tests directory, they seems a bit outdated?
Another idea - some tests worth running multiple times (like low-level ecdsa/eddsa key generation/signing), probably cmocka allow it. Or just update tests themselves.

dewyatt:
> some tests worth running multiple times 

@ni4 I believe @catap added some support for this in 6d550494ce503d2ee8c555c11349d740df336b95. Looking at it now though, I think it needs some work. I would expect the loop to just call `cmocka_run_group_tests` multiple times and do an early exit if `ret != 0`. It seems like the result isn't currently used. Anyways, maybe you can build on that.

dewyatt:
> I can do this when will be updating cmocka with CLI tests.

@ni4 Also I think it would be helpful if you could do this in a separate PR, that ought to be quicker and easier to review.

> Probably we should also remove current contents of tests directory, they seems a bit outdated?

I think there was some discussion about this and @ronaldtse had some input I can't recall.

ni4:
@dewyatt Sure. Previous two PRs were huge and messy due to large amount of breaking changes, which then were merged together with new live changes which used older codebase. I don't like such PRs as well.

zgyarmati:
As a sidenote here, but maybe it worth a separate ticket: it also would be nice from the packaging point of view to get all of tests (both cmocka and CLI tests) integrated into automake and invoked by a single call of `make check`.

ronaldtse:
[...]

Issue #294, opened 2017-07-11
Tests for CLI


flowher:
We need test cases that test tools which are called from command line. It seems to me that some tests are already available in ``/tests`` directory - maybe this could be good starting point.

flowher:
@MohitKumarAgniotri It's just FYI - I think it would fall somewhere under yours responsibility

ronaldtse:
I think @ni4 was going to do something about it, he already had a python script.

ni4:
@ronaldtse Yeah, I planned to add some tests for the CLI once finished with gluing together rnp + rnpkeys.  So we can assign this task to me.

ronaldtse:
Thanks @ni4 !

ni4:
@ronaldtse I think we can close this issue since we already have tests for CLI? They needs to be updated to cover some specific areas, but this issue could be considered as resolved.

@dewyatt Could you please integrate ./cli_tests.py with Travis? Should I create an issue for that?

ronaldtse:
Thanks @ni4, and please do create an issue for @dewyatt on Travis integration !

dewyatt:
> @dewyatt Could you please integrate ./cli_tests.py with Travis? Should I create an issue for that?

I think the challenge here will be that Travis CI is based on Ubuntu Trusty and we'll only have gpg 1.4 and 2.0 available. I think these tests are designed around 2.1, so we'll have to figure something out.

PR #314, opened 2017-07-18
Restructure src cmocka

ni4:
This PR moves cmocka tests to src/tests, and rename files, fixing #304
/tests/ folder with .at tests will be removed as part of the #294


ronaldtse:
Very very nice. Thanks @ni4 !

Issue #192, opened 2017-06-08
Provide API documentation


zgyarmati:
Partially depends on #167.
 Currently the API of the lib documented only in the src/lib/librnp.3 file and don't have any autogenerating infrastructure in place. Instead of this we need a source-code comment based, autogenerated  API documentation. 
For this we can use [doxygen](http://www.stack.nl/~dimitri/doxygen/)  or with a slightly more modern approac [cldoc](https://github.com/jessevdk/cldoc). The disadvantage of the latter one is that it's not available in most the distros (yet), but we can still integrate it into our build process. 
So which one should be used?
 After decided which one we want to use, i'm happy to take the coordination of setting up the auto-generation, although obviously wrinting the API comments themselves should be a team effort.

dewyatt:
A lot of the internal API is documented with doxygen (at least the parts inherited from OpenPGP-SDK).
But I know some of the doxygen comments are misleading and out of date.

cldoc certainly looks promising but I do have some concerns there: https://github.com/jessevdk/cldoc/issues/11
If that's still a valid issue, I'm not sure the output would be grouped in a meaningful way. We should look into it maybe.

zgyarmati:
Indeed, i shouldn't have imply that there are no existing doxygen document comments, on the other hand i more prefer here to the public API, mainly because of the #173 .
 I didn't spot this bug in cldoc (i've used only for C++), and anyway, i don't have a strong position about using it, so i'm totally fine if we work in doxygen.

ronaldtse:
It seems that jessevdk/cldoc#11 is not (or not going to be) fixed. Maybe we should stick with doxygen for now and could migrate to another solution afterwards?

dewyatt:
@flowher mentioned:

> Except for static functions, I think comments should go to header files as those files are delivered to the end user (especially for API functions).

There are pros and cons to both sides of this so we should discuss and standardize

### Documentation in Header Files

#### Pros

* Users that browse the public header files can read the doxygen comments.

#### Cons

* An excess of documentation in the headers can make it more difficult for users to find the actual functions/types/etc.
* The point of using doxygen is to generate HTML/etc documentation, so users don't need to read header files to figure out what's going on.
* Changing a comment in a header can trigger a rebuild of the entire source tree, depending on the code base.

### Documentation in Source Files

#### Pros

* When a developer updates a function, I think they are much more likely to update the corresponding doxygen comment if it is right above the code.
[...]

PR #319, opened 2017-07-19
Added example for doxygen comment

ni4:
This PR updates Developers Guide with sample doxygen comment as it was discussed in #192.
All comments, including related to English language usage, are welcome.

@flowher I saw you like lines of stars/dashes in comments, imho they are not visually appealing. However if you insist on them we can run a voting.

### DIFF
--- a/Developers Guide.md
+++ b/Developers Guide.md
@@ -373,11 +373,11 @@ Comments should use doxygen markdown style, like the following example:
  *  After an empty line you may add detailed description in case it is needed. You may put 
  *  details about the memory allocation, what happens if function fails and so on.
  *
- *  @param [in] param1 first parameter, null-terminated string which should not be NULL
- *  @param [in] param2 integer, some number representing something
- *  @param [in, out] size number of bytes available to store in buffer
- *  @param [out] buffer buffer to store results, may be NULL. In this case size can be used to 
- *                      obtain the required buffer length
+ *  @param param1 first parameter, null-terminated string which should not be NULL
+ *  @param param2 integer, some number representing something
+ *  @param size number of bytes available to store in buffer
+ *  @param buffer buffer to store results, may be NULL. In this case size can be used to 
+ *                obtain the required buffer length
  *  @return 0 if operation succeeds, or error code otherwise. If operation succeeds then buffer 
  *          is populated with the resulting data, and size contains the length of this data.
  *          if error code is E_BUF_TOOSMALL then size will contain the required size to store
--- a/Developers Guide.md
+++ b/Developers Guide.md
@@ -356,14 +356,19 @@ Do not:
 
 ### Documentation
 Documentation is done in Doxygen comments format, which must be put in header files.
-Exception are static or not exposed to the public API functions - they should be documented in 
-the code.
+
+Exception are static or having only definition functions - it is not required to document them,
+however if they are documented then this should be done in the source file and using the @private tag.
 
 Comments should use doxygen markdown style, like the following example:
 
 ```c
-/** @brief brief description of the sample function which does something
- *         Which may be continued here
+/** Some comments regarding the file purpose, like 'PGP packet parsing utilities'
+ *  @file
+ */
+
+/** brief description of the sample function which does something, keyword 'brief' is ommitted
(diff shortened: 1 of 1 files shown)


Developers Guide.md
```diff
@@ -354,6 +354,33 @@ Do not:
...
+ *  @param [in] param2 integer, some number representing something
+ *  @param [in, out] size number of bytes available to store in buffer
+ *  @param [out] buffer buffer to store results, may be NULL. In this case size can be used to 
+ *                      obtain the required buffer length
+ *  @return 0 if operation succeeds, or error code otherwise. If operation succeeds then buffer 
+ *          is populated with the resulting data, and size contains the length of this data.
+ *          if error code is E_BUF_TOOSMALL then size will contain the required size to store
+ *          the result
+ **/
+uint32_t
```
flowher: type for error codes is called rnp_error

Developers Guide.md
```diff
@@ -354,6 +354,33 @@ Do not:
...
 * Do not use `pragma`, and try to avoid `__attribute__` as well.
 
+### Documentation
+Documentation is done in Doxygen comments format, which must be put in header files.
+Exception are static or not exposed to the public API functions - they should be documented in 
+the code.
+
+Comments should use doxygen markdown style, like the following example:
+
+```c
```
flowher: 'c' like comment :) maybe not needed?

dewyatt:
Great! Some minor suggestions:

1. I think we may as well drop the `@brief` markup since we're using javadoc-style comments, we can just turn on `JAVADOC_AUTOBRIEF` and use:

```c
/** your brief description here
 *
 *  Details here
[...]
 */
```

2. > Exception are static or not exposed to the public API functions - they should be documented in the code.

Can you also add a note to add an `@internal` tag for these?

3. We should add a minimal per-file header as well, something similar to this:

```c
/** PGP packet parsing utilities
 *  @file
 */
```

Developers Guide.md
```diff
@@ -354,6 +354,33 @@ Do not:
   ```
 * Do not use `pragma`, and try to avoid `__attribute__` as well.
 
+### Documentation
+Documentation is done in Doxygen comments format, which must be put in header files.
+Exception are static or not exposed to the public API functions - they should be documented in
```
flowher: As per our voting result, please remove "or not exposed to the public API functions". You can use instead "not having declaration"  or "having only definition" (generally I mean those which are not in .h for some reason).

Developers Guide.md
ni4: ok

Developers Guide.md
ni4: It's just to have a nice coloring of the function definition below :)

flowher: ah! wow didn't know this feature. cool

Developers Guide.md
ni4: Ok.

ni4:
@dewyatt, @flowher please see the updated documentation, it seems I didn't miss any of your suggestions. GitHub PR reviews/comments sometimes hard to follow.

dewyatt:
Any thoughts on dropping the param direction (`[in]/[out]`) piece?
It kind of seems redundant in C.
```c
rnp_result some_func(int abc, // [in]
                     int *def, // [out]
                     const char *s);   // [in]
```
I guess the only time it isn't clear is when you have `[in,out]`, which seems pretty rare.

Anyways, LGTM. This is sort of a policy update I guess so let's get @ronaldtse to approve.

ronaldtse:
`[in] / [out]` might not be necessary, if the param description already describes what that does. In the case of `[in, out]`, it would actually be much cleaner for a function not to have an in & out argument, but simply in or out.

ni4:
@ronaldtse  @dewyatt  Agreed on [in, out]. Removed them, so if there are no any other suggestions let's merge this.

dewyatt:
> it would actually be much cleaner for a function not to have an argument used both for in & out. Maybe we should specify this when we drop the requirement for [in] / [out].

Definitely agree, botan's FFI interface has some in/out parameters and it has puzzled me before (partially a documentation issue there though).

PR #320, opened 2017-07-20
Come up with consistent return values: Step 2

flowher:
* I've tracked all occurences RNP_EOF and carefully change functions returning those error codes to ``rnp_result``.
* After this step all functions (except rnp.h) returning RNP_OK or RNP_FAIL were automatically changed to bool functions. They were either returning 0 or 1, so this step was  safe (BTW: I'm not using ``bool`` because I like it, but more because it safe way to convert consistent return codes). So now we have either functions that return ``bool`` or ``rnp_result`` (and still some functions returning -1, 0, 1... ). But we go towards more consistent error codes.
*  I've done some research and it seems using ``typedef enum`` doesn't cause problems with ABI compatibility if we enforce the enum to be represented as int. So now ``rnp_result`` is defined as ``typedef enum {..} rnp_result`` so that declaration of type and definition of error codes is very close. To be super safe, it is checked if ``rnp_result`` is same size as ``int`` at compile time. This also addresses  (kind of?) discussion on "Why not to use just int instead of uint32_t?" by @ni4  from previous code review (see [here](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B) and interesting discussion/answer of "Bremar" [here](https://stackoverflow.com/questions/20809669/should-enum-never-be-used-in-an-api). Also ABI specs give some informations)
* ``constants.h`` is moved to ``rnp_defs.h`` and new ``rnp_types.h`` file is created to store type definitions.
* there are only two files that are using RNP_OK and RNP_FAIL macros. those macros were moved to temporary ``rnp_obsolete_defs.h`` file which will be removed in next step (please don't include this file in other files).
* some minor bugs were fixed (related to the fact that we were mixing int and unsigned int)

Hope I havn't introduced new bugs.


dewyatt:
> I've done some research and it seems using typedef enum doesn't cause problems with ABI compatibility if we enforce the enum to be represented as int.

So I think you're suggesting we force the size of the enum to equal `sizeof(int)` by adding a final enum constant equal to `INT_MAX`.

Why not just use `int`? What is the advantage to the enum typedef?

It seems we gain nothing and just hide the fact that we're using an int. Plus it's a bit of a nuisance for some language interop stuff due to the lack of a simple definite size for enums.

> To be super safe, it is checked if rnp_result is same size as int at compile time.

`static_assert` is not C99.

> This also addresses (kind of?) discussion on "Why not to use just int instead of uint32_t?" by ni4

I have to agree with @ni4, I can't see any reason to not use a simple `typedef int`. The C99 standard states that enum constants are `an integer constant expression that has a value representable as an int.`.

> (see here and interesting discussion/answer of "Bremar" here. Also ABI specs give some informations)

This assumes that compilers strictly adhere to platform ABIs, which isn't really true. This is why ABIs tend to end up with little footnotes like `some implementations of C permit enums larger than an int.`. Compilers can get a bit crazy in the embedded world, for example.

Anyways, personally I'm not seeing any technical reason that we can't do the typedef safely, but I can only see disadvantages to it.

flowher:
I don't think you have any compiler that will use other type for such enum. This would cause other problems. I know one project that uses typedefed enum to return errors [here](https://android.googlesource.com/platform/hardware/libhardware/+/master/include/hardware/keymaster2.h)

include/rnp.h
```diff
@@ -33,8 +33,10 @@
 
 #include <stddef.h>
 #include <stdbool.h>
+
 #include "packet.h"
-#include "constants.h"
+#include "rnp_defs.h"
+#include "rnp_types.h"
```
catap: Maybe use `#include <rnp_def.h>`?

flowher: Good by me. Done

ni4:
Regarding int vs uint32_t - basically, the easier type the better. But I'm leaving this discussion to you, for me int just looks slightly better but uint32_t will work as well.

flowher:
as for me it doesn't really matter if it's uint, enum or int. I don't see any technical issue with any of those. But I would definitely prefer to use typedef.

src/lib/utils.h
```diff
@@ -45,6 +48,46 @@
...
+
+#ifndef __UNCONST
+#define __UNCONST(a) ((void *) (unsigned long) (const void *) (a))
+#endif
+
+/* Portable way to convert bits to bytes */
 #define BITS_TO_BYTES(b) (((b) + (CHAR_BIT - 1)) / CHAR_BIT)
 
+/* number of elements in an array */
+#define array_size(a) (sizeof(a) / sizeof(*(a)))
```
dewyatt: Can you restore this to uppercase to fit the convention?

flowher: Done. Thanks

PR #322, opened 2017-07-20
Tests framework improvements

dewyatt:


dewyatt:
Hmm I see the libtool wrapper causes some issues with getting the correct path to the data directory. I'll see if that can be improved.

dewyatt:
This should be OK now. I'm trying to get some tests+data in that will use this, before I finish up the initial subkey stuff in the next few days here.

ronaldtse:
LGTM

PR #325, opened 2017-07-22
Add simple test for pgp_parse

dewyatt:
A few small things here:
* A generic linked list, which I initially added for something else, but found it useful here. I'll have to add some separate official tests for it in the future.
* A very simple test for `pgp_parse` and some of the supporting items.
* A .gpg keyring for testing.


dewyatt:
I think I'll end up doing some small updates on this, so maybe don't review just yet.
(btw this is tangential to subkey support, which is still my current focus)

catap:
@dewyatt Why haven't you use `LIST_*` from `sys/queue.h`?

dewyatt:
@catap 
1. Portability (we have enough work to do there already).
2. I don't really like that it requires adding structure members to whatever you want a list of. Like in the test added, I would end up creating some new structure just to hold the pgp_content_enum tag. At least that's the impression I got at a glance, but I didn't look that closely because of the above.

catap:
@dewyatt ok, I got your point. But I afraid implement stuctures likes a list because it request some time to debug it.

Anyway, if it works, it's fine :) but I preferer to use something standard likes `sys/queue.h`.

dewyatt:
@catap I totally understand and I didn't really want to reinvent the wheel but it seemed like a quick solution after looking at our options.. We just don't have anything like the STL in C++. I think the Linux and BSD versions of `sys/queue.h` are completely different and Windows has none of course, so I skipped it as an option.

Anyways, I went ahead and added some tests for it at least.

(And I think I've finished making minor changes to this PR, so feel free to review)

catap:
@dewyatt well... Linux/BSD version of `sys/queue.h` has littel differences but I'm 99% sure that both of them has same interface for list.

But anyway, your point about windows is a trump card :)

catap:
@dewyatt I see only one strange — you added secring.gpg that never used.

dewyatt:
@catap Correct, I plan to use it later,but didn't want to spend too much time on these parsing tests. I'll likely use it for some subkey tests soon.

catap:
@dewyatt Ok, I haven't see anything criminal here.

anyway, I would like re-read list implementation tomorrow or at monday morning because it's complicated and may have something :)

src/lib/list.c
```diff
@@ -0,0 +1,255 @@
...
+            free(*lst);
+            *lst = NULL;
+        }
+        return NULL;
+    }
+    if (data) {
+        memset(item, 0, sizeof(*item));
+        memcpy(get_data_ptr(item), data, data_size);
+    } else {
+        memset(item, 0, sizeof(*item) + data_size);
```
catap: I think that `list_do_insert` should return an error when it has `data_size` != 0 and `data` is `NULL`.

src/lib/list.h
```diff
@@ -0,0 +1,218 @@
...
+ *  while (item) {
+ *      printf("Found %d\n", *(int*)item);
+ *      item = list_find_next(item, &one, sizeof(one));
+ *  }
+ *  @endcode
+ */
+
+typedef struct list_head *list;
+typedef struct list_head  list_head;
+typedef struct list_item  list_item;
```
catap: maybe use `rnp_` suffix?

Also, `typedef struct list_head *list` may confuse and I think better to use `typedef struct list_head list`.

src/lib/list.c
ni4: It may be useful if somebody would like to store some NULL pointers inside, for instance to separate some sets of char*

catap: got it, thanks!

PR #326, opened 2017-07-23
Rename pgp_subpacket_t and add tag field

dewyatt:
* Rename pgp_subpacket_t to pgp_rawpacket_t. The naming on this is very misleading and it confuses me every time I see it.
This structure never had anything to do with signature subpackets, it's just a raw chunk of data. 
In OpenPGP-SDK this was called ops_packet_t, which probably made more sense. Unfortunately, we use pgp_packet_t to refer to what OpenPGP-SDK called ops_parser_content_t (parsed pkts).

* Add tag field to pgp_rawpacket_t
I have plans to use this and I think it makes sense to toss it into the structure since it's already been parsed out.

### DIFF
--- a/src/lib/key_store.c
+++ b/src/lib/key_store.c
@@ -453,7 +453,7 @@ rnp_key_store_add_key(pgp_io_t *       io,
         if (newkey->packets == NULL) {
             return false;
         }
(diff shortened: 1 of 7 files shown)


ni4:
Thanks, @dewyatt . This change was definitely needed, I was also surprised with pgp_subpacket_t while working on #256.

PR #327, opened 2017-07-23
CLI tests: added test groups and basic cli/perf tests

### LINKED ISSUE #149: Performance benchmark suite for rnp
We should have a performance benchmark suite that compares performance of rnp to GnuPG / GPGME to prevent regressions.

From @dewyatt : 
> We could also launch gpg externally as well and do tests to ensure we're generating compatible keys, things like that. Assuming it has sane exit codes, which it probably does. (Or test with gpgme, but that would probably be more effort)

ni4:
This PR is not the final set of CLI tests as in issue #294 but mostly infrastructure : it includes everything needed to later add more python CLI and performance tests, separates test by groups, and adds some example RSA key generation test with checks via --list-packets call, regexps and so on.
Personally I like how easily we can add complicated checks with python using this infrastructure.
Also it fixes #149 partially.
Created this PR now so we can discuss the infrastructure overall, before I added a lot of tests.

So now rnp_tests without parameters runs lib tests, as it was before. With cli or perf parameter it will run CLI and performance tests using the python scripts.
However this will run only in src/tests directory since python scripts are not copied to /usr/local/bin
@dewyatt after your changes with get_data_dir it will work only in src/tests anyway since realpath requires data subfolder to exist. Or did I miss something during the merge?


src/tests/rnp_tests.c
```diff
@@ -156,13 +145,132 @@ main(int argc, char *argv[])
...
     }
 
-    int ret = 0;
-    for (int i = 0; i < iteration; i++) {
-        printf("Iteration %d\n", i);
+    if (iterations == 0) {
         ret = cmocka_run_group_tests(tests, setup_test_group, teardown_test_group);
-        if (ret != 0) {
-            break;
+    } else {
```
catap: Why did you return this logic?

src/tests/rnp_tests.c
```diff
@@ -156,13 +145,132 @@ main(int argc, char *argv[])
...
+            printf("Wrong parameter: %s\n", argv[1]);
+            print_usage();
+            return 1;
+        }
+    } else if (argc > 2) {
+        printf("Wrong parameter count.\n");
+        print_usage();
+        return 1;
+    }
+
```
catap: I think you should add way to run `cli` tests over travis-ci. Because right now it runs only `lib` test.

dewyatt:
> @dewyatt after your changes with get_data_dir it will work only in src/tests anyway since realpath requires data subfolder to exist. Or did I miss something during the merge?

You should be able to run it from anywhere, with absolute or relative paths. The data directory does have to exist, but not as a subdir of the working directory or anything like that. Let me know if this does not work for you in current master, and what OS, etc. I did use an extension that I thought was fairly widely provided (`realpath` with a `NULL` second arg), maybe that's causing problems.

dewyatt:
I really think these should be separate and not integrated into the cmocka tests at all. I mean, the cmocka code here is just calling `system()` and running the appropriate python scripts anyways.

src/tests/rnp_tests.c
ronaldtse: I agree the CLI tests should be run in Travis in addition to `make check`.

ronaldtse:
Agree with @dewyatt that the CLI test suite be separate from CMocka tests.

ni4:
@dewyatt I switched back to master, using macOS at the moment. Trying to run rnp_tests from /usr/local/bin after make install, getting input to realpath: "/usr/local/bin/../data".
And realpath returns NULL since "The realpath() function will resolve both absolute and relative paths and return the absolute pathname corresponding to file_name.  All components of file_name must exist when realpath() is called.".
Running rnp_tests from Git repository rnp/src/tests works fine.

ni4:
@ronaldtse @dewyatt Ok, CLI tests could be easily separated from clocks/moved to other folder/whatever else. Anybody have good idea where to put it/how to run? Add just some python script like rnptests_cli.py /rnptests_perf.py ? And add them to travis of course.

src/tests/rnp_tests.c
ni4: Possibly it was error during the merge, but I didn't want to change anything around this. What's wrong with this logic?

dewyatt:
> @dewyatt I switched back to master, using macOS at the moment. Trying to run rnp_tests from /usr/local/bin after make install

Ahh ok I did not realize `rnp_tests` was even installed during `make install` and never considered this usage. I don't really think it should be installed at all as it's not something an end-user should be using. I would say it should only ever reside within a git repo. Plus if you stick to running it via `src/tests/rnp_tests` the libtool wrappers *should* ensure that you are running it with the libraries in that git repo, not some globally installed ones. Thoughts?

What I would propose is:

- [ ] Similar to the above, rework `find_utility` et al, to test the executables in the git repo, not those that are in PATH.
- [ ] Remove .c/.h files from this PR as we'll be better off just using python and calling it directly.
- [ ] (At some point I think we should remove `rnp_tests` from `make install`).

Anyways, that's my feeling on this.

(Also, don't worry about the travis-ci stuff, I can add that quickly)

ronaldtse:
Agree with @dewyatt on the steps. We should also add rnp_tests and the CLI test suite to `make check` In addition to Travis so we could use it in packaging tools easily.

src/tests/rnp_tests.c
catap: @ni4 this logic was removed by @dewyatt :)

ni4: Ah, I see. Most likely did it by mistake during the merge. Thanks for pointing on this. Anyway, it seems that changes to the *.c code will be reverted so only python scripts left.

ni4:
Ok, I prepare the changes in separate PR - it would be easier to drop all *.c/*.h changes and copy py files from this branch, then reverting all commits and making sure nothing went wrong.

@catap @dewyatt I created issue #328 to add `make check` target, I guess it is something we need to do? 
Unfortunately I'm not quite familiar with Makefiles/autotools, however if nobody else have time for that I can do it.

catap:
@ni4 I take care of autohell stuff ;)

catap:
@ni4 anyway you can add your python script to test framwork like I did it at #331

ronaldtse:
@ni4 should this be merged soon?

ni4:
@ronaldtse I'll create new PR for CLI tests and delete this one, hopefully later today.

ronaldtse:
Look forward to it @ni4 !

ni4:
I'm closing this PR since it was superseded by the PR #355

Issue #127, opened 2017-05-26
Support SM2 


ronaldtse:
SM2 is a family of elliptic curve algorithms described in [draft-shen-sm2-ecdsa-02](https://tools.ietf.org/html/draft-shen-sm2-ecdsa-02), and includes ECDSA and public key encryption algorithms.

This task is to implement support for SM2 in rnp (and in conjunction with SM3), and depends on implementation of SM2 within Botan.

Relevant OIDs (http://gmssl.org/docs/oid.html):

* 1.2.156.10197.1.301, SM2 Elliptic Curve Cryptography
* 1.2.156.10197.1.301.1, SM2-1 Digital Siganture Algorithm
* 1.2.156.10197.1.301.2, SM2-2 Key Exchange Protocol
* 1.2.156.10197.1.301.3, SM2-3 Public Key Encryption
* 1.2.156.10197.1.501, SM2 Signing with SM3
* 1.2.156.10197.6.1.1.3, SM2 Standard

ronaldtse:
The best public information I was able to find is this link:
http://blog.csdn.net/arlaichin/article/details/23708155

ronaldtse:
I believe this is completed, right @randombit ? Or is this not selectable from the CLI on key generation?

randombit:
@ronaldtse Encryption is not supported yet, PR for that coming today I think

ronaldtse:
@randombit should we close this?

Ref: https://github.com/riboseinc/rfc-openpgp-oscca/pull/2

ronaldtse:
@randombit can we close this now?

randombit:
@ronaldtse I believe everything is done here but let me double check.

ronaldtse:
Thanks @randombit !

PR #332, opened 2017-07-26
Add support for SM2 signatures

randombit:
For #127 SM2 keys are currently always generated using the recommended SM2 256-bit curve and SM3 hash since if someone wants SM2 keys that's probably what they want. But it is perfectly reasonable to use the signature scheme over NIST P-384 or P-521 curves as well and Botan supports that, so there is some accommodation in `sm2.c` for larger curves. (But not in the `rnpkeys` UI, which hardcodes the curve and hash).

I defined SM2 using PKA=99 instead of in the reserved 100-110 range because `packet-print.c` and co already "support" that range. So adding SM2 at 100 would require first removing the handling for the reserved codepoint, then adding it back later on if SM2 moved outside that range to an official value. However I'm not sure if any of the code for handling the PKA_RESERVED codepoints makes any sense - all it is doing is explicitly saying "I don't support this" but that  same codepath should be followed for any unknown/unsupported algo not special-cased to just algos in the reserved range. There isn't any reason I can think of to treat an unknown algo with PKA=105 any differently from an unknown algo with PKA=250. So I'd vote to just remove the code+enums for the reserved ranges entirely.


src/lib/signature.c
```diff
@@ -249,6 +250,41 @@ ecdsa_sign(pgp_hash_t *            hash,
...
+    if (!pgp_write(output, &hashbuf[0], 2))
+        return false;
+
+    /* write signature to buf */
+    if (pgp_sm2_sign_hash(&sig, hashbuf, hashsize, prv_key, pub_key) != PGP_E_OK) {
+        return false;
+    }
+
+    /* convert and write the sig out to memory */
+    bool ret = !!pgp_write_mpi(output, sig.r);
```
dewyatt: We shouldn't really need the `!!` idiom since 1) ret is a _Bool and the compiler will emit instructions that handle this and 2) `pgp_write_mpi` currently returns a _Bool.

src/lib/sm2.c
```diff
@@ -0,0 +1,271 @@
...
+
+    if (botan_privkey_create(&pr_key, "SM2_Sig", ec_curves[curve].botan_name, rng)) {
+        goto end;
+    }
+
+    if (botan_privkey_export_pubkey(&pu_key, pr_key)) {
+        goto end;
+    }
+
+    // Crash if seckey is null. It's clean and easy to debug design
```
dewyatt: I feel like we ought to start adding some debug assertions for things like this, rather than crashing. But that's a conversation for another day.

PR #334, opened 2017-07-29
Add some supporting signature functions.

dewyatt:
These are just some supporting functions, this PR doesn't add anything that actually uses them.
I have a few thousand lines of code that add subkey loading+generation, better key flag support, and some user preferences support. At this point it's probably easiest to just pick it apart and add the supporting pieces first, since it certainly won't merge.

(And I'm not planning to add any docs at this time, but will when time permits)


ronaldtse:
I like how the relevant function names are now grouped under `pgp_sig_` too.

PR #337, opened 2017-07-29
Misc keyring fixes

dewyatt:
This is *mostly* fixes for stuff inherited from netpgp, nothing too controversial.

### DIFF
--- a/src/lib/crypto.h
+++ b/src/lib/crypto.h
@@ -63,6 +63,7 @@
 #include "packet-parse.h"
 #include "symmetric.h"
 #include "bn.h"
+#include <rekey/rnp_key_store.h>
 
 #define PGP_MIN_HASH_SIZE 16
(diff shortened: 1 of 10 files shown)


src/lib/crypto.h
```diff
@@ -63,6 +63,7 @@
 #include "packet-parse.h"
 #include "symmetric.h"
 #include "bn.h"
+#include <rekey/rnp_key_store.h>
```
catap: I have no idea why you include this header. Anyway, it is just style-remark, so, you can ignore it.

dewyatt: `crypto.h` uses `rnp_key_store_t`, the only reason that include wasn't in there before is because it was indirectly pulled in from `packet.h` (which included it for `pgp_parse_and_accumulate`, which I removed).
:)

catap: @dewyatt I see. Anyway from this commit it wasn't clear ;)

PR #342, opened 2017-07-30
Abstract key generation a bit

dewyatt:
Depends on: #341 

This is an intermediate step I took when working on subkey generation (with some updates).

The last two commits are the ones relevant for this PR.
(Actually the middle commit is sort of unrelated, but is used in some future fixes)

The diff is kind of hard to read, so you're probably better off just checking out `pgp_generate_seckey` and `pgp_generate_keypair` in https://github.com/riboseinc/rnp/blob/efe2cad0aabacde3b88bf366efef60e5ccbe7fc5/src/lib/crypto.c

### DIFF
--- a/src/lib/crypto.c
+++ b/src/lib/crypto.c
@@ -234,12 +234,118 @@ pgp_elgamal_encrypt_mpi(const uint8_t *          encoded_m_buf,
     return true;
 }
 
+bool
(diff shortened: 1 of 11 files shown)


ronaldtse:
Does look cleaner to me, thanks @dewyatt

PR #343, opened 2017-07-31
Removes duplicated rnp.h

flowher:
``rnp.h`` file was duplicated. One was in the ./include and and another one in ./include/rnp/. I think it is safe to remove the one in ./include, isn't it?

When doing ``make install``, I'm getting
```
/usr/bin/install: will not overwrite just-created '/usr/local/include/rnp.h' with 'rnp/rnp.h'
make[2]: *** [Makefile:358: install-includeHEADERS] Error 1
``` 

Hence the PR


flowher:
Nop, I'm wrong!

Issue #25, opened 2017-03-26
Adopt RFC4880bis-01 draft


ronaldtse:
Need to describe the implementation gaps with the [RFC4880bis-01 draft](https://tools.ietf.org/html/draft-ietf-openpgp-rfc4880bis-01), and figure out priority to implement them.

ni4:
Not relevant for a while.

PR #344, opened 2017-08-01
ECDH support

flowher:
This patch implements support for encryption with ECDH as specified in RFC 4880 bis 01.

Key can be generated with ``rnpkey --generate-key -expert``. Key can use any of NIST curves : P-256, P-384 or P-521.
Encryption/decryption can be done as usual with ``rnp --encrypt/--decrypt``. To do round trip one may use:
```
rnp --encrypt file.in --output file.out
rnp --decrypt file.out --output file.res

diff file.in file.res
```

Nevertheless, one more stepis needed in the development to support compatibility with ``gpg``. Currently we have 2 problems:
* GnuPG always expects that ECDH key is a subkey of ECDSA key. As it is not a case currently in ``rnp`` (and not part of RFC 4880 spec) it means we can't (yet) import key generated by ``rnp`` to ``gpg``. 
* There is also a problem in ``rnp`` when encryption key is a subkey. When we have a situation when master key is signature key and encryption key is a subkey of this key (like ECDSA and ECDH as subkey), then rnp wrongly chooses  ECDSA key to do encryption.

Those both problems require additional development and will be fixed in separated step.

### DIFF
--- a/src/lib/ecdh.c
+++ b/src/lib/ecdh.c
@@ -214,10 +214,10 @@ pgp_ecdh_encrypt_pkcs5(const uint8_t *const     session_key,
     // See 13.5 of RFC 4880 for definition of other_info_size
     const size_t other_info_size = (pubkey->ec.curve == PGP_CURVE_NIST_P_256) ? 54 : 51;
     const size_t key_len = BITS_TO_BYTES(curve_desc->bitlen) * 2 + 1;
-    const size_t kek_len = pgp_key_size(pubkey->kdf.wrap_alg);
+    const size_t kek_len = pgp_key_size(pubkey->key_wrap_alg);
 
     size_t tmp_len = kdf_other_info_serialize(
-      other_info, curve_desc, fingerprint, pubkey->kdf.hash, pubkey->kdf.wrap_alg);
+      other_info, curve_desc, fingerprint, pubkey->kdf_hash_alg, pubkey->key_wrap_alg);
 
     if (tmp_len != other_info_size) {
         RNP_LOG("Serialization of other info failed");
@@ -240,7 +240,7 @@ pgp_ecdh_encrypt_pkcs5(const uint8_t *const     session_key,
                      curve_desc,
                      pubkey->ec.point->mp,
                      eph_prv_key,
-                     pubkey->kdf.hash)) {
+                     pubkey->kdf_hash_alg)) {
         RNP_LOG("KEK computation failed");
         goto end;
     }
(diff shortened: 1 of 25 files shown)


flowher:
xref: #27 #25 #18

flowher:
hmm... job fails with no reason when building botan

dewyatt:
> hmm... job fails with no reason when building botan

Two separate issues with the build, should be good now.

> There is also a problem in rnp when encryption key is a subkey. When we have a situation when master key is signature key and encryption key is a subkey of this key (like ECDSA and ECDH as subkey), then rnp wrongly chooses ECDSA key to do encryption.

I have some fixes for this as part of my work on subkeys and key flags, PR coming soon.

flowher:
Thanks !

ronaldtse:
@flowher minor conflicts, could we fix this up and merge it? OK with @dewyatt and @randombit ?

flowher:
@ronaldtse  Fully aware. It's minor stuff, I'm rebasing and preparing next commit

ronaldtse:
We're pending one more review to merge. Thanks!

randombit:
All looks good to me. OpenPGP's use of PKCS5 padding inside the ECDH keywrapping seems kind of weird to me - even with the length hidden doesn't an attacker know exactly what the keylength must be, thanks to the outer algorithm id? So obfuscating it doesn't seem helpful. Anyway +1 on writing the unpadding code in a const time way. Re the comment, is always hard to be sure what a compiler will or won't do, but I checked and gcc 7.1 on x86-64 does compile `unpad_pkcs7` without input-depdendent loops. But Clang 4.0 vectorizes the function O_o and inserts a number of input-dependent jumps. It may be hard to fix this in a portable way. I checked how `gpg` handles this situation ... it just doesn't check the padding bytes, looks at the last byte of the data and strips off so many bytes. Oops. (This is canonically *not* the right way to do that...)

src/lib/ecdh.c
```diff
@@ -0,0 +1,370 @@
...
+}
+
+static bool
+unpad_pkcs7(uint8_t *buf, size_t buf_len, size_t *offset)
+{
+    uint8_t        err = 0;
+    const uint8_t  pad_byte = buf[buf_len - 1];
+    const uint32_t pad_begin = buf_len - pad_byte;
+
+    if (!buf || !offset || !buf_len) {
```
dewyatt: These checks are a bit late in the game

flowher: argh... thnx. will fix

flowher:
@randombit thanks for review. Indeed, that's good point - seems like kind of bug in a spec. Algo used for session key is known, so... hmm.. no idea. Need to look at IETF mail archive, but indeed I don't see what's a value of PKCS5.
Regarding constant time  (damn, that's interesting topic). I think it was BearSSL which had something for <, >, == etc. I would like to look at it more for my personal interest. P. Schwabe made me recently realize that even if you have constant time C code, CPU may screw it because you can't assume that absolutely all instructions are constant time (especially on ARM). I'm not expert with it, just trying my best (hoping to be one day).

randombit:
@flowher IIRC BoringSSL also has some code to handle comparisons in a const-time way that might be worth checking out.

PR #354, opened 2017-08-05
Subkey generation / rework

dewyatt:
EDIT: I'm continuing to split some of these fixes in to separate PRs.

I don't know if this will end up merged, as it only really works with GPG/KBX keystore formats (no G10/GPG21). I've had to rework this branch too many times with all the recent PRs, so I figure I better at least get input.

It also reworks the key listings to match gpg more.

```
$ src/rnpkeys/rnpkeys --gen-key --keystore-format=GPG
Generating a new key...
Enter passphrase for new primary key: 
Repeat passphrase for new primary key: 
Enter passphrase for new subkey: 
Repeat passphrase for new subkey: 
pub 2048/RSA (Encrypt or Sign) a1c91f6a6800dc5d 2017-08-05 [SC] 
                 aea9 1b0a ec72 2023 cd4f 1666 a1c9 1f6a 6800 dc5d 
uid              RSA (Encrypt or Sign) 2048-bit key <daniel@localhost>
sub 2048/RSA (Encrypt or Sign) 90ef443358269873 2017-08-05 [E] 
                 bb4d b248 66f4 d76e dbe6 a738 90ef 4433 5826 9873 
```

```
$ src/rnpkeys/rnpkeys --list-keys --keystore-format=GPG
2 keys found

pub   2048/RSA (Encrypt or Sign) a1c91f6a6800dc5d 2017-08-05 [SC]
      aea91b0aec722023cd4f1666a1c91f6a6800dc5d
uid           RSA (Encrypt or Sign) 2048-bit key <daniel@localhost>
sub   2048/RSA (Encrypt or Sign) 90ef443358269873 2017-08-05 [E]
      bb4db24866f4d76edbe6a73890ef443358269873

$ src/rnpkeys/rnpkeys --list-sigs --keystore-format=GPG
2 keys found

pub   2048/RSA (Encrypt or Sign) a1c91f6a6800dc5d 2017-08-05 [SC]
      aea91b0aec722023cd4f1666a1c91f6a6800dc5d
uid           RSA (Encrypt or Sign) 2048-bit key <daniel@localhost>
sig           a1c91f6a6800dc5d 2017-08-05 RSA (Encrypt or Sign) 2048-bit key <daniel@localhost>
sub   2048/RSA (Encrypt or Sign) 90ef443358269873 2017-08-05 [E]
      bb4db24866f4d76edbe6a73890ef443358269873
```

### Pros
* Full key+subkey generation. I haven't actually tested "only primary key" generation or "only subkey generation", but generating both works well, is compatible with gpg, etc.
* Fixes round-trip reading+writing out unencrypted .gpg/.asc keys. (We still can't generate and write these, but that may not matter)
* Fixes incorrect transferable key writing. Transferable keys are only supposed
  to contain certain packets.
* Fixes all key MPIs leaking on key store free (probably inherited from netpgp)
### Cons
* As mentioned, this really only supports .gpg/.asc/.kbx. (And since KBX is never used without G10 with GPG, KBX alone doesn't count for much)
* rnpkeys expert mode would need some work.
* To save time, I used a lot of bools and don't return error codes back to callers (I did use RNP_LOG a lot though)

### DIFF
--- a/include/rekey/rnp_key_store.h
+++ b/include/rekey/rnp_key_store.h
@@ -159,7 +159,7 @@ bool rnp_key_store_json(pgp_io_t *, const rnp_key_store_t *, json_object *, cons
 
 bool rnp_key_store_append_keyring(rnp_key_store_t *, rnp_key_store_t *);
 
-bool rnp_key_store_add_key(pgp_io_t *, rnp_key_store_t *, pgp_key_t *, pgp_content_enum);
+bool rnp_key_store_add_key(pgp_io_t *, rnp_key_store_t *, pgp_key_t *);
 bool rnp_key_store_add_keydata(
   pgp_io_t *, rnp_key_store_t *, pgp_keydata_key_t *, pgp_key_t **, pgp_content_enum);
 
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -80,7 +80,7 @@ bool  rnp_find_key(rnp_t *, const char *);
 char *rnp_get_key(rnp_t *, const char *, const char *);
 char *rnp_export_key(rnp_t *, const char *);
 int   rnp_import_key(rnp_t *, char *);
-int   rnp_generate_key(rnp_t *, const char *);
+int   rnp_generate_key(rnp_t *);
 int   rnp_secret_count(rnp_t *);
 int   rnp_public_count(rnp_t *);
 
--- a/src/lib/Makefile.am
+++ b/src/lib/Makefile.am
@@ -16,6 +16,7 @@ librnp_la_SOURCES	= \
 	ecdsa.c \
 	eddsa.c \
 	elgamal.c \
+	generate-key.c \
 	hash.c \
 	list.c \
 	misc.c \
--- a/src/lib/crypto.c
+++ b/src/lib/crypto.c
@@ -494,8 +494,11 @@ Encrypt a file
 \return true if OK
 */
 bool
-pgp_encrypt_file(
-  rnp_ctx_t *ctx, pgp_io_t *io, const char *infile, const char *outfile, const pgp_key_t *key)
+pgp_encrypt_file(rnp_ctx_t *         ctx,
+                 pgp_io_t *          io,
+                 const char *        infile,
+                 const char *        outfile,
+                 const pgp_pubkey_t *pubkey)
 {
     pgp_output_t *output;
     pgp_memory_t *inmem;
@@ -523,7 +526,7 @@ pgp_encrypt_file(
     }
 
     /* Push the encrypted writer */
-    if (!pgp_push_enc_se_ip(output, key, ctx->ealg)) {
+    if (!pgp_push_enc_se_ip(output, pubkey, ctx->ealg)) {
         pgp_memory_free(inmem);
         return false;
     }
@@ -540,11 +543,11 @@ pgp_encrypt_file(
 
 /* encrypt the contents of the input buffer, and return the mem structure */
 pgp_memory_t *
-pgp_encrypt_buf(rnp_ctx_t *      ctx,
-                pgp_io_t *       io,
-                const void *     input,
-                const size_t     insize,
-                const pgp_key_t *pubkey)
+pgp_encrypt_buf(rnp_ctx_t *         ctx,
+                pgp_io_t *          io,
+                const void *        input,
+                const size_t        insize,
+                const pgp_pubkey_t *pubkey)
 {
     pgp_output_t *output;
     pgp_memory_t *outmem;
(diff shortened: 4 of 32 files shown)


dewyatt:
I guess I'll fix the build once this 3GB docker image downloads (sheesh).
I think I see what I botched there.

dewyatt:
Ok it turns out I just got lucky and uncovered an unrelated issue (#356).

Anyways...I should mention I do plan to see if I can pull some of these fixes out into separate PRs.
Some of them can definitely be pulled out, but I know a few of them can't be without major changes to master.

dewyatt:
CI will fail until #364 is fixed and included here, since I check the return values of pgp_write_sig in this branch.

ronaldtse:
Looks like we're ready to go! @dewyatt could you help rebase this so we can fast-forward the commits? Or are you splitting in to PRs?

dewyatt:
@ronaldtse I'm wondering if it might make sense to wait on #368 and then I can look at whether I can get G10/GPG21 support re-enabled in this branch before merge (I've disabled some or all tests that use that keystore format in this branch, and I don't like to do that of course).

(On splitting it up: I think it's probably been split apart as much as it can be, everything remaining is likely tightly coupled by necessity.)

ronaldtse:
@dewyatt It would be great if we can get #368 solved before, it would just require your review once more now that @catap is using higher level functions in botan 😉

catap:
@ronaldtse also I think we should simplified `symmetric.c` if we switched to use high-level botan functions.

ronaldtse:
@catap are there high-level botan functions for `symmetric.c`? I'm surprised since @randombit implemented that...

catap:
@ronaldtse for example we have `pgp_cipher_cfb_encrypt` and `pgp_cipher_cfb_decrypt` that is low-level API and botan has high-level API for this.

I'm not 100% sure that we can switched here to botan implementation but for me it looks so.

dewyatt:
I would bet that Jack used the lower-level `botan_block_cipher_` interface for CFB because of the CFB resync quirkiness in the RFC (the RFC kind of has its own variant of CFB I believe).

catap:
@dewyatt maybe, let ask @randombit !

flowher:
Indeed, the CFB for openpgp is a bit (different)[https://eprint.iacr.org/2005/033.pdf], because of this resyncing at each decryption/encryption. So I wouldn't be totally surprised if we need treat it in special way.

ni4:
Agree with @flowher - OpenPGP uses a specific CFB mode which is not used anywhere else as far as I remember. So we'll need our custom implementation or ask @randombit to implement it as part of botan. Anyway I don't see any problem in this - it works, and doesn't use much code lines.

ronaldtse:
@ni4 @flowher I think we should just keep this CFB mode in rnp instead of Botan which is a generic cryptographic library, given all the problems we know of this mode.

ronaldtse:
@dewyatt since this PR is good can we rebase and merge it? It already has the required number of approvals. Don't want to carry it for too long. Thanks!

catap:
@flowher thanks. But I feel that I'm still misunderstanding something.

I see the different between OpenPGP's CBC and standard CBC from your articles.

But [here](https://github.com/riboseinc/rnp/blob/d08f532696646c726b6a2dc0a6386502b41bf82d/src/lib/symmetric.c#L130) I see:
```c
int
pgp_cipher_cfb_encrypt(pgp_crypt_t *crypt, uint8_t *out, const uint8_t *in, size_t bytes)
{
    for (size_t i = 0; i < bytes; ++i) {
        if (crypt->num == 0) {
            botan_block_cipher_encrypt_blocks(
              crypt->block_cipher_obj, crypt->iv, crypt->iv, 1);
        }
        out[i] = in[i] ^ crypt->iv[crypt->num];
        crypt->iv[crypt->num] = out[i];

        crypt->num = (crypt->num + 1) % crypt->blocksize;
    }
    return 0;
}
```
and it looks for me as 100% standard CBC.

Am I wrong?

dewyatt:
> @dewyatt since this PR is good can we rebase and merge it? It already has the required number of approvals. Don't want to carry it for too long. Thanks!

@ronaldtse Sure I'll get this done momentarily. I do have some small additions and tests on this locally but will just do a separate PR. I'll get this rebased today and will do a last check on any issues with recent commits in master before merge.

flowher:
@catap I think problem is ``pgp_cipher_cfb_resync``.

flowher:
@ronaldtse This CFB mode is specific to openpgp (I don't think this variations is used widely somewhere else), so keeping it in ``rnp`` makes sense for me. One thing we definitely should do is to add to rnp is some testing (KAT + negative tests). Not sure if it is easy to find test vectors for this mode, but I think BoncyCastle (https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/crypto/modes/PGPCFBBlockCipher.java) and cryptography.io supports this implementation so we can generate vectors somehow.

Issue #353, opened 2017-08-05
Decryption of GPG encrypted file fails for 8-byte block size ciphers


ni4:
## Description

Trying to encrypt the file with IDEA, 3DES, CAST5, BLOWFISH makes RNP fail with error message attached below. All of them have 8 byte block size, and all 16-byte ciphers work fine.
It could be something simple like disabled ciphers in botan, however it could be related to the block size.

## Steps to Reproduce

1. `gpg -e --cipher-algo CAST5 sample.txt`
2. `rnp --decrypt sample.txt.gpg`

## Expected Behavior
File is decrypted successfully

## Actual Behavior

```
packet-parse.c:2498: Unknown, Packet was not consumed
packet-parse.c:424: PGP_E_R_READ_FAILED, Read failed
packet-parse.c:3338: PGP_E_P_UNKNOWN_TAG, Unknown content tag 0xa
```

ni4:
@dewyatt @randombit Maybe you have quick thoughts on this?

dewyatt:
@ni4 Not able to reproduce on master, can you post the file? (`cat sample.txt.gpg | gpg --enarmor`)

ronaldtse:
@ni4 we should have this tested in the CLI tests so we know for sure what exactly is failing and under what condition it fails?

ni4:
@ronaldtse If it works on @dewyatt system then it is probably related to macOS  or my setup.
Command lines to quickly reproduce it (given that rnp and gpg are installed):
```
mkdir temp
cd temp
mkdir .rnp
mkdir .gpg
rnpkeys --homedir .rnp --generate-key --userid=cast5
echo "Hello, world" > hello.txt
gpg --homedir .gpg --import .rnp/pubring.gpg
gpg --homedir .gpg --cipher-algo CAST5 -e -r cast5 hello.txt
rnp --homedir .rnp --decrypt hello.txt.gpg
```

ronaldtse:
@ni4 I know you're working on the PR for the CLI test suite, could you incorporate this test in there so it would be tested in Travis too? Thanks!

ni4:
@ronaldtse Sure, it is already there. I'm testing all symmetric algorithms while encrypting with GPG side, and when it would be available as option - do the same from the RNP side.

ni4:
This issue was fixed in some commit which did not reference it, so I'm closing it now.

PR #355, opened 2017-08-05
Standalone CLI python tests

ni4:
As it was discussed in the previous PR #327 we agreed to have CLI tests as standalone python script(s).

This pull request resolves this by adding two python scripts in /src/tests:
`./cli_tests.py` - script which should be run with `[rnp | rnpkeys | all]` parameter, and perform testing of the corresponding utility. Now the following tests are available:
- RSA key generation, import/export to GPG
- encryption/decryption against GPG
- signing/verification against GPG
All tests use different options (symmetric algos, key bits, compression, etc), however they would be extended soon as well as number of tests.

`./perf_tests.py` - some basic at the moment test which compares performance of RNP against the GPG for the small and large file encryption. It will be extended later since now more focus was on `cli_tests.py`

Both scripts should be run from `/src/tests/`


dewyatt:
I think I managed to run into that issue you mentioned in #353 while running cli_tests.py.

Anyways, my first impression is that this looks great!

I guess there's some comparisons that maybe aren't valid yet, like rnp doesn't support compression, etc. But it's pretty nice to have the support there for that comparison.

src/tests/cli_tests.py
```diff
@@ -0,0 +1,454 @@
+#!/usr/bin/python
```
dewyatt: On some newer OSes python == python3, so I usually go with `#!/usr/bin/env python2` (or even `#!/usr/bin/python2`).
Is this OK or does it cause any issues on macOS?

ni4: On macOS I have only `/usr/bin/python2.7` and `/usr/local/bin/python2`. So, maybe, when integrating this to Travis we should better run it as 'python2 cli_tests.py all'? Or there are other ideas how to make it to work everywhere.

ni4:
@dewyatt Hm, `rnp` at least supports decompression. Also `rnp` by default compresses data (and we actually don't have a way to enable/disable compression or change the compression level yet):
```
# off=0 ctb=c1 tag=1 hlen=3 plen=268 new-ctb
:pubkey enc packet: version 3, algo 1, keyid D78F04BE3AF65209
	data: [2048 bits]
# off=271 ctb=d2 tag=18 hlen=2 plen=71 new-ctb
:encrypted data packet:
	length: 71
	mdc_method: 2
# off=284 ctb=c8 tag=8 hlen=2 plen=36 new-ctb
:compressed packet: algo=2
# off=287 ctb=cb tag=11 hlen=2 plen=27 new-ctb
:literal data packet:
	mode b (62), created 1502000813, name="hello.txt",
	raw data: 12 bytes
```

src/tests/cli_tests.py
ronaldtse: @ni4 `#!/usr/bin/env python2` works with my macOS which has Python 2 installed via `brew`.

ronaldtse:
Thank you @ni4 ! This is what we really need and gives a good framework for adding more tests.

dewyatt:
@ni4 I just meant that we don't have comparable compression support, like selecting which alg, varying compression levels, etc.

PR #356, opened 2017-08-06
tests: Fix random failures in ecdsa_signverify_success.

dewyatt:
The message was uninitialized garbage and when it happened to be
all zeroes, the signing process would (rightfully) fail in botan.

### DIFF
--- a/src/tests/cipher.c
+++ b/src/tests/cipher.c
@@ -362,7 +362,6 @@ void
 ecdsa_signverify_success(void **state)
 {
     rnp_test_state_t *rstate = *state;
-    uint8_t           message[64];
 
     struct curve {
         pgp_curve_t id;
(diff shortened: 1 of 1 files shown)


ronaldtse:
@dewyatt great find -- I'm think it would make sense to insert a comment on why we're initializing to `{0xAA}` there?

dewyatt:
Actually I wonder if this is a bug in botan (or an undocumented limitation). AFAICT ECDSA ought to be able to reliably sign&verify a digest of all zeros (however impossibly rare that may be). I'll file a bug and see what comes of it (regardless, this PR is still a valid fix).

ronaldtse:
@randombit could probably take a look here, not being able to sign all zeros might not be an intended feature.

flowher:
It was deliberately uninitialized. There is no reason why botan would not sign a garbage.

dewyatt:
> It was deliberately uninitialized. There is no reason why botan would not sign a garbage or zeros (seems like a bug in botan?). I would prefer we don't merge this PR.

I would *really* prefer not having any tests relying on uninitialized memory. I know it's a source of randomness, but it's just plain lazy when we have botan's RNG and other sources that are actually guaranteed to be random.

Now that I realize it's being used as a source of random data though, I do agree on not merging this and instead implementing something that does not used uninitialized memory.

dewyatt:
Jack did confirm that this is likely a bug in botan.

Anyways, I feel *very* strongly that using uninitialized memory in tests is not worth the tradeoffs of dealing with the mass of valgrind errors. We should be using fixed test vectors and/or RNG sources (ideally both).

Closing this and will file an issue.

PR #357, opened 2017-08-06
Fix EDDSA random failures.

### LINKED ISSUE #300: rnp_test_eddsa fails after numerous runs
## Description

After running 100-1000 iterations of rnp_test_eddsa it fails.
Most likely this depends on some parameters of randomly generated key - leading/final zeroes, high bits or so on.

## Steps to Reproduce

1. Checkout the branch ni4-eddsa-test-failure
2. make, make install, rnp_tests

## Expected Behavior

All 10000 iterations of rnp_test_eddsa run correctly.

## Actual Behavior

Test fails randomly after 100-1000 iterations.

dewyatt:
Fixes #300 

@flowher I think you were looking at this but I ran into it while looking at this other thing, so I just went ahead. Can you double check the logic here? It looks like it's just the same issue as that r/s BN<->array conversion.

### DIFF
--- a/src/lib/eddsa.c
+++ b/src/lib/eddsa.c
@@ -134,7 +134,7 @@ pgp_eddsa_sign_hash(BIGNUM *                r,
     botan_pk_op_sign_t sign_op = NULL;
     botan_rng_t        rng = NULL;
     int                result = -1;
-    uint8_t            bn_buf[64];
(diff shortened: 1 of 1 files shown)


ronaldtse:
LGTM and thanks for finding and fixing such a obscure bug!

Issue #251, opened 2017-06-25
Single instance of RNG


flowher:
## Description

RNP performs call to ``botan_rng_init`` function in many places of the code in order to initialize source of random numbers. This isn't really needed - it impacts code clarity, causes memory leaks when RNG object is not destroyed correctly and may impact performance.  

I propose ``rnp`` initializes RNG only once and makes it available via some sort of handle - global, part of a ``rnp_t`` structure or ``pgp_random`` function (solution must be thread safe). 

If rnp uses  some sort of DRBG (like HMAC_DRBG) then read from device like ``/dev/urandom`` can be limited to one and done only when DRBG is seeded (also reseeding must be taken into account).

randombit:
Sounds good to me. Re performance, it should not be so much an issue because currently `botan_rng_init` is always called with NULL param, which means it defaults to reading `/dev/urandom`, and Botan uses a single persistent handle to the system RNG among all such RNG objects. However if rnp wanted to use a DRBG, the seeding overhead would be quite high due to the repeated initializations so there sharing a single object would be a big win.

Seems better to have the RNG be part of the `rnp_t` structure, since a global would cause problems for applications that use rnp from multiple threads.

Another possible advantage to using a DRBG instead of the system PRNG - it is possible then to derive private keys deterministicly from a passphrase (by using the passphrase as the DRBG seed, and running the key generation algo). This can be useful for brainwallet-type scenarios where a user wants to be able to recreate a private key later on without storing it anywhere. Actually at one point I tried to hack this functionality into `gpg` but the code did not make it easy to do.

ronaldtse:
@randombit key generation via passphrase is definitely very useful especially when provided by a lib/CLI using a DRBG. Any interested in getting this moving?

flowher:
I'm interested in it. after ecdh is done

ronaldtse:
It's all yours @flowher !

ronaldtse:
@flowher sounds like we finally can get this going 👍

PR #363, opened 2017-08-06
Tests: Introduces helper function for getting random data in efficient way

flowher:
* `get_random`:
	* On a first call: Initializes global context for botan HMAC_DRBG
	  which is seeded with random source from system
	* retrieves some random numbers

* `destroy_global_rng` Is used to destroy global object. Function
  is called after all tests are finished

* Reduces number of false-possitive warrnings reported by
  valgrind from 47K to 0 when calling ecdsa_signverify_success test


flowher:
#251

ronaldtse:
LGTM, @dewyatt any comments?

src/tests/support.c
```diff
@@ -405,3 +410,33 @@ set_default_rsa_key_desc(rnp_keygen_desc_t *key_desc, pgp_hash_alg_t hashalg)
     key_desc->crypto.rsa.modulus_bit_len = 1024;
     key_desc->crypto.hash_alg = hashalg;
 }
+
+bool
+get_random(uint8_t *data, size_t len)
+{
+    bool ret = false;
+
+    static bool is_rng_initialized = false;
```
dewyatt: I think it's better to just make this a global up there next to `global_rng`, so that we're not hiding non-local state here (plus we have a rule in the dev guide about non-const static locals).

src/tests/support.c
```diff
@@ -405,3 +410,33 @@ set_default_rsa_key_desc(rnp_keygen_desc_t *key_desc, pgp_hash_alg_t hashalg)
...
+    }
+    ret = true;
+end:
+    return ret;
+}
+
+void
+destroy_global_rng()
+{
+    (void) botan_rng_destroy(global_rng);
```
dewyatt: Not a bad idea to add `global_rng = NULL;` just to be thorough.

dewyatt:
LGTM other than the two minor things.

src/tests/support.c
flowher: actually flag is not needed

dewyatt: That's a good point, you can just check for NULL

flowher:
Thnx for review

Issue #364, opened 2017-08-06
pgp_ecdsa_sign_hash fails certain digest/curve combinations


dewyatt:
## Description

This test would actually be failing currently if we were checking the return value of `pgp_sig_write` in a certain place (PR coming), but it goes undetected because we fail to check that return value (though botan complains).

From what I can see, the logic here is causing botan to fail: https://github.com/riboseinc/rnp/blob/20cb9ad367cf414b9678b703b84a82fef38f634b/src/lib/ecdsa.c#L231
Botan wants to write the full 64 bytes, while this indicates a buffer size of 48 for this particular test.
Or something like that, I didn't look very closely.

## Steps to Reproduce

Look for the output `in botan_pk_op_sign_finish exception Invalid argument Encoding error: encode_1363: n is too large to encode properly` during test runs.

flowher:
'r' and 's' is a value modulo multiplicative order of a base point. On the first sight code in rnp seems to be OK. 
Thanks for reporting it - I'll investigate that further.

flowher:
Code in ``rnp`` seems to be correct - all tests work correctly when botan is compiled with rfc6979 module disabled (``./configure --disable-mod=rfc6979``). Short term it could be a solution, nevertheless I would prefer we don't disable this module in long term.

I need to do quick review of RFC6979 before proposing patch to botan.

ronaldtse:
Thanks @flowher. @randombit any thoughts on this one?

flowher:
I'm working on a fix

flowher:
hmmm... It might also be that ``botan`` expects that message (digest) to be signed can't be longer than ``Signature_with_EMSA::max_input_bits()`` and then it means rnp needs to cut off few bytes from digest before signing.

flowher:
Patch proposed https://github.com/randombit/botan/pull/1147. Not sure it's preferred way but seems OK according to spec, tests are OK and ... let's keep ball rolling.

flowher:
Ok, finally after looking at botan documentation about PK_Signer - I think it's a bug in a ``rnp``. Rnp calls botan after digest of the message to sign has been produced - so we need to use "Raw" signing (we can't use EMSA). In this mode "if the message is too big, the signing operation will fail." So RNP needs to make sure that length of the digest is correct.

PR #365, opened 2017-08-06
Check return value of pgp_sig_write.

dewyatt:
This should fail in CI as it depends on https://github.com/riboseinc/rnp/issues/364 being fixed first.

### DIFF
--- a/src/lib/pgp-key.c
+++ b/src/lib/pgp-key.c
@@ -577,12 +577,13 @@ pgp_add_rawpacket(pgp_key_t *key, const pgp_rawpacket_t *packet)
 bool
 pgp_add_selfsigned_userid(pgp_key_t *key, const uint8_t *userid)
 {
-    struct pgp_create_sig_t *sig;
+    struct pgp_create_sig_t *sig = NULL;
     pgp_rawpacket_t          sigpacket;
     struct pgp_memory_t *    mem_userid = NULL;
     pgp_output_t *           useridoutput = NULL;
     struct pgp_memory_t *    mem_sig = NULL;
     pgp_output_t *           sigoutput = NULL;
+    bool                     ok = false;
 
     /*
      * create signature packet for this userid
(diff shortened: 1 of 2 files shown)


flowher:
LGTM. Code is correct and I would propose comment failing test and merge this change.

src/lib/pgp-key.c
```diff
@@ -604,10 +605,13 @@ pgp_add_selfsigned_userid(pgp_key_t *key, const uint8_t *userid)
     pgp_sig_end_hashed_subpkts(sig);
 
     if (!pgp_setup_memory_write(NULL, &sigoutput, &mem_sig, 128)) {
-        (void) fprintf(stderr, "can't setup memory write\n");
+        RNP_LOG("failed to set up memory write");
+        goto end;
+    }
+    if (!pgp_sig_write(sigoutput, sig, &key->key.seckey.pubkey, &key->key.seckey)) {
+        RNP_LOG("failed to write sig");
         return false;
```
flowher: Please also goto end here

flowher: Should go to end and free memory before leaving

dewyatt: Good catch

dewyatt:
> I would propose temporarily comment failing test and merge this change (after it is fixed). Please don't remove failing test as it also correct (and useful).

I was thinking we could just leave this open and you could push to this branch when you've had time to figure a fix, that way we would have CI tests passing before merging.

flowher:
Ok, works for me. I'll be looking at it today, so should be done rather quickly

flowher:
Code in ``rnp`` seems to be correct - all tests work correctly when botan is compiled with rfc6979 module disabled (``./configure --disable-mod=rfc6979``). Short term it could be a solution, nevertheless I would prefer we don't disable this module in long term.

I need to do quick review of RFC6979 before proposing patch to botan.

ronaldtse:
Now that #364 is fixed should this be merged @dewyatt @flowher ?

dewyatt:
Rebased on master and all tests are passing in CI, so IMO it's good to merge.

flowher:
Seems this works after #370. I propose we merge it now

ronaldtse:
Thank you guys!

PR #368, opened 2017-08-08
Third step of librekey with G10

catap:
Hey,

Here I finish implementing decoding and encoding protected private key with default settings of GnuPG.

In TODO list:
- "openpgp-native".

### DIFF
--- a/src/lib/packet.h
+++ b/src/lib/packet.h
@@ -439,11 +439,11 @@ typedef enum {
 } pgp_symm_alg_t;
 
 typedef enum {
-    PGP_BLOCK_CIPHER_MODE_NONE = 0,
-    PGP_BLOCK_CIPHER_MODE_CFB = 1,
-    PGP_BLOCK_CIPHER_MODE_CBC = 2,
-    PGP_BLOCK_CIPHER_MODE_OCB = 3,
-} pgp_block_cipher_t;
+    PGP_CIPHER_MODE_NONE = 0,
+    PGP_CIPHER_MODE_CFB = 1,
+    PGP_CIPHER_MODE_CBC = 2,
+    PGP_CIPHER_MODE_OCB = 3,
+} pgp_cipher_t;
 
 typedef struct symmetric_key_t {
     pgp_symm_alg_t type;
@@ -573,7 +573,7 @@ typedef enum {
     PGP_S2KS_ITERATED_AND_SALTED = 3
 } pgp_s2k_specifier_t;
 
-#define PGP_SA_DEFAULT_BLOCK_CIPHER_MODE PGP_BLOCK_CIPHER_MODE_CFB
+#define PGP_SA_DEFAULT_CIPHER_MODE PGP_CIPHER_MODE_CFB
 
 void pgp_calc_mdc_hash(
   const uint8_t *, const size_t, const uint8_t *, const unsigned, uint8_t *);
@@ -593,9 +593,9 @@ typedef struct pgp_seckey_t {
     pgp_pubkey_t        pubkey; /* public key */
     pgp_s2k_usage_t     s2k_usage;
     pgp_s2k_specifier_t s2k_specifier;
-    pgp_symm_alg_t      alg;               /* symmetric alg */
-    pgp_block_cipher_t  block_cipher_mode; /* block cipher mode */
-    pgp_hash_alg_t      hash_alg;          /* hash algorithm */
+    pgp_symm_alg_t      alg;         /* symmetric alg */
+    pgp_cipher_t        cipher_mode; /* block cipher mode */
+    pgp_hash_alg_t      hash_alg;    /* hash algorithm */
     uint8_t             salt[PGP_SALT_SIZE];
     unsigned            s2k_iterations;
     uint8_t             iv[PGP_MAX_BLOCK_SIZE];
--- a/src/lib/packet.h
+++ b/src/lib/packet.h
@@ -438,6 +438,13 @@ typedef enum {
     PGP_SA_SM4 = 105 /* RNP extension - SM4 */
 } pgp_symm_alg_t;
 
+typedef enum {
+    PGP_BLOCK_CIPHER_MODE_NONE = 0,
+    PGP_BLOCK_CIPHER_MODE_CFB = 1,
+    PGP_BLOCK_CIPHER_MODE_CBC = 2,
+    PGP_BLOCK_CIPHER_MODE_OCB = 3,
+} pgp_block_cipher_t;
+
 typedef struct symmetric_key_t {
     pgp_symm_alg_t type;
     uint8_t        key[PGP_MAX_KEY_SIZE];
@@ -566,9 +573,30 @@ typedef enum {
     PGP_S2KS_ITERATED_AND_SALTED = 3
 } pgp_s2k_specifier_t;
 
+#define PGP_SA_DEFAULT_BLOCK_CIPHER_MODE PGP_BLOCK_CIPHER_MODE_CFB
+
 void pgp_calc_mdc_hash(
   const uint8_t *, const size_t, const uint8_t *, const unsigned, uint8_t *);
 unsigned pgp_is_hash_alg_supported(const pgp_hash_alg_t *);
+
+/* Maximum block size for symmetric crypto */
+#define PGP_MAX_BLOCK_SIZE 16
+
+/* Maximum key size for symmetric crypto */
+#define PGP_MAX_KEY_SIZE 32
+
+/* Salt size for hashing */
+#define PGP_SALT_SIZE 8
+
+#define PGP_NONCE_SIZE 12
+
+#define PGP_PROTECTED_AT_SIZE 15
+
+typedef struct pgp_key_t pgp_key_t;
+
+struct pgp_seckey_t;
+
+typedef struct pgp_seckey_t *pgp_seckey_decrypt_t(const pgp_key_t *key, FILE *passfp);
 
 /** pgp_seckey_t
  */
@@ -576,11 +604,13 @@ typedef struct pgp_seckey_t {
     pgp_pubkey_t        pubkey; /* public key */
     pgp_s2k_usage_t     s2k_usage;
     pgp_s2k_specifier_t s2k_specifier;
-    pgp_symm_alg_t      alg;      /* symmetric alg */
-    pgp_hash_alg_t      hash_alg; /* hash algorithm */
+    pgp_symm_alg_t      alg;               /* symmetric alg */
+    pgp_block_cipher_t  block_cipher_mode; /* block cipher mode */
+    pgp_hash_alg_t      hash_alg;          /* hash algorithm */
     uint8_t             salt[PGP_SALT_SIZE];
     unsigned            s2k_iterations;
     uint8_t             iv[PGP_MAX_BLOCK_SIZE];
+    uint8_t             nonce[PGP_NONCE_SIZE];
     union {
         pgp_rsa_seckey_t     rsa;
         pgp_dsa_seckey_t     dsa;
@@ -589,6 +619,11 @@ typedef struct pgp_seckey_t {
     } key;
     unsigned checksum;
     uint8_t *checkhash;
+
+    size_t                encrypted_len;
+    uint8_t *             encrypted;
+    pgp_seckey_decrypt_t *decrypt_cb;
+    const char *          protected_at[PGP_PROTECTED_AT_SIZE];
 } pgp_seckey_t;
 
 /** Signature Type.
@@ -1024,7 +1059,7 @@ typedef struct pgp_subsig_t {
 } pgp_subsig_t;
 
 /* describes a user's key */
-typedef struct pgp_key_t {
+struct pgp_key_t {
     DYNARRAY(uint8_t *, uid);          /* array of user ids */
     DYNARRAY(pgp_rawpacket_t, packet); /* array of raw packets */
     DYNARRAY(pgp_subsig_t, subsig);    /* array of signature subkeys */
@@ -1040,7 +1075,7 @@ typedef struct pgp_key_t {
     uint8_t           revoked;    /* key has been revoked */
     pgp_revoke_t      revocation; /* revocation reason */
     symmetric_key_t   session_key;
-} pgp_key_t;
+};
 
 /* structure used to hold context of key generation */
 typedef struct rnp_keygen_crypto_params_t {
--- a/src/lib/packet.h
+++ b/src/lib/packet.h
@@ -443,7 +443,7 @@ typedef enum {
     PGP_CIPHER_MODE_CFB = 1,
     PGP_CIPHER_MODE_CBC = 2,
     PGP_CIPHER_MODE_OCB = 3,
-} pgp_cipher_t;
+} pgp_cipher_mode_t;
 
 typedef struct symmetric_key_t {
     pgp_symm_alg_t type;
@@ -594,7 +594,7 @@ typedef struct pgp_seckey_t {
     pgp_s2k_usage_t     s2k_usage;
     pgp_s2k_specifier_t s2k_specifier;
     pgp_symm_alg_t      alg;         /* symmetric alg */
-    pgp_cipher_t        cipher_mode; /* block cipher mode */
+    pgp_cipher_mode_t   cipher_mode; /* block cipher mode */
     pgp_hash_alg_t      hash_alg;    /* hash algorithm */
     uint8_t             salt[PGP_SALT_SIZE];
     unsigned            s2k_iterations;
--- a/src/lib/packet.h
+++ b/src/lib/packet.h
@@ -579,17 +579,6 @@ void pgp_calc_mdc_hash(
   const uint8_t *, const size_t, const uint8_t *, const unsigned, uint8_t *);
 unsigned pgp_is_hash_alg_supported(const pgp_hash_alg_t *);
 
-/* Maximum block size for symmetric crypto */
-#define PGP_MAX_BLOCK_SIZE 16
-
-/* Maximum key size for symmetric crypto */
-#define PGP_MAX_KEY_SIZE 32
-
-/* Salt size for hashing */
-#define PGP_SALT_SIZE 8
-
-#define PGP_NONCE_SIZE 12
-
 #define PGP_PROTECTED_AT_SIZE 15
 
 typedef struct pgp_key_t pgp_key_t;
@@ -610,7 +599,6 @@ typedef struct pgp_seckey_t {
     uint8_t             salt[PGP_SALT_SIZE];
     unsigned            s2k_iterations;
     uint8_t             iv[PGP_MAX_BLOCK_SIZE];
-    uint8_t             nonce[PGP_NONCE_SIZE];
     union {
         pgp_rsa_seckey_t     rsa;
         pgp_dsa_seckey_t     dsa;
--- a/src/lib/packet.h
+++ b/src/lib/packet.h
@@ -611,7 +611,7 @@ typedef struct pgp_seckey_t {
     size_t                encrypted_len;
     uint8_t *             encrypted;
     pgp_seckey_decrypt_t *decrypt_cb;
-    const char *          protected_at[PGP_PROTECTED_AT_SIZE];
+    const char *          protected_at[PGP_PROTECTED_AT_SIZE + 1]; // keep 1 byte for \0 and padding
 } pgp_seckey_t;
 
 /** Signature Type.
--- a/src/lib/pgp-key.c
+++ b/src/lib/pgp-key.c
@@ -55,6 +55,7 @@
 #include <rnp/rnp_sdk.h>
 #include "readerwriter.h"
 #include "validate.h"
+#include "packet.h"
 
 #include <stdio.h>
 #include <string.h>
@@ -405,24 +406,12 @@ decrypt_cb_empty(const pgp_packet_t *pkt, pgp_cbdata_t *cbinfo)
     }
 }
 
-/**
-\ingroup Core_Keys
-\brief Decrypts secret key from given key with given passphrase
-\param key Key from which to get secret key
-\param passphrase Passphrase to use to decrypt secret key
-\return secret key
-*/
 pgp_seckey_t *
-pgp_decrypt_seckey(const pgp_key_t *key, FILE *passfp)
+pgp_decrypt_seckey_parser(const pgp_key_t *key, FILE *passfp)
 {
     pgp_stream_t *stream;
     const int     printerrors = 1;
     decrypt_t     decrypt;
-
-    // key hasn't got raw packets, so, we can't decrypt it
-    if (key->packetc == 0) {
-        return (pgp_seckey_t *) &key->key.seckey;
-    }
 
     /* XXX first try with an empty passphrase */
     (void) memset(&decrypt, 0x0, sizeof(decrypt));
@@ -455,6 +444,23 @@ pgp_decrypt_seckey(const pgp_key_t *key, FILE *passfp)
     pgp_parse(stream, !printerrors);
     pgp_stream_delete(stream);
     return decrypt.seckey;
+}
+
+/**
+\ingroup Core_Keys
+\brief Decrypts secret key from given key with given passphrase
+\param key Key from which to get secret key
+\param passphrase Passphrase to use to decrypt secret key
+\return secret key
+*/
+pgp_seckey_t *
+pgp_decrypt_seckey(const pgp_key_t *key, FILE *passfp)
+{
+    if (key->key.seckey.decrypt_cb == NULL) {
+        return NULL;
+    }
+
+    return key->key.seckey.decrypt_cb(key, passfp);
 }
 
 /**
--- a/src/lib/pgp-key.c
+++ b/src/lib/pgp-key.c
@@ -429,7 +429,9 @@ pgp_decrypt_seckey(const pgp_key_t *key, FILE *passfp)
 
     decrypt.key = key;
     stream = pgp_new(sizeof(*stream));
-    pgp_key_reader_set(stream, key);
+    if (!pgp_key_reader_set(stream, key)) {
+        return NULL;
+    }
     pgp_set_callback(stream, decrypt_cb_empty, &decrypt);
     stream->readinfo.accumulate = 1;
     pgp_parse(stream, !printerrors);
@@ -445,7 +447,9 @@ pgp_decrypt_seckey(const pgp_key_t *key, FILE *passfp)
     /* ask for a passphrase */
     decrypt.passfp = passfp;
     stream = pgp_new(sizeof(*stream));
-    pgp_key_reader_set(stream, key);
+    if (!pgp_key_reader_set(stream, key)) {
+        return NULL;
+    }
     pgp_set_callback(stream, decrypt_cb, &decrypt);
     stream->readinfo.accumulate = 1;
     pgp_parse(stream, !printerrors);
--- a/src/librekey/key_store_g10.c
+++ b/src/librekey/key_store_g10.c
@@ -150,7 +150,7 @@ parse_sexp(s_exp_t *s_exp, const char **r_bytes, size_t *r_length)
     size_t      length = *r_length;
     const char *bytes = *r_bytes;
 
-    s_exp_t new_s_exp = {};
+    s_exp_t new_s_exp = {0};
 
     if (bytes == NULL || length == 0) {
         fprintf(stderr, "empty s-exp\n");
@@ -484,7 +484,7 @@ g10_decrypt_seckey(const pgp_key_t *key, FILE *passfp)
     unsigned       keysize;
     uint8_t *      decrypted;
     pgp_seckey_t * seckey;
-    s_exp_t        s_exp = {};
+    s_exp_t        s_exp = {0};
     size_t         output_written = 0;
     size_t         input_consumed = 0;
     botan_cipher_t decrypt;
@@ -783,7 +783,7 @@ rnp_key_store_g10_from_mem(pgp_io_t *       io,
                            rnp_key_store_t *key_store,
                            pgp_memory_t *   memory)
 {
-    s_exp_t     s_exp = {};
+    s_exp_t     s_exp = {0};
     size_t      length = memory->length;
     const char *bytes = (const char *) memory->buf;
 
@@ -1125,7 +1125,7 @@ static bool
 write_protected_seckey(s_exp_t *s_exp, pgp_seckey_t *key, const uint8_t *passphrase)
 {
     char *         type;
-    s_exp_t        raw_s_exp = {};
+    s_exp_t        raw_s_exp = {0};
     s_exp_t *      sub_s_exp, *sub_sub_s_exp, *sub_sub_sub_s_exp;
     botan_cipher_t encrypt;
     uint8_t        derived_key[PGP_MAX_KEY_SIZE];
@@ -1413,7 +1413,7 @@ rnp_key_store_g10_key_to_mem(pgp_io_t *     io,
                              pgp_memory_t * memory)
 {
     bool     rc;
-    s_exp_t  s_exp = {};
+    s_exp_t  s_exp = {0};
     s_exp_t *sub_s_exp;
 
     if (!add_sub_sexp_to_sexp(&s_exp, &sub_s_exp)) {
--- a/src/librekey/key_store_g10.c
+++ b/src/librekey/key_store_g10.c
@@ -35,6 +35,9 @@
 #include "key_store_g10.h"
 
 #include "crypto/bn.h"
+#include "crypto/s2k.h"
+#include "symmetric.h"
+#include "readerwriter.h"
 
 typedef struct {
     size_t   len;
@@ -224,20 +227,45 @@ parse_sexp(s_exp_t *s_exp, const char **r_bytes, size_t *r_length)
     return true;
 }
 
-static s_exp_block_t *
+#define STR_HELPER(x) #x
+#define STR(x) STR_HELPER(x)
+
+static unsigned
+block_to_unsigned(s_exp_block_t *block)
+{
+    char s[sizeof(STR(UINT_MAX)) + 1];
+    if (block->len >= sizeof(s)) {
+        return UINT_MAX;
+    }
+
+    memcpy(s, block->bytes, block->len);
+    return (unsigned int) atoi(s);
+}
+
+static bool
+add_unsigned_block_to_sexp(s_exp_t *s_exp, unsigned u)
+{
+    char s[sizeof(STR(UINT_MAX)) + 1];
+    snprintf(s, sizeof(s), "%u", u);
+    return add_block_to_sexp(s_exp, (uint8_t *) s, strlen(s));
+}
+
+static s_exp_t *
 lookup_variable(s_exp_t *s_exp, const char *name)
 {
+    size_t name_len = strlen(name);
     for (int i = 0; i < s_exp->sub_s_expc; i++) {
-        if (s_exp->sub_s_exps[i].blockc != 2) {
+        if (s_exp->sub_s_exps[i].blockc < 2) {
             fprintf(stderr,
-                    "Expected 2 block (<name> <value>) but has: %d\n",
+                    "Expected 2 or more blocks (<name> <value> [<value2>]+) but has: %d\n",
                     s_exp->sub_s_exps[i].blockc);
             return NULL;
         }
-        if (!strncmp(name,
+        if (name_len == s_exp->sub_s_exps[i].blocks[0].len &&
+            !strncmp(name,
                      (const char *) s_exp->sub_s_exps[i].blocks[0].bytes,
                      s_exp->sub_s_exps[i].blocks[0].len)) {
-            return &s_exp->sub_s_exps[i].blocks[1];
+            return &s_exp->sub_s_exps[i];
         }
     }
     fprintf(stderr, "Haven't got variable '%s'\n", name);
@@ -247,14 +275,14 @@ lookup_variable(s_exp_t *s_exp, const char *name)
 static BIGNUM *
 read_bignum(s_exp_t *s_exp, const char *name)
 {
-    s_exp_block_t *var = lookup_variable(s_exp, name);
+    s_exp_t *var = lookup_variable(s_exp, name);
     if (var == NULL) {
         return NULL;
     }
 
-    BIGNUM *res = PGPV_BN_bin2bn(var->bytes, (int) var->len, NULL);
+    BIGNUM *res = PGPV_BN_bin2bn(var->blocks[1].bytes, (int) var->blocks[1].len, NULL);
     if (res == NULL) {
-        char *buf = malloc((var->len * 3) + 1);
+        char *buf = malloc((var->blocks[1].len * 3) + 1);
         if (buf == NULL) {
             fprintf(stderr, "Can't allocate memory\n");
             return NULL;
@@ -262,7 +290,7 @@ read_bignum(s_exp_t *s_exp, const char *name)
         fprintf(stderr,
                 "Can't convert variable '%s' to bignum. The value is: '%s'\n",
                 name,
-                rnp_strhexdump(buf, var->bytes, var->len, ""));
+                rnp_strhexdump(buf, var->blocks[1].bytes, var->blocks[1].len, ""));
     }
     return res;
 }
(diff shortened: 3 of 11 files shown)


dewyatt:
It's probably a good idea to have the crypto specialists work on this (that's what they're here for).
My untrained eye sees a KDF with a salt of all zeros. I know there's also a heap overflow.

I wonder if it would be possible to use botan's higher-level cipher interfaces (`botan_cipher_*` rather than the low-level `botan_block_cipher_*`) here, and let those handle CBC/OCB (plus I think AEAD is used in this file format)?

ronaldtse:
Thanks @catap ! Agree with @dewyatt that we should rely on botan for encryption modes.

flowher:
I agree that we should reuse botan as much as possible.

catap:
Well... I agree to it, beause I had some legal question about OBC and just use botan is easy way.

Ok, will push the patch today.

catap:
@dewyatt nope, this format doesn't use AEAD. Only AES-128 with CBC or OCB with SHA-1 hashing.

catap:
Pushed.

Well, looks like it's missed only `openpgp-native` format.

src/lib/packet.h
```diff
@@ -565,21 +572,44 @@ typedef enum {
...
 unsigned pgp_is_hash_alg_supported(const pgp_hash_alg_t *);
 
+/* Maximum block size for symmetric crypto */
+#define PGP_MAX_BLOCK_SIZE 16
+
+/* Maximum key size for symmetric crypto */
+#define PGP_MAX_KEY_SIZE 32
+
+/* Salt size for hashing */
+#define PGP_SALT_SIZE 8
```
dewyatt: I think you accidentally duplicated a few of theses defines in packet.h. `PGP_MAX_BLOCK_SIZE` `PGP_MAX_KEY_SIZE` `PGP_SALT_SIZE` are duplicated.

src/lib/packet.h
```diff
@@ -565,21 +572,44 @@ typedef enum {
...
     pgp_s2k_specifier_t s2k_specifier;
-    pgp_symm_alg_t      alg;      /* symmetric alg */
-    pgp_hash_alg_t      hash_alg; /* hash algorithm */
+    pgp_symm_alg_t      alg;               /* symmetric alg */
+    pgp_block_cipher_t  block_cipher_mode; /* block cipher mode */
+    pgp_hash_alg_t      hash_alg;          /* hash algorithm */
     uint8_t             salt[PGP_SALT_SIZE];
     unsigned            s2k_iterations;
     uint8_t             iv[PGP_MAX_BLOCK_SIZE];
+    uint8_t             nonce[PGP_NONCE_SIZE];
```
dewyatt: We don't really need a separate nonce field, just use the iv field. For OCB it will be 12 bytes, for CBC use the block size.

src/lib/symmetric.c
```diff
@@ -161,6 +161,42 @@ pgp_cipher_cfb_decrypt(pgp_crypt_t *crypt, uint8_t *out, const uint8_t *in, size
     return 0;
 }
 
+int
+pgp_cipher_cbc_encrypt(pgp_crypt_t *crypt, uint8_t *out, const uint8_t *in, size_t bytes)
```
dewyatt: Better to remove these two functions now that they're not used.

src/librekey/key_store_g10.c
```diff
@@ -779,6 +1118,286 @@ write_seckey(s_exp_t *s_exp, pgp_seckey_t *key)
...
+
+    raw = pgp_memory_new();
+    if (raw == NULL) {
+        (void) fprintf(stderr, "Can't allocate memory\n");
+        destroy_s_exp(&raw_s_exp);
+        return false;
+    }
+
+    // allocated some memory
+    pgp_memory_init(raw, 1024);
```
dewyatt: Seems like an arbitrary size, might be a better way.

src/librekey/key_store_g10.c
```diff
@@ -779,6 +1118,286 @@ write_seckey(s_exp_t *s_exp, pgp_seckey_t *key)
...
+
+        if (botan_cipher_set_key(encrypt, derived_key, keysize)) {
+            free(encrypted);
+            destroy_s_exp(&raw_s_exp);
+            pgp_memory_free(raw);
+            botan_cipher_destroy(encrypt);
+            return false;
+        }
+
+        if (botan_cipher_start(encrypt, key->iv, PGP_MAX_BLOCK_SIZE)) {
```
dewyatt: Shouldn't be MAX_BLOCK_SIZE, use the block size of the cipher.

src/librekey/key_store_g10.c
```diff
@@ -808,17 +1427,53 @@ rnp_key_store_g10_key_to_mem(pgp_io_t *     io,
...
+            // force switch to AES with CBC and SHA-1
+            if (key->key.seckey.alg != PGP_SA_AES_128) {
+                key->key.seckey.alg = PGP_SA_AES_128;
+            }
+
+            if (key->key.seckey.block_cipher_mode != PGP_BLOCK_CIPHER_MODE_CBC) {
+                key->key.seckey.block_cipher_mode = PGP_BLOCK_CIPHER_MODE_CBC;
+            }
+
+            if (key->key.seckey.hash_alg != PGP_HASH_SHA1) {
```
dewyatt: These 3 conditionals seem odd as they're equivalent to just an assignment.

src/lib/packet.h
```diff
@@ -565,21 +572,44 @@ typedef enum {
...
+/* Maximum block size for symmetric crypto */
+#define PGP_MAX_BLOCK_SIZE 16
+
+/* Maximum key size for symmetric crypto */
+#define PGP_MAX_KEY_SIZE 32
+
+/* Salt size for hashing */
+#define PGP_SALT_SIZE 8
+
+#define PGP_NONCE_SIZE 12
```
dewyatt: Probably best to move this to `key_store_g10.c` and name it something like `G10_OCB_NONCE_SIZE` (or even `G10_OCB_IV_SIZE`) since it's not part of the RFC.

dewyatt:
Salt is currently still zero and half the IV is zero (w/default cbc). I would randomize both fully inside `write_protected_seckey`.

src/lib/packet.h
catap: Yay. Thanks!

src/librekey/key_store_g10.c
catap: Yes, it is a hack. And this hack has a reason: to save any algorithm/cipher, I should implement `opengpg-native` mode of encoding that I haven't done yet.

So, I see few diffferent ways:
 - remove it and disable G10 for default key settings;
 - update default settings;
 - keep it as is.

Anyway, I will remove it as next step with `openpgp-native`.

src/librekey/key_store_g10.c
```diff
@@ -779,6 +1121,281 @@ write_seckey(s_exp_t *s_exp, pgp_seckey_t *key)
...
+    }
+
+    switch (key->block_cipher_mode) {
+    case PGP_BLOCK_CIPHER_MODE_CBC:
+        if (key->alg == PGP_SA_AES_128) {
+            type = "openpgp-s2k3-sha1-aes-cbc";
+        } else if (key->alg == PGP_SA_AES_256) {
+            type = "openpgp-s2k3-sha1-aes256-cbc";
+        } else {
+            fprintf(stderr, "Unsupported algorithm: %d\n", key->alg);
```
dewyatt: `return false;`

src/librekey/key_store_g10.c
```diff
@@ -779,6 +1121,281 @@ write_seckey(s_exp_t *s_exp, pgp_seckey_t *key)
...
+        }
+        break;
+
+    case PGP_BLOCK_CIPHER_MODE_OCB:
+        if (key->alg == PGP_SA_AES_128) {
+            type = "openpgp-s2k3-ocb-aes";
+        } else {
+            fprintf(stderr, "Unsupported algorithm: %d\n", key->alg);
```
dewyatt: `return false;`

src/librekey/key_store_g10.c
catap: Yay!

src/librekey/key_store_g10.c
catap: Yay!

dewyatt:
Key generation doesn't seem to work after the first key for some reason.
```
$ rm -rf ~/.rnp
$ src/rnpkeys/rnpkeys --gen-key --keystore-format=GPG21
$ ls ~/.rnp/private-keys-v1.d/
830950e1f78a8a817deb7c9b050bf564e4e21af5.key
$ src/rnpkeys/rnpkeys --gen-key --keystore-format=GPG21 (bad exit code)
$ ls ~/.rnp/private-keys-v1.d/
830950e1f78a8a817deb7c9b050bf564e4e21af5.key (key not saved)
```

src/librekey/key_store_g10.c
dewyatt: Ok I see the use of it as a temporary workaround. But I guess I meant that it's equivalent to:

```c
// force switch to AES with CBC and SHA-1
key->key.seckey.alg = PGP_SA_AES_128;
key->key.seckey.block_cipher_mode = PGP_BLOCK_CIPHER_MODE_CBC;
key->key.seckey.hash_alg = PGP_HASH_SHA1;
```

Since it's temporary though, it doesn't really matter.

catap:
@dewyatt thanks! Good catch.

src/lib/packet.h
```diff
@@ -435,6 +435,13 @@ typedef enum {
     PGP_SA_SM4 = 105 /* RNP extension - SM4 */
 } pgp_symm_alg_t;
 
+typedef enum {
+    PGP_BLOCK_CIPHER_MODE_NONE = 0,
+    PGP_BLOCK_CIPHER_MODE_CFB = 1,
+    PGP_BLOCK_CIPHER_MODE_CBC = 2,
+    PGP_BLOCK_CIPHER_MODE_OCB = 3,
+} pgp_block_cipher_t;
```
dewyatt: I think the name `pgp_cipher_mode_t` would be slightly more accurate here. And then `PGP_CIPHER_MODE_CFB`, etc.

src/librekey/key_store_g10.c
```diff
@@ -779,6 +1121,291 @@ write_seckey(s_exp_t *s_exp, pgp_seckey_t *key)
     return true;
 }
 
+static bool
+write_protected_seckey(s_exp_t *s_exp, pgp_seckey_t *key, const uint8_t *passphrase)
+{
+    char *         type;
+    s_exp_t        raw_s_exp = {};
```
dewyatt: I'm thinking we shouldn't introduce more GNU extensions into the code base if we can avoid it.

dewyatt:
It looks like this isn't quite compatible with GPG 2.1. 

```
$ gpg --version
gpg (GnuPG) 2.1.22
[...]
```

```
$ src/rnpkeys/rnpkeys --gen-key --keystore-format=GPG21
$ echo test | gpg --homedir=~/.rnp --sign --armor
gpg: /home/daniel/.rnp/trustdb.gpg: trustdb created
gpg: no default secret key: No secret key
gpg: signing failed: No secret key
```

It doesn't find the secret key because the hex portion of the filename has to be uppercase (so I guess don't use `rnp_strhexdump` as is currently done).

Beyond that it appears there's another issue:

```
$ cd ~/.rnp/private-keys-v1.d
$ mv 3acb4cef69a3d2b8b29551728b07a2adc2c2af59.key 3ACB4CEF69A3D2B8B29551728B07A2ADC2C2AF59.key
$ echo test | gpg --homedir=~/.rnp --sign --armor
gpg: signing failed: Invalid S-expression
-----BEGIN PGP MESSAGE-----

gpg: signing failed: Invalid S-expression

```

src/librekey/key_store_g10.c
```diff
@@ -441,6 +475,308 @@ parse_seckey(pgp_keydata_key_t *keydata, s_exp_t *s_exp, pgp_pubkey_alg_t alg)
...
+    char           pass[MAX_PASSPHRASE_LENGTH];
+    unsigned       keysize;
+    uint8_t *      decrypted;
+    pgp_seckey_t * seckey;
+    s_exp_t        s_exp = {};
+    size_t         output_written = 0;
+    size_t         input_consumed = 0;
+    botan_cipher_t decrypt;
+
+    if (key->key.seckey.encrypted_len <= 0) {
```
flowher: I think it should be == 0

src/librekey/key_store_g10.c
```diff
@@ -441,6 +475,308 @@ parse_seckey(pgp_keydata_key_t *keydata, s_exp_t *s_exp, pgp_pubkey_alg_t alg)
...
+        fprintf(stderr, "Hasn't got encrypted data!\n");
+        return NULL;
+    }
+
+    if (pgp_getpassphrase(passfp, pass, sizeof(pass)) == 0) {
+        pass[0] = '\0';
+    }
```
flowher: "else { return NULL; }" ??

src/librekey/key_store_g10.c
```diff
@@ -441,6 +475,308 @@ parse_seckey(pgp_keydata_key_t *keydata, s_exp_t *s_exp, pgp_pubkey_alg_t alg)
...
+
+    if (pgp_s2k_iterated(key->key.seckey.hash_alg,
+                         derived_key,
+                         keysize,
+                         passphrase,
+                         key->key.seckey.salt,
+                         key->key.seckey.s2k_iterations)) {
+        (void) fprintf(stderr, "pgp_s2k_iterated failed\n");
+        return false;
```
flowher: you should also pgp_forget and free before returning false

src/librekey/key_store_g10.c
```diff
@@ -441,6 +475,308 @@ parse_seckey(pgp_keydata_key_t *keydata, s_exp_t *s_exp, pgp_pubkey_alg_t alg)
...
+    }
+
+    pgp_forget(passphrase, strlen(passphrase));
```
flowher: free passphrase?

catap:
@dewyatt thanks, I think that I changed it.

src/librekey/key_store_g10.c
```diff
@@ -779,14 +1121,299 @@ write_seckey(s_exp_t *s_exp, pgp_seckey_t *key)
...
+            pgp_memory_release(&raw);
+            botan_cipher_destroy(encrypt);
+            return false;
+        }
+
+        break;
+
+    case PGP_CIPHER_MODE_OCB:
+        if (botan_cipher_init(&encrypt, "AES-128/OCB", BOTAN_CIPHER_INIT_FLAG_ENCRYPT)) {
+            (void) fprintf(stderr, "botan_cipher_init failed\n");
```
flowher: a lot of copy-pasta here. Isn't it better to "goto error" end handle this all in one place?

src/lib/pgp-key.c
```diff
@@ -415,7 +406,9 @@ pgp_decrypt_seckey(const pgp_key_t *key, FILE *passfp)
     /* ask for a passphrase */
     decrypt.passfp = passfp;
     stream = pgp_new(sizeof(*stream));
-    pgp_key_reader_set(stream, key);
+    if (!pgp_key_reader_set(stream, key)) {
+        return NULL;
```
catap: @dewyatt, @flowher am I wrong that this function has leaks by design — it never makes `free` for stream?

src/librekey/key_store_g10.c
catap: Why?

src/lib/pgp-key.c
dewyatt: These were being worked on in https://github.com/riboseinc/rnp/issues/350

catap: @dewyatt shall I revert https://github.com/riboseinc/rnp/pull/368/commits/549ed8e940b9693cd50edafe5f930e003e9ca1f8 ?

dewyatt: Preferably, yes

catap: @dewyatt done

src/librekey/key_store_g10.c
catap: Thanks. Good point.

src/librekey/key_store_g10.c
flowher: I would say because you should stop if pgp_getpassphrase fails, isn't ?

catap:
@dewyatt, @flowher ping!

src/librekey/key_store_g10.c
catap: @flowher to be honest: this function [never return false](https://github.com/riboseinc/rnp/blob/bb010474d9b4a710af58cfef775b5147be9b49a6/src/lib/reader.c#L145) and it is a copy-and-paste from [here](https://github.com/riboseinc/rnp/blob/7d06d4569e458b49c7080a1ad9959dcf2c6f8f87/src/lib/pgp-key.c#L320)

dewyatt:
I think it might help to add a little lookup table to get rid of some of these switch statements. Something like:

```diff
diff --git a/src/librekey/key_store_g10.c b/src/librekey/key_store_g10.c
index c3c54a1..3e51004 100644
--- a/src/librekey/key_store_g10.c
+++ b/src/librekey/key_store_g10.c
@@ -475,6 +475,31 @@ parse_seckey(pgp_seckey_t *seckey, s_exp_t *s_exp, pgp_pubkey_alg_t alg, bool ha
     return true;
 }
 
+typedef struct format_info {
+    pgp_symm_alg_t    cipher;
+    pgp_cipher_mode_t cipher_mode;
+    const char *      botan_cipher_name;
+    const char *      g10_type;
+    size_t            iv_size;
+} format_info;
+
+static const format_info formats[] = {
+  {PGP_SA_AES_128, PGP_CIPHER_MODE_CBC, "AES-128/CBC", "openpgp-s2k3-sha1-aes-cbc", G10_CBC_IV_SIZE},
+  {PGP_SA_AES_256, PGP_CIPHER_MODE_CBC, "AES-256/CBC", "openpgp-s2k3-sha1-aes256-cbc", G10_CBC_IV_SIZE},
+  {PGP_SA_AES_128, PGP_CIPHER_MODE_OCB, "AES-128/OCB", "openpgp-s2k3-ocb-aes", G10_OCB_IV_SIZE}
+};
+
+static const format_info *
+find_format(pgp_symm_alg_t cipher, pgp_cipher_mode_t mode)
+{
+    for (size_t i = 0; i < ARRAY_SIZE(formats); i++) {
+        if (formats[i].cipher == cipher && formats[i].cipher_mode == mode) {
+            return &formats[i];
+        }
+    }
+    return NULL;
+}
+
 static pgp_seckey_t *
 g10_decrypt_seckey(const pgp_key_t *key, FILE *passfp)
 {
```

src/librekey/key_store_g10.c
```diff
@@ -779,14 +1130,263 @@ write_seckey(s_exp_t *s_exp, pgp_seckey_t *key)
     return true;
 }
 
+static bool
+write_protected_seckey(s_exp_t *s_exp, pgp_seckey_t *key, const uint8_t *passphrase)
+{
+    char *         type;
```
dewyatt: Can you initialize this to NULL, my version of GCC refuses to build otherwise.

src/librekey/key_store_g10.c
```diff
@@ -779,14 +1130,263 @@ write_seckey(s_exp_t *s_exp, pgp_seckey_t *key)
...
+    s_exp_t        raw_s_exp = {0};
+    s_exp_t *      sub_s_exp, *sub_sub_s_exp, *sub_sub_sub_s_exp;
+    botan_cipher_t encrypt;
```
dewyatt: Init to NULL here as well, CI is failing to build otherwise.

ronaldtse:
We're almost there! Once this is done we should merge in #354...

catap:
@dewyatt thanks for idea of lookup table. Yes, it is better than switch-spagetti.

@ronaldtse I think I fixed everything. Let's ask @dewyatt and @flowher to think a while and if everything is fine — you can merge it.

src/librekey/key_store_g10.c
```diff
@@ -779,14 +1164,218 @@ write_seckey(s_exp_t *s_exp, pgp_seckey_t *key)
...
+    size_t             input_consumed = 0;
+
+    if (key->s2k_specifier != PGP_S2KS_ITERATED_AND_SALTED) {
+        fprintf(stderr, "s2k should be iterated and salted\n");
+        return false;
+    }
+
+    // if we had encrypted block, don't renecrypt
+    if (key->encrypted_len > 0 && key->encrypted != NULL) {
+        goto write;
```
dewyatt: This would crash I believe.

catap: Yes, you're right. I rebased this commit after first push, but second push was without `-f`. So... github declined it.

Anyway, I pushed one more time and it should be fixed.

src/librekey/key_store_g10.c
```diff
@@ -779,14 +1164,218 @@ write_seckey(s_exp_t *s_exp, pgp_seckey_t *key)
...
+
+    if (!add_string_block_to_sexp(sub_s_exp, "protected-at")) {
+        return false;
+    }
+
+    if (!add_block_to_sexp(sub_s_exp, (uint8_t *) key->protected_at, PGP_PROTECTED_AT_SIZE)) {
+        return false;
+    }
+
+    return true;
```
dewyatt: This would leak things though. I would change the `error` label to a `done` label that frees everything, and replace all the `return true`/`return false` with `goto done`. Something like what `pgp_eddsa_sign_hash` and a few other functions do: https://github.com/riboseinc/rnp/blob/c7c3d1973156ff0bc82c8b969a2ed0613f7a2fa5/src/lib/eddsa.c#L174

dewyatt:
I think we lost the uppercase .key name fix in here somewhere.

dewyatt:
GPG21 still refuses to use the private key, saying `Invalid S-expression`. I would really like to get it compatible before merge.

src/librekey/key_store_g10.c
```diff
@@ -441,13 +527,305 @@ parse_seckey(pgp_keydata_key_t *keydata, s_exp_t *s_exp, pgp_pubkey_alg_t alg)
...
+        hexdump(stderr, "encrypted", key->key.seckey.encrypted, key->key.seckey.encrypted_len);
+    }
+
+    decrypted = malloc(key->key.seckey.encrypted_len);
+    if (decrypted == NULL) {
+        (void) fprintf(stderr, "can't allocate memory\n");
+        return false;
+    }
+
+    switch (key->key.seckey.cipher_mode) {
```
dewyatt: Should the formats table be used here? It seems to assume AES-128 (no 256) below as well.

src/librekey/key_store_g10.c
```diff
@@ -441,13 +527,305 @@ parse_seckey(pgp_keydata_key_t *keydata, s_exp_t *s_exp, pgp_pubkey_alg_t alg)
...
+    seckey->s2k_iterations = block_to_unsigned(&alg->blocks[2]);
+    if (seckey->s2k_iterations == UINT8_MAX) {
+        fprintf(stderr,
+                "Wrong numbers of iteration, %.*s\n",
+                (int) alg->blocks[2].len,
+                alg->blocks[2].bytes);
+        return false;
+    }
+
+    if (seckey->cipher_mode == PGP_CIPHER_MODE_CBC) {
```
dewyatt: Should be able to simplify this with the table, right?

```diff
diff --git a/src/librekey/key_store_g10.c b/src/librekey/key_store_g10.c
index b031a30..6a554fc 100644
--- a/src/librekey/key_store_g10.c
+++ b/src/librekey/key_store_g10.c
@@ -759,27 +759,14 @@ parse_protected_seckey(pgp_seckey_t *seckey, s_exp_t *s_exp)
         return false;
     }
 
-    if (seckey->cipher_mode == PGP_CIPHER_MODE_CBC) {
-        if (encrypted->blocks[0].len != G10_CBC_IV_SIZE) {
-            fprintf(stderr,
-                    "Wrong IV size, should be %d but %d\n",
-                    G10_CBC_IV_SIZE,
-                    (int) encrypted->blocks[0].len);
-            return false;
-        }
-
-        memcpy(seckey->iv, encrypted->blocks[0].bytes, encrypted->blocks[0].len);
-    } else if (seckey->cipher_mode == PGP_CIPHER_MODE_OCB) {
-        if (encrypted->blocks[0].len != G10_OCB_NONCE_SIZE) {
-            fprintf(stderr,
-                    "Wrong nonce size, should be %d but %d\n",
-                    G10_OCB_NONCE_SIZE,
-                    (int) encrypted->blocks[0].len);
-            return false;
-        }
-
-        memcpy(seckey->iv, encrypted->blocks[0].bytes, encrypted->blocks[0].len);
+    if (encrypted->blocks[0].len != format->iv_size) {
+        fprintf(stderr,
+                "Wrong IV size, should be %lu but %d\n",
+                format->iv_size,
+                (int) encrypted->blocks[0].len);
+        return false;
     }
+    memcpy(seckey->iv, encrypted->blocks[0].bytes, encrypted->blocks[0].len);
 
     if (var->blockc != 3) {
         fprintf(stderr, "Hasn't got encrypted blob\n");
+}
```

src/librekey/key_store_g10.c
dewyatt: It does return false when fgets fails (when using passfp and EOF is hit, for example) so it's better to fail here. Also since `pgp_getpassphrase` returns a bool, you could probably just do something like `if (!pgp_getpassphrase(...)) { return false; }`.

flowher: fgets will return false if error occurs or there is nothing to read from stream. But - in a big picture - what does it mean if there is "nothing to read from stream". Is it error condition or it means no password (empty file)? 
Personally I would treat it as an error and return false at this point.

src/lib/packet.h
```diff
@@ -565,18 +572,29 @@ typedef enum {
...
+
+typedef struct pgp_key_t pgp_key_t;
+
+typedef struct pgp_seckey_t pgp_seckey_t;
+
+typedef pgp_seckey_t *pgp_seckey_decrypt_t(const pgp_key_t *key, FILE *passfp);
+
 /** pgp_seckey_t
  */
-typedef struct pgp_seckey_t {
```
dewyatt: Why split these `typedef`s from the struct definitions?

catap:
@dewyatt I agree that it should work with GnuPG21. I'm on digging why it is and will fix table lookup after I'd finish it.

src/lib/packet.h
catap: Because I added to `pgp_seckey_t` field `pgp_seckey_decrypt_t *decrypt_cb` that returns `pgp_seckey_t`;

src/librekey/key_store_g10.c
```diff
@@ -35,6 +35,13 @@
 #include "key_store_g10.h"
 
 #include "bn.h"
+#include "s2k.h"
+#include "symmetric.h"
+#include "readerwriter.h"
+
+#define G10_CBC_IV_SIZE 16
```
flowher: you can put those defines in crypto.h. they shouldn't be prefixed with G10_ as IV/Nonce size doesn't depend on key store version (it always will be 16 or 12)

flowher: you can put those defines in crypto.h. they shouldn't be prefixed with G10_ (better idea is to prefix it with AES_) as IV/Nonce size doesn't depend on key store version (it always will be 16 or 12)

src/librekey/key_store_g10.c
```diff
@@ -49,6 +56,58 @@ struct s_exp {
...
+    pgp_hash_alg_t    hash_alg;
+    const char *      botan_cipher_name;
+    const char *      g10_type;
+    size_t            iv_size;
+} format_info;
+
+static const format_info formats[] = {{PGP_SA_AES_128,
+                                       PGP_CIPHER_MODE_CBC,
+                                       PGP_HASH_SHA1,
+                                       "AES-128/CBC",
```
flowher: For CBC I would add paddinging explicitly as this implementation depends on it. So this line should be "AES-128/CBC/PKCS7" (same for 256 key)

src/librekey/key_store_g10.c
```diff
@@ -441,13 +527,305 @@ parse_seckey(pgp_keydata_key_t *keydata, s_exp_t *s_exp, pgp_pubkey_alg_t alg)
...
+        fprintf(stderr,
+                "Wrong salt size, should be %d but %d\n",
+                PGP_SALT_SIZE,
+                (int) alg->blocks[1].len);
+        return false;
+    }
+
+    memcpy(seckey->salt, alg->blocks[1].bytes, alg->blocks[1].len);
+    seckey->s2k_iterations = block_to_unsigned(&alg->blocks[2]);
+    if (seckey->s2k_iterations == UINT8_MAX) {
```
flowher: Hmm... I think it should it be seckey->s2k_iterations == UINT_MAX instead of UINT8_MAX (see line 294)

src/librekey/key_store_g10.c
dewyatt: G10 is basically being used as the file format name here I believe, so it does kind of make sense in the name IMO.

src/lib/packet.h
dewyatt: I think it's better to keep the struct and typedef together as it's a pretty strong convention in the code I see. You can always pre-declare the struct and use something like this:

```diff
diff --git a/src/lib/packet.h b/src/lib/packet.h
index 095996d..5b23279 100644
--- a/src/lib/packet.h
+++ b/src/lib/packet.h
@@ -582,13 +582,12 @@ unsigned pgp_is_hash_alg_supported(const pgp_hash_alg_t *);
 
 typedef struct pgp_key_t pgp_key_t;
 
-typedef struct pgp_seckey_t pgp_seckey_t;
-
-typedef pgp_seckey_t *pgp_seckey_decrypt_t(const pgp_key_t *key, FILE *passfp);
+struct pgp_seckey_t;
+typedef struct pgp_seckey_t *pgp_seckey_decrypt_t(const pgp_key_t *key, FILE *passfp);
 
 /** pgp_seckey_t
  */
-struct pgp_seckey_t {
+typedef struct pgp_seckey_t {
     pgp_pubkey_t        pubkey; /* public key */
     pgp_s2k_usage_t     s2k_usage;
     pgp_s2k_specifier_t s2k_specifier;
@@ -611,7 +610,7 @@ struct pgp_seckey_t {
     uint8_t *             encrypted;
     pgp_seckey_decrypt_t *decrypt_cb;
     const char *protected_at[PGP_PROTECTED_AT_SIZE + 1]; // keep 1 byte for \0 and padding
-};
+} pgp_seckey_t;
 
 /** Signature Type.
  * OpenPGP defines different signature types that allow giving
```

src/librekey/key_store_g10.c
dewyatt: Nice catch.

ronaldtse:
@catap hoping to merge this soon with the fixing of the above comments 😉

dewyatt:
@catap Did you have a chance to further investigate the GPG21 incompatibility?

I saw the s-exprs are a bit different but didn't really look into it beyond that.

```
$ /usr/lib/gnupg/gpg-protect-tool ~/.rnp/private-keys-v1.d/032C2D2A410B3BDEA1047D7A063C03C0680B2E03.key 
(protected-private-key 
 (rsa 
  (n #...keydata...#)
  (e #010001#)
  (protected openpgp-s2k3-sha1-aes-cbc #...keydata...#
   (#...IV...# 
    (sha1 #...salt...# "...iterations...")
    )
   )
  (protected-at "20170817T024444")
  )
 )
```

Compared to a key generated by gpg 2.1:

```
$ /usr/lib/gnupg/gpg-protect-tool ~/.gnupg/private-keys-v1.d/560BD3EB9BA6ACF099302B8F3A9B2BC6E3DAAB49.key 
(protected-private-key 
 (rsa 
  (n #...keydata...#)
  (e #010001#)
  (protected openpgp-s2k3-sha1-aes-cbc
   (
    (sha1 #...salt...# "...iterations...")
    #...IV...#)
   #...keydata...#)
  (protected-at "20170810T191747")
  )
 )

```

src/librekey/key_store_g10.c
catap: Just for your information, by default botan has PKCS7 padding and for case `disable padding` I should use `AES-128/CBC/NoPadding`.

catap: Just for your information, looks like that by default botan has PKCS7 padding and for case `disable padding` I should use `AES-128/CBC/NoPadding`.

dewyatt: It wouldn't hurt to be explicit and add PKCS7 though, it is admittedly better than relying on the hidden default.

catap: @dewyatt sure, it is JFI :)

anyway I agree that better keep it as is.

src/librekey/key_store_g10.c
catap: Thanks!

ronaldtse:
Following this PR it seems we really should separate out librekey and librepgp (OpenPGP packet parsing) with separate tests. It would be much more maintainable that way. Maybe even with separate repos.

src/lib/packet.h
catap: It's question of style... but ok, fixed

catap:
@dewyatt yes, you was right. But it was part of issue. Another one was: it should has hash value that I missed.

Anyway, it works on my laptop.

Can you confirm it?

Also, @dewyatt and @flowher I fixed all remarks. If it's fine let rebase it to last master and merge.

dewyatt:
It seems like gpg 2.1.23 doesn't like the s-exprs and gpg 2.1.15 doesn't like our KBX.

```
$ rm -rf ~/.rnp
$ src/rnpkeys/rnpkeys --gen-key --keystore-format=GPG21
```

gpg 2.1.23 (Arch)
```
$ gpg --homedir=~/.rnp --list-keys
/home/daniel/.rnp/pubring.kbx
-----------------------------
pub   rsa2048 2017-08-18 [SCEA]
      7C46D7A71D0292148A787787D79F711705F9685B
uid           [ unknown] RSA (Encrypt or Sign) 2048-bit key <daniel@localhost>

$  echo test | gpg --homedir=~/.rnp --sign --armor
gpg: signing failed: Invalid S-expression
-----BEGIN PGP MESSAGE-----

gpg: signing failed: Invalid S-expression
```

gpg 2.1.15 (Ubuntu Zesty)
```
$ docker run -ti --rm -v $HOME/.rnp:/root/.gnupg ubuntu:zesty gpg --list-keys
gpg: WARNING: unsafe ownership on homedir '/root/.gnupg'
gpg: parse_keyblock_image: more signatures than found in the meta data
gpg: keydb_get_keyblock failed: Invalid keyring
```

catap:
@dewyatt strange, it works on my mac with GnuPG 2.1.23:
```sh
➜  rnp git:(librekey) ✗ rm -rf ~/.rnp                                  
➜  rnp git:(librekey) ✗ ./src/rnpkeys/rnpkeys --format=GPG21 --generate-key
[pgp_mem_readfile() misc.c:750] can't open "/Users/catap/.rnp/pubring.gpg"
cannot read pub keyring
Generating a new key...
signature  2048/RSA (Encrypt or Sign) 217ea6c27a19736e 2017-08-18 [] 
Key fingerprint: a625 93fc f3a6 292e e45a 02c1 217e a6c2 7a19 736e 
uid              RSA (Encrypt or Sign) 2048-bit key <catap@localhost>
Enter passphrase for 217ea6c27a19736e: 
Repeat passphrase for 217ea6c27a19736e: 
➜  rnp git:(librekey) ✗ gpg --version                                      
gpg (GnuPG) 2.1.23
libgcrypt 1.8.0
Copyright (C) 2017 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Home: /Users/catap/.gnupg
Supported algorithms:
Pubkey: RSA, ELG, DSA, ECDH, ECDSA, EDDSA
Cipher: IDEA, 3DES, CAST5, BLOWFISH, AES, AES192, AES256, TWOFISH,
        CAMELLIA128, CAMELLIA192, CAMELLIA256
Hash: SHA1, RIPEMD160, SHA256, SHA384, SHA512, SHA224
Compression: Uncompressed, ZIP, ZLIB, BZIP2
➜  rnp git:(librekey) ✗ echo test | gpg --homedir=~/.rnp --sign --armor    
gpg: starting migration from earlier GnuPG versions
gpg: porting secret keys from '/Users/catap/.rnp/secring.gpg' to gpg-agent
gpg: key 217EA6C27A19736E: secret key imported
gpg: migration succeeded
gpg: /Users/catap/.rnp/trustdb.gpg: trustdb created
-----BEGIN PGP MESSAGE-----

owEBUgGt/pANAwAIASF+psJ6GXNuAcsLYgBZlu7ndGVzdAqJATMEAAEIAB0WIQSm
JZP886YpLuRaAsEhfqbCehlzbgUCWZbu5wAKCRAhfqbCehlzbgjrCAClA2OCSotZ
HZomUEudxWSJrWb6FiLP9cRhNegr7s3RhL4dV37PmFEM//Ah1fKmD5+hGPWrZPNB
CRUZAehrHERHRpiPyg+FSdiXaZ/b7Ik56c7K32OUtlGnvsA1NfHqi7opJYpx1pRk
N2gI3ezb+0uZ//MZYAVhASWrQ5Io1PUEobpDhp1N4FbICMmK+9dpVN+DGvOe26JZ
POkPLNGECr6iJMo5M6LFMlft/aEs3VFXL8z4RzzUhLDCIcN4ntfE1gRKc+NaHhR/
KHIvzYyQz7Z4sm0NBjTIz66jjoWAQFCSi/LS4w97dxmzQ0NTchLJf2DjLYnGqm3A
aNamlvubCBNq
=D/jc
-----END PGP MESSAGE-----
➜  rnp git:(librekey) ✗ 
```

catap:
@ronaldtse moving it as separated repository is a good idea, but it request a lot of pgp low-level stuff and I feel that before we should make very-very big refactoring.

I tried do it as volunteer and after 4-5 hours of hell I give up and realised that it costs about of month.

dewyatt:
@catap I think you mixed up the argument names (`--format` vs `--keystore-format`), so it's importing the .gpg. We need to get better argument validation on our CLI utils. :(

> @ronaldtse moving it as separated repository is a good idea, but it request a lot of pgp low-level stuff and I feel that before we should make very-very big refactoring.

Agreed, it would be really nice to get everything separated but would also be a *lot* of work, probably requiring building a bunch of new abstractions, etc. I do think we have made some progress towards this, but it still remains a fairly significant undertaking.

catap:
@dewyatt, nope, I didn't. I used this command to debug it.

```sh
➜  rnp git:(librekey) ✗ gpg --homedir=~/.rnp --list-keys
/Users/catap/.rnp/pubring.gpg
-----------------------------
pub   rsa2048 2017-08-18 [SCEA]
      A62593FCF3A6292EE45A02C1217EA6C27A19736E
uid           [ unknown] RSA (Encrypt or Sign) 2048-bit key <catap@localhost>

➜  rnp git:(librekey) ✗ find ~/.rnp                     
/Users/catap/.rnp
/Users/catap/.rnp/.gpg-v21-migrated
/Users/catap/.rnp/private-keys-v1.d
/Users/catap/.rnp/private-keys-v1.d/C2EB77E4701B4EF0AE7728A27F4C29A3DC0DDA46.key
/Users/catap/.rnp/pubring.gpg
/Users/catap/.rnp/S.gpg-agent
/Users/catap/.rnp/S.gpg-agent.browser
/Users/catap/.rnp/S.gpg-agent.extra
/Users/catap/.rnp/S.gpg-agent.ssh
/Users/catap/.rnp/S.scdaemon
/Users/catap/.rnp/secring.gpg
/Users/catap/.rnp/trustdb.gpg
➜  rnp git:(librekey) ✗ 
```

catap:
Hm, a lot of gpg files. Digging.

catap:
@dewyatt confirmed, I mixed up the name. Ok, going back to hell.

catap:
@dewyatt ok, found this issue. Can you try because it works on me with right CLI arguments:

```sh
➜  rnp git:(librekey) ✗ echo test | gpg --homedir=~/.rnp --sign --armor
-----BEGIN PGP MESSAGE-----

owEBUgGt/pANAwAIARbiyWoUmB5LAcsLYgBZlvlmdGVzdAqJATMEAAEIAB0WIQQn
QX3I41Cxw6E82/sW4slqFJgeSwUCWZb5ZgAKCRAW4slqFJgeS0adB/9tmNt+YRqe
kmqkkR6AN51zxX8NMSdDbdyPi9OLJboiSEqVNe52CYXm7sqDsbQFzH9+4ta6koh9
im+11ZVYhYiDRcOwGv9NV6IRTseDlhF3irv+PwDWwj1rBrDSmbx597asWDF73J2k
9SMOMznmFs9GBfeU26CH7N/1GH0LYj/3NB230rYc/uJpsutqALepV1G8NtkqEfZj
NFvFmf9/GxtCkR8v7furgVf8ls+i7QLUTwvvmXO1H7BFhlh7aKgqmaHr01ullxST
5gMWu6hVdCnMf1wEEHFTDR7lAO+N0SKf2nluhmZ9k0E/hQvJpCb1SZNtL6obu6ao
Y7RDb2b0mVCW
=VDja
-----END PGP MESSAGE-----
➜  rnp git:(librekey) ✗ find ~/.rnp
/Users/catap/.rnp
/Users/catap/.rnp/private-keys-v1.d
/Users/catap/.rnp/private-keys-v1.d/DE68D62B35A4B6608C06445AE73CC021977F0B5B.key
/Users/catap/.rnp/pubring.kbx
/Users/catap/.rnp/S.gpg-agent
/Users/catap/.rnp/S.gpg-agent.browser
/Users/catap/.rnp/S.gpg-agent.extra
/Users/catap/.rnp/S.gpg-agent.ssh
/Users/catap/.rnp/S.scdaemon
/Users/catap/.rnp/trustdb.gpg
➜  rnp git:(librekey) ✗ 
```

dewyatt:
Excellent, it's working for me with gpg 2.1.23 now!

(gpg 2.1.15 still complains about our KBX, but I'll file a separate issue)

So, rebase time I guess.

catap:
rebased

dewyatt:
LGTM and merging.
I don't think the leaks in https://github.com/riboseinc/rnp/pull/368#discussion_r132817409 were addressed but it can be done later (and IMO it wouldn't hurt to collapse those conditionals into a single one, for readability).

dewyatt:
It looks like one thing was broken in the rebase, will do a PR and add a test.

Issue #367, opened 2017-08-08
When generating ECDSA signaure deterministicly same hash algorithm must be used for HMAC_DRBG and signing


henrydcase:
## Description

By default ``rnp`` uses RFC6979 for ECDSA signature generation. Currently we call ``botan_pk_op_sign_create(&signer, key, "Raw", 0)`` in order to create ECDSA signer. This call will always initialize HMAC_DRBG to SHA-512.

As per 3.3 of RFC 6979 we should initialize HMAC_DRBG with same hash algorithm as the one used for signing.

SHA2-512 seems to be safe default, but would be better to explicitly set an algorithm.

flowher:
Alright. With this feature in botan we don't need anymore  #370. I'll take care of that

randombit:
Fixed by #729

PR #369, opened 2017-08-09
Use same hash algorithm for HMAC_DRBG in RFC6797 as for message sign

flowher:
For details, please see GH #367 

I think it will also solve GH #364


flowher:
argh... no. EMSA1 will hash the digest again. That's not what we want.

PR #370, opened 2017-08-09
ECDSA Reduce message size to base point order before signing

### LINKED ISSUE #364: pgp_ecdsa_sign_hash fails certain digest/curve combinations

flowher:
Fixes #364 (see ticket for more details)

### DIFF
--- a/src/lib/ecdsa.c
+++ b/src/lib/ecdsa.c
@@ -199,19 +199,23 @@ pgp_ecdsa_sign_hash(pgp_ecc_sig_t *         sign,
                     const pgp_ecc_seckey_t *seckey,
                     const pgp_ecc_pubkey_t *pubkey)
 {
-    botan_pk_op_sign_t signer = NULL;
-    botan_privkey_t    key = NULL;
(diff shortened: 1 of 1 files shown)


dewyatt:
Would be nice to have randombit's input too, but otherwise I think we ought to merge.

ronaldtse:
True. @dewyatt could you approve and merge?

PR #377, opened 2017-08-13
Change key flags JSON to be an array of strings.

### LINKED ISSUE #35: Key flags are not really handled

dewyatt:
Just a minor change to address [this comment](https://github.com/riboseinc/rnp/pull/285#issuecomment-313824599). I didn't do proper error handling at the call site here because that whole function needs a lot of rework with regards to that, and there are much higher priority items in rnp.

I also renamed the "flags" entry in the JSON to "key flags" to fit better with f10588d25a11e737aacd0a5e14115cc4d2cf30f7.

I think we can close #35 with this.

### DIFF
--- a/src/lib/packet-print.c
+++ b/src/lib/packet-print.c
@@ -589,6 +589,30 @@ format_key_usage(char *buffer, size_t size, uint8_t flags)
     return true;
 }
 
+static bool
+format_key_usage_json(json_object *arr, uint8_t flags)
+{
+    static const pgp_bit_map_t flags_map[] = {
+      {PGP_KF_ENCRYPT, "encrypt"},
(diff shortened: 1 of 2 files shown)


ronaldtse:
@ni4 @flowher any comments here? If not maybe we should just merge it. Thanks!

PR #382, opened 2017-08-14
Fix #374 warning.

### LINKED ISSUE #374: WARNING: signature made with encryption key
## Description

This warning is printed based on a check that is no longer valid after the subkey rework. The check should probably be changed to something like `if (!pgp_key_can_sign(...))` and the message updated as well.

https://github.com/riboseinc/rnp/blob/ab7546e9f6d9a115f25e15ed9bdca8b83c0506a3/src/lib/validate.c#L307

https://github.com/riboseinc/rnp/blob/ab7546e9f6d9a115f25e15ed9bdca8b83c0506a3/src/lib/validate.c#L488

## Steps to Reproduce

1. src/rnp/rnp --verify somefile.gpg

## Expected Behavior

No warning, unless the key that made the signature has key flags indicating it should not be signing things.

## Actual Behavior

An inaccurate warning is printed.

dewyatt:


ronaldtse:
@dewyatt looks like `sigkey` is now unused and causing the build error. Remove them? Thanks!

ronaldtse:
We should also align the build command `[-Werror,-Wunused-variable]`

dewyatt:
Ah I failed to stage those hunks. Fixed.

dewyatt:
> We should also align the build command [-Werror,-Wunused-variable] for our Makefile so the same behavior happens on CI and local.

@ronaldtse Was just checking back on this. CI just uses autoreconf, ./configure, make. So it should generally match (sanitize builds have some extra flags in there of course). Let me know if you are seeing something different. Some of the CI builds use clang, some use gcc, and obviously they aren't always going to match our local versions either so that may be what you are seeing.

ronaldtse:
Thanks for checking. I thought the build process must have been somewhat different because the unused variables caused a build error in CI due to those flags, but probably not on a development machine. As long as they are aligned it's good.

Issue #39, opened 2017-03-28
Reboot passphrase handling (was: Passphrase communication is inconsistent)


dewyatt:
The mechanisms for providing passphrases seem a bit inconsistent.

The netpgp_* layers have a "pass-fd" variable that can be used to communicate a file descriptor where a passphrase can be written by client software when needed.

The lower-level pgp_* layers typically have a passfp parameter that is a file pointer where a passphrase can be written.

Additionally there are some outliers like:

```c
pgp_memory_t *
pgp_decrypt_buf(pgp_io_t *io,
                        const void *input,
                        const size_t insize,
                        pgp_keyring_t *secring,
                        pgp_keyring_t *pubring,
                        const unsigned use_armour,
                        const unsigned sshkeys,
                        void *passfp,
                        int numtries,
                        pgp_cbfunc_t *getpassfunc)
```

I'm not sure why this function has both a passfp and a getpassfunc callback parameter. On the ruby side I've not had any issues with sending a nil getpassfunc parameter and just using passfp.

Personally I think a callback mechanism might make the most sense for providing a passphrase, like gpgme does. A callback also clearly communicates that a passphrase is required, unlike the unidirectional pipe pass-fd/passfp.

(Side note: pgp_parse does use a callback PGP_GET_PASSPHRASE, but everything else generally does not)

ronaldtse:
It would be the easiest if we consolidate all of these functions (that could take passphrases) to only support only one passphrase passing method.

For example, we can implement `pass-fd` or `passfp` using the callback method anyway.

What do the others think? Pinging expert @randombit , @shaunwheelhouse @zgyarmati .

(`pgp_decypt_buf` itself seems to take way too many unnecessary arguments too.)

ronaldtse:
In addition, in places that passphrases are checked, the error code should be consistently returned to the user indicating passphrase mismatch.

zgyarmati:
Ah, i made a typo in my commit msg, so reopening this, sorry for the noise!

ronaldtse:
@riboseinc/rnp Let's use this to settle on a "correct" passphrase handling structure that is solid and secure.

dewyatt:
Done!

Issue #165, opened 2017-06-01
Audit for `pgp_forget()` to make sure passphrases are not kept


ronaldtse:
From @catap : 

> Made an audit because sometime we doesn't call pgp_forget() for passphase for example here:
> https://github.com/riboseinc/rnp/blob/a771df4524dca6c23d5e1977e29de18ceb6f8605/src/lib/packet-parse.c#L1129

Related to #39

ni4:
Already fixed.

PR #386, opened 2017-08-14
Ensure pgp_forget always zeros things

### LINKED ISSUE #263: find_passphrase is not guaranteed to clear sensitive data

dewyatt:
Fixes #263 

Two things here:

* Use `explicit_bzero`/`explicit_memset`/`memset_s` in `pgp_forget` to ensure that the crucial piece of code is not optimized out. I pulled in a public domain implementation of `explicit_bzero` from libressl as a fallback. It's built as a separate static lib with no optimization (you'll probably see both -O2 and -O0 on the command line but the last option takes effect - room for improvement there). Most BSDs and Linux w/glibc >= 2.25 have `explicit_bzero` or `explicit_memset` available. I *think* macOS has `memset_s`. For Windows we'll probably eventually want to use `SecureZeroMemory`. None of these functions are very portable, but it's a start.
* Add a few more `pgp_forget` calls in to `find_passphrase`. The `find_passphrase` function is still a bit wonky and could definitely be improved but I'm thinking I may rework it as part of #39.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -16,7 +16,6 @@ AC_REVISION([m4_esyscmd_s([git describe --always])])
 AC_DEFINE([GIT_REVISION],["m4_esyscmd_s([git describe --always])"],
 [Git revision generated when autoreconf was called])
 
-
 AS_SHELL_SANITIZE
 
 AC_CONFIG_SRCDIR([src/rnp/rnp.c])
@@ -91,6 +90,18 @@ AC_TYPE_UINT32_T
 AC_TYPE_UINT64_T
 AC_TYPE_UINT8_T
 
+# Check for a secure memory wipe implementation.
+AC_CHECK_FUNCS([explicit_bzero explicit_memset memset_s])
+AS_IF([test $ac_cv_func_explicit_bzero = no &&
+       test $ac_cv_func_explicit_memset = no &&
+       test $ac_cv_func_memset_s = no],
+      [
+        AM_CONDITIONAL([USE_BUNDLED_EXPLICIT_BZERO], [true])
+        AC_DEFINE([USE_BUNDLED_EXPLICIT_BZERO], [1], [use the bundled explicit_bzero])
+      ], [
+        AM_CONDITIONAL([USE_BUNDLED_EXPLICIT_BZERO], [false])
+      ])
+
 # Check for Botan
 AX_CHECK_BOTAN(, [AC_MSG_ERROR([Missing Botan])])
 AX_CHECK_BOTAN_DEFINES()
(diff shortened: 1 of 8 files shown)


dewyatt:
Also I didn't spend the time to add any tests for the fallback implementation, but we really ought to soon.  OpenBSD has [some tests](http://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/src/regress/lib/libc/explicit_bzero/explicit_bzero.c?rev=1.6&content-type=text/plain) (related [article](https://plus.google.com/+MatthewDempsky/posts/KQHFBouxurX)) for it that we may eventually want to consider.

Some additional safeguards are mentioned [here](https://github.com/libressl-portable/openbsd/issues/5) along with some valid warnings.

ronaldtse:
@dewyatt could you help make the issues for the remaining TODOs so we don't forget them?

It's probably a good time to revisit #165 to ensure all the passphrases are forgotten too.

flowher:
LGTM

flowher:
Interesting. I'm often use the approach from here (if I know the compiler) (https://www.securecoding.cert.org/confluence/display/CINT/MSC06-C.+Beware+of+compiler+optimizations).

Having those tests ported would be really good think, as it's kind of tricky to ensure this implementation always works.

dewyatt:
> @dewyatt could you help make the issues for the remaining TODOs so we don't forget them?

Yes, will do.

> Having those tests ported would be really good think, as it's kind of tricky to ensure this implementation always works.

Definitely, it's vital to test this or it can just give us a false sense of security. Filed https://github.com/riboseinc/rnp/issues/389. I'm just in a bit of a rush as I'm working on a few different pieces of rnp right now but I will circle back to add those tests if nobody else does it soon.

PR #391, opened 2017-08-15
Use botan_scrub_mem in pgp_forget.

dewyatt:
Ref: https://github.com/riboseinc/rnp/issues/387 and https://github.com/riboseinc/rnp/pull/386

This is basically a revert of PR 386 and then the additions to use `botan_scrub_mem` instead, as pointed out by @randombit in https://github.com/riboseinc/rnp/issues/387#issuecomment-322536112. I remember seeing this function but apparently completely forgot about it.

Pros:

* Much simpler.

Cons:

* `botan_scrub_mem`/`secure_scrub_memory` doesn't have very thorough [tests](https://github.com/randombit/botan/blob/7ab9ab26ab19c26428932954fea47efa284a77d8/src/tests/test_ffi.cpp#L54) when compared to the [tests](http://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/src/regress/lib/libc/explicit_bzero/explicit_bzero.c?rev=1.6&content-type=text/plain) of the OpenBSD implementation I had pulled in. But I'm sure that will be improved, given a bit of time.


randombit:
@dewyatt Those explicit_bzero tests look really thorough, thanks for pointing those out. This is a fairly hard function to test well since the whole point is that for any conforming C program it is indistinguishable from using plain `memset` (in that if you look at the buffer afterwards to confirm it was zeroed, then the memset is side-effecting and the compiler won't elide it.)

Issue #309, opened 2017-07-17
Extract out PGP packet parsing code to separate library


ronaldtse:
## Description

PGP packet parsing is well-defined and the code can be self-contained.

This task is to extract out PGP parsing code to a separate library, e.g., `librepgp`.

Maybe these functions should be prefixed with `repgp_`?

flowher:
Ok, ``repgp_`` it's gonna be

ronaldtse:
Or:
1. `librepgp`
2. `libpgp`
3. `librgp` (`librpg` is already used)

Please vote: 😄  for 1, 🎉  for 2, ❤️  for 3.

ronaldtse:
Without broader participation it looks like `librepgp` is the likely candidate 😉

PR #392, opened 2017-08-15
Move crypto implementation to separated directory

flowher:
Towards #309 .

Mostly cleanup, patch doesn't change any functionality
* Moves crypto algorithms implementation/wrapping to separated directory
* Removes ``pgp_parse_hash_`` function definitions as it's not needed
* Removes ``rnpdigest.h`` file


ronaldtse:
Very welcome change @flowher . Thanks!

PR #395, opened 2017-08-16
Add support for SM2 encryption

randombit:
Adds support for SM2 encryption. (#127)

The SM2 ciphertext has a specific form (it's a triplet of an ECC point, the masked plaintext, and a SM3 hash that serves as an authenticator). However here it's just formatted as a BIGNUM holding an opaque block of bits. Which I think is fairly consistent with OpenPGPs approach to such things (eg using a single BIGNUM to hold ECC points).

Testing on the command line, keygen and encrypt/decrypt round trip work ok. Of course there is no other implementation to test this against. so maybe I made some dumb mistake.

I changed `rnpkeys` so SM2 signature key get an encryption subkey. I'm not totally sure this is the correct approach.

### DIFF
--- a/src/lib/crypto/sm2.c
+++ b/src/lib/crypto/sm2.c
@@ -36,10 +36,8 @@
 #include "utils.h"
 #include "utils.h"
 
-extern ec_curve_desc_t ec_curves[PGP_CURVE_MAX];
-
 pgp_errcode_t
-pgp_sm2_genkeypair(pgp_seckey_t *seckey, pgp_curve_t curve)
+pgp_sm2_genkeypair(pgp_seckey_t *seckey, pgp_curve_t curve_id)
 {
     /**
      * Keeps "0x04 || x || y"
@@ -48,14 +46,21 @@ pgp_sm2_genkeypair(pgp_seckey_t *seckey, pgp_curve_t curve)
      * P-521 is biggest supported curve for SM2, though ordinarilly
      * it uses the 256-bit SM2 recommended curve.
      */
+    const ec_curve_desc_t *curve = get_curve_desc(curve_id);
+
     uint8_t         point_bytes[BITS_TO_BYTES(521) * 2 + 1] = {0};
-    const size_t    filed_byte_size = BITS_TO_BYTES(ec_curves[curve].bitlen);
     botan_privkey_t pr_key = NULL;
     botan_pubkey_t  pu_key = NULL;
     botan_rng_t     rng = NULL;
     BIGNUM *        public_x = NULL;
     BIGNUM *        public_y = NULL;
     pgp_errcode_t   ret = PGP_E_C_KEY_GENERATION_FAILED;
+
+    if (curve == NULL) {
+        goto end;
+    }
+
+    const size_t filed_byte_size = BITS_TO_BYTES(curve->bitlen);
 
     if (botan_rng_init(&rng, NULL)) {
         goto end;
@@ -65,7 +70,7 @@ pgp_sm2_genkeypair(pgp_seckey_t *seckey, pgp_curve_t curve)
     * SM2 encryption and signature keys share the same form, only difference
     * is the OID used for the X.509 encoding (which is not used by OpenPGP).
     */
-    if (botan_privkey_create(&pr_key, "SM2_Sig", ec_curves[curve].botan_name, rng)) {
+    if (botan_privkey_create(&pr_key, "SM2_Sig", curve->botan_name, rng)) {
         goto end;
     }
 
@@ -99,6 +104,7 @@ pgp_sm2_genkeypair(pgp_seckey_t *seckey, pgp_curve_t curve)
     const size_t y_bytes = BN_num_bytes(public_y);
 
     // Safety check
+
     if ((x_bytes > filed_byte_size) || (y_bytes > filed_byte_size)) {
         RNP_LOG("Key generation failed");
         goto end;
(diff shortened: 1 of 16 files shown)


src/lib/crypto/sm2.c
```diff
@@ -269,3 +273,124 @@ pgp_sm2_verify_hash(const pgp_ecc_sig_t *   sign,
...
+        goto done;
+    }
+
+    if (botan_mp_init(&public_x) || botan_mp_init(&public_y) ||
+        botan_mp_from_bin(public_x, &point_bytes[1], point_len) ||
+        botan_mp_from_bin(public_y, &point_bytes[1 + point_len], point_len)) {
+        goto done;
+    }
+
+    const char *curve_name = ec_curves[pubkey->curve].botan_name;
```
flowher: Can you use get_ec_curve() instead? This will crash if pubkey->curve has unexpected value

src/lib/crypto/sm2.c
```diff
@@ -269,3 +273,124 @@ pgp_sm2_verify_hash(const pgp_ecc_sig_t *   sign,
...
+                const pgp_ecc_seckey_t *privkey,
+                const pgp_ecc_pubkey_t *pubkey)
+{
+    botan_pk_op_decrypt_t decrypt_op = NULL;
+    botan_privkey_t       key = NULL;
+    botan_rng_t           rng = NULL;
+    pgp_errcode_t         retval = PGP_E_FAIL;
+
+    if (botan_privkey_load_sm2_enc(
+          &key, privkey->x->mp, ec_curves[pubkey->curve].botan_name)) {
```
flowher: Again, better to use ``get_curve_desc``

src/lib/crypto/sm2.c
```diff
@@ -269,3 +273,124 @@ pgp_sm2_verify_hash(const pgp_ecc_sig_t *   sign,
...
+done:
+    free(ctext_buf);
+    botan_pk_op_encrypt_destroy(enc_op);
+    botan_pubkey_destroy(sm2_key);
+    botan_rng_destroy(rng);
+
+    return retval;
+}
+
+pgp_errcode_t
```
flowher: I'm trying to remove ``pgp_errorcode_t`` type. Can you use ``rnp_result`` instead? (see rnp_def.h for return values)

randombit: Changed all declarations in sm2.h to return `rnp_result`, and updated callers

src/lib/crypto/sm2.c
randombit: Didn't know about this func, much better - fixed

ronaldtse:
LGTM, we should probably add some CLI tests regarding the generation of SM2 keys and roundtripping?

ni4:
@ronaldtse I created #399.

ronaldtse:
Thanks @ni4 !

src/lib/packet.h
```diff
@@ -377,6 +377,7 @@ typedef enum {
                                * IETF-S/MIME) */
     PGP_PKA_EDDSA = 22,       /* EdDSA from draft-ietf-openpgp-rfc4880bis */
 
+    PGP_PKA_SM2_ENCRYPT = 98, /* SM2 encryption */
```
dewyatt: I had a late question, are we sure it's best to add SM2 encryption as a separate alg? If SM2 can be used for both sign+encrypt+etc, we really only need one entry, right?

It would be similar to RSA, ignoring the deprecated 0x2 and 0x3, there's just one entry and it can be used for everything.

ronaldtse: SM2 has separate modes for DSA and Public Key Encryption, it might be clearer in implementation to not mix them up?

randombit: Yeah I wasn't so sure about this point either. In general, best practice is to not use the same key for both encryption + signing because sometimes there are strange cross-effects that can occur. But it's not really clear from the docs that I have seen what the convention is for SM2. One hint however is that there are distinct OIDs assigned for SM2 encryption, signature, and key exchange - that to me suggests that each keypair is only supposed to be used for one of the operations.

dewyatt: I definitely get that and it makes sense, but I'm wondering if we should be thinking about this at a higher level. Like if I changed the question to: What would RFC 4880 9.1 actually look like if SM2 were added by the WG?

I get the impression that usage-specific PK algorithm identifiers are deprecated in the RFC in favor of key flags, so I'm not sure if they would add it separately like this.

From RFC 4880:

>    There are algorithm types for RSA Sign-Only, and RSA Encrypt-Only keys.  These types are deprecated.  The "key flags" subpacket in a signature is a much better way to express the same idea, and generalizes it to all algorithms.

Also, sort of a different point, but if we look at how RSA is treated in gpg, for example, it defaults to the best practice of generating a primary signing key and an encrypting subkey. So that's the default but it does *permit* you to generate an RSA key that both signs+encrypts. It's up to the user to ultimately decide and accept the risk.

Anyways, I don't have a strong opinion on it, it just sort of jumped out at me as slightly divergent.

randombit: @dewyatt That definitely all makes sense to me. FTR I'm not strongly in favor of one approach vs the other. @ronaldtse your take?

ronaldtse:
I agree too. Let's use one ID for SM2. We just want to be sure we can generate the best practice subkey structure :-)

randombit:
OK I'll make that change and have SM2 keys be a signing primary + encryption subkey by default.

ronaldtse:
@randombit just to be sure have we consolidated the IDs for SM2? Thanks!

randombit:
@ronaldtse Not yet, I should have a patch for that today.

Issue #238, opened 2017-06-20
Implement file encryption with password


ni4:
## Description

OpenPGP standard allows file encryption with password(s), without using the public keys.

ronaldtse:
I know both botan and OpenSSL offer utility functions for standalone file encryption, but maybe we can make it simpler to use?

ni4:
I'm not sure what encryption standard is used by botan/OpenSSL (CMS?). But at least OpenPGP password-encrypted document can be decrypted by GnuPG on any *nix system or by any PGP software or library (PGP Desktop, whatever else) on Windows.

ronaldtse:
@ni4 could you take on this task? We could use this on a project by end of this month. Thanks!

ni4:
@ronaldtse Ok, sure.

ronaldtse:
@ni4 just to confirm that stream processing will mean (given a block size of b, available bytes of a, and a total length of n inside a packet):
* `0 < a < b`  => queued
* `b <= a < n`   => process block (encrypt / decrypt)
* fully available packet (`a == n`) => process packet (encrypt / decrypt)

Right?

ni4:
@ronaldtse What do you mean under the block size?
Under the streamed processing I mean that data is read from the source, processed, and written out in one pass and using the constant-size cache (say, 32kb). It doesn't depend on packet sizes/partial data lengths and type of the input - the only requirement for the input is to be able to read data, and inform about the end of data. So, after initialization, in cycle we read some amount of data from input, then decrypt, then decompress, and then write to the output.

ronaldtse:
Ah I got confused, thanks @ni4 .

ni4:
Done in #463

PR #401, opened 2017-08-17
Cleanups in symmetric cipher code

randombit:
The big change here is replacing the combination of `pgp_crypt_any`, `pgp_decrypt_init`, `pgp_cipher_set_iv` and `pgp_cipher_set_key` with the combination function `pgp_cipher_start` which does all of these operations on one go. Also while I could still not make the cipher structure opaque I at least change all code to not touch any fields in it except (in a couple of spots) `alg`.

Also removed several completely unused functions `pgp_decrypt_se`, `pgp_encrypt_se`, `pgp_write_symm_enc_data` and  `pgp_cipher_block_decrypt`, and removed `pgp_cipher_block_encrypt` since it was only called in one spot that was accessing internal structures of the cipher struct, so I hide the operation behind a new function `pgp_cipher_cfb_resync_v2` (which could use a better name).

More changes in this area coming soon but this seems like a big enough change to start.

Important question: I have long been confused by `iv`, `siv`, and `civ` arrays in the `pgp_crypt_t` struct. Logically, only one feedback register is required to implement CFB. What are the other two for? Now having removed several (already unused/never called) functions, it becomes clear that the only function which actually use `siv` or `civ` is `pgp_cipher_cfb_resync`. That is to say, resync moves a register around, but does not modify the register that is actually being used during encryption/decryption (which is `iv`). Which suggests to me that symmetric crypto is not matching OpenPGP format. Anyhow this PR shouldn't make the problem any worse than it already is.

### DIFF
--- a/src/lib/crypto.c
+++ b/src/lib/crypto.c
@@ -173,7 +173,7 @@ pgp_decrypt_decode_mpi(uint8_t *           buf,
     case PGP_PKA_SM2_ENCRYPT:
         BN_bn2bin(encmpi, encmpibuf);
 
-        size_t out_len = buflen;
+        size_t     out_len = buflen;
         rnp_result err = pgp_sm2_decrypt(buf,
                                          &out_len,
                                          encmpibuf,
--- a/src/lib/generate-key.c
+++ b/src/lib/generate-key.c
@@ -286,7 +286,7 @@ pgp_generate_primary_key(rnp_keygen_primary_desc_t *desc,
     bool          ok = false;
     pgp_output_t *output = NULL;
     pgp_memory_t *mem = NULL;
-    pgp_seckey_t seckey;
+    pgp_seckey_t  seckey;
 
     memset(&seckey, 0, sizeof(seckey));
 
--- a/src/lib/packet-create.c
+++ b/src/lib/packet-create.c
@@ -443,7 +443,9 @@ write_seckey_body(const pgp_seckey_t *key, const uint8_t *passphrase, pgp_output
         RNP_LOG("Memory allcoation failed");
         return false;
     }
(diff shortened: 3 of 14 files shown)


dewyatt:
> removed several completely unused functions pgp_decrypt_se, pgp_encrypt_se, pgp_write_symm_enc_data

I'm thinking these were part of an implementation of https://github.com/riboseinc/rnp/issues/238 that was never completed in OpenPGP-SDK.

> pgp_cipher_cfb_resync [...] suggests to me that symmetric crypto is not matching OpenPGP format

I think you're right that `pgp_cipher_cfb_resync` is broken and a test I did seems to confirm that. AFAICT the impact of this right now is that we can't decrypt/use encrypted V3 keys. It also means we can't decrypt any of the older RFC tag 9 encrypted data. I *think* that's mostly used for #238 (`gpg -c` with versions before 2.1, after that SEIP/MDC tag 18 is used, which doesn't use the CFB resync) so we kind of already don't support it.

dewyatt:
(I'll file a separate issue about the above so as not to detract from this PR)

ronaldtse:
Thanks @randombit !

randombit:
BTW I checked the initial import of NetPGP in 89d0d6e957 and it does seem to have the same issue with CFB resync.

dewyatt:
> BTW I checked the initial import of NetPGP in 89d0d6e and it does seem to have the same issue with CFB resync.

Same here (I think I checked a few commits after that). I didn't check further back in CVS or anything, I do kind of wonder if it was always broken, but I guess it doesn't matter. It might be worth confirming the issue in NetBSD and reporting it upstream at some point.

PR #402, opened 2017-08-17
Removes memory leaks from compress.c

flowher:
* Adds header file for compress.h
* Cleans up compress.c to remove memory leaks
* Adds test which checks received packets after decompression


ronaldtse:
Thank you @flowher !

PR #409, opened 2017-08-19
Introduces librepgp and moves type definitions closer to implementation

flowher:
This patch:

- introduces initial version of ``librepgp`` - library with files moved from ``lib``. Work is ongoing

- moves types closer to implementation that uses those types. 
Motivation here is following - if compilation unit needs to use "dh.c" it should only need to ``#include`` "dh.h" and link with dh.o. In current state, each compilation unit needs to include ``packet.h`` which contains a lot of not always needed stuff. This creates dependencies between compilation units which start to become quite complicated. 
The other problem that we have - currently some public include files contain ``#include "packet.h"`` and ``packet.h`` includes a lot of ``.h`` files which contain implementation details that not necessarily we should be exporting as public header. Long term goal is to fix this situation. In future PRs I'll remove ``packet.h``, move some other bits and pieces closer to implementation and move common types into ``types.h``. This will improve (but not completely fix) this situation. Real solution IMHO is to introduce opaque data types.
 
- moves definitions from RFC 4880 to separated file.
My idea here is that ``librepgp`` will kind of "own" all definitions from RFC (like ID of the algorithms, modes, etc..). New file ``rnp_repgp_def.h`` contains all those definitions (moved from ``packet.h``) and   
must not include any other file, so that it can be kept independent from ``rnp`` internals.

Important for reviewers - this patch is kind of big and boring to review, but it doesn't provide any new implementation. It's just moving code in different places (maybe few function changed their names). 

Other changes: 
 * Moves files for packet printing and parsing to librepgp
 * Adds repgp_show_packet_tag to librepgp interface
 * Move PGPV_BIGNUM definition in right place
 * Decouple packet.h and dsa.h
 * Decouple packet.h and rsa.h
 * Decouple packet.h and elgamal.h
 * Moves DEFAULT_ values to rnp_def.h
 * Moves all defines from RFC to rnp_repgp_def.h


dewyatt:
This looks ok, but just a few things:

* There's a regression here with `pgp_export_key`. I *think* that's the only one.
* Leading underscores should generally be avoided. (See the C99 standard or [CERT guidelines](https://www.securecoding.cert.org/confluence/display/c/DCL37-C.+Do+not+declare+or+define+a+reserved+identifier), for example).
* The prefix `RNP_REPGP` used in a few places seems strange to me if rnp and librepgp will eventually split.

ronaldtse:
@flowher Great start with the restructure!

Agree with @dewyatt to remove `RNP_` from librepgp code and also filenames e.g., (`repgp/rnp_repgp.h`, `repgp/rnp_repgp_def.h`).

catap:
@agreed with @dewyatt remarks and yes, it looks like a very good start!

flowher:
Thanks guys for the comments! I think all of them addressed now.

PR #413, opened 2017-08-21
G10 keystore doesn't keep a key metadata likes `key_flags`.

catap:
So we can't be sure does this key can be used to signing.

This commit just switched to using pubring that is GPG or KBX and keeps metadata.

### DIFF
--- a/src/lib/rnp.c
+++ b/src/lib/rnp.c
@@ -1141,6 +1141,7 @@ rnp_sign_file(rnp_ctx_t * ctx,
     int                 attempts;
     int                 ret;
     int                 i;
+    unsigned            from;
 
     io = ctx->rnp->io;
     if (f == NULL) {
@@ -1154,6 +1155,12 @@ rnp_sign_file(rnp_ctx_t * ctx,
     if (!pgp_key_can_sign(keypair) &&
         !(keypair = find_suitable_subkey(keypair, PGP_KF_SIGN))) {
         RNP_LOG("this key can not sign");
+        return RNP_FAIL;
+    }
+    // key exist and might be used to sign, trying get it from secring
+    from = 0;
+    if ((keypair = rnp_key_store_get_key_by_id(
+           io, ctx->rnp->secring, keypair->keyid, &from, NULL)) == NULL) {
         return RNP_FAIL;
     }
     ret = RNP_OK;
@@ -1270,6 +1277,7 @@ rnp_sign_memory(rnp_ctx_t * ctx,
     int                 attempts;
     int                 ret;
     int                 i;
+    unsigned            from;
 
     io = ctx->rnp->io;
     if (mem == NULL) {
@@ -1282,6 +1290,12 @@ rnp_sign_memory(rnp_ctx_t * ctx,
     if (!pgp_key_can_sign(keypair) &&
         !(keypair = find_suitable_subkey(keypair, PGP_KF_SIGN))) {
         RNP_LOG("this key can not sign");
+        return RNP_FAIL;
+    }
+    // key exist and might be used to sign, trying get it from secring
+    from = 0;
+    if ((keypair = rnp_key_store_get_key_by_id(
+           io, ctx->rnp->secring, keypair->keyid, &from, NULL)) == NULL) {
         return RNP_FAIL;
     }
     ret = RNP_OK;
--- a/src/lib/rnp.c
+++ b/src/lib/rnp.c
@@ -1148,7 +1148,7 @@ rnp_sign_file(rnp_ctx_t * ctx,
         return RNP_FAIL;
     }
     /* get key with which to sign */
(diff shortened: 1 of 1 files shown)


ronaldtse:
@dewyatt @ni4 any concerns with this? Would like to move forward. Thanks!

dewyatt:
It would just create an infinite loop, and if it somehow made it past that, it would try to sign with a pubkey and probably crash. I wish we had `cli_tests.py` ready for CI.

catap:
@dewyatt good point, thanks. Will fix it soon.

ronaldtse:
`cli_tests.py` are now available -- any updates @catap @dewyatt ? Just don't want this to get stale. Thanks!

ni4:
@catap I'm not sure that it is a good idea to check pubring for signing keys - we should definitely use secring. And we should not care about the key flags at all - having the secret key we could set any flags and re-sign the key. So, basically, the owner of the secret key can do whatever he wants. Recipient of the message, having only the public key, should care about the key flags and check whether sender was allowed to sign with that key.

dewyatt:
> I haven't got your point about infinity loop. How is it possible?

I did experience an infinite loop before but maybe I had some local changes or something, can't reproduce now. It does still attempt to sign with the pubkey though.

> I'm not sure that it is a good idea to check pubring for signing keys - we should definitely use secring. 

I don't see any issues with searching the pubring for an appropriate key. This is exactly what GPG 2.1+ does, because the (G10/`private-keys-v1.d`) secring doesn't have the metadata duplicated like it used to with the .gpg format. So gpg searches the pubring.kbx for an appropriate key, then confirms with the agent that there is a corresponding seckey.

> And we should not care about the key flags at all - having the secret key we could set any flags and re-sign the key. So, basically, the owner of the secret key can do whatever he wants. Recipient of the message, having only the public key, should care about the key flags and check whether sender was allowed to sign with that key.

I have to disagree, I think part of the usefulness of key flags is preventing the user from shooting themselves in the foot. The user should be guided towards best practices, like using separate keys for encrypting+signing. They can always alter the key flags (once implemented), but this is a very explicit action that the user has to take.

catap:
Anyway, I agree with @dewyatt 's points about keeping and using pubring as much as we can.

But @dewyatt I can't see anyway to reproduce this infinity loop. So, shall we merge this small fix?

ni4:
@dewyatt Now I agree. I was thinking about the keyrings just as storage for key data, not from the user's point of view. 
@catap So you are welcome to merge this :)

ronaldtse:
@catap could you help fix the 2 failures? Thanks!

```
signature  1024/RSA (Encrypt or Sign) fc7ace6caa412c02 2017-08-23 [SC] 
                 6170 7837 822e d5cc 6b71 d1d8 fc7a ce6c aa41 2c02 
uid              sigtest_SHA1
[pgp_key_is_locked() pgp-key.c:772] key is not a secret key
pgp_sig_write: null rsa.d
[  ERROR   ] --- 0 == 0
[   LINE   ] --- generatekey.c:116: error: Failure!
Generating a new key...
pub 1024/RSA (Encrypt or Sign) fc7ace6caa412c02 2017-08-23 [SC] 
                 6170 7837 822e d5cc 6b71 d1d8 fc7a ce6c aa41 2c02 
uid              sigtest_SHA1
sub 1024/RSA (Encrypt or Sign) cc60c34581ec5118 2017-08-23 [E] 
                 9c15 2082 ca6c a207 0628 90ac cc60 c345 81ec 5118 
[  FAILED  ] rnpkeys_generatekey_testSignature
```

And
```
signature  1024/RSA (Encrypt or Sign) 7bc6709b15c23a4a 2017-07-20 [SC] 
                 e95a 3cbf 583a a80a 2ccc 53aa 7bc6 709b 15c2 3a4a 
uid              key0-uid0
uid              key0-uid1
uid              key0-uid2
[pgp_key_is_locked() pgp-key.c:772] key is not a secret key
pgp_sig_write: null rsa.d
signature  1024/RSA (Encrypt or Sign) 7bc6709b15c23a4a 2017-07-20 [SC] 
                 e95a 3cbf 583a a80a 2ccc 53aa 7bc6 709b 15c2 3a4a 
uid              key0-uid0
uid              key0-uid1
uid              key0-uid2
[pgp_key_is_locked() pgp-key.c:772] key is not a secret key
pgp_sig_write: null rsa.d
[  ERROR   ] --- 0 == 0
[   LINE   ] --- key-unlock.c:150: error: Failure!
[  FAILED  ] test_key_unlock_pgp
```

ronaldtse:
Maybe @dewyatt could hint on the problem.

catap:
@ronaldtse ooops, sorry, I missed it.

Anyway, it fixed and pushed. I also rebased it to last master.

ronaldtse:
@dewyatt could you confirm this is good to go? Thanks!

dewyatt:
LGTM

dewyatt:
Ah I just realized we should have used key grip (not keyid) for this for it to work with G10.

Issue #403, opened 2017-08-17
CFB resync is broken


dewyatt:
## Description

See https://github.com/riboseinc/rnp/pull/401 for more discussion.

AFAICT this isn't something that broke recently and may have always been broken in netpgp.

## Steps to Reproduce

One impact of this is that encrypted V3 keys are not usable. To demonstrate, use the keys below (password 'test').

1. `gpg --dearmor < secring.gpg.asc > ~/.rnp/secring.gpg && gpg --dearmor < pubring.gpg.asc > ~/.rnp/pubring.gpg`
2. `echo hi | src/rnp/rnp --sign` (won't work)
3. `echo hi | gpg --homedir=~/.rnp --sign` (will work)

`secring.gpg.asc`
```
-----BEGIN PGP ARMORED FILE-----
Comment: Use "gpg --dearmor" for unpacking

lQGxAzvqRosAAAEDeNMKLJMJQeGC2RG5Nec6R2mzC12N1wGLiYYJCsmSQd1Y8mht
A2Sc+4k/q5+l6GHtfqUR/RTCIIudAZUzrQVIMhHDKF+5de9lsE5QxQS1u43QGVCb
/9IYrOLOizYQ2pkBtD9LCrf7W2DccMEkpQKD8QAFE/8DAwKCmfCK/R7eN9JRAyab
Z9omBAN3pmFiDt0EYNc8dLkVxDKTGr8RNzhN35VUyj4VYrPPbOTsMBbLCLOpdCBF
pphXdGEQI/iDtzznfvXnDvMWxqFCZtAZJkH0J8Z+vEN33Rx0zflmuOs8DMuGRxUn
lk2YQKmveafKnl2k5AxoHt6i4FucAbzofwcVmX7Gb9UW9PK4NSN19WIsnCG2/CL7
kTvoCdMze9fNrpAcfNhJxEgi0H7RO0f1bb5Sh9G8IAG8BogNO/oGBwyc+5C60h80
svQd1xnO5ehnOzcs2ZlskiKorVE3389m3HO59bupxZMVxdL25K/j1ysBsd/9LGgp
loM+OgKGfkaWZkm1HALO7nYoqotutb4xJkDLNw9lYCxRm9oAUQKZ4WtY5W+5BeaV
fzSC6LQRcGdwMi42LjMtdGVzdC1rZXk=
=WcmK
-----END PGP ARMORED FILE-----
```

`pubring.gpg.asc`
```
[...]

PR #414, opened 2017-08-21
Fix decryption of V3 keys

randombit:
This took way longer than I expected because I figured it was just an issue of implementing CFB resync correctly. It turned out there were also two bugs in `reader.c` that meant reading a V3 encrypted key almost certainly never worked (including in NetPGP, based on the history).

First, the resync call happened in the wrong time, which screwed up the IVs. Additionally, when copying the unencrypted v3 MPI lengths, it copied them into an offset. So what ended up happening is RNP would interpret the first 2 bytes of the decrypted modulus (still around in the buffer at offset 0) as the length of the following MPI. This would fail due to either being a seemingly invalid length or else cause corruptions. So it seems like there is no way this could have worked.

Anyway I confirmed that with this change we can correctly decrypt the key in #403 and sign a message that gpg will verify.

CFB resync for tag=9 messages may still be broken, it doesn't seem like `rnp` supports symmetrically encrypted messages?

See #403 and #401

### DIFF
--- a/src/lib/symmetric.c
+++ b/src/lib/symmetric.c
@@ -127,7 +127,6 @@ pgp_cipher_start(pgp_crypt_t *crypt, pgp_symm_alg_t alg, const uint8_t *key, con
     if (cipher_name == NULL)
         return false;
 
-    crypt->offset = 0;
     crypt->alg = alg;
     crypt->blocksize = pgp_block_size(alg);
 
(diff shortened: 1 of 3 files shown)


ronaldtse:
Thank you @randombit for the major step forward!

@ni4 would you be able to add a corresponding CLI test for this case mentioned in #403 ?

dewyatt:
Great work, must have been fun to debug.

ni4:
@ronaldtse Ok, created #419

flowher:
LGTM

PR #415, opened 2017-08-21
Add passphrase provider framework and key locking/unlocking

dewyatt:
Refs: https://github.com/riboseinc/rnp/issues/39 https://github.com/riboseinc/rnp/issues/396

I apologize in advance for the large PR with a single commit.
The main thing here is the addition of a passphrase provider concept, and also key unlocking/locking.
The interfaces were added to librepgp and two providers were added to librnp.

Some random notes:
* A passphrase provider is just a callback and a userdata void pointer.
* There is also a passphrase context, which gives some information on why a passphrase is required. (Like we're generating a key, signing, ...).
* I was able to keep all the tests virtually the same. The `rnp_passphrase_provider_file` provider is used which reads from a pipe as before. No modifications required for `cli_tests.py all`.
* For CLI, the passphrase provider is by default `rnp_passphrase_provider_stdin`. I also added a replacement for `getpass`. Please test (esp. macOS).
* I split `user_input_fp` apart from passphrase handling. So `user_input_fp` is only used for rnpkeys expert mode. There is a separate `passfp` for passphrases specifically.
* I also added the key unlocking/locking functionality as discussed. There is a large test for it too. It's a bit too long, but can be split apart later.
* I removed the CT_ENCRYPTED_*KEY psuedo types and replaced them with a boolean, mostly to simplify some things. (Also renamed the G10 encrypted fields to encrypted_data to accommodate this).

Note: I shuffled things around at the last minute and somewhat hackishly resolved a circular dependency. Can be improved later.

### DIFF
--- a/src/lib/pass-provider.c
+++ b/src/lib/pass-provider.c
@@ -1,0 +1,155 @@
+/*
+ * Copyright (c) 2017, [Ribose Inc](https://www.ribose.com).
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1.  Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *
+ * 2.  Redistributions in binary form must reproduce the above copyright notice,
+ *     this list of conditions and the following disclaimer in the documentation
+ *     and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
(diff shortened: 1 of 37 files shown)


dewyatt:
I should add that it would help if someone with macOS would specifically test commands like `src/rnp/rnp --sign` (which would test `rnp_getpass`).

flowher:
Those circular dependencies are really a pain.

src/lib/pgp-key.c
```diff
@@ -703,24 +690,25 @@ pgp_key_init(pgp_key_t *key, const pgp_content_enum type)
...
+               const pgp_passphrase_provider_t *passphrase_provider)
 {
     pgp_output_t *output;
     pgp_memory_t *mem;
     char *        cp;
 
     RNP_USED(io);
     if (!pgp_setup_memory_write(NULL, &output, &mem, 128)) {
-        RNP_LOG("Can't setup memory write");
+        (void) fprintf(io->errs, "can't setup memory write\n");
```
flowher: There is also RNP_LOG_FD for such cases

src/lib/pass-provider.c
```diff
@@ -0,0 +1,155 @@
...
+#include <stdio.h>
+#include <string.h>
+#include <termios.h>
+
+#include <packet.h>
+#include <pgp-key.h>
+#include <rnp/rnp_sdk.h>
+
+bool
+rnp_getpass(const char *prompt, char *buffer, size_t size)
```
flowher: Should be static

src/lib/pgp-key.c
```diff
@@ -776,3 +764,78 @@ pgp_pk_alg_capabilities(pgp_pubkey_alg_t alg)
...
+        return false;
+    }
+    return key->key.seckey.encrypted;
+}
+
+bool
+pgp_key_unlock(pgp_key_t *key, const pgp_passphrase_provider_t *provider)
+{
+    pgp_seckey_t *decrypted_seckey = NULL;
+    bool          ok = false;
```
flowher: ``ok`` doesn't seem really useful here. ``decrypted_seckey`` will be set in case it was correctly decrypted and won't be set in case key isn't locked.

dewyatt: Right, this function used to be much longer, split most of it off into `pgp_decrypt_seckey`. Could definitely simplify this now.

src/lib/pass-provider.c
dewyatt: Yep, originally had this in `misc.c`, can be static now.

src/lib/pgp-key.c
dewyatt: I know, old code.

ronaldtse:
Thanks @dewyatt for making this possible. So many people have been waiting for this 👍 

Let's expedite this if possible.

PR #416, opened 2017-08-21
Moves commonly used ECC functions to separated file

flowher:
* Files ``ec.c/ec.h`` will contain functions/types/data that's commonly used by other ECC implementation in RNP
* Moves uncompressed EC key generation to ``ec.c`` and removes code duplication


ronaldtse:
LGTM @flowher could we change the copyright header of `ec.c/h` to the Ribose-only one since this code isn't inherited from NetPGP? Thanks!

flowher:
@ronaldtse Thanks for checking that. Done now

Issue #359, opened 2017-08-06
CLI should allow selection of compression parameters


ronaldtse:
## Description

From @ni4 #294:

> rnp supports decompression and by default compresses data, but we actually don't have a way to enable/disable compression or change the compression level yet

```sh
# off=0 ctb=c1 tag=1 hlen=3 plen=268 new-ctb
:pubkey enc packet: version 3, algo 1, keyid D78F04BE3AF65209
	data: [2048 bits]
# off=271 ctb=d2 tag=18 hlen=2 plen=71 new-ctb
:encrypted data packet:
	length: 71
	mdc_method: 2
# off=284 ctb=c8 tag=8 hlen=2 plen=36 new-ctb
:compressed packet: algo=2
# off=287 ctb=cb tag=11 hlen=2 plen=27 new-ctb
:literal data packet:
	mode b (62), created 1502000813, name="hello.txt",
	raw data: 12 bytes
```

rnp already implements ZIP, BZIP and ZLIB.

Relevant GPG options:
```
-z n
Set compression level to n. A value of 0 for n disables compression. Default is to use the default compression level of zlib (normally 6).

--compress-algo n
Use compress algorithm n. Default is 2 which is RFC1950 compression. You may use 1 to use the old zlib version (RFC1951) which is used by PGP. The default algorithm may give better results because the window size is not limited to 8K. If this is not used the OpenPGP behavior is used, i.e. the compression algorithm is selected from the preferences; note, that this can't be done if you do not encrypt the data.
```

dewyatt:
> rnp already implements ZIP, BZIP and ZLIB.

Just a note, we do support all of these for decompression, but for compression we only support zlib.

Adding support for setting the zlib compression level shouldn't be difficult: https://github.com/riboseinc/rnp/blob/342fb12583ad98b45160460a3041790b8525ab25/src/lib/compress.c#L422

ronaldtse:
@dewyatt Just created an issue on that 👍

ni4:
I can take this and update CLI tests accordingly if nobody against it

ronaldtse:
@ni4 please proceed, much appreciated!

Issue #360, opened 2017-08-06
Implement BZIP and ZIP OpenPGP compression


ronaldtse:
## Description

From @dewyatt #359
> ... we do support all of these [ZLIB, BZIP and ZIP] for decompression, but for compression we only support zlib.

Adding support for setting the zlib compression level shouldn't be difficult: https://github.com/riboseinc/rnp/blob/342fb12583ad98b45160460a3041790b8525ab25/src/lib/compress.c#L422

This is also related to #359 which is CLI support for selecting these algorithms and setting compression parameters.

randombit:
Should be fixed with #417, the CLI support is #359

PR #417, opened 2017-08-22
Add support for bzip2 and zip (deflate) compression

randombit:
Also fixes bzip2 compression which called the wrong shutdown function `BZ2_bzDecompressInit` instead of `BZ2_bzDecompressEnd` - and so always crashed... Also fix what appears to be an off by one in the decompression code that caused bzip2 (but not zlib) to fail.

I've updated the test to check compression roundtrips, but this does not update the CLI (#359). I don't see a good way to pass any arguments/options to `encrypt_se_ip_writer` so maybe some more refactoring is required to get that to work.

### DIFF
--- a/src/lib/compress.c
+++ b/src/lib/compress.c
@@ -83,7 +83,7 @@ __RCSID("$NetBSD: compress.c,v 1.23 2012/03/05 02:20:18 christos Exp $");
 #include "memory.h"
 #include "writer.h"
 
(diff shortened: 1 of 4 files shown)


ronaldtse:
Thanks @randombit !

ronaldtse:
@randombit this solves #360 right?

ni4:
Thanks, @randombit. I'll proceed with #359 as soon as possible.

PR #418, opened 2017-08-22
Add GPG 2.1 to CI and run cli_tests.py

dewyatt:
Fixes https://github.com/riboseinc/rnp/issues/394#issuecomment-323879018

I also realized my mistake of setting CORES based on /proc/cpuinfo in Travis CI. The container's /proc/cpuinfo matches the host, of course, so sometimes we were trying to build botan and other things with 32 cores in a travis-ci container (which resulted in processes being killed). Now defaults to 2 cores and can be overridden.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ env:
     - BOTAN_INSTALL="$HOME/builds/botan-install"
     - CMOCKA_INSTALL="$HOME/builds/cmocka-install"
     - JSONC_INSTALL="$HOME/builds/jsonc-install"
+    - GPG21_INSTALL="$HOME/builds/gpg21-install"
     - LOGNAME="Travis"
     - CLANG_FORMAT_DIFF="clang-format-diff-4.0"
 
@@ -46,6 +47,7 @@ cache:
     - "$BOTAN_INSTALL"
     - "$CMOCKA_INSTALL"
     - "$JSONC_INSTALL"
+    - "$GPG21_INSTALL"
 
 install:
   - ./ci/install.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,11 @@ branches:
 
 env:
   global:
-    - BOTAN_INSTALL="$HOME/builds/botan-install"
-    - CMOCKA_INSTALL="$HOME/builds/cmocka-install"
-    - JSONC_INSTALL="$HOME/builds/jsonc-install"
-    - GPG21_INSTALL="$HOME/builds/gpg21-install"
+    - LOCAL_BUILDS="$HOME/local-builds"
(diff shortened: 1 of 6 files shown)


ronaldtse:
@dewyatt could we use a separate build directory that gets removed after the CI run than just `~/builds`?

The runs on Jenkins reuses this directory across separate jobs and I right now I have to specify `rm` after runs.

ronaldtse:
Or at least extract that path and make it configurable through ENV. Thanks!

dewyatt:
@ronaldtse Sure thing.

@ni4 Ideas on the failure here?: https://travis-ci.org/riboseinc/rnp/jobs/267022584

dewyatt:
@ronaldtse Not sure if I completely addressed your issue there. I see the jenkins build failed?

ronaldtse:
I've attempted a fix, could you push something again to trigger it?

dewyatt:
Not sure why but `cli_tests.py` seems to randomly fail on removing some temporary directory.
Might just need to add `ignore_errors=True`, but would be nice to know the underlying cause.

(I think we can just close & reopen to trigger build again)

dewyatt:
Oh right that won't work for Jenkins. Ok one sec.

ronaldtse:
Thanks, that should work - it's polling every minute so you might need to wait for half more minute...

ronaldtse:
@dewyatt it ran but failed here:
```sh
Running in /tmp/tmpdOFCAx
Cannot find utility gpg. Exiting.
```

We need to manually point to the `gpg` executable since it's specified inside `$LOCAL_BUILDS`? It looks like it works on Travis...

ronaldtse:
Ah it failed much earlier. There's a line to call `gpg` to import the key but there's no gpg installed in the build machine. Will fix this.

dewyatt:
@ni4 It looks like there's a race condition where gpg-agent sometimes leaves a root-owned socket in the temporary directory used for the tests. I think the best thing to do is to rework `cli_tests.py` so that each test gets its own directory (or a group of tests) and then it's safer to have `ignore_errors=True` when purging things. Thoughts?

ronaldtse:
Yay! Jenkins works!

```sh
Run 143 tests, 143 succeeded and 0 failed.
```

ronaldtse:
@dewyatt @ni4 yes indeed, running GnuPG in quick succession will cause lingering sockets which can be solved with separating the `.gpg` home directories which is what we do currently. Thus the need for rnp.

ronaldtse:
@ni4 this is passing now, could you approve / merge? Thanks!

ni4:
@dewyatt @ronaldtse Ok, I'll update script to not fail in clear_keyrings within the next PR 
Also I'm thinking about updating it with some more logging.

PR #420, opened 2017-08-22
Removal of packet.h and cleanup

flowher:
Finishes work started in #409

* Functions defined in packet.h were moved to dedicated header files
* Moves some types closer to implementation
* Types used commonly across were moved from packet.h to types.h.

Sorry for big commit but it has to go in one shot


flowher:
Hmmm.. . fails because of ``OSError: [Errno 2] No such file or directory: '/tmp/tmp_E1tDH/.gpg/S.gpg-agent.ssh'`` ... weird.

dewyatt:
> Hmmm.. . fails because of OSError: [Errno 2] No such file or directory: '/tmp/tmp_E1tDH/.gpg/S.gpg-agent.ssh' ... weird.

Don't worry about it for now: https://github.com/riboseinc/rnp/pull/418#issuecomment-323978472

It will fail randomly.

dewyatt:
On the topic of cleanup, at some point we ought to move all these `pgp_memory_t` functions out of `misc.c` and into a new `memory.c` (especially since we already have a `memory.h`).

randombit:
Looks good to me

flowher:
Okis - merging then, because I need it for next steps.
Indeed there is more cleanup needed, but this can be done in smaller steps.

Issue #412, opened 2017-08-20
Notes after trying on performance testing.


catap:
I've tried to run very simple and dirty performance testings to compare performance rnp (last master) and GnuPG 2.1.23.

All tests run on my laptop with a lot of running background application and I didn't re-run it 5-7 times to makes clear result. Anyway it shows the picture and makes a detection to optimisation.

Short summary:
 - rnp slower than GnuPG for 30%;
 - rnp can't encrypt files bigger than 4Gb (because it uses a `unsigned int` as file size, so 4Gb is 0 and bigger file will be just corrupted ( @ronaldtse example of issue that I mentored before ;) );
 - rnp uses huge size of memory (about 4x size of raw file at encryption: 3gb file and 11.54gb memory used);
 - rnp crashed when I try to decrypt 1gb file.

Test restults (all times in seconds, used same key):
rnp:
- 128mb: 10.16
- 1gb: 77.07
- 4gb: failed without any error(!), just created a file with size 331b

gpg:
- 128mb: 7.862
- 1gb: 41.51
- 4gb: 167.61

CPU using at encrypting 1gb file:

```
Total: 7474 samples
    5843  78.2%  78.2%     5843  78.2% __platform_memccpy
     816  10.9%  89.1%     3795  50.8% _deflateCopy
     256   3.4%  92.5%      256   3.4% Botan::SHA_160::sse2_compress_n
     229   3.1%  95.6%      231   3.1% _inflateMark
     121   1.6%  97.2%      252   3.4% _pgp_cipher_cfb_encrypt
     103   1.4%  98.6%      103   1.4% _CAST_encrypt
      53   0.7%  99.3%      836  11.2% _deflateSetDictionary
      11   0.1%  99.4%       11   0.1% _zError
       9   0.1%  99.6%      111   1.5% _CAST_ecb_encrypt
       9   0.1%  99.7%      132   1.8% _botan_block_cipher_encrypt_blocks
       8   0.1%  99.8%      122   1.6% _EVP_EncryptUpdate
       5   0.1%  99.9%        5   0.1% __kernelrpc_mach_vm_deallocate_trap
       4   0.1%  99.9%      116   1.6% _EVP_cast5_ecb
       2   0.0%  99.9%        2   0.0% _botan_block_cipher_decrypt_blocks
       1   0.0%  99.9%      133   1.8% Botan::mgf1_mask
       1   0.0% 100.0%        1   0.0% _CAST_set_key
[...]

PR #421, opened 2017-08-23
One OpenPGP block can't be bigger than 0xffffffff, so, switched to streaming if it required

catap:
First step of #412.

Each OpenPGP block can be up to 4Gb size and in case if someone try to encrypt file >= 4Gb we should use the parital encoding.

Good new: we have this code and looks like it works.

Bad news:
 - we haven't got decryption code :(
 - sometime GnuPG can't decode the file that rnp provides.

But I'm using 'crypto-container' (gpg encrypted archives, from few gigabytes up to handreds gigabytes) and I can say that GnuPG not at all time can decrypt bigger file that it created :)

### DIFF
--- a/src/lib/crypto.c
+++ b/src/lib/crypto.c
@@ -473,17 +473,20 @@ pgp_encrypt_file(rnp_ctx_t *         ctx,
     }
 
     /* Push the encrypted writer */
-    if (!pgp_push_enc_se_ip(output, pubkey, ctx->ealg)) {
+    if (!pgp_push_enc_se_ip(output, pubkey, ctx->ealg, pgp_mem_len(inmem))) {
         pgp_memory_free(inmem);
         return false;
     }
 
     /* This does the writing */
-    pgp_write(output, pgp_mem_data(inmem), (unsigned) pgp_mem_len(inmem));
+    if (!pgp_write(output, pgp_mem_data(inmem), pgp_mem_len(inmem))) {
+        pgp_memory_free(inmem);
+        return false;
+    }
 
     /* tidy up */
-    pgp_memory_free(inmem);
     pgp_teardown_file_write(output, fd_out);
+    pgp_memory_free(inmem);
 
     return true;
 }
@@ -516,10 +519,18 @@ pgp_encrypt_buf(rnp_ctx_t *         ctx,
     }
 
     /* Push the encrypted writer */
-    pgp_push_enc_se_ip(output, pubkey, ctx->ealg);
+    if (!pgp_push_enc_se_ip(output, pubkey, ctx->ealg, insize)) {
+        pgp_writer_close(output);
+        pgp_output_delete(output);
+        return false;
+    }
 
     /* This does the writing */
-    pgp_write(output, input, (unsigned) insize);
+    if (!pgp_write(output, input, insize)) {
+        pgp_writer_close(output);
+        pgp_output_delete(output);
+        return false;
+    }
 
     /* tidy up */
     pgp_writer_close(output);
--- a/src/lib/symmetric.c
+++ b/src/lib/symmetric.c
@@ -124,8 +124,10 @@ pgp_cipher_start(pgp_crypt_t *crypt, pgp_symm_alg_t alg, const uint8_t *key, con
     memset(crypt, 0x0, sizeof(*crypt));
 
     const char *cipher_name = pgp_sa_to_botan_string(alg);
-    if (cipher_name == NULL)
+    if (cipher_name == NULL) {
+        fprintf(stderr, "Unsupported algorithm: %d\n", alg);
         return false;
+    }
 
     crypt->alg = alg;
     crypt->blocksize = pgp_block_size(alg);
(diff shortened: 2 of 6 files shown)


catap:
and as side effect I decreased memory using a littel bit. Now it requested about 3x file size for large file :) (before it was about 4x)

catap:
Yay, tests crashed on very strange error:
```
OSError: [Errno 2] No such file or directory: '/tmp/tmphx5L2C/.gpg/S.gpg-agent.ssh'
```

@dewyatt @ni4 do you know that is it?

ronaldtse:
@catap yes that's a problem with GnuPG conflicting on socket usage due to the way tests are run. @ni4 is going to fix that by using separate `homedir`s for GnuPG. So I think it's good to go!

ni4:
@catap Partial encoding was used when data is compressed, and this didn't work well for armored messages. I'm preparing the fix for this and other problems with armor/cleartext signing later today.

catap:
@ni4 thanks! I haven't use armored  encoding to be honest :)

src/lib/crypto.c
```diff
@@ -518,10 +521,14 @@ pgp_encrypt_buf(rnp_ctx_t *         ctx,
     }
 
     /* Push the encrypted writer */
-    pgp_push_enc_se_ip(output, pubkey, ctx->ealg);
+    if (!pgp_push_enc_se_ip(output, pubkey, ctx->ealg, insize)) {
+        return false;
```
flowher: pgp_setup_memory_write will allocate ``output``, so here we will leak memory. 
I would suggest set ``output=NULL`` and ``outmem=NULL, have ``goto end`` here and have ``end`` label just before leaving this function (see for example ``ecdsa.c``). We also need to clear outmem in case of error

src/lib/symmetric.c
```diff
@@ -124,8 +124,10 @@ pgp_cipher_start(pgp_crypt_t *crypt, pgp_symm_alg_t alg, const uint8_t *key, con
     memset(crypt, 0x0, sizeof(*crypt));
 
     const char *cipher_name = pgp_sa_to_botan_string(alg);
-    if (cipher_name == NULL)
+    if (cipher_name == NULL) {
+        fprintf(stderr, "Unsupported algorithm: %d\n", alg);
```
flowher: Can you use RNP_LOG() instead?

src/lib/writer.c
```diff
@@ -871,12 +873,20 @@ static void encrypt_se_ip_destroyer(pgp_writer_t *);
...
+pgp_push_enc_se_ip(pgp_output_t *      output,
+                   const pgp_pubkey_t *pubkey,
+                   pgp_symm_alg_t      cipher,
+                   size_t              len)
 {
     pgp_pk_sesskey_t *encrypted_pk_sesskey;
     encrypt_se_ip_t * se_ip;
     pgp_crypt_t *     encrypted;
 
+    /* one packet can't be bigger than 0xffffffff, so, switch to streaming encruption*/
```
flowher: encruption -> encryption

src/lib/writer.c
```diff
@@ -961,23 +971,21 @@ encrypt_se_ip_writer(const uint8_t *src,
     }
 
     /* create compressed packet from literal data packet */
-    if (!pgp_writez(zoutput, pgp_mem_data(litmem), (unsigned) pgp_mem_len(litmem),
-                    PGP_C_ZLIB, 6)) {
+    if (!pgp_writez(
+          zoutput, pgp_mem_data(litmem), (unsigned) pgp_mem_len(litmem), PGP_C_ZLIB, 6)) {
```
flowher: I think this cast to ``unsigned`` can be removed

src/lib/writer.c
```diff
@@ -1019,6 +1027,11 @@ pgp_write_se_ip_pktset(pgp_output_t * output,
     }
     bufsize = preamblesize + len + mdcsize;
 
+    if (bufsize > UINT_MAX) {
+        fprintf(stderr, "Buffer should be less than 4gb\n");
```
flowher: RNP_LOG gives better output

ronaldtse:
@catap could we resolve @flowher 's comments and merge? Thanks!

ronaldtse:
@catap looks like we can fix up @flowher 's comments and we can merge?

catap:
@ronaldtse thanks, I forgot about this :( Will fix it right now.

src/lib/symmetric.c
catap: `symetric.c` hasn't got any of using `RNP_LOG`, so, I think the right way is rewrite all `fprintf` to `RNP_LOG`

catap:
@ronaldtse, @flowher fixed!

catap:
@flowher I think you can merge it ;)

flowher:
Ah.. sorry. Indeed, CI is OK. Done now

ronaldtse:
Thank you guys! 👏

PR #422, opened 2017-08-23
Use a single ID for SM2 encryption+signatures

randombit:
See discussion in #395

### DIFF
--- a/include/repgp/repgp_def.h
+++ b/include/repgp/repgp_def.h
@@ -266,8 +266,7 @@ typedef enum {
                                * IETF-S/MIME) */
     PGP_PKA_EDDSA = 22,       /* EdDSA from draft-ietf-openpgp-rfc4880bis */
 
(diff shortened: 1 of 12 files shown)


ronaldtse:
@randombit thanks for this! For SM2 both encryption and signatures it requires a hash: do we allow the specification of that or is it locked to SM3 right now? Ideally we'd like to have it configurable.

randombit:
@ronaldtse Currently in Botan SM2 is completely hardcoded to use SM3 for hashing. I could make this configurable, there just didn't seem any point because SM2 without SM3 is AIUI non-conforming with the Chinese standard, and I didn't see that anyone would be interested in using SM2 signature or encryption but not caring about conformance.

ronaldtse:
@randombit it would be great if we can make it configurable. The only reason SM3 is used with SM2 is because it is allowed by OSCCA, as you know. But once they have newer hash standards, those can be used with SM2 as well, so making it configurable would future-proof it.

PR #426, opened 2017-08-24
Fix up KBX keystore format for subkeys.

dewyatt:
I missed this in https://github.com/riboseinc/rnp/pull/354.

Also fixes the keyid offset in KBX, which was off by a bit.

I did notice that we don't really support V3 keys in KBX right now. They require a bit of extra work (see [here](https://github.com/gpg/gnupg/blob/5c8fe5480964c282540c051b538e812851988422/kbx/keybox-blob.c#L598)). We should probably reject them somewhere in the save functions for kbx.


ronaldtse:
Any thoughts @ni4 @flowher ?

Issue #400, opened 2017-08-17
Rarely reproducible problem with RSA keys during decryption/verification


ni4:
## Description

Rarely after number of runs of `./cli_tests.py` encryption/decryption or signing/verification tests fails, and only in one case (i.e. with the same rnp-generated key one round of encryption/decryption or signing/verification fails and then works fine).
I suspect some 'edge case' like symmetric key wrapping/unwrapping/bit boundary or so on problem.

This should be reproduced and further investigated.

ni4:
This can be reproduced with the following python CLI test sample:

```python
def rnp_sign_bug_hunt():
    for i in range(0, 100):
        rnp_genkey_rsa(KEY_SIGN_GPG)
        gpg_import_pubring()
        gpg_import_secring()
        
        for j in range(0, 30):
            run_test(rnp_signing_gpg_to_rnp, 15000, 0, 1)
        clear_keyrings()
```

It happens after number of runs with the same key, so it could be related to some wrong key calculations. Signature is generated by GPG, and verification fails in both GPG and RNP. So it could be even bug in the GPG.

dewyatt:
This does seem like a gpg bug (or two).
I created a couple of repos with reduced cases if it helps (and unencrypted keys, as the key material may be relevant here):
https://github.com/dewyatt/gpg-bug-decrypting
https://github.com/dewyatt/gpg-bug-signing

They both succeed most of the time, but will usually fail within a minute or so. The decryption bug repo is particularly interesting because it uses static data (which it randomly fails to decrypt).
Using gpg1 instead makes both issues disappear so it seems like a regression (maybe related to gpg-agent? or libgcrypt?).

I wasn't able to reproduce the issue with keys that were not generated by rnp. Anyways, good luck on this @ni4!

ni4:
Thanks @dewyatt for your investigations.
Today I also spent some time on this and found out the reason.
RFC 4880 tells that p < q for RSA key, but we ignored this fact. And looks like GPG relies on this during signing/decryption (there is faster calculation algorithm based on p, q and u).
I'm preparing PR right now.

PR #427, opened 2017-08-25
Added compression parameters selection to CLI

### LINKED ISSUE #359: CLI should allow selection of compression parameters

ni4:
- Added --zlib, --zip and --bzip2, -z 0..9 parameters to the rnp commandline and implemented their handling. Since compression now is used only during encryption it works only in this case.
- Fixed hanging bug in bzip2 decompression
- Updated CLI tests with more compression/decompression cases
- (hopefully) Fixed possible random CLI tests crash in Travis
- fixes #359

### DIFF
--- a/src/tests/cli_tests.py
+++ b/src/tests/cli_tests.py
@@ -11,6 +11,7 @@ import subprocess
 import re
 import random
 import string
+import time
 from subprocess import Popen, PIPE
 from cli_common import find_utility, run_proc, pswd_pipe, rnp_file_path, random_text, file_text
 import cli_common
--- a/src/tests/cli_tests.py
+++ b/src/tests/cli_tests.py
@@ -26,6 +26,11 @@ RMWORKDIR = False
 DEBUG = False
 TESTS_SUCCEEDED = []
 TESTS_FAILED = []
+TEST_WORKFILES = []
+# Key userids
+KEY_ENCRYPT = 'encryption@rnp'
+KEY_SIGN_RNP = 'signing@rnp'
+KEY_SIGN_GPG = 'signing@gpg'
 
 RE_RSA_KEY = r'(?s)^' \
 r'# .*' \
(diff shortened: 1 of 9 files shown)


ni4:
Looks like I'm lucky enough to hit #400
And another check hanged during the detached signature validation

src/tests/cli_tests.py
```diff
@@ -144,18 +152,44 @@ def raise_err(msg, log = None):
         print log
     raise NameError(msg)
```
dewyatt: Just a nitpick but `NameError` isn't really an appropriate error to raise here. Best practice would be to add a small class that derives from `Exception` and takes that optional log parameter (and override `__str__` to print that out if provided).

ni4: Thanks, @dewyatt . It was just a quick pick, will fix it to work in a better way at some point.

ni4:
I fixed hang during the signature verification, however still have #400 in one of the checks. 
Also updated exception throwing according to @dewyatt 's suggestion.
It seems that it would be a good idea to show less output with --debug so more tests could be run.

dewyatt:
> It seems that it would be a good idea to show less output with --debug so more tests could be run.

I don't think we're anywhere near the travis limits on this, and the debug output really does help so I would prefer we keep it. You may hit the limit on the web interface, but you can always download the raw log (click on "Raw Log").

Some other minor suggestions for later:

* You may be interested in using the `--with-colons` option for `gpg --list-keys`, very easy to parse in a robust manner.
* It would be nice if we added a `prefix=rnp-cli` (or similar) to the `mkdtemp` call. Otherwise, when debugging, we end up with a bunch of directory names like `/tmp/tmpw0nT6G/` which are pretty generic sounding (it feels safer to `rm -rf /tmp/rnp-*`, rather than `rm -rf /tmp/tmp*`).

ronaldtse:
Agree with both suggestions from @dewyatt . The last one is super easy to change.

@ni4 look like we can we rebase this and merge? This will solve #359

ni4:
@ronaldtse I fixed prefix for mkdtemp, and now we know why tests are failing (see PR #429 ). So now this PR can be safely rebased and merged. Or should I do this?

ronaldtse:
Thanks for finding that issue out with the RSA keys. @ni4 it would be great if you could rebase and merge it 👍

ni4:
@ronaldtse Done, except I forgot to squash, sorry.

ronaldtse:
No problem at all @ni4, thanks!

PR #428, opened 2017-08-27
Copy primary crypto params to subkey crypto params in non-expert mode.

dewyatt:
Otherwise you may do something like:
```
rnpkeys --gen-key --numbits 1024
```
And you'll get a primary key 1024-bit RSA key but a 2048-bit (the default) subkey.


ronaldtse:
Thanks @dewyatt and @ni4 !

PR #429, opened 2017-08-27
Ensure that p < q in the generated RSA key.

### LINKED ISSUE #400: Rarely reproducible problem with RSA keys during decryption/verification

ni4:
As per RFC 4880 p should be less then q for the RSA key. Otherwise generated key could behave wrong with GnuPG because of some optimized calculations which most likely utilize this fact.
This fixes #400.

### DIFF
--- a/src/lib/crypto/rsa.c
+++ b/src/lib/crypto/rsa.c
@@ -301,7 +301,8 @@ pgp_genkey_rsa(pgp_seckey_t *seckey, size_t numbits)
 {
     botan_privkey_t rsa_key = NULL;
     botan_rng_t     rng = NULL;
-    int             ret = 0;
+    int             ret = 0, cmp;
+    BIGNUM *        t;
 
     seckey->pubkey.key.rsa.n = BN_new();
     seckey->pubkey.key.rsa.e = BN_new();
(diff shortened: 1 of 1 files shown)


ni4:
JFYI: Travis fails, but the failure reason is already fixed in #427.

ronaldtse:
Very nice @ni4! Just wondering if we could add a test for this in case someone removes it without understanding the problem?

dewyatt:
Nice find, I see libgcrypt does the [same](https://github.com/gpg/libgcrypt/blob/208aba6f9a0475ba049f5a66fe02cf9a6214a887/cipher/rsa.c#L283).
I think it may be simpler to use `botan_mp_swap` here though.

dewyatt:
> Just wondering if we could add a test for this in case someone removes it without understanding the problem?

I think we'll be OK since he added a comment explaining the code.

ronaldtse:
Sure @dewyatt. Looks like just a minor change with swap and we're good to go.

ni4:
@ronaldtse @dewyatt Ok, np. I'll fix it to botan_mp_swap soon. Actually we can add some test case to check generated key's parameters for all generated key types.

ni4:
@ronaldtse @dewyatt Done. Finally green checkmarks :)
@randombit Btw, do you have any idea whether we need to optimize RSA decryption/signing using the p, q, u? Or does botan implement this already? (However we definitely do not pass u to the botan).

randombit:
@ni4 Botan already implements the RSA-CRT optimization. The `u` param ends up being recomputed internally.

ronaldtse:
Thanks @ni4 @randombit , going to merge now.

PR #430, opened 2017-08-30
Provide OCB license.

ronaldtse:
Allow us to thank Prof. Phillip Rogaway for granting an OCB license to `rnp` users! 👍

A quote from him:
> My real reward [...], has always been in the pleasure of doing crypto and seeing a bit of
it get into the world.

OCB should be made default in `rnp`.


ronaldtse:
Thank you guys! 👍

PR #444, opened 2017-09-07
fixed buffer overflow issue when setting homedir with long string

jjr840430:


src/rnp/rnpcfg.c
```diff
@@ -426,6 +426,9 @@ rnp_path_compose(const char *dir, const char *subdir, const char *filename, char
     }
 
     /* concatenating dir, subdir and filename */
+    if (strlen(dir) > MAXPATHLEN - 1)
+        return false;
```
dewyatt: Can you add braces to these one-line conditionals? ([Dev Guide](https://github.com/riboseinc/rnp/blob/master/Developers%20Guide.md#general-guidelines))

src/rnp/rnpcfg.c
```diff
@@ -434,6 +437,9 @@ rnp_path_compose(const char *dir, const char *subdir, const char *filename, char
             res[pos++] = '/';
         }
 
+        if (strlen(subdir) + pos > MAXPATHLEN - 1)
```
dewyatt: Same here

src/rnp/rnpcfg.c
```diff
@@ -442,6 +448,9 @@ rnp_path_compose(const char *dir, const char *subdir, const char *filename, char
         res[pos++] = '/';
     }
 
+    if (strlen(filename) + pos > MAXPATHLEN - 1)
```
dewyatt: And here

ni4:
@dewyatt Thanks for pointing on this, right now I was thinking about PATH_MAX/MAXPATHLEN. It's the question of architecture - should we stick to some constant thing and to which one, PATH_MAX or MAXPATHLEN? Are there any best practices and if yes - what do they say?

dewyatt:
> @dewyatt Thanks for pointing on this, right now I was thinking about PATH_MAX/MAXPATHLEN. It's the question of architecture - should we stick to some constant thing and to which one, PATH_MAX or MAXPATHLEN? Are there any best practices and if yes - what do they say?

Honestly, my understanding is that generally none of these arbitrary limits are reliable. I believe POSIX says that a system may define a uniform PATH_MAX, for example, but typically this isn't valid because it may depend on the filesystem and other variables.

http://insanecoding.blogspot.com/2007/11/pathmax-simply-isnt.html
https://www.gnu.org/software/libc/manual/html_node/Limits-for-Files.html

I would bet that the proper solution is to use only dynamic allocation and avoid any constant limits. Anyways, probably needs more investigation at some point.

jjr840430:
@dewyatt @ni4 Thanks for your reviews. Can you please check again?

dewyatt:
@ni4 I've noticed we have some random failures in cli_tests.py like this:
https://travis-ci.org/riboseinc/rnp/jobs/273777467

I've seen it at least 3 times I think now. Any ideas? I don't want to burn time on this at the moment, so I've just been restarting builds when I encounter it.

jjr840430:
Can anyone help me to solve failure of codecov checks?

ni4:
@dewyatt I don't see any failures on the log you've linked - Run 419 tests, 419 succeeded and 0 failed.

dewyatt:
@jjr840430
> Can anyone help me to solve failure of codecov checks?

Don't worry about them.

dewyatt:
@ni4 
> @dewyatt I don't see any failures on the log you've linked - Run 419 tests, 419 succeeded and 0 failed.

I restarted the build and it succeeded as expected. Unfortunately, travis doesn't seem to keep old build logs for PRs and the s3 archive link expired. Anyways, I'll try to repro it on one of my open PRs and let you know.

ronaldtse:
Merging this. Thanks!

PR #445, opened 2017-09-07
Correct SM2 signature ID

randombit:
This passed "Raw" but for SM2 the parameter is the associated userid, not the hash. Instead leave it empty.

Remove .ecdsa sig type from union and use .ecc everywhere since the same format is used for all ECC signature methods (ECDSA, SM2, EDDSA)

Inline hash_and_check_sig into its only caller, minor simplification.


ronaldtse:
Thanks @randombit !

PR #446, opened 2017-09-07
Support arbitrary hashes for SM2 encryption KDF/integrity check

randombit:
Also appends a byte to the encrypted packet that specifies what hash was used, which matches the draft text.

The actual call in pgp_create_pk_sesskey still hardcodes SM3, this should I guess take the hash from the key's hash preferences?

Requires recent Botan with https://github.com/randombit/botan/pull/1188. I will clear the Travis cache so it gets rebuilt.

### DIFF
--- a/src/tests/cipher.c
+++ b/src/tests/cipher.c
@@ -610,11 +610,11 @@ sm2_roundtrip(void **state)
 {
     rnp_test_state_t *rstate = *state;
 
-    uint8_t key[32] = {0};
+    uint8_t key[27] = {0};
     pgp_random(key, sizeof(key));
 
     uint8_t ctext_buf[1024];
-    uint8_t decrypted[32];
+    uint8_t decrypted[27];
 
     const rnp_keygen_crypto_params_t key_desc = {.key_alg = PGP_PKA_SM2,
                                                  .hash_alg = PGP_HASH_SM3,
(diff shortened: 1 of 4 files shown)


randombit:
Hmm test is failing need to investigate this.

randombit:
Oh duh I forgot to clear the cache.

src/tests/cipher.c
```diff
@@ -633,20 +633,29 @@ sm2_roundtrip(void **state)
     const pgp_ecc_pubkey_t *pub_ecc = &pub_key->key.ecc;
     const pgp_ecc_seckey_t *sec_ecc = &sec_key->key.ecc;
 
-    size_t        ctext_size = sizeof(ctext_buf);
-    pgp_errcode_t enc_result =
-      pgp_sm2_encrypt(ctext_buf, &ctext_size, key, sizeof(key), pub_ecc);
-    rnp_assert_int_equal(rstate, enc_result, PGP_E_OK);
+    uint8_t hashes[] = {PGP_HASH_SM3, PGP_HASH_SHA256, PGP_HASH_SHA512, 0};
```
dewyatt: Maybe instead of the sentinel, use `ARRAY_SIZE` below (since we do that in some other places). Regardless, LGTM.

randombit: Fixed

dewyatt:
Should be good to merge here, Jenkins probably needs a newer version of botan to pass.

ronaldtse:
Thanks guys, @dewyatt is correct that the botan version is cached on Jenkins. Will fix this today and also ensure the build log is uploaded as a gist.

PR #450, opened 2017-09-10
Improve key lock test

dewyatt:
This just makes this test a little more thorough.
I also moved the passphrase callbacks to support because I used them in some other tests (future PR).

### DIFF
--- a/src/tests/key-unlock.c
+++ b/src/tests/key-unlock.c
@@ -116,8 +116,18 @@ test_key_unlock_pgp(void **state)
     rnp_ctx_free(&ctx);
 
     // grab the signing key to unlock
+    key = NULL;
     rnp_assert_true(rstate,
                     rnp_key_store_get_key_by_name(rnp.io, rnp.secring, keyids[0], &key));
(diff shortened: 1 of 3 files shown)


dewyatt:
@ni4 Here's an example: https://travis-ci.org/riboseinc/rnp/jobs/273893046

ronaldtse:
@dewyatt you mean this sort of failure?

```sh
[pgp_mem_readfile() misc.c:750] can't open "/tmp/rnpctmpoi4TKO/.rnp/pubring.gpg"
```

dewyatt:
> @dewyatt you mean this sort of failure?

Not quite, there's a regex match failure for generated keys in cli_tests.py that seems to come up occasionally.

As far as the `pgp_mem_readfile` messages, IMO we should probably eliminate those from that function. The messages should probably be logged at the call site, where there is enough context to determine whether something is normal or an error. (Sometimes it's perfectly OK for a file to not exist)

ronaldtse:
Agree. New issue for `pgp_mem_readfile` messages.

PR #452, opened 2017-09-11
G10 rework and key protect/unprotect

dewyatt:
* Reworked G10 key store code a bit to align with the pgp code and function again after the subkey reworking. I did sacrifice G10 public-key support, only to save myself some time while reworking this. AFAIK this does not affect compatibility with gpg at all. So currently the valid keystore formats are GPG and GPG21 only. (cmocka tests for GPG21 have been re-enabled)
* Added key protect/unprotect.
* Fixed some bugs here and there. ~~G10 keys had keyids that varied based on the current time at which they were loaded, for example.~~ EDIT: Actually this case was handled previously, but there were some other fields that weren't right and caused keyids to be inaccurate in some cases before.

Just a quick overview of the current state (most of this is in the comments):
* A `pgp_key_t` always has the raw storage form of the key in `pgp_key_t->packets[0]`. The format of this is `pgp_key_t->format` (really it's either PGP packets or G10).
* lock/unlock - a locked key has NULL secret MPIs in the seckey structure. An unlocked key has populated secret MPIs in the seckey structure. Unlocking a key involves using the stored key in `pgp_key_t->packets[0]`. Unlocking a protected key requires a passphrase. Locking/unlocking doesn't make a lot of sense for unprotected keys since the secret MPIs would still be exposed via `pgp_key_t->packets[0]`.
* protect/unprotect - a protected key has an encrypted key stored in `pgp_key_t->packets[0]`. An unprotected key has an unencrypted key in `pgp_key_t->packets[0]`. Unprotecting a key that is protected requires a passphrase. Protecting a key requires unlocking it first (I only did this to avoid the confusion of the passphrase provider being called twice, easy to change).
* Note that keys are generated and written unencrypted after the first commit and up until the key protect/unprotect commit (at which point they are switched back to being written out encrypted).

Known issues (not introduced here AFAIK):
* The defkey setting for G10 is likely incorrect, it relies on the order of `readdir`. So signing with a GPG21 keystore format typically requires `--userid` to hint at which key you actually want.
* When we are searching for corresponding keys between pubring/secring, we really should be using the key grip and not the keyid (which is more likely to collide and slightly less reliable in general).

TODOs:
* The parameters for key protection are not currently customizable. For example, the symmetric algorithm and s2k iteration count are currently hardcoded.

### DIFF
--- a/include/rekey/rnp_key_store.h
+++ b/include/rekey/rnp_key_store.h
@@ -111,13 +111,13 @@ typedef struct {
     uint32_t blob_created_at;
 } kbx_pgp_blob_t;
 
-enum key_store_format_t {
+typedef enum key_store_format_t {
     UNKNOW_KEY_STORE = 0,
     GPG_KEY_STORE,
     SSH_KEY_STORE,
     KBX_KEY_STORE,
     G10_KEY_STORE,
-};
+} key_store_format_t;
 
 #define RNP_KEYSTORE_GPG "GPG" /* GPG keystore format */
 #define RNP_KEYSTORE_KBX "KBX" /* KBX keystore format */
@@ -177,10 +177,7 @@ bool rnp_key_store_get_key_by_name(pgp_io_t *,
 bool rnp_key_store_get_next_key_by_name(
   pgp_io_t *, const rnp_key_store_t *, const char *, unsigned *, const pgp_key_t **);
 
-bool rnp_key_store_get_key_grip(pgp_pubkey_t *, uint8_t *);
-bool rnp_key_store_get_key_by_grip(pgp_io_t *,
-                                   const rnp_key_store_t *,
-                                   const uint8_t *,
-                                   pgp_pubkey_t **);
+bool       rnp_key_store_get_key_grip(pgp_pubkey_t *, uint8_t *);
+pgp_key_t *rnp_key_store_get_key_by_grip(pgp_io_t *, rnp_key_store_t *, const uint8_t *);
 
 #endif /* KEY_STORE_H_ */
--- a/include/rekey/rnp_key_store.h
+++ b/include/rekey/rnp_key_store.h
@@ -38,9 +38,9 @@
 #include <stdbool.h>
 
 #include "memory.h"
-#include "pgp-key.h"
 
 typedef struct rnp_t rnp_t;
+typedef struct pgp_key_t pgp_key_t;
 
 typedef enum {
     KBX_EMPTY_BLOB = 0,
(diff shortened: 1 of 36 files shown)


dewyatt:
(Forgot I need to make some updates for clang to build, in progress...)

ronaldtse:
@dewyatt LGTM, could you help post the two mentioned issues?

1.
> I did sacrifice G10 public-key support, only to save myself some time while reworking this. 

2.
> The parameters for key protection are not currently customizable. For example, the symmetric algorithm and s2k iteration count are currently hardcoded.

ronaldtse:
Jenkins build failing due to botan not the newest release. Will fix.

dewyatt:
> The parameters for key protection are not currently customizable. For example, the symmetric algorithm and s2k iteration count are currently hardcoded.

Actually I just realized I forgot to do one thing here so I'm going to see if I can tackle this issue real quick. I meant to remove the now-unused fields from crypto params, but I guess I forgot.

dewyatt:
Best to hold off on further review for a day or so as I've noticed at least one mistake I made here (I guess I've been rushing things too much). Let me get this corrected and do another self-review.

catap:
@dewyatt sure and thanks for protect/unprotect keys, I think I can totally finished G10 issue :)

dewyatt:
This ended up a lot larger than I planned (sorry!).

I realized I mixed up the s2k hash alg field with another, so I had to get that fixed. I'm pretty sure I've seen others make similar mistakes so I thought it might be good to just break s2k+protection pieces out into separate structs to be explicit about what they are used for. It helps with some of our currently confusing generic field names like `seckey->alg` (the symmetric algorithm for encrypting the seckey), which would now be `seckey->protection.symm_alg`. It does make for some overly long/verbose field accesses but I think using a local shortcut pointer makes that less of an issue.

I added in the ability to specify the protection parameters at the API level. At the CLI level we can only specify the symmetric alg and the hash alg (same as before, with `--cipher` and `--hash`). (Can't specify cipher mode or iteration count). I'll create an issue for that if we don't already have one.

dewyatt:
I'm going to add some minor commits on top here that just improve usability a bit:
* Print the keys out earlier during key generation, so you know what key you're protecting.
* Adjust spacing a bit for stdin passphrase provider.
* Use pgp_key_t pointer in passphrase ctx. (Now possible after the separation of generation from protection)

(I would do these in a separate PR but they depend on this stuff anyways)

dewyatt:
(Rebased on master to make Travis CI happy)

ronaldtse:
@ni4 @catap one more approval and we can merge 👍

ni4:
@ronaldtse Done. Are there any open questions/suggestions, or I may merge right now?

ronaldtse:
I'll leave you the joy of rebase/merging 👍

PR #455, opened 2017-09-12
FFI Interface

randombit:
Not for merging - none of the APIs added here are implemented yet.

Opening this PR to discuss the API both in terms of what operations are useful/desired and how we want them to look. This is basically me typing out the first things that came into my head :)

Definitely missing is anything for handling key stores.

Goals, as I understand it:
 - Provide all the operations needed for useful applications, in a somewhat atomic form.
 - Be easy to call from FFI (for Ruby/Go/Python/whatever): opaque structs, and ideally a single header (this is nice for langs like Swift that can actually parse a C header to do FFI)
 - Anything else?

I'm also going to look over some other PGP APIs (including gpgme, Golang's x/crypto/openpgp, python-gnupg, others as I find them) to figure out what additional operations are relevant/useful. I think one reasonable sub-goal would be to define interfaces sufficient to implement all of the functionality in the `rnp` and `rnpkeys` tools, without using the original NetPGP interface in `rnp.h` at all, though rewriting `rnp` and co is not part of the agenda of this branch.


include/rnp/rnp2.h
```diff
@@ -0,0 +1,241 @@
...
+* @param buf the serialized key (either binary or ASCII formats accepted)
+* @param buf_len length of buf in bytes
+*/
+rnp_result rnp_public_key_load(rnp_public_key *key, const uint8_t buf[], size_t buf_len);
+
+/**
+* @param key the public key
+* @param uid on success will be set to the UID of the key
+*/
+rnp_result rnp_public_key_get_uid(rnp_public_key key, char **uid);
```
dewyatt: Maybe this would return the most recent uid? I think we'll need some more APIs to deal with keys that have multiple uids.

include/rnp/rnp2.h
```diff
@@ -0,0 +1,241 @@
...
+
+/**
+* @param key the public key
+* @param fprint on success will be set to the fingerprint of the key
+*/
+rnp_result rnp_public_key_get_fprint(rnp_public_key key, char **fprint);
```
dewyatt: May want to provide a similar API for keyids and key grips too.

include/rnp/rnp2.h
```diff
@@ -0,0 +1,241 @@
...
+typedef int (*rnp_passphrase_cb)(void *          app_ctx,
+                                 rnp_private_key key,
+                                 char            buf[],
+                                 size_t          buf_len);
+
+/*
+* Load a private key into memory. Initially it will be encrypted,
+* assuming the input is encrypted.
+*/
+rnp_result rnp_private_key_load(rnp_private_key *key, const uint8_t buf[], size_t buf_len);
```
dewyatt: I wonder how we should handle different formats here. Typically it would be PGP or G10. We could just auto-detect internally (G10 starts and ends with parentheses, so it's not like it's difficult to guess).

include/rnp/rnp2.h
```diff
@@ -0,0 +1,241 @@
...
+
+rnp_result rnp_private_key_is_decrypted(rnp_private_key key);
+
+rnp_result rnp_private_key_decrypt(rnp_private_key key, rnp_passphrase_cb cb, void *app_ctx);
```
dewyatt: How would the concepts map over? I mean in the current API we have lock/unlock and protect/unprotect.

include/rnp/rnp2.h
```diff
@@ -0,0 +1,241 @@
...
+                            const char *     userid,
+                            pgp_pubkey_alg_t prim_alg,
+                            pgp_pubkey_alg_t sub_alg,
+                            pgp_hash_alg_t   signature_hash,
+                            uint32_t         expiration);
+
+rnp_result rnp_private_set_hash_prefs(rnp_private_key key,
+                                      const uint8_t   hash_prefs[],
+                                      size_t          len_hash_prefs);
+/* TODO other set_*_prefs plus get_*_prefs */
```
dewyatt: This stuff is going to be tough I think. All these prefs are set inside signature subpackets so we would need to locate the appropriate signature (which userid?) if it exists, remove it, and generate a new signature that has everything that was previously there and additionally the new prefs. I think.

include/rnp/rnp2.h
```diff
@@ -0,0 +1,241 @@
...
+* @param output_len will be set to the length of output
+* @param cipher_used if non-null will be set to the cipher used
+* to encrypt the message.
+*/
+rnp_result rnp_decrypt(rnp_private_key key,
+                       const uint8_t   msg[],
+                       size_t          msg_len,
+                       uint8_t **      output,
+                       size_t *        output_len,
+                       pgp_symm_alg_t *cipher_used);
```
dewyatt: This one may be more difficult than this. The user may not know what key was used to encrypt something and instead would want the API to search some keyring for the appropriate seckey (and possibly request password for that seckey if needed).

include/rnp/rnp2.h
```diff
@@ -0,0 +1,241 @@
...
+* the hash algorithm used to create the signature will be returned.
+* @return RNP_SUCCESS if valid otherwise error
+*/
+rnp_result rnp_verify_detached(rnp_public_key  key,
+                               const uint8_t   msg[],
+                               size_t          msg_len,
+                               const uint8_t   sig[],
+                               size_t          sig_len,
+                               pgp_hash_alg_t *hash_used);
+
```
dewyatt: This API may not be acceptable for detached verification. We should be able to validate a detached sig of a 1TB file, for example, without reading the entire thing into a memory buffer.

include/rnp/rnp2.h
```diff
@@ -0,0 +1,241 @@
...
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+/*
+* Currently we use some definitions/enums from other others, including
+* - rnp_result
+* - pgp_hash_alg_t
+* - pgp_pubkey_alg_t
+*
```
dewyatt: What do you think about using strings instead of some of the enums? Like for hash, "SHA-256" or something.

I guess maybe I'm *slightly* biased against enums for FFI because I have to do [this](https://github.com/riboseinc/ruby-rnp/blob/master/lib/rnp/lowlevel/enums.rb) sort of error-prone duplication (which is, of course, now out of date).

include/rnp/rnp2.h
dewyatt: I guess we could require clients to mmap, but I think it may be an awkward requirement to have some python/ruby script doing some platform-specific file memory mapping operation. Thoughts?

include/rnp/rnp2.h
```diff
@@ -0,0 +1,241 @@
...
+
+/* Operations on public keys */
+
+/**
+* Load a public key
+* @param key a pointer, on success will be assigned to the new object
+* @param buf the serialized key (either binary or ASCII formats accepted)
+* @param buf_len length of buf in bytes
+*/
+rnp_result rnp_public_key_load(rnp_public_key *key, const uint8_t buf[], size_t buf_len);
```
dewyatt: Similar to the one for loading a private key, the format may need to be a parameter here, right? I mean it could be:
* KBX
* PGP
* PGP (armored) (could auto-detect this though)

include/rnp/rnp2.h
ni4: Agree on this - we should have the way to return uid by index. Probably it would be handier to leave this one to return first/default uid, and add rnp_public_key_get_uid_ex(key, uid, int idx, int *count).

include/rnp/rnp2.h
```diff
@@ -0,0 +1,241 @@
...
+
+/**
+* @param sub_alg may be 0 (NONE) to disable subkeys
+*/
+rnp_result rnp_generate_key(rnp_private_key *key,
+                            const char *     userid,
+                            pgp_pubkey_alg_t prim_alg,
+                            pgp_pubkey_alg_t sub_alg,
+                            pgp_hash_alg_t   signature_hash,
+                            uint32_t         expiration);
```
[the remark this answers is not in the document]
ni4: What about have multiple functions like rnp_generate_key_ecdsa(), rnp_generate_key_rsa(keybits, subkeybits) and so on?

include/rnp/rnp2.h
ni4: Why not to specify file pathes/some kind of stream handles here? Will this work over FFI?

include/rnp/rnp2.h
ni4: Agree, this could require some callbacks.

include/rnp/rnp2.h
randombit: Yes makes sense

include/rnp/rnp2.h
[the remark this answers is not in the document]
randombit: My thought here was if sub_alg == NONE then it would not generate a subkey, otherwise it would. But, since a key can legitimately have more than one subkey, it would probably be better to have generate_key + add_subkey operations. (Plus I guess also remove_subkey, not totally sure if that's useful).

And yes this should provide a way to specify key params. I'm torn on a `generate_key` per algo type. One the one hand it makes things a little simpler in each API, but it causes a combinatoric explosion given we already suppport RSA, DSA, ECDSA, ECDH, EdDSA, ElGamal, and SM2. Yikes. I don't know what the best approach is here.

I wanted to avoid a union as is used in `rnp_keygen_crypto_params_t` but `union`s in FFI seem kind of messy/dangerous. Maybe this is fine after all.

include/rnp/rnp2.h
randombit: I'm good with strings for this kind of thing, avoids forcing the FFI library to know more details of the PGP format than it must. Nobody should need to know that SHA-256 == 8 but librnp itself.

include/rnp/rnp2.h
randombit: Yeah good point, especially as it seems Ruby doesn't even have `mmap` built in. I think this function as it is would be useful for a lot of cases, but we'll need something else to deal with very large files.

We could pass a `FILE*` or file descriptor to read from. But file descriptor is Unix specific, and I'm not sure if creating a stdio `FILE*` for passing to a C function exists in Ruby or Python (in a brief search I couldn't find anything on this for either lang).

So I guess file path it is?

include/rnp/rnp2.h
randombit: Maybe we don't need this right away.

include/rnp/rnp2.h
dewyatt: > We could pass a FILE* or file descriptor to read from. But file descriptor is Unix specific, and I'm not sure if creating a stdio FILE* for passing to a C function exists in Ruby or Python (in a brief search I couldn't find anything on this for either lang).

It's not hard to use `fopen` [via ruby-ffi](https://github.com/ffi/ffi/blob/c52ceb173288073a321a94b401243ff40110d910/spec/ffi/memorypointer_spec.rb#L74) and I've done it in the past, but I would have concerns about inadvertently passing a `FILE*` created in one libc over to a library that may be using a different libc. Like what if my ruby code makes a reasonable assumption that `libc` is being used, so it uses `fopen` from `libc`...but it turns out `librnp` is using `musl`. I wouldn't expect it's standardized enough to do anything but crash, though I admit I haven't looked.

> So I guess file path it is?

Seems like it may be the simplest way.

include/rnp/rnp2.h
ronaldtse: Fully agree!

include/rnp/rnp2.h
ronaldtse: I think file path is the preferred approach, but we also need to make sure the FFI libraries handle files streaming properly, otherwise the improvements @ni4 made in memory usage could be annulled, say if the ruby-rnp reads in the whole file and passes it to librnp. Maybe some documentation could help.

include/rnp/rnp2.h
ronaldtse: Yes we need callbacks 👍

include/rnp/rnp2.h
ronaldtse: Agree that we can do this later.

include/rnp/rnp2.h
ronaldtse: @dewyatt could you propose a modified structure here?

include/rnp/rnp2.h
ronaldtse: Or we could extract out an additional function called `rnp_detect_key_format`? Too much magic could make like difficult.

include/rnp/rnp2.h
ronaldtse: Agree with @dewyatt about more APIs here and with @ni4 about allowing random access.

include/rnp/rnp2.h
ronaldtse: Agree, there should be functions to get basic information from a public / private key.

include/rnp/rnp2.h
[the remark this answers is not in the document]
ronaldtse: I think we should separate the API for generating subkeys since there could be more than one. Specifically, `generate_key` (main key and subkeys) and `add/remove_subkey` should be separated so we could set key params and preferences in a simple way.

include/rnp/rnp2.h
ronaldtse: We could also add a separate `detect` function here, I don't think the FFI library is expected to know all that information.

include/rnp/rnp2.h
ni4: This could be addressed (not right now of course :) ) with addition of public key signing/revoking methods. We need it at some point to allow public key cross-signing/building web of trust/supporting key servers, don't we?

randombit:
Updated this and a lot of implementation bits are now in. Still a few functions not implemented including detached signatures (need to modify some code in the lib for those), and some functionality that is almost certainly required doesn't even have interface definitions (eg no way to get the fingerprint of a key right now).

I removed the public_key/private_key notions because it turned out to be quite hard to express in terms of how RNP is structured internally. Eg, there is no good way to copy a `pgp_key_t` to another `pgp_key_t`. Anyway it got messy quickly and I was bogged down in that for a long time. So now all operations are expressed in terms of a keyring, and that seems to work fairly well.

@ronaldtse @dewyatt take a look and let me know what you think.

dewyatt:
This is looking better. We actually both ended up doing some work independently on this (I did some yesterday). [Gist here](https://gist.github.com/dewyatt/38918db2a399893e9908e67fa0affa5e) but I'll probably just add comments on this PR.

I came to the same conclusion on the public/private separation. I don't think It's really rnp's internal structure that prevents it, it's more PGP itself.

include/rnp/rnp2.h
```diff
@@ -0,0 +1,234 @@
...
+#include <stdint.h>
+
+/*
+* Currently we use some definitions/enums from other others, including
+* - rnp_result_t
+*
+* It might be good to consolidate these definitions into one spot, or
+* else replicate them here in new enums and make this header freestanding.
+*/
+#include <rnp/rnp_def.h>
```
dewyatt: We should get rid of this and just include whatever we need here.
(I also ended up adding an `rnpi` prefix to everything as a lazy way of avoiding naming conflicts, but probably not necessary.)

include/rnp/rnp2.h
```diff
@@ -0,0 +1,234 @@
...
+* @return 0 on success, or any other value to stop decryption.
+*/
+typedef int (*rnp_passphrase_cb)(void *      app_ctx,
+                                 const char *pgp_context,
+                                 char        buf[],
+                                 size_t      buf_len);
+
+/* Operations on key rings */
+
+rnp_result_t rnp_keyring_open(rnp_keyring_t *   keyring,
```
dewyatt: You may want to check the stuff I did in my gist for these interfaces (maybe). I'm not sure this is flexible enough.

include/rnp/rnp2.h
```diff
@@ -0,0 +1,234 @@
...
+* Load an armored public key into the keyring
+* @param key a pointer, on success will be assigned to the new object
+* @param key the PGP armored public key
+*/
+rnp_result_t rnp_insert_armored_public_key(rnp_keyring_t keyring, const char *key);
+
+/**
+* Generate a key and add it to the keyring
+*/
+rnp_result_t rnp_generate_private_key(rnp_keyring_t keyring,
```
dewyatt: I'm kind of wondering if we would be better off just doing a quick implementation of https://github.com/riboseinc/rnp/issues/318 for this.

I started brainstorming on it a bit earlier today.
It would be pretty easy I think to pass some JSON to a key generation function. Would just need to merge in some defaults, check for unused object attributes (error), etc.

```javascript
// generate both a primary and subkey
{
"primary": {
    "type": "ECDSA",
    "curve": "secp256r1",
    "usage": ["sign"]
 },
 "subkey": {
    "type": "RSA",
    "length": 4096
 }
}

// generate just a subkey
{
"primary": {
    "grip": "...",
 },
 "subkey": {
    "type": "RSA",
    "length": 4096
 }
}
```

Anyways, it's worth considering I think. Because even with all these parameters here, we are still missing certain things (like key usage).

Also, since we have separated key generation from key protection, the passphrase parameters shouldn't be here.

src/lib/pgp-key.c
```diff
@@ -834,16 +834,39 @@ pgp_key_protect(pgp_key_t *                      key,
...
+        return false;
+    }
+
+    bool ret = pgp_key_protect_passphrase(key, format, protection, passphrase);
+    pgp_forget(passphrase, sizeof(passphrase));
+    return ret;
+}
+
+bool
+pgp_key_protect_passphrase(pgp_key_t *                  key,
```
dewyatt: I actually considered having a passphrase parameter (instead of a provider) when I implemented `pgp_key_protect` 
 originally. I introduced the passphrase provider concept to deal with cases where a passphrase *may* be required, and key protection is a case where a passphrase *is* definitively always required, so it might make sense for it to just take a passphrase parameter.

include/rnp/rnp2.h
dewyatt: (Also I certainly wouldn't mind tackling this particular piece if you're thinking "I have enough work to do")

include/rnp/rnp_def.h
```diff
@@ -77,6 +77,7 @@ enum {
     RNP_ERROR_NOT_SUPPORTED,
     RNP_ERROR_OUT_OF_MEMORY,
     RNP_ERROR_SHORT_BUFFER,
+    RNP_ERROR_NULL_POINTER,
```
dewyatt: You may want to just use `RNP_ERROR_BAD_PARAMETERS` (which IMO would be better named `RNP_ERROR_INVALID_ARGS` or something, but maybe that's just the US)

include/rnp/rnp_def.h
```diff
@@ -88,6 +89,11 @@ enum {
     RNP_ERROR_MAC_INVALID,
     RNP_ERROR_SIGNATURE_INVALID,
     RNP_ERROR_KEY_GENERATION,
+    RNP_ERROR_KEY_NOT_FOUND,
+    RNP_ERROR_NO_SUITABLE_KEY_FOR_OPERATION,
```
dewyatt: (Not important but `RNP_ERROR_NO_SUITABLE_KEY` seems descriptive enough)

dewyatt:
We should also add an FFI function to get a descriptive error message for an `rnp_result_t` at some point.

include/rnp/rnp_def.h
randombit: Yeah could go either way on this, I kind of like `NULL_POINTER` since it explicitly calls out the issue vs `BAD_PARAMS` which could be referring to anything. But no strong feelings on it.

include/rnp/rnp_def.h
randombit: Better, thanks

include/rnp/rnp2.h
randombit: Only thing we need is rnp_result_t and I hadn't realized that's actually just a uint32_t, not an enum. So all that's needed is an equivalent typedef at the top of this header and that makes rnp2.h freestanding which is definitely better.

include/rnp/rnp2.h
randombit: Yeah I wasn't so happy with that. There are a lot of optional params and etc. So maybe JSON is the way to go here. If you want to take this part on feel free :)

include/rnp/rnp2.h
randombit: I think the way you have it is better

randombit:
Yes, useful - added rnp_result_to_string

include/rnp/rnp_def.h
dewyatt: Yeah I think I agree with you after looking again, there's a useful distinction here.

include/rnp/rnp2.h
dewyatt: Ok I'll get started on that piece today, I think it'll be pretty quick.

include/rnp/rnp2.h
dewyatt: Ok I'll get this updated in a bit, I think we can merge some ideas here as I think I'm seeing some improvements on both sides.

include/rnp/rnp2.h
```diff
@@ -0,0 +1,283 @@
...
+                      const char *  hash_fn,
+                      bool          clearsign,
+                      bool          armor,
+                      const uint8_t msg[],
+                      size_t        msg_len,
+                      uint8_t **    sig,
+                      size_t *      sig_len);
+
+/**
+* Generate an embedded signature (the output will include the
```
ni4: This should be something like 'Generated detached signature (the signature will be saved in separate from input file)'

include/rnp/rnp2.h
```diff
@@ -0,0 +1,283 @@
...
+* @param hash_fn the hash function to use
+* @param msg the message to sign
+* @param msg_len the length of msg in bytes
+* @param sig on success, the output signature buffer
+* @param sig_len on success, the length of sig in bytes
+*/
+rnp_result_t rnp_sign(rnp_keyring_t keyring,
+                      const char *  ident,
+                      const char *  hash_fn,
+                      bool          clearsign,
```
ni4: Imho it would be logical to separate cleartext signing from 'normal' signing since they are a bit different things. Also such interface will not allow signing with multiple keys.

include/rnp/rnp2.h
```diff
@@ -0,0 +1,283 @@
...
+/**
+* Verify a signature with embedded message
+* @param key the keyring
+* @param sig the signature
+* @param sig_len length of signature in bytes
+* @param msg on succes, output buffer to the message
+* @param msg_len on success, length of msg in bytes
+*/
+
+rnp_result_t rnp_verify(
```
ni4: I believe we should add some 'keyneeded' callback instead of keyring here and in decryp() function. So we could search for the key (on PGP servers, multiple files, database, etc) instead of loading the whole known bunch of keys into memory.

include/rnp/rnp2.h
```diff
@@ -0,0 +1,283 @@
...
+* @return RNP_SUCCESS if valid otherwise error
+*/
+rnp_result_t rnp_verify_detached_file(rnp_keyring_t keyring,
+                                      const char *  file_path,
+                                      const uint8_t sig[],
+                                      size_t        sig_len);
+
+/* Encryption/decryption operations */
+
+rnp_result_t rnp_encrypt(rnp_keyring_t keyring,
```
ni4: Same as with sign() call - what about multiple recipients?

include/rnp/rnp2.h
dewyatt: I forgot to update all of these functions...

include/rnp/rnp2.h
dewyatt: > Imho it would be logical to separate cleartext signing from 'normal' signing since they are a bit different things.

Maybe, I'll take a look.

> Also such interface will not allow signing with multiple keys.

We don't really support this internally at the moment though. In the future we could add some kind of "sign operation" concept and "add signer" "add data" operations, etc. We're not aiming to get everything done perfectly on the first round here.

include/rnp/rnp2.h
dewyatt: I think the key store may eventually evolve to provide this functionality per the discussion in https://github.com/riboseinc/rnp/issues/351#issuecomment-320001558., though we're a long way from that currently.

I think that adding a "storage provider/backend" concept to the key store might be a potential way of handling it, for example. Then a "find key" operation on a key store could query the database backend, etc. (Probably better discussed in the linked issue)

@ronaldtse For an immediate solution, would a 'keyneeded' callback (for verify and decrypt operations) be necessary for what you need?

include/rnp/rnp2.h
dewyatt: Same deal, we don't really support it internally yet, so until then...

ni4: Actually, now I started moving pk-encryption/decryption into the stream architecture, and plan to add multiple-recipient encryption during that. So, soon we'll support it :-)

dewyatt: @ronaldtse Is this something Ribose will need in the initial FFI/ruby-rnp? (If not, then we can leave this for later)

randombit: Well it is easy to support at the API level for now, pass `const char* receipients[], size_t recipients_len` and for now just error if more than 1 is specified.

randombit:
@dewyatt Changes look good to me I will start implementation of some of the other rnp_key_ APIs and keep pushing here.

Also re the comment in the header, I think the fingerprint and co is probably best exposed as a hex string, at least that seems more convenient to me. In the odd case where someone actually needs the binary value (maybe for a database lookup?) they can just decode it themselves.

include/rnp/rnp2.h
dewyatt: Actually looking closer at some of these interfaces I see other similar issues and it may be easiest to get this callback in place. Let me see what we can do.

randombit:
@dewyatt Do you have commits active on this branch? I'd like to rebase + force push to resolve the conflict with master but don't want to screw things up on your side.

dewyatt:
@randombit No worries, go for it!

randombit:
Rebased, added yet more functions, still not building though - need to get back to having working tests.

dewyatt:
> Rebased, added yet more functions, still not building though - need to get back to having working tests.

I've got some local work I'll get integrated tomorrow and will try to get that fixed up too.

src/lib/rnp2.c
```diff
@@ -897,7 +902,7 @@ rnp_generate_private_key(rnp_keyring_t keyring,
 }
 
 rnp_result_t
-rnp_key_free(rnp_key_t *key)
+rnp_key_free(rnp_key_t key)
```
dewyatt: If we have `rnp_key_free(rnp_key_t *key)` then we can NULL the pointer out on the caller's side with `*key = NULL` after freeing. Do you feel it's not really worth it?

include/rnp/rnp2.h
ronaldtse: @dewyatt yes the callback would work already. The "storage provider" / "backend" concept is exactly what we need.

src/lib/rnp2.c
randombit: Fine either way, I guess nulling it out will probably save someone a crash someday though.

dewyatt: It does seem like the chances of it being beneficial are pretty low and it makes the API slightly more difficult to use, so your way is probably better.

dewyatt:
I've got FFI key generation w/JSON done, seems to be working well. I meant to get it cleaned up and pushed today but I spent more time chasing bugs than planned and it's getting late, so I'll continue tomorrow.

The current schema looks like (for generating both a primary and sub):

```json
{
  "primary": {
      "type": "ECDSA",
      "curve": "NIST P-256",
      "userid": "test0",
      "usage": ["sign"],
      "key expiration": 0,
      "preferences" : {
        "hashes": ["SHA512", "SHA256"],
        "ciphers": ["AES256", "AES128"],
        "compression": ["zlib", "bzip2"],
        "key server": "hkp://pgp.mit.edu"
      }
   },
   "subkey": {
      "type": "RSA",
      "length": 1024
   }
}
```

randombit:
@dewyatt I added back a simplified keygen code just for testing purposes. Much easier to verify functionality when the tests work :) We can change it to use the JSON-based generator once that lands.

dewyatt:
Definitely a WIP, but I pushed some stuff. I"ll probably break out all the supporting utils into a separate PR I guess.

I need to think a bit more on some of these interfaces.

@randombit Let me know if it's OK to rebase. (Also, unrelated, but am I the only one seeing the sm2_roundtrip test failing in master?)

ni4:
@dewyatt sm2_roundtrip fails on Mac as well. Didn't debug it yet anyway.

randombit:
@dewyatt Rebase away

randombit:
@dewyatt @ni4 SM2 tests are failing due to assumption on the size of ciphertext which changed recently b/c in Botan we switched to using the ASN.1 format for the ciphertext, which is what the newer SM2 standard has. My original code followed the IETF draft. I'll post a PR with a fix in a bit.

randombit:
@dewyatt Rebased

randombit:
The rebases are getting increasingly hairy, I'm wondering if we should merge this to master as as somewhat provisional/WIP change, and then continue improvements in a new PR.

randombit:
LOL warring rebases

dewyatt:
haha oops, missed that message!

dewyatt:
> The rebases are getting increasingly hairy, I'm wondering if we should merge this to master as as somewhat provisional/WIP change, and then continue improvements in a new PR.

Agreed we should get this merged. This PR doesn't really change any existing functionality anyways.

ronaldtse:
@dewyatt could you approve and merge this for the new WIP PR? Thanks! 👍

PR #457, opened 2017-09-13
tests: Provide explicit path for gpg to cli_tests.py.

dewyatt:
It looks like CI isn't always finding the correct gpg, for whatever reason, so it's likely safest to just provide an explicit path.
I just used the environment here, but an arg would certainly work as well. It still falls back to the normal search.


dewyatt:
This should stabilize CI again.

PR #459, opened 2017-09-15
Clean up inclusion of botan/ffi.h

randombit:
All super piddly stuff actually but I noticed it while looking at something else and it bugged me.

Basically don't include `<botan/ffi.h>` in any headers, just in the files that directly use it. Changes ECDH interface slightly (taking BIGNUM instead of botan_mp_t)

Removes `new_BN_take_mp` because it was only used in one place and the code is simpler without it.

The BN layer is due for a pruning, a lot of functionality in there was just to implement ElGamal etc and is no longer necessary. But not really a problem so ignoring that here.

### DIFF
--- a/src/lib/crypto/ecdh.c
+++ b/src/lib/crypto/ecdh.c
@@ -202,7 +202,7 @@ pgp_ecdh_encrypt_pkcs5(const uint8_t *const     session_key,
                        size_t                   session_key_len,
                        uint8_t *                wrapped_key,
                        size_t *                 wrapped_key_len,
-                       botan_mp_t               ephemeral_key,
+                       BIGNUM *                 ephemeral_key,
                        const pgp_ecdh_pubkey_t *pubkey,
                        const pgp_fingerprint_t *fingerprint)
 {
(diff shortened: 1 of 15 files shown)


src/lib/crypto/ecdh.c
```diff
@@ -215,7 +215,7 @@ pgp_ecdh_encrypt_pkcs5(const uint8_t *const     session_key,
     uint8_t *       tmp_buf = NULL;
     uint8_t         kek[32] = {0}; // Size of SHA-256 or smaller
 
-    if ((session_key_len > OBFUSCATED_KEY_SIZE) || !ephemeral_key || !pubkey || !wrapped_key ||
+    if ((session_key_len > OBFUSCATED_KEY_SIZE) || !ephemeral_key->mp || !pubkey || !wrapped_key ||
```
flowher: I think we should be checking both here - "!ephemeral_key || !ephemeral_key->mp"

src/lib/crypto/ecdh.c
```diff
@@ -312,7 +312,7 @@ pgp_ecdh_decrypt_pkcs5(uint8_t *                session_key,
                        size_t *                 session_key_len,
                        uint8_t *                wrapped_key,
                        size_t                   wrapped_key_len,
-                       const botan_mp_t         ephemeral_key,
+                       const BIGNUM *           ephemeral_key,
```
flowher: With this change we should check against NULL at line 328

ronaldtse:
@randombit could you help check the comments by @flowher so we can merge this? Thanks!

randombit:
Rebased and added null checks @flowher can you take a second look?

flowher:
Thanks! LGTM

PR #460, opened 2017-09-15
Add secret subkey printing to list-packets.

dewyatt:
Noticed this wasn't supported, so just a quick little update here.


ronaldtse:
Thanks @dewyatt !

PR #462, opened 2017-09-16
Check during build-time if botan was build with required algorithms

flowher:
Update for list of required algorithms


ronaldtse:
Looks fine except the line is really long. Any thoughts @randombit ?

PR #463, opened 2017-09-17
Streamed symmetric encryption

ni4:
**This is a forward PR - call for discussion, with ongoing updates.**

It doesn't make a lot of changes into existing codebase (mostly additions to some types), however adds a new architectural solution for streamed data parsing and processing/writing. And adds feature for symmetric (password) encryption/decryption for the files as the demonstration of this architecture.
Currently it may be checked with `rnp --sym-decrypt` and `rnp -c` commands. `--sym-decrypt` is only a temporary solution - later on `--decrypt` will be able to process the source, distinguishing whether it is encrypted with the password or with the key.

**Parsing  the input data.**

Instead of having separate calls for `decrypt()`, `verify()`, checking for armoring and so on everything could be done in a single `process_pgp_source` call. It autodetects stream contents, and builds correct chain of `pgp_source_t` (see below), which will do all the work. When chain of sources is built then processing is as easy as reading chunks from the topmost source and writing to the output `pgp_dest_t`

`pgp_source_t` is the abstract object, which main functions are reading from something. Also it has cache so reads are cached which improves performance. Reading is done via `src_read()` and `src_peek()` calls. The last one just looks forward for the data, without dismissing it from the source, which makes parsing easier.

It has quite easy implementation for files/stdin, however for encrypted pgp stream, partial data stream, compressed stream and so on it is more complicated, and has additional fields - like list of keys/passwords it is encrypted to, compression algorithm, and so on.

**Writing output data.**
Writing is done via chain of `pgp_dest_t` objects. For instance, for symmetric encryption it would be:
- file
- encrypting stream (which may contain partial-length writing stream)
- compressing stream (which also may contain partial-length writing stream)
- literal data stream (together with partial-length writing stream)

So, when chain of dests is built and initialized then data processing is as easy as reading chunk from the stdin/file and writing to the last stream (literal data). 

**What can be easily done with this new architecture.**
- processing of streams of any kind and size with constant memory overhead. The only requirement is that it must be readable, and able to report the end of stream. Requirement for output - is to be writeable and closable.
- easy processing of new combinations of streams (like just compressed file, encrypted and signed file, encrypted with multiple passwords file, encrypted with multiple keys and passwords file, etc)
- simple addition of new compression/encryption algorithms
- easy rewriting of base64 encoding/decoding data (it is not implemented yet, but will be soon).

**Open questions:**
- where it stream-*.c sources should be? librepgp or rnp?
- to which error reporting define or function do we stick to? I used fprintf(stderr, ..), but this should be replaced.
- most likely it will be needed to add caching in dst_write, as it is done in src_read
- most likely I missed some `static` functions and definitely didn't `clang-format`

After discussing and approving of these changes I can easily move existing public key stream encryption to the new architecture, as well as signing/verification.


src/librepgp/stream-parse.h
```diff
@@ -0,0 +1,112 @@
...
+/* statically preallocated cache for sources. Not used for input filters */
+typedef struct pgp_source_cache_t {
+    uint8_t buf[PGP_INPUT_CACHE_SIZE];
+    unsigned pos;
+    unsigned len;
+} pgp_source_cache_t;
+
+typedef struct pgp_source_t {
+    pgp_source_read_func_t  *readfunc;
+    pgp_source_close_func_t *closefunc;
```
dewyatt: Minor detail but instead of `readfunc`/`closefunc` for field names, I think `read`/`close` or `reader`/`closer` would be fine (like how we currently have `pgp_writer_func_t *writer`). It's obvious from context at the call sites that it's a function, so it seems overly verbose maybe.

src/librepgp/stream-parse.h
```diff
@@ -0,0 +1,112 @@
...
+    pgp_source_type_t        type;
+
+    int64_t             size;  /* size of the data if available, 0 otherwise */
+    int64_t             read;  /* number of bytes read from the stream via src_read. Do not confuse with number of bytes as returned via the readfunc since data may be cached */
```
dewyatt: Just curious why `size` and `read` are signed?

dewyatt:
@ni4 I saw armored symmetric encryption doesn't work?
```
$ src/rnp/rnp -c --armor
rnp_encrypt_stream: encryption failed with error code 26843545
```

ni4:
@dewyatt Yeah, I didn't implement armoring/dearmoring streams yet. Hope to update PR later today or tomorrow.

src/librepgp/stream-parse.h
ni4: Agree on this, will rename.

src/librepgp/stream-parse.h
ni4: Basically, don't know. Just quick-typed. Will change, thanks.

src/librepgp/stream-write.c
```diff
@@ -0,0 +1,943 @@
...
+    /* stack of the streams would be as following:
+       [armoring stream] - if armoring is enabled
+       encrypting stream, partial writing stream
+       [compressing stream, partial writing stream] - if compression is enabled
+       literal data stream, partial writing stream
+    */
+    uint8_t       readbuf[PGP_INPUT_CACHE_SIZE];
+    ssize_t       read;
+    pgp_dest_t    dests[4];
+    int           destc = 0;
```
dewyatt: The `dests`+`destc` approach feels flimsy, it might be better to do something similar to the current `pgp_writer_push` approach..

dewyatt:
1. How would `pgp_source_t`/`pgp_source_read_func_t` relate to `pgp_stream_t` / `pgp_reader_t`? Would it replace them? What are the advantages (I did already read the above)?

    Same for `pgp_dest_t`/`pgp_dest_write_func_t`, etc.

2. > When chain of sources is built then processing is as easy as reading chunks from the topmost source and writing to the output pgp_dest_t

    Do we not already have this though? I mean we currently do things like 
    `pgp_reader_push_dearmour`+`pgp_reader_push_decrypt`, etc.

3. > pgp_source_t is the abstract object, which main functions are reading from something. Also it has cache so reads are cached which improves performance. 

    Well this sort of thing has to be quantified of course. The OS has its own caches, libc has its own (though I see we're using syscalls here), and this adds another. We should probably measure whether it is of real benefit.

4. Maybe I'm missing it, but I didn't see anything that would signal "we need more data". If I think of an application streaming something in over the network, for example, I expect the library to maybe have 1) a way of indicating that more data is needed to complete processing and 2) a way to add more data to be processed.

I'll continue looking through it later, so maybe that will clear up my understanding of the above. Mostly I just want to get a clear understanding of the benefits that this provides and how it relates to existing stuff.

src/librepgp/stream-write.c
ni4: Internally dests have pointers to the next one. This array is to have access from the calling function to the whole list of dests. For instance, if some of them was failed to initialize and didn't put pointer to the correct place then we will not be able to dealloc things.

ni4:
@dewyatt 
1. Yes, it is going to replace later on `pgp_stream_t` and all corresponding stuff. The reason for that is that currently all those code is quite mixed up, non-isolated in functionality and hard to maintain. For instance, why should `zlib_compressed_data_reader` should card about `z->region->indeterminate`, and so on. Within new approach all code, related to one pgp stream, will not care about packet formats, and upper and lower-level streams. 
The same about `pgp_dest_t` - switching between partial and indeterminate/fixed packet length would take just 2 lines of code in one place.

2. The same as 1 - we have something similar, but currently everything mixed up together, and kinda packet-oriented - whole chunk of processed data is written to the memory and then re-processed. See for instance `pgp_writez`

3. Caching here is to cache write/read requests - so even if some stream in the middle will output data by 1-2 bytes then writing to the file will be done in larger chunks. 

4. Signal for the 'more data' would read() call on bottom-layer source. It is not required to return all the data, may delay till some data appear, and so on. This could be easily redesigned (for instance, some callbacks added to the handler) in case we will need it.

src/lib/defs.h
```diff
@@ -136,6 +136,26 @@
...
+            } else {                                                                  \
+                (void) memset(&__newarrc[str->arr##vsize * sizeof(*str->arr##s)],     \
+                              0x0,                                                    \
+                              (__newsize - str->arr##vsize) * sizeof(*str->arr##s));  \
+                str->arr##s = __newarr;                                               \
+                str->arr##vsize = __newsize;                                          \
+            }                                                                         \
+        }                                                                             \
+    } while (/*CONSTCOND*/ 0)
+
```
catap: Maybe do not duplicate code and rewrite `EXPAND_ARRAY` to use a new macros?

src/lib/pass-provider.c
```diff
@@ -100,15 +100,30 @@ rnp_passphrase_provider_stdin(const pgp_passphrase_ctx_t *ctx,
...
+
+    if ((ctx->op != PGP_OP_DECRYPT_SYM) && (ctx->op != PGP_OP_ENCRYPT_SYM)) {
+        rnp_strhexdump(keyidhex, ctx->key->keyid, PGP_KEY_ID_SIZE, "");
+        snprintf(target, sizeof(target), "key 0x%s", keyidhex);
+    }
 start:
-    snprintf(prompt, sizeof(prompt), "Enter passphrase for %s: ", target);
+    if (ctx->op == PGP_OP_DECRYPT_SYM) {
+        snprintf(prompt, sizeof(prompt), "Enter passphrase to decrypt data: ");
+    } else if (ctx->op == PGP_OP_ENCRYPT_SYM) {
```
catap: I love `switch` and `case` to this but it is a style question.

src/lib/rnp.c
```diff
@@ -1147,6 +1149,148 @@ rnp_decrypt_file(rnp_ctx_t *ctx, const char *f, const char *out)
...
+int
+rnp_process_stream(rnp_ctx_t *ctx, const char *in, const char *out)
+{
+    pgp_source_t               src;
+    rnp_result_t               err;
+    pgp_parse_handler_t *      handler = NULL;
+    pgp_parse_handler_param_t *param = NULL;
+    int                        result;
+
+    if (in && (strlen(in) > sizeof(param->in))) {
```
catap: I think that's using `PATH_MAX` make this code easy to read but it also style question

src/lib/rnp.c
```diff
@@ -1147,6 +1149,148 @@ rnp_decrypt_file(rnp_ctx_t *ctx, const char *f, const char *out)
...
+        result = RNP_OK;
+    } else {
+        (void) fprintf(stderr, "processing failed: error 0x%x\n", err);
+        result = RNP_FAIL;
+    }
+
+finish:
+    src_close(&src);
+    free(handler);
+    free(param);
```
catap: you can `free` NULL pointer here

src/lib/symmetric.c
```diff
@@ -193,6 +193,7 @@ pgp_cipher_finish(pgp_crypt_t *crypt)
     return 0;
 }
 
+/* we rely on fact that in and out could be the same */
```
catap: yes, it is CFB cipher ;)

src/librepgp/stream-common.c
```diff
@@ -0,0 +1,445 @@
...
+        read = cache->len - cache->pos;
+        if (read >= len) {
+            memcpy(buf, &cache->buf[cache->pos], len);
+            cache->pos += len;
+            goto finish;
+        } else {
+            memcpy(buf, &cache->buf[cache->pos], read);
+            cache->pos += read;
+            buf = (uint8_t *) buf + read;
+            left = len - read;
```
catap: and if `read` returns something less than `0` that's happened?

src/librepgp/stream-common.c
```diff
@@ -0,0 +1,445 @@
...
+
+    if (!init_source_cache(src, sizeof(pgp_source_file_param_t))) {
+        close(fd);
+        return RNP_ERROR_OUT_OF_MEMORY;
+    }
+
+    param = src->param;
+    param->fd = fd;
+    src->read = file_src_read;
+    src->close = file_src_close;
```
catap: Just an idea: maybe use a `mmap` here makes the code easy?

src/lib/rnp.c
dewyatt: Nothing wrong with calling free on a NULL pointer.

src/lib/defs.h
ni4: I did this PR as non-breaking so added this macro instead of rewriting. Actually, most likely, we'll get rid of this define at some point in future at all.

src/lib/pass-provider.c
ni4: Most likely when this get to more then 3 branches it will be converted to the switch :)

src/lib/rnp.c
ni4: After reading some stuff about PATH_MAX i think we should not rely on it. Actually, better would be to use char * instead of pre-allocated arrays.

src/lib/rnp.c
ni4: Yeah, free(NULL) works pretty well.

src/librepgp/stream-common.c
ni4: mmap has some disadvantages, especially for large files: https://stackoverflow.com/questions/258091/when-should-i-use-mmap-for-file-access
Here we just don't need it, since streamed reading is enough.

src/librepgp/stream-common.c
ni4: `read` cannot be less than 0 because cache->len is always greater or equal than cache->pos, by design.

catap: ok, but why have you use `ssize_t` and don't use `size_t` in this case?

src/librepgp/stream-common.c
catap: @ni4 at Linux it works a littel different because Linux won't put a mmaped file to swap :) because it is on disk already.

anyway, rnp use mmap at another point and here my idea was make code similar.

src/lib/rnp.c
catap: yes, it is safe but looks strange for me. But I agree that it is a style question

src/librepgp/stream-common.c
ni4: Because later on it will be used to call src->read, which returns ssize_t. In this case cache->len and cache->pos are small values, so mixing together signed/unsigned will not loose precision.

src/lib/rnp.c
dewyatt: A particular implementation isn't really relevant, the C standard itself says `if ptr is a null pointer, no action occurs` to cover causes just like this.

ronaldtse:
@ni4 is armored en/decryption implemented? Want to merge this ASAP. Thanks!

ni4:
@ronaldtse I implemented dearmoring stream (i.e. decryption), and need some time to finish the armoring. Most likely it will be on Monday, it's much easier then dearmoring.

ni4:
@ronaldtse Armouring/dearmouring is done, however it seems there is some rarely-reproducible issue. I will try to reproduce it locally.

Also probably it worth some more reviews/checking it with fresh eye.

src/lib/rnp.c
```diff
@@ -1223,6 +1226,187 @@ rnp_decrypt_file(rnp_ctx_t *ctx, const char *f, const char *out)
...
+    handler->passphrase_provider = &ctx->rnp->passphrase_provider;
+    handler->dest_provider = rnp_parse_handler_dest;
+    handler->param = param;
+
+    err = process_pgp_source(handler, &src);
+    if (err == RNP_SUCCESS) {
+        result = RNP_OK;
+    } else {
+        (void) fprintf(stderr, "processing failed: error 0x%x\n", err);
+        result = RNP_FAIL;
```
flowher: You shouldn't use RNP_FAIL anymore

ni4: @flowher Thanks, will fix this.

catap:
(for me it's good and ready for merge but I haven't make a deep reading)

ni4:
@catap So you are welcome to approve it :) 
I fixed bug mentioned in the previous comment. Also this PR is non-destructive - i.e. it adds new functionality, but doesn't change any existing.

catap:
@ni4 done, but UI of github sometimes drives me crazy — I should re-request of review for myself to change decision.

ronaldtse:
Any further comments @dewyatt and @flowher ? 

According to tests by @ni4, the performance with this PR exceeds GnuPG. Anyone interested in making some tests to directly show this? 👍

src/librepgp/stream-armour.c
```diff
@@ -0,0 +1,987 @@
...
+    uint8_t  decbuf[ARMOURED_BLOCK_SIZE + 4]; /* decoded 6-bit values */
+    uint8_t *bufptr = buf;                    /* for better readability below */
+    uint8_t *bptr, *bend;                     /* pointer to input data in b64buf */
+    uint8_t *dptr, *dend, *pend; /* pointers to decoded data in decbuf: working pointer, last
+                                    available byte, last byte to process */
+    uint8_t  bval;
+    uint32_t b24;
+    ssize_t  read;
+    ssize_t  left = len;
+    int      eqcount; /* number of '=' at the end of base64 stream */
```
dewyatt: Can you initialize this to 0, my compiler believes it can be used uninitialized (seems potentially true).

```
stream-armour.c:382:19: error: ‘eqcount’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
```

src/librepgp/stream-common.c
```diff
@@ -0,0 +1,456 @@
...
+rnp_result_t
+init_file_src(pgp_source_t *src, const char *path)
+{
+    int                      fd;
+    struct stat              st;
+    pgp_source_file_param_t *param;
+
+    if (stat(path, &st) != 0) {
+        (void) fprintf(stderr, "can't stat \"%s\"\n", path);
+        return PGP_E_R_READ_FAILED;
```
dewyatt: Wrong set of error codes here.

src/librepgp/stream-common.c
```diff
@@ -0,0 +1,456 @@
...
+    }
+
+    param = src->param;
+    param->fd = fd;
+    src->read = file_src_read;
+    src->close = file_src_close;
+    src->type = PGP_STREAM_FILE;
+    src->size = st.st_size;
+    src->readb = 0;
+    src->knownsize = 1;
```
dewyatt: Not a big deal but I'm curious if this flag could just be replaced with a `src->size == 0` check.

src/librepgp/stream-parse.c
```diff
@@ -0,0 +1,1312 @@
...
+    return read;
+}
+
+void
+encrypted_src_close(pgp_source_t *src)
+{
+    pgp_source_encrypted_param_t *param = src->param;
+    if (param) {
+        FREE_ARRAY(param, symenc);
+        FREE_ARRAY(param, pubenc);
```
dewyatt: I think we ought to scrub each of these as well.

dewyatt:
It would probably take a week to do a thorough review, so I can't reliably say LGTM, but the code generally looks well-written to me.

I'd like to see a reduction in the code duplication but it's not a big deal at the moment.

We may want to add some kind of prefix on the public function names like `init_file_src`, `src_read`, etc. (may depend on where this code should live, as mentioned by ni4 above).

Overall I'm good with it since this PR is just adding new functionality and not yet converting existing things to this framework. I'd like to see some static analysis and fuzzing at some point in the future since it's a lot of tricky code.

src/librepgp/stream-armour.c
ni4: Sure, fixed. Actually it will not be used not initialized but it relies on flag eofb64, which seems not to be clear for compiler.

src/librepgp/stream-common.c
ni4: Thanks for noticing. Fixed.

src/librepgp/stream-common.c
ni4: Zero size logically is also a valid input size :) I removed knownsize - added it in advance, but didn't make use of it anywhere. Thanks.

src/librepgp/stream-parse.c
ni4: Do you mean we should change DYN_ARRAY to list, or.. ?

ni4:
@dewyatt Thanks for review. I fixed things you mentioned in comments. Actually, if questions or ideas arise, we easily could introduce changes in interfaces later on.

ronaldtse:
Thank you guys, let's merge this and post further issues about this separately!

src/librepgp/stream-parse.c
dewyatt: No I meant that we should maybe scrub the symmetric keys with `pgp_forget`.

ni4: Ah, got it. I don't keep symmetric keys in `pgp_sk_sesskey_t`, the idea about it is that it keeps just information from the packet itself. All keys and decryption data will go to the decrypting stream (we need just one correct session key to decrypt the stream).

However It seems I forgot to `pgp_forget` in some other places, thanks.

PR #464, opened 2017-09-17
Code consistency: rename type rnp_result to rnp_result_t

flowher:


ni4:
@dewyatt Maybe this is not the best place to ask, but what do you think about the recently added s2k type? Most likely we should rename it to s2k_t as well?

dewyatt:
> @dewyatt Maybe this is not the best place to ask, but what do you think about the recently added s2k type? Most likely we should rename it to s2k_t as well?

Yep, I must have rushed and missed that. There's a couple more in types.h we need to rename as well.

PR #467, opened 2017-09-19
Add FreeBSD build and move installation steps to before_install.sh

ronaldtse:


ronaldtse:
Build related, going to just merge first so we also get FreeBSD builds.

PR #468, opened 2017-09-21
Add support for adding additional userids to keys, fix key importing

dewyatt:
Ok I had to expand this PR and fix an unrelated issue or CI would be broken.

So this:
* Fixes key importing (https://github.com/riboseinc/rnp/issues/21). I've only tested .gpg format. It would likely (should) fail for imports that require conversions (like importing a G10 key into a GPG key store).
    * Support secret keys.
    * Actually save to disk.
* Adds some support+tests for adding additional userids to keys.

### DIFF
--- a/include/repgp/repgp_def.h
+++ b/include/repgp/repgp_def.h
@@ -189,7 +189,6 @@ typedef enum {
     PGP_PARSER_PACKET_END = 0x103,
 
     /* signature subpackets (0x200-2ff) (type+0x200) */
-    /* only those we can parse are listed here */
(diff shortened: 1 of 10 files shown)


dewyatt:
Ah I have to fix something...

dewyatt:
Ok I managed to uncover an unrelated issue that wasn't being caught before, so I added a fix in for that. This should be good now.

dewyatt:
One TODO (I'll create an issue) here that I'm aware of is:
* Support importing just a subkey(s). The current code here only supports the case where you are importing either 1) a primary key or 2) a primary key with subkey(s). DYNARRAY complicates all of this.

PR #471, opened 2017-09-26
Tests for repgp API

flowher:
Adds tests for repgp API


ronaldtse:
@flowher your email box is full... 😉

flowher:
@ronaldtse Thanks! Need to check the server. You can try contact@amongbytes.com

ronaldtse:
@flowher I did, and it's full too 😉

flowher:
hmmmm... seems like bigger problem. krzysiek.kwiatkowski@gmail.com will work then

src/tests/pgp-parse.c
```diff
@@ -224,3 +228,246 @@ pgp_compress_roundtrip(void **state)
...
+    char    input_file[1024] = {0};
+    size_t  out_buf_size = sizeof(out_buf);
+    uint8_t in_buf[1024] = {0};
+    size_t  in_buf_size = sizeof(in_buf);
+    char    plaintext[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890\n";
+
+    /* -------------------------------------------------------------------------
+        Setup keystore
+         This text was encrypted with keys stored in keyrings/1/..
+       -------------------------------------------------------------------------*/
```
dewyatt: Can you use one of the existing commenting styles in the code base, so we can be consistent?

src/lib/rnp.c
```diff
@@ -1590,13 +1590,15 @@ rnp_decrypt_memory(
                           attempts,
                           &ctx->rnp->passphrase_provider);
     if (mem == NULL) {
-        return -1;
+        return RNP_ERROR_BAD_PARAMETERS;
     }
     // TODO: we should actually fail if output buffer is too small
```
dewyatt: Doesn't it make more sense to just fail like this TODO mentions (rather than partially decrypt)?
(Also we're supposed to avoid in/out parameters like outsize is here.)

flowher: Yes, this is actually a bug which I wanted to address in separated PR. But maybe quick and dirty fix here followed by real fix later is also good idea.

dewyatt:
I think there's a few unnecessary casts like:

```c
paths_concat((char *) input_file, sizeof(input_file), rstate->data_dir, "encrypted_text.gpg\0", NULL);
```

But not a big concern.

PR #472, opened 2017-09-27
Add -Wmissing-declarations -Wmissing-prototypes to GCC, fix bugs

randombit:
Mostly failure to `#include` the header with the matching definition, but also a few functions that turned out to be completely unused, and a few cases (in test support.h) where the declaration did not match the definition (just by returning bool vs int, which is why nothing noticed the problem).

Unused funcs: `pgp_elgamal_encrypt_mpi` (are we missing ElGamal encryption?) and `rnp_cfg_check_homedir`

Also, unrelated, type some void* pointers in `rnp_t` and `pgp_io_t`.


src/rnpkeys/tui.c
```diff
@@ -2,6 +2,7 @@
 #include "crypto.h"
 #include "crypto/ec.h"
 #include "rnp/rnpcfg.h"
+#include "rnpkeys.h"
```
flowher: why it is now needed?

src/lib/misc.h
```diff
@@ -32,10 +32,15 @@
...
+
 int pgp_keyid(uint8_t *, const size_t, const pgp_pubkey_t *);
 
 int pgp_fingerprint(pgp_fingerprint_t *, const pgp_pubkey_t *);
 
 void pgp_forget(void *, size_t);
 
-#endif
\ No newline at end of file
+int hash_string(pgp_hash_t *hash, const uint8_t *buf, uint32_t len);
```
flowher: hash.h is probably better place for this

src/rnp/rnpcfg.c
```diff
@@ -346,6 +346,7 @@ rnp_cfg_get_pswdtries(rnp_cfg_t *cfg)
     }
 }
 
+#if 0
```
flowher: I would prefer we remove this if it's not used. we can still get it from github history if needed

src/rnpkeys/tui.c
randombit: Declaration of `rnp_generate_key_expert_mode`

src/lib/misc.h
randombit: Only use of this function is for creating key fingerprints. The PGP fingerprint code is in `misc.c` (where `hash_string` is defined). Only reason it must be declared is it is also used in `key_store_ssh.c` for SSH fingerprints.

Probably best thing is to move `ssh_fingerprint` and `pgp_fingerprint` out of their respective files into some shared `fingerprint.c` and then all of these helper functions can just be static in that one file.

src/rnp/rnpcfg.c
randombit: Yeah probably better, if everyone agrees we don't need them right now / in immediate future.

randombit:
Travis build is failing with

```
stream-common.c:226:1: error: no previous prototype for function 'file_src_read' [-Werror,-Wmissing-prototypes]
file_src_read(pgp_source_t *src, void *buf, size_t len)
^
stream-common.c:238:1: error: no previous prototype for function 'file_src_close' [-Werror,-Wmissing-prototypes]
file_src_close(pgp_source_t *src)
^
stream-common.c:345:1: error: no previous prototype for function 'file_dst_write' [-Werror,-Wmissing-prototypes]
file_dst_write(pgp_dest_t *dst, const void *buf, size_t len)
^
stream-common.c:368:1: error: no previous prototype for function 'file_dst_close' [-Werror,-Wmissing-prototypes]
file_dst_close(pgp_dest_t *dst, bool discard)
^
```

But there is no stream-common.c in my source tree, and `grep -r file_dst .` returns nothing. WTF?

randombit:
Ah I needed to rebase, ok

src/lib/misc.h
flowher: ah, ok I see

dewyatt:
> pgp_elgamal_encrypt_mpi (are we missing ElGamal encryption?) 

It's from netpgp, was replaced with `pgp_elgamal_public_encrypt_pkcs1` at some point it seems.

Definitely +1 on removing unused code, rather than commenting it out.

randombit:
OK stopping with 7f9b976 here, will wait for Travis to catch up and then merge if all looks ok.

Issue #473, opened 2017-09-28
Review for `pgp_forget` calls from #463


ronaldtse:
## Description

https://github.com/riboseinc/rnp/pull/463#discussion_r141321416

@ni4: 
> It seems I forgot to pgp_forget in some other places.

Review #463 to ensure `pgp_forget` is called whenever necessary.

ni4:
Resolved via #474

PR #474, opened 2017-09-28
Review pgp_forget usage in #463, some s2k refactoring

ni4:
This PR:
- adds `pgp_forget` in some places, fixing #473 
- renames pgp_s2k to pgp_s2k_t
- removes some s2k-related code duplication by using new function pgp_s2k_derive_key


randombit:
This looks good to me.

ronaldtse:
@randombit could we have an explicit approval so we can merge this? Thanks 👍

PR #476, opened 2017-09-28
Fix rnp_strcasecmp.

dewyatt:
I encountered a tricky bug while working on FFI and had to track this down.

Previously, strings like "test1" and "test198281939823" would compare equal. I ran some tests on random strings of random lengths for 10 min or so and confirmed this produces correct results (compared to a known-good impl).

### DIFF
--- a/src/lib/misc.c
+++ b/src/lib/misc.c
@@ -913,7 +913,7 @@ rnp_strcasecmp(const char *s1, const char *s2)
 {
     int n;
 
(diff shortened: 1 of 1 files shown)


dewyatt:
> What about adjusting to return +1/-1 if n is non-zero?

Personally I think it may be slightly better to match existing strcasecmp implementations, to avoid any surprises later on.

ni4:
@dewyatt Ah, sorry. Thought strncasecmp returns -1/1, mixed with Objective-C approach.

PR #478, opened 2017-09-30
Fix SM2 test

randombit:
It expected the ciphertext size the match that from the IETF draft, but Botan has changed (in  https://github.com/randombit/botan/pull/1218) to use the ASN.1 formatting instead.


ronaldtse:
Thanks @randombit ! Could you also help update the draft to reflect the ASN.1 adoption?

PR #479, opened 2017-10-02
Key provider interface

ni4:
While moving PK encryption to the streamed approach I got to the point where I need the key provider, similar to the password provider. So decided to add it in separate PR for easier discussion and updates.

Please review it and tell if you have any suggestions regarding the interface. Implementation is just basic and probably will be updated later on.

### DIFF
--- a/src/lib/key-provider.c
+++ b/src/lib/key-provider.c
@@ -29,9 +29,9 @@
 #include <rekey/rnp_key_store.h>
 
 bool
-pgp_key_needed(const pgp_key_provider_t *   provider,
-               const pgp_key_request_ctx_t *ctx,
-               pgp_key_t **                 key)
+pgp_request_key(const pgp_key_provider_t *   provider,
+                const pgp_key_request_ctx_t *ctx,
+                pgp_key_t **                 key)
 {
     if (!provider || !provider->callback || !ctx || !key) {
         return false;
--- a/src/lib/key-provider.c
+++ b/src/lib/key-provider.c
@@ -43,29 +43,29 @@ bool
 rnp_key_provider_keyring(const pgp_key_request_ctx_t *ctx, pgp_key_t **key, void *userdata)
 {
     rnp_t *    rnp = (rnp_t *) userdata;
-    pgp_key_t *pkey, *skey;
+    pgp_key_t *ks_key;
     unsigned   from = 0;
 
     *key = NULL;
     if (ctx->stype == PGP_KEY_SEARCH_KEYID) {
         if (ctx->secret) {
-            skey =
+            ks_key =
               rnp_key_store_get_key_by_id(rnp->io, rnp->secring, ctx->search.id, &from, NULL);
-            if (skey && pgp_key_unlock(skey, ctx->pass_provider)) {
-                *key = skey;
+            if (ks_key && pgp_key_unlock(ks_key, ctx->pass_provider)) {
+                *key = ks_key;
             }
         } else {
-            pkey =
+            ks_key =
               rnp_key_store_get_key_by_id(rnp->io, rnp->pubring, ctx->search.id, &from, NULL);
-            if (pkey) {
-                *key = pkey;
+            if (ks_key) {
+                *key = ks_key;
             } else {
                 /* searching for public key in secret keyring as well */
                 from = 0;
-                skey = rnp_key_store_get_key_by_id(
+                ks_key = rnp_key_store_get_key_by_id(
                   rnp->io, rnp->pubring, ctx->search.id, &from, NULL);
-                if (skey) {
-                    *key = skey;
+                if (ks_key) {
+                    *key = ks_key;
                 }
             }
         }
--- a/src/lib/key-provider.c
+++ b/src/lib/key-provider.c
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2017, [Ribose Inc](https://www.ribose.com).
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1.  Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *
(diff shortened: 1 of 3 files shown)


src/lib/key-provider.c
```diff
@@ -0,0 +1,75 @@
...
+                /* searching for public key in secret keyring as well */
+                from = 0;
+                skey = rnp_key_store_get_key_by_id(
+                  rnp->io, rnp->pubring, ctx->search.id, &from, NULL);
+                if (skey) {
+                    *key = skey;
+                }
+            }
+        }
+    }
```
catap: I think you can use only only one `pgp_key_t *` here. Am I wrong?

ni4: Right, thanks.

dewyatt:
:thumbsup: This is pretty much exactly what I was thinking about adding later on.

I think I have some minor comments I'll add.

src/lib/key-provider.c
```diff
@@ -0,0 +1,75 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+ * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "key-provider.h"
+#include "pgp-key.h"
+#include <rekey/rnp_key_store.h>
+
+bool
+pgp_key_needed(const pgp_key_provider_t *   provider,
```
dewyatt: I think `pgp_request_key` is a slightly more accurate name, and aligns better with `pgp_request_passphrase`. (`pgp_key_needed` sounds like its checking to see if a key is needed, rather than actively retrieving one)

src/lib/key-provider.c
```diff
@@ -0,0 +1,75 @@
...
+    rnp_t *    rnp = (rnp_t *) userdata;
+    pgp_key_t *ks_key;
+    unsigned   from = 0;
+
+    *key = NULL;
+    if (ctx->stype == PGP_KEY_SEARCH_KEYID) {
+        if (ctx->secret) {
+            ks_key =
+              rnp_key_store_get_key_by_id(rnp->io, rnp->secring, ctx->search.id, &from, NULL);
+            if (ks_key && pgp_key_unlock(ks_key, ctx->pass_provider)) {
```
dewyatt: I'm thinking we'll want to ensure this provider is only used for rnp's CLI. It seems perfect for the CLI, I just want to be sure we don't silently unlock keys without API/FFI clients explicitly requesting it.

src/lib/key-provider.c
```diff
@@ -0,0 +1,75 @@
...
+
+bool
+rnp_key_provider_keyring(const pgp_key_request_ctx_t *ctx, pgp_key_t **key, void *userdata)
+{
+    rnp_t *    rnp = (rnp_t *) userdata;
+    pgp_key_t *ks_key;
+    unsigned   from = 0;
+
+    *key = NULL;
+    if (ctx->stype == PGP_KEY_SEARCH_KEYID) {
```
dewyatt: Are we leaving the others for another time?

src/lib/key-provider.c
```diff
@@ -0,0 +1,75 @@
...
+               pgp_key_t **                 key)
+{
+    if (!provider || !provider->callback || !ctx || !key) {
+        return false;
+    }
+    return provider->callback(ctx, key, provider->userdata);
+}
+
+bool
+rnp_key_provider_keyring(const pgp_key_request_ctx_t *ctx, pgp_key_t **key, void *userdata)
```
dewyatt: I wonder if it be better to just call this `rnp_key_provider`, since it takes an `rnp_t` as `userdata`.

src/lib/key-provider.h
```diff
@@ -0,0 +1,80 @@
...
+typedef struct pgp_key_t pgp_key_t;
+
+typedef enum {
+    PGP_KEY_SEARCH_KEYID,
+    PGP_KEY_SEARCH_GRIP,
+    PGP_KEY_SEARCH_USERID
+} pgp_key_search_t;
+
+typedef struct pgp_key_request_ctx_t {
+    pgp_passphrase_provider_t *pass_provider;
```
dewyatt: I think it's better to remove this field and use the passphrase provider within `rnp_t`. This structure should contain the minimal amount of info for retrieving a key.

ronaldtse:
Thanks @ni4 and @catap @dewyatt for the comments! @ni4 once we can get the comments sorted let's merge it!

src/lib/key-provider.c
ni4: Agree

src/lib/key-provider.c
ni4: I added keyring as the description of source. I.e. further providers would be called `rnp_key_provider_database`, `rnp_key_provider_ssh`, `rnp_key_provider_file` (if we use file with the key instead of keyrings), and so on.

src/lib/key-provider.c
ni4: Yeah, I Implement userid search a bit later, together with other code. Wanted to discuss interface as a first step.

src/lib/key-provider.c
ni4: Key will be unlocked only if private key is needed, and actually it will not be possible to unlock it without the passphrase. Probably we can move unlock() call out from here to the calling code, but will this change anything?

src/lib/key-provider.h
ni4: I added rnp_t parameter only because we do not have good separated keyring object. 
Currently streamed methods do not make any use of rnp_t, and the only information they need is password provider and key provider. So imho it is better to do it in this way, later on separating and abstracting keyring object out of the rnp_t. What do you think?

ni4: Hm, got to another point. Callee needs to distinguish whether key is unavailable or password is incorrect. So I think we should remove unlocking from this code at all - it will be done by the callee.

ni4:
@dewyatt Please see the updated PR. I decided to remove pass provider at all, so no automatic unlocking is done. This should address all the previous comments.

PR #481, opened 2017-10-03
Expand FFI json keygen tests, some other misc things

dewyatt:


dewyatt:
> LGTM on first glance but we really should update the README on how to use the JSON key gen interface.

You reminded me to update the doxygen too, so I did that at least. Maybe we'll end up adding an example application to show usage of the FFI. I can definitely add that to ruby-rnp when we get there as well.

ronaldtse:
Just hoping for one more approval to merge this. Thanks guys!

PR #482, opened 2017-10-06
Add ffi symmetric encryption.

dewyatt:
Looking for feedback on this.

Basically the idea is:
* Add an "input" and "output" concept, which can currently be files, memory (not impl), or callbacks (for custom needs).
* Add a symmetric encryption operation concept, with methods to set parameters and execute the operation, etc.

The tests aren't very thorough since the decryption side isn't actually implemented (so I've manually verified instead).

I was thinking we may want to do something similar for signing, encrypting, etc. With signing, for example, we could have an `rnp_op_sign_add_signer(const char *identifier_type, const char *identifier)` (Note that we do technically need to keep track of how a key is located - by userid, keyid, etc, to properly use preferences).

### DIFF
--- a/include/rnp/rnp2.h
+++ b/include/rnp/rnp2.h
@@ -51,6 +51,15 @@ const char *rnp_result_to_string(rnp_result_t result);
 typedef struct rnp_keyring_st *    rnp_keyring_t;
 typedef struct rnp_key_st *        rnp_key_t;
 typedef struct rnp_op_generate_st *rnp_op_generate_t;
+typedef struct rnp_input_st *      rnp_input_t;
+typedef struct rnp_output_st *     rnp_output_t;
+typedef struct rnp_op_symenc_st *  rnp_op_symenc_t;
+
+/* Callbacks */
+typedef ssize_t rnp_input_reader_t(void *app_ctx, void *buf, size_t len);
+typedef void rnp_input_closer_t(void *app_ctx);
+typedef int rnp_output_writer_t(void *app_ctx, const void *buf, size_t len);
+typedef void rnp_output_closer_t(void *app_ctx, bool discard);
 
 /**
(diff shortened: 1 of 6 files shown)


ni4:
@dewyatt Basically, I would not separate symmetric encryption from the public-key encryption since it's the same packet/stream. And actually it is possible to encrypt stream both with password(s) and key(s). Imho would be enough to have one `encrypt` operation and methods `encrypt_add_passphrase()`,  `encrypt_add_recipient()`.

I am currently implementing support for this by adding fields `list recipients` and `unsigned passphrasec` (passphrase count) to the `rnp_ctx_t`, and propagating them to the `init_encrypted_dest`.

The same could be done later on  for the signing stream - `list signers`

dewyatt:
> I would not separate symmetric encryption from the public-key encryption since it's the same packet/stream. And actually it is possible to encrypt stream both with password(s) and key(s). 

Good point, I forgot that these can be combined.

> have one encrypt operation and methods encrypt_add_passphrase(), encrypt_add_recipient().

Yes that does sound like a reasonable approach. I'll make some adjustments to fit that interface.

> I am currently implementing support for this by adding fields list recipients and unsigned passphrasec (passphrase count) to the rnp_ctx_t, and propagating them to the init_encrypted_dest.

Any idea on the timeline for this?

> The same could be done later on for the signing stream - list signers

That's my next target and I had the same plan there. (and just allowing one signer for now, as discussed)

ni4:
@dewyatt I think I'll publish 'forward' PR tomorrow. Code is almost ready, however I will need some time for writing CLI tests and testing.

dewyatt:
Ok I think I'll close this and see if I can rework it into a general encrypt (or finish up some of these other interfaces I've left unimplemented).

PR #485, opened 2017-10-07
Streamed PK-based encryption and decryption

### LINKED ISSUE #424: Implement streamed data processing
## Description

Within current implementation data is processed as a whole and processing results are stored in memory for later processing. So memory usage is a multiply of input size, which is wrong - it must be (and could be) constant except some rare circumstances (when we don't have one-pass signature? don't remember for sure right now).

So we need to switch to streamed processing, so input data will be handle like the following:
- input file is read to memory by chunks (say, 64k blocks)
- it is parsed and according to contents or operation data processors are created
- and this chunk of data passes through all processors up to the output file/memory/stdout

Say, for instance, for file encryption it could be like this:
- file_processor - processor which starts the chain and reads file by chunks in cycle
- data_hash_calculator for MDC
- data_compressor
- data_encryptor
- file_writer

For PGP data processing it would be a bit different since we don't know the packet contents from the beginning, so should insert some 'decision makers' which parse data and create processors on the go.

It could be divided into the following parts:

- [x] Implement streamed processing for symmetric encryption and decryption
- [x] Implement armouring/dearmouring streams
- [x] Move existing PK encryption and decryption to streamed processing
- [x] Implement streamed signing and verification of attached signatures
- [x] Implement streamed signing/verification for detached signatures
- [x] Implement streamed cleartext signing/verification

ni4:
This PR completely moves encrypt and decrypt commands (including password-only encryption) to the streamed approach.
Also it makes rnp ready for multiple-recipient encryption and decryption, just some minor CLI modifications are required.
It doesn't remove old encrypt/decrypt functionality - I think it would be better to remove it when signing/verification and cleartext operations will be updated to use streamed approach as well.

It is the part of the issue #424

### DIFF
--- a/src/librepgp/stream-parse.c
+++ b/src/librepgp/stream-parse.c
@@ -44,6 +44,9 @@
 #include "types.h"
 #include "symmetric.h"
 #include "crypto/s2k.h"
+#include "crypto/sm2.h"
+#include "fingerprint.h"
+#include "pgp-key.h"
 #include "signature.h"
 #ifdef HAVE_ZLIB_H
 #include <zlib.h>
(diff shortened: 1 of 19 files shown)


src/rnp/rnp.c
```diff
@@ -393,24 +381,13 @@ rnp_cmd(rnp_cfg_t *cfg, rnp_t *rnp, int cmd, char *f)
...
-        if (f) {
-            const char *out = rnp_cfg_get(cfg, CFG_OUTFILE);
-            repgp_set_input(io, create_filepath_handle(f));
-            repgp_set_output(io, create_filepath_handle(out));
-        } else {
-            repgp_set_input(io, create_stdin_handle());
-            repgp_set_output(io,
-                             create_buffer_handle((size_t) rnp_cfg_getint(cfg, CFG_MAXALLOC)));
-        }
-        ret = (RNP_SUCCESS == repgp_decrypt(&ctx, io));
```
flowher: I would rather see ``rnp_process_stream`` called by ``repgp_decrypt``.  Also implementation from ``init_stdin_src`` needs to be moved to ``create_stdin_handle`` as it's obviously duplicated now.

src/librepgp/stream-parse.c
```diff
@@ -568,114 +503,183 @@ encrypted_src_close(pgp_source_t *src)
...
+        err = pgp_sm2_decrypt(decbuf,
+                              &declen,
+                              sesskey->params.sm2.m,
+                              sesskey->params.sm2.mlen,
+                              &seckey->key.ecc,
+                              &seckey->pubkey.key.ecc);
+
+        if (err != RNP_SUCCESS) {
+            (void) fprintf(
+              stderr, "%s: SM2 decryption failure, error %x\n", __func__, (int) err);
```
flowher: ``RNP_LOG`` please (or at least same output)

src/librepgp/stream-parse.c
```diff
@@ -568,114 +503,183 @@ encrypted_src_close(pgp_source_t *src)
...
+                                     &declen,
+                                     sesskey->params.ecdh.m,
+                                     sesskey->params.ecdh.mlen,
+                                     ecdh_p,
+                                     &seckey->key.ecc,
+                                     &seckey->pubkey.key.ecdh,
+                                     &fingerprint);
+        BN_free(ecdh_p);
+
+        if ((err != RNP_SUCCESS) || (declen > INT_MAX)) {
```
flowher: check on ``declen`` is not needed

src/librepgp/stream-parse.c
```diff
@@ -568,114 +503,183 @@ encrypted_src_close(pgp_source_t *src)
...
+        return true;
+    } else {
+        return false;
+    }
+}
 
-    switch (skey->s2k.specifier) {
-    case PGP_S2KS_SIMPLE:
+static bool
+encrypted_try_key(pgp_source_t *src, pgp_pk_sesskey_pkt_t *sesskey, pgp_key_t *key)
```
flowher: This function is partially copy paste of ```pgp_create_pk_sesskey()```. Can you reuse the code which already exist?

ni4: Yeah, I borrowed code from old places. The reason I don't reuse old functions is that they use other input/output model and don't follow the 'dump entities' logic, i.e. packet struct is just a parsed OpenPGP entity without any additional data like decrypted keys, flags, and so on. Once transition to the streams is finished all old code and old functions will be removed.

flowher: ok. I see, that's just for transition time

src/librepgp/stream-parse.c
ronaldtse: Agree

src/rnp/rnp.c
ronaldtse: @ni4 what's your plan regarding `librepgp` API in the streamed approach?

src/librepgp/stream-parse.c
ni4: Thanks, fixed.

src/librepgp/stream-parse.c
ni4: Agree, added #490

src/librepgp/stream-parse.c
ni4: Okay, will update the PR

src/librepgp/stream-parse.c
```diff
@@ -1026,22 +1006,54 @@ init_encrypted_src(pgp_processing_ctx_t *ctx, pgp_source_t *src, pgp_source_t *r
...
+        keyctx.stype = PGP_KEY_SEARCH_KEYID;
+
+        for (int i = 0; i < param->pubencc; i++) {
+            memcpy(keyctx.search.id, param->pubencs[i].key_id, sizeof(keyctx.search.id));
+            /* Get the key if any */
+            if (!pgp_request_key(ctx->handler.key_provider, &keyctx, &seckey)) {
+                continue;
+            }
+            /* Unlock key */
+            if (!pgp_key_unlock(seckey, ctx->handler.passphrase_provider)) {
```
dewyatt: Since this isn't really a CLI-layer function, I would rather we didn't use `pgp_key_unlock`.
Normally what I would do is decrypt the seckey into a separate temporary variable and dispose of it at the end, like:
https://github.com/riboseinc/rnp/blob/master/src/lib/generate-key.c#L484

My intention for the use of `pgp_key_unlock` was always mainly FFI/API clients (and CLI would be OK too).
The general idea is:

* call `sign()` (as an example)
  * the key is already unlocked, so it is used as-is, or:
  * the key is locked, so it is temporarily decrypted and used, then disposed of

So in the case of a protected+locked key (should be the most common case), the decrypted MPIs are only exposed for a minimum window of time.

src/librepgp/stream-parse.c
```diff
@@ -1063,8 +1076,11 @@ init_encrypted_src(pgp_processing_ctx_t *ctx, pgp_source_t *src, pgp_source_t *r
                 goto finish;
             }
         } while (1);
-    } else {
-        errcode = RNP_ERROR_NOT_SUPPORTED;
+    }
+
+    if (!have_key) {
+        (void) fprintf(stderr, "%s: failed to obtain decrypting key or password\n", __func__);
```
dewyatt: We should try to return an appropriate error code here. Currently this logs to the console when I use an incorrect passphrase, for example. I believe this function would be used by FFI clients, so ideally we would not do that here, but at the call site.

src/rnp/rnp.c
ni4: @ronaldtse @flowher I think at first I will finish all further stream-based code and then we'll decide how to better glue it with repgp. Actually, it will be just one call for data parsing - rnp_process_stream, outputting decrypted/verified data and giving other information (like signatures) via the callback.

src/librepgp/stream-parse.c
ni4: Thanks. Looks like I mislooked the purpose of unlock or was in hurry. Will fix this.

ni4: @dewyatt `pgp_seckey_free_secret_mpis` doesn't scrub/zero mpi contents, just calls `botan_mp_destroy`. Does it clear the contents?

src/rnp/rnp.c
ronaldtse: 👍

src/librepgp/stream-parse.c
dewyatt: @ni4 Generally yes `botan_mp_destroy` will scrub that memory. See [here](https://botan.randombit.net/manual/side_channels.html#memory-allocation). The MP FFI interface is backed by botan's `BigInt`, which uses `secure_vector`, so that memory is scrubbed when freed.

ni4:
@dewyatt @flowher I addressed all your comments in the last commits. Do you have any further suggestions?

flowher:
Nop, this looks GTM

PR #493, opened 2017-10-11
tests: Copy test data to each individual test home.

dewyatt:
I needed this for some tests I was writing.
Closes https://github.com/riboseinc/rnp/issues/338 but leaves room for improvement as noted here and there.

### DIFF
--- a/src/tests/support.c
+++ b/src/tests/support.c
@@ -180,6 +180,18 @@ delete_recursively(const char *path)
     nftw(path, remove_cb, 64, FTW_DEPTH | FTW_PHYS);
 }
 
+void copy_recursively(const char *src, const char *dst) {
+    /* sanity check, we should only be copying things to /tmp/ */
+    assert_int_equal(strncmp(dst, "/tmp/", 5), 0);
+    assert_true(strlen(dst) > 5);
+
+
+    // TODO: maybe use fts or something less hacky
+    char buf[2048];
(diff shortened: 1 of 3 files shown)


ronaldtse:
@dewyatt the macOS build fails due to:

```
cp: the -R and -r options may not be specified together.
Could not run the test - check test fixtures
[  ERROR   ] ecdh_decryptionNegativeCases
...
```

Could we have a more "platform-independent" way (or add some platform detection) of calling `cp`?

src/tests/support.c
```diff
@@ -180,6 +180,18 @@ delete_recursively(const char *path)
...
 
+void copy_recursively(const char *src, const char *dst) {
+    /* sanity check, we should only be copying things to /tmp/ */
+    assert_int_equal(strncmp(dst, "/tmp/", 5), 0);
+    assert_true(strlen(dst) > 5);
+
+
+    // TODO: maybe use fts or something less hacky
+    char buf[2048];
+    snprintf(buf, sizeof(buf), "/bin/cp -ar '%s' '%s'", src, dst);
```
ronaldtse: Fails on macOS:
```
cp: the -R and -r options may not be specified together.
Could not run the test - check test fixtures
[  ERROR   ] test_ffi_detect_key_format
```

dewyatt: Should be OK now, the `-r` was redundant...

ni4:
@ronaldtse I think this can be merge on your approval.

ronaldtse:
Thanks guys!

PR #494, opened 2017-10-12
Check pgp_build_pubkey return value (and callers).

dewyatt:
I ran into a tricky issue with some local changes, and I saw tests aren't catching it, so this is just trying to improve some error checking in this area.

(`signature.c` still has a lot of room for improvement on error checking, but I'm just doing this one piece since it caused me some trouble.)


ronaldtse:
Codecov seems to be doing something stupid here...

ronaldtse:
Didn’t realize I already approved this. @ni4 @flowher any further comments or we can merge this?

PR #495, opened 2017-10-14
FFI WIP

dewyatt:
This is definitely a work-in-progress, and not intended to be merged.
This is what I came up with after rethinking some things.

(Note that I ripped out some functionality just to speed up the prototyping here and get things building+testing again.)

It's a bit less flexible than before but it's generally intentional.

I was thinking this and something similar to what I was doing in https://github.com/riboseinc/rnp/pull/482/files for enc/sig would be a decent API.

### DIFF
--- a/include/rnp/rnp2.h
+++ b/include/rnp/rnp2.h
@@ -34,49 +34,82 @@ extern "C" {
 #endif
 
 /**
-* Function return type. 0 == SUCCESS, all other values indicate an error.
-*/
+ * Function return type. 0 == SUCCESS, all other values indicate an error.
+ */
 typedef uint32_t rnp_result_t;
 
 #define RNP_EXPORT_FLAG_ARMORED (1U << 0)
 
 /**
-* Return a constant string describing the result code
-*/
+ * Return a constant string describing the result code
+ */
 const char *rnp_result_to_string(rnp_result_t result);
 
 /*
-* Opaque structures
-*/
-typedef struct rnp_keyring_st *rnp_keyring_t;
-typedef struct rnp_key_st *    rnp_key_t;
+ * Opaque structures
+ */
+typedef struct rnp_ffi_st *       rnp_ffi_t;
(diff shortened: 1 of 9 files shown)


dewyatt:
Also I did consider changing the key loading/saving interface to be more like:
```c
rnp_result_t rnp_load_keys_from_path(rnp_ffi_t ffi, const char *path);
rnp_result_t rnp_save_keys_to_path(rnp_ffi_t ffi, const char *pubpath, const char *secpath);
```

And letting the library internally sort them between the pub/sec rings. It's a bit more work though so I abandoned it for now.

ronaldtse:
Any updates @dewyatt or should we merge this first?

dewyatt:
I wouldn't merge it, I'm still actively working on it, I mostly just wanted to put this here for any feedback.

ronaldtse:
@dewyatt any updates on the FFI? Thanks!

dewyatt:
@ronaldtse Last week wasn't as productive as planned, but I'm making good progress so far this week. I'm just making some internal changes that should enable a less awkward FFI in some places right now.

dewyatt:
@ronaldtse Despite delays, I think I've made some good progress on this. I'm aiming to finish decryption in the next 1-2 days and then continue this PR.

After that I think signing+verification remains. I'll probably have to settle for the old non-streaming API on that since I know ni4 is actively working on the streaming one.

ronaldtse:
Thanks for the update @dewyatt . @ni4 perhaps could you let us know the status of streaming API work on signing/verification?

dewyatt:
This still needs some work but probably worth merging as it's kind of intermediary.

Some doc comments need updates and some interfaces are likely to be improved. Things are still in flux a bit.

You can mostly get a good sense of the API by looking at the `test_ffi_encrypt_pass` test. I need to add a test for mixed pk+pass encryption as well.

Still need to add signing/verification as well, and will target the new streaming API when that is ready.

catap:
I agree with @ronaldtse that better to merge it now ;)

PR #505, opened 2017-10-16
Passphrase command line parameter

ni4:
This PR adds --passphrase command line parameter handling, and introduces 'string' password provider. Also it updates performance test with some fixes and improvements.
It was added for easier handling of passphrase in performance and CLI tests, however most likely users will make use of it as well. Despite the fact of quite low security properties of such approach.

### DIFF
--- a/src/tests/cli_perf.py
+++ b/src/tests/cli_perf.py
@@ -7,7 +7,8 @@ import os
 import shutil
 import subprocess
 from timeit import default_timer as perf_timer
-from cli_common import find_utility, run_proc, pswd_pipe, rnp_file_path, random_text, file_text, size_to_readable, raise_err
+from cli_common import find_utility, run_proc, run_proc_fast, pswd_pipe, rnp_file_path, random_text, file_text, size_to_readable, raise_err
+import cli_common
 
 RNP = ''
 RNPK = ''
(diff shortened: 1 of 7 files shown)


ronaldtse:
@ni4 Maybe the problem with the performance discrepancy is that RNP is run first, so it warms the file cache for the later run of GnuPG. We could solve it by first warming the cache manually by reading it and then execute RNP/GnuPG?

src/tests/cli_perf.py
```diff
@@ -54,7 +55,7 @@ def setup():
     ret, out, err = run_proc(GPG, ['--batch', '--passphrase', '', '--homedir', GPGDIR, '--import', path.join(RNPDIR, 'pubring.gpg'), path.join(RNPDIR, 'secring.gpg')])
 
     # Generating small file for tests
-    SMALLSIZE = 3312;
+    SMALLSIZE = 33120;
```
flowher: is it intended?

ni4: Nope, changed during some testing. Thanks for noticing.

ni4:
@ronaldtse I tried to change their order and that didn't make any impact on comparison. Even if cache has some impact then it may happen only on first iteration, while on small files tests run 100 times.

ni4:
Got 2 approvals, so merging.

ronaldtse:
Got it, thanks @ni4 !

Issue #419, opened 2017-08-22
Create CLI test case for CFB resync/loading of v3 keys


ni4:
## Description

It's not a good tone to write description like this but see  the #414 and #403, #401  for the info.

ronaldtse:
A perfect description 😉

ni4:
For a while we have cmocka tests (like `test_stream_key_decrypt`, `test_stream_key_encrypt`) for new key encryption/decryption, so those test cfb resync as well.
So I think this issue may be closed safely.

PR #509, opened 2017-10-17
CFB resync clarification and tests update

### LINKED ISSUE #440: Optimize CFB mode usage
## Description

Current CFB mode implementation processes input data by byte, which is quite slow.
It should be optimized to work at least on per-word level.
Other optimization would be to process the whole chunk on botan level so botan_encrypt_block call overhead is not applied for each 8-16 bytes.

ni4:
Within this PR:
- CFB resync clarified to single method, removed unneeded overhead from symmetric crypto
- updated cmocka test with decryption of v3 keys and corresponding keyring from PGP 2.6.3
- updated CLI tests with non-mdc (tag 9) encryption by GPG and different s2k modes usage

Closes #440, #419


src/lib/symmetric.c
```diff
@@ -255,17 +239,17 @@ pgp_cipher_cfb_encrypt(pgp_crypt_t *crypt, uint8_t *out, const uint8_t *in, size
         memcpy(crypt->iv, iv64, blsize);
     }
 
-    /* filling prev_iv */
-    memcpy(crypt->prev_iv, crypt->iv, blsize);
-    botan_block_cipher_encrypt_blocks(crypt->obj, crypt->iv, crypt->iv, 1);
-    crypt->remaining = blsize;
+    if (bytes) {
```
ronaldtse: Shortcut this to `return 0` if there are no `bytes`?

src/lib/symmetric.c
```diff
@@ -338,18 +322,18 @@ pgp_cipher_cfb_decrypt(pgp_crypt_t *crypt, uint8_t *out, const uint8_t *in, size
...
-    crypt->remaining = blsize;
-
-    /* decrypting tail */
-    while (bytes) {
-        uint8_t c = *in++;
-        *out++ = c ^ crypt->iv[blsize - crypt->remaining];
-        crypt->iv[blsize - crypt->remaining] = c;
-        crypt->remaining--;
-        bytes--;
+    if (bytes) {
```
ronaldtse: Shortcut this?

ni4:
@ronaldtse Done

ronaldtse:
@dewyatt @flowher are we ready to merge this? Thanks!

dewyatt:
I probably can't spare the time to validate *all* the logic changes in-depth at the moment, but overall it looks like what I would expect from [Jack's comment](https://github.com/riboseinc/rnp/issues/440#issuecomment-326947690) and it has tests (thank you!), so I would say merge.

PR #511, opened 2017-10-18
(rnp) Adds option symmetric to --dearmor

flowher:
Option that makes it possible to convert from binary format to ASCII Armor. This functionality was mainly implemented in order to test performance of pure base64 implementation, but I found it may be useful for anybody who wants to convert from one format to another (for whatever reason).

Usage:
```
rnp --enarmor=msg input > middle
rnp --dearmor middle > output 

> diff input output
>
```

### DIFF
--- a/src/librepgp/stream-armour.c
+++ b/src/librepgp/stream-armour.c
@@ -493,24 +493,35 @@ find_armour_header(const char *buf, size_t len, size_t *hdrlen)
     return NULL;
 }
 
-static pgp_armoured_msg_t
-armour_message_type(const char *hdr, size_t len)
+pgp_armoured_msg_t
+armour_str_to_data_type(const char *str, size_t len)
 {
-    if (!strncmp(hdr, "BEGIN PGP MESSAGE", len)) {
-        return PGP_ARMOURED_MESSAGE;
-    } else if (!strncmp(hdr, "BEGIN PGP PUBLIC KEY BLOCK", len) ||
-               !strncmp(hdr, "BEGIN PGP PUBLIC KEY", len)) {
-        return PGP_ARMOURED_PUBLIC_KEY;
-    } else if (!strncmp(hdr, "BEGIN PGP SECRET KEY BLOCK", len) ||
-               !strncmp(hdr, "BEGIN PGP SECRET KEY", len)) {
-        return PGP_ARMOURED_SECRET_KEY;
-    } else if (!strncmp(hdr, "BEGIN PGP SIGNATURE", len)) {
-        return PGP_ARMOURED_SIGNATURE;
-    } else if (!strncmp(hdr, "BEGIN PGP SIGNED MESSAGE", len)) {
-        return PGP_ARMOURED_CLEARTEXT;
-    } else {
+    if (!str) {
         return PGP_ARMOURED_UNKNOWN;
     }
+
+    if (!strncmp(str, "BEGIN PGP MESSAGE", len) || !strncmp("msg", str, len)) {
+        return PGP_ARMOURED_MESSAGE;
+    }
+
+    if (!strncmp(str, "BEGIN PGP PUBLIC KEY BLOCK", len) ||
+        !strncmp(str, "BEGIN PGP PUBLIC KEY", len) || !strncmp("pubkey", str, len)) {
+        return PGP_ARMOURED_PUBLIC_KEY;
+    }
+
+    if (!strncmp(str, "BEGIN PGP SECRET KEY BLOCK", len) ||
+        !strncmp(str, "BEGIN PGP SECRET KEY", len) || !strncmp("seckey", str, len)) {
+        return PGP_ARMOURED_SECRET_KEY;
+    }
+
+    if (!strncmp(str, "BEGIN PGP SIGNATURE", len) || !strncmp("sign", str, len)) {
+        return PGP_ARMOURED_SIGNATURE;
+    }
+
+    if (!strncmp(str, "BEGIN PGP SIGNED MESSAGE", len)) {
+        return PGP_ARMOURED_CLEARTEXT;
+    }
+    return PGP_ARMOURED_UNKNOWN;
 }
 
 static bool
@@ -536,7 +547,7 @@ armour_parse_header(pgp_source_t *src)
         RNP_LOG("extra data before the header line");
     }
 
-    param->type = armour_message_type(armhdr + 5, armhdrlen - 10);
+    param->type = armour_str_to_data_type(armhdr + 5, armhdrlen - 10);
     if (param->type == PGP_ARMOURED_UNKNOWN) {
         RNP_LOG("unknown armour header");
         return false;
@@ -904,7 +915,7 @@ init_armoured_dst(pgp_dest_t *dst, pgp_dest_t *writedst, pgp_armoured_msg_t msgt
     param->llen = 76; /* must be multiple of 4 */
 
     if (!armour_message_header(param->type, false, hdr)) {
-        RNP_LOG("unknown message type");
+        RNP_LOG("unknown data type");
         ret = RNP_ERROR_BAD_PARAMETERS;
         goto finish;
     }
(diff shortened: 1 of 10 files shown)


src/librepgp/stream-armour.c
```diff
@@ -493,24 +493,27 @@ find_armour_header(const char *buf, size_t len, size_t *hdrlen)
...
-    if (!strncmp(hdr, "BEGIN PGP MESSAGE", len)) {
+    if (!str) {
+        return PGP_ARMOURED_UNKNOWN;
+    }
+
+    if (!strncmp(str, "BEGIN PGP MESSAGE", len) || !strncmp("msg", str, len)) {
         return PGP_ARMOURED_MESSAGE;
-    } else if (!strncmp(hdr, "BEGIN PGP PUBLIC KEY BLOCK", len) ||
-               !strncmp(hdr, "BEGIN PGP PUBLIC KEY", len)) {
+    } else if (!strncmp(str, "BEGIN PGP PUBLIC KEY BLOCK", len) ||
```
ronaldtse: We can shortcut all of these `else if` Into just `if` since they are all returned.

ronaldtse:
Build failure due to armoring code change. 

```
  File "cli_tests.py", line 891, in <module>
    run_tests()
  File "cli_tests.py", line 865, in run_tests
    run_rnp_tests()
  File "cli_tests.py", line 807, in run_rnp_tests
    rnp_armour()
  File "cli_tests.py", line 791, in rnp_armour
    compare_files(dst_beg, dst_fin, "RNP armour/dearmour test failed")
  File "cli_tests.py", line 129, in compare_files
    if file_text(src) != file_text(dst):
  File "/home/travis/build/riboseinc/rnp/src/tests/cli_common.py", line 59, in file_text
    with open(path, 'r') as f:
IOError: [Errno 2] No such file or directory: '/tmp/rnpctmp34Y0K3/beg.dst
```

src/rnp/rnp.c
```diff
@@ -48,23 +48,27 @@
...
-                           "\t--decrypt [--output=file] [options] files... OR\n\n"
+                           "\t--decrypt [--output=file] [options] files... OR\n"
                            "\t--sign [--detach] [--hash=alg] [--output=file]\n"
                            "\t\t[options] files... OR\n"
                            "\t--verify [options] files... OR\n"
                            "\t--cat [--output=file] [options] files... OR\n"
                            "\t--clearsign [--output=file] [options] files... OR\n"
                            "\t--list-packets [options] OR\n"
                            "\t--dearmor [--output=file] file OR\n"
+                           "\t--enarmor=<msg|pubkey|seckey|sign|text> \n"
```
ni4: <msg|pubkey|seckey|sign|text> looks different from other commands description.
Also, 'text' means cleartext then it is not armoring at all.
Also --enarmor is a command, not the option - it would be more logical to make command '--enarmor' with additional option '--header', which could be 'PUBLIC KEY', 'SECRET KEY', or anything else. By default it should be 'MESSAGE'.

src/rnp/rnpcfg.h
```diff
@@ -32,10 +32,13 @@
 /* cfg variables known by rnp */
 #define CFG_OVERWRITE "overwrite" /* overwrite output file if it is already exist or fail */
 #define CFG_ARMOUR "armour"       /* armour output data or not */
-#define CFG_DETACHED "detached"   /* produce the detached signature */
-#define CFG_OUTFILE "outfile"     /* name/path of the output file */
-#define CFG_RESULTS "results"     /* name/path for results, not used right now */
-#define CFG_MAXALLOC "maxalloc"   /* maximum memory allocation during the reading from stdin */
+#define CFG_ARMOUR_DATA_TYPE                                                      \
+    "armour_type"               /* armour data type, used with ``enrmour`` option \
```
ni4: enarmor, not enrmour?

src/tests/cli_tests.py
```diff
@@ -777,6 +777,21 @@ def rnp_misc_operations():
     run_test(rnp_encryption_no_mdc)
     run_test(rnp_encryption_s2k)
 
+def rnp_armour():
+    src_beg,dst_beg = reg_workfiles('beg','.src','.dst')
```
ni4: Maybe add some more test cases with different message sizes and headers?
Actually, armouring itself will be tested within encryption/decryption and signing/verification. But this could test for validity of headers and so on.

ni4:
@flowher Main change I would suggest is regarding the separating command from headers parameter. For most usages default 'PGP MESSAGE' would be enough.

@ronaldtse Should we stick to commont 'armor' or British 'armour' :) ? I think we should change all occurrences to single word version so code looks consistent.

src/rnp/rnp.c
flowher: Yes, other commands do not take optional argument. Here the data type is not an option - it is required. 
I actually had implementation with '--header', but it doesn't seem to be good idea. User needs to provide data type, it's not an option. Also I don't think default is good idea - we can't guess what kind of data user tries to convert in this case. Instead, we need to force user to decide.

src/rnp/rnpcfg.h
flowher: same thing apparently

src/tests/cli_tests.py
flowher: yeah, I'll work on tests. for some reason they are broken on CI but working localy

src/rnp/rnp.c
ni4: Forcing user to do something doesn't seem to be a good idea. We can just peek first few bytes, check the packet type, and make correct decision about the data contents. And leave '--header' option for the user in case he wants more precise control over the process. Anyway, this command is not critical for RNP functionality.

src/rnp/rnpcfg.h
ni4: I meant missed letter 'a' in the word :)

ronaldtse:
Let’s go with “armor” since it’s very hard for Americans to spell “armour” :wink:

src/rnp/rnpcfg.h
flowher: argh... I see :)

src/rnp/rnp.c
flowher: Ah, you mean rnp should find out by itself type of the data or let the user decide. In such, yes, it would be good improvement to this. It's just not implemented now, as well, it indeed is not critical functionality (so no reason to complicate a code).

dewyatt: Just my opinion but I feel that separating the options into `--enarmor` and `--header` would be rather confusing to the user. They are intrinsically linked and that is crystal clear when they are coupled in `--enarmor=msg`.

I haven't looked at the difficulty of guessing the format, but personally the way I would probably do this is I would have `--enarmor` try to guess (or force a default even), but allow overrides like `--enarmor=msg` to force a type as well.

src/tests/cli_tests.py
dewyatt: It fails in CI because rnp considers it a fatal error when the pubring can't be read [here](https://github.com/riboseinc/rnp/blob/master/src/rnp/rnp.c#L744). We don't need keys to enarmor so I would say that's a bug.

(Also since your `run_proc` invocations don't include `'--homedir', RNPDIR`, this will mkdir ~/.rnp which isn't ideal for a test.)

src/rnp/rnp.c
ni4: @flowher From rfc4880:
```
   BEGIN PGP MESSAGE
       Used for signed, encrypted, or compressed files.

   BEGIN PGP PUBLIC KEY BLOCK
       Used for armoring public keys.

   BEGIN PGP PRIVATE KEY BLOCK
       Used for armoring private keys.

   BEGIN PGP SIGNATURE
       Used for detached signatures, OpenPGP/MIME signatures, and
       cleartext signatures.  Note that PGP 2.x uses BEGIN PGP MESSAGE
       for detached signatures.
```

So basically you should just check first byte of the stream, use get_packet_type() from `stream-parse.c`, and for tag 2 use 'SIGNATURE', tag 5 - 'SECRET KEY', tag 6 - 'PUBLIC KEY', and 'MESSAGE' for all other tags. If you don't have time for this now - I may do it later on, just add an issue.

Also probably later we may wish to add `--base64` or `--raw` modifier, like gnupg does, to just encode/decode to raw base64, without headers and crc.

flowher: ok, thanks. I would say, let's keep it stupid simple until we really needed.

src/tests/cli_tests.py
flowher: Aha! Thnx. Indeed that's a bug - we don't need to read keys in this case (or call keystore for anything else).

src/rnp/rnp.c
flowher: Just FYI: GnuPG wraps in "-----BEGIN PGP ARMORED FILE----" when called with "--enarmor" (which kind of sucks)

ronaldtse:
@ni4 @dewyatt good to go?

PR #517, opened 2017-10-19
Don't require from user to provide ``--hash`` for signing with secp384r1 and secp521r1

flowher:
It seems from some time, we require user to provide a hash algo to be used when signing with ECDSA. For some cases it may be good idea, but for majority of cases this process should be automatic.

With this fix ``rnp`` will chose hash which can be used with given curve securely.


ronaldtse:
Agree with @dewyatt , we should also have a better interface to deal with "recipient preferred" vs "sender preferred".

PR #519, opened 2017-10-23
Add --overwrite option to rnp

ni4:
Before now rnp was overwriting output file by default, and without any option to change this behavior.
So this PR:
- adds `--overwrite` option, which will allow to overwrite output by default
- fixes weird double fd opening during file signing (should be typo)
- updates handling of output stream creation with asking user whether to overwrite file/new filename


flowher:
"Before now rnp was overwriting output file by default", is it really true? Or do you mean that it wasn't overwriting?

src/rnp/rnpcfg.c
```diff
@@ -306,17 +306,12 @@ bool
...
-        if ((strcmp(val, "true") == 0) || (strcmp(val, "True") == 0)) {
-            return true;
-        } else if (atoi(val) > 0) {
-            return true;
-        } else
-            return false;
-    } else {
-        return false;
+    
+    if (val && ((strcmp(val, "true") == 0) || (strcmp(val, "True") == 0) || (atoi(val) > 0))) {
```
flowher: better to use ``tolower()`` and then compare to "true"

flowher: easier to read if "return (val && ((strcmp(val, "true") == 0) || (strcmp(val, "True") == 0) || (atoi(val) > 0))) "

src/lib/misc.c
```diff
@@ -1134,3 +1134,15 @@ rnp_strlwr(char *s)
...
+rnp_strip_eol(char *s)
+{
+    size_t len = strlen(s);
+
+    while (len > 0 && (s[len - 1] == '\n' || s[len - 1] == '\r')) {
+        s[--len] = '\0';
+    }
+
+    return s;
+}
```
flowher: \n

src/lib/rnp.c
```diff
@@ -1200,11 +1200,116 @@ rnp_decrypt_file(rnp_ctx_t *ctx, const char *f, const char *out)
...
+ *  @param dst Allocated dest structure to put result in
+ *  @param out Output filename/path as provided by user. For NULL or '-' stdout will be used.
+ *  @return true on success or false otherwise
+ **/
+
+static bool
+rnp_initialize_output(rnp_ctx_t *ctx, pgp_dest_t *dst, const char *out)
+{
+    char newname[PATH_MAX];
+    bool is_stdout = out == NULL || out[0] == '\0' || strcmp(out, "-") == 0;
```
flowher: can you put () around those conditions. It hard to read this way

flowher:
LGTM, some minor comments added

src/rnp/rnpcfg.c
ronaldtse: Agree.

ronaldtse: Agree.

ni4: Used strcasecmp - otherwise would need to check val, allocate buffer for lowercase copy, deallocate, etc, etc.

ni4: Done.

src/lib/misc.c
ni4: Ok

src/lib/rnp.c
ni4: Ok.

ni4:
@flowher @ronaldtse Thanks for comments, fixed.

> "Before now rnp was overwriting output file by default", is it really true? Or do you mean that it wasn't overwriting?

Yeah, it was. overwriting without the prompt. It has hardcoded setting overwrite to 1, so I moved the same to rnpcfg that time.

ni4:
@flowher Any other comments, or this may be merged?

src/lib/misc.c
```diff
@@ -1134,3 +1134,15 @@ rnp_strlwr(char *s)
     }
     return s;
 }
+
+char *
+rnp_strip_eol(char *s)
+{
+    size_t len = strlen(s);
+
+    while (len > 0 && (s[len - 1] == '\n' || s[len - 1] == '\r')) {
```
flowher: () here also please

include/rnp/rnp_sdk.h
```diff
@@ -71,4 +71,6 @@ bool rnp_hex_decode(const char *hex, uint8_t *buf, size_t buf_len);
 
 char *rnp_strlwr(char *s);
 
+char *rnp_strip_eol(char *s);
```
flowher: Please add comment what this function really does

flowher:
Two additional comments. Go ahead and merge after they're done

src/lib/misc.c
ni4: Done

include/rnp/rnp_sdk.h
ni4: Done

Issue #456, opened 2017-09-13
Consolidate output filename handling and file overwrite


ni4:
## Description

Currently during decryption/encryption/signing/verification we by default overwrite output file if it already exists, and use input file name without .gpg/.asc extension for output file name if it is not specified. Also we do not handle file name stored in literal data packet.
We should fix it to use the same behavior everywhere.

It is better to implement this after #424 - streamed data processing.
It will add the idea of output destination handler.

## Draft of output filename detection logic

#Decryption/verification:
1. If exact file name is specified via --output file parameter then we should use it.
2. If filename parameter is present in literal data packet then use it (however, we should take care with path handling - not overwrite /etc/passwd and so on).
3. If filename is present in literal data packet and equals to _CONSOLE then we should add prompt like 'This message is marked 'for your eyes only', should we display it right now? or so on)
4. Otherwise write to stdout (or if output is equal to '-').

If we are writing to the file then we should check whether it exists already, and if it is true, and there is no --overwrite parameter, then we should ask 'File filename.dat already exists. Should we overwrite it? y/N', and if 'N' - then ask for the new file name.

#Encryption/signing
1. If exact file name is specified via --output parameter then we should use it (or stdout if output is -)
2. If we are producing armored output then we should add/change to .asc extension
3. If we are producing binary output - then we should add/change to .pgp extension (.gpg? .rnp?)
4. Cleartext signing - which extension should we use?

If output file exists already then we should ask the user.

## Related issues

#220 Implement proper handling of 'for your eyes only' flag
#221 rnp: ask whether to overwrite output file if it is exist already

ni4:
@ronaldtse @dewyatt @flowher What do you think? I implemented such logic in PR #571.
In short:
- when we transform file from raw one, then without `--output` we add .asc/.sig/.pgp extension (file.dat -> file.dat.pgp)
- when we process pgp file, then write to stdout if no `--output` option is given

ronaldtse:
@ni4 I agree with the approach. This saves a lot of confusion!

flowher:
@ni4 I also agree

PR #520, opened 2017-10-23
Enarmor/dearmor commands improvements

ni4:
This PR improves some things within the enarmor/dearmor commands:
- adds automatic guessing of the message type
- refactored to better match the architecture:
  - separated to use two functions, armor/dearmor (currently it looks like a lot of code duplication, but with PR #519 it will look much better)
  - moved msgtype parameter to rnp_ctx_t
  - moved msgtype parameter parsing to the CLI level

### DIFF
--- a/src/lib/rnp.c
+++ b/src/lib/rnp.c
@@ -1428,7 +1428,7 @@ rnp_armor_stream(rnp_ctx_t *ctx, bool armor, const char *in, const char *out)
     pgp_dest_t        dst;
     rnp_result_t      result;
     pgp_armored_msg_t msgtype;
-    
+
     if (!rnp_initialize_input(ctx, &src, in)) {
         RNP_LOG("failed to initialize reading");
         return RNP_ERROR_READ;
@@ -1445,7 +1445,7 @@ rnp_armor_stream(rnp_ctx_t *ctx, bool armor, const char *in, const char *out)
         if (msgtype == PGP_ARMORED_UNKNOWN) {
             msgtype = rnp_armor_guess_type(&src);
         }
-    
+
         result = rnp_armor_source(&src, &dst, msgtype);
     } else {
         result = rnp_dearmor_source(&src, &dst);
--- a/src/lib/rnp.c
+++ b/src/lib/rnp.c
@@ -1422,8 +1422,7 @@ finish:
 }
 
 rnp_result_t
-rnp_armor_stream(
-  rnp_ctx_t *ctx, const char *in, const char *out, bool is_armor, unsigned data_type)
+rnp_armor_stream(rnp_ctx_t *ctx, const char *in, const char *out)
 {
     pgp_source_t src;
     pgp_dest_t   dst;
@@ -1440,12 +1439,43 @@ rnp_armor_stream(
         return RNP_ERROR_WRITE;
     }
 
-    result = is_armor ? rnp_armor_source(&src, &dst, (pgp_armored_msg_t) data_type) :
-                        rnp_dearmor_source(&src, &dst);
+    if ((msgt = (pgp_armored_msg_t) ctx->armortype) == PGP_ARMORED_UNKNOWN) {
+        msgt = rnp_armor_guess_type(&src);
+    }
+
+    result = rnp_armor_source(&src, &dst, msgt);
 
     if (result != RNP_SUCCESS) {
         RNP_LOG("error code 0x%x", result);
-        return result;
+    }
+
+    src_close(&src);
+    dst_close(&dst, result != RNP_SUCCESS);
+    return result;
+}
+
+rnp_result_t
+rnp_dearmor_stream(rnp_ctx_t *ctx, const char *in, const char *out)
+{
+    pgp_source_t src;
+    pgp_dest_t   dst;
+    rnp_result_t result;
+
+    if (!rnp_initialize_input(ctx, &src, in)) {
+        RNP_LOG("failed to initialize reading");
+        return RNP_ERROR_READ;
+    }
+
+    if (!rnp_initialize_output(ctx, &dst, out)) {
+        RNP_LOG("failed to initialize writing");
+        src_close(&src);
+        return RNP_ERROR_WRITE;
+    }
+
+    result = rnp_dearmor_source(&src, &dst);
+
+    if (result != RNP_SUCCESS) {
+        RNP_LOG("error code 0x%x", result);
     }
 
     src_close(&src);
--- a/src/lib/rnp.c
+++ b/src/lib/rnp.c
@@ -1424,9 +1424,10 @@ finish:
 rnp_result_t
 rnp_armor_stream(rnp_ctx_t *ctx, const char *in, const char *out)
 {
-    pgp_source_t src;
-    pgp_dest_t   dst;
-    rnp_result_t result;
+    pgp_source_t      src;
+    pgp_dest_t        dst;
+    rnp_result_t      result;
+    pgp_armored_msg_t msgt;
 
     if (!rnp_initialize_input(ctx, &src, in)) {
         RNP_LOG("failed to initialize reading");
--- a/src/lib/rnp.c
+++ b/src/lib/rnp.c
@@ -1422,13 +1422,13 @@ finish:
 }
 
 rnp_result_t
-rnp_armor_stream(rnp_ctx_t *ctx, const char *in, const char *out)
+rnp_armor_stream(rnp_ctx_t *ctx, bool armor, const char *in, const char *out)
 {
     pgp_source_t      src;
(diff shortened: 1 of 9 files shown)


src/rnp/rnp.c
```diff
@@ -496,12 +493,31 @@ setoption(rnp_cfg_t *cfg, int *cmd, int val, char *arg)
...
         *cmd = val;
-        rnp_cfg_setint(
-          cfg, CFG_ARMOR_DATA_TYPE, armor_str_to_data_type(arg, arg ? strlen(arg) : 0));
+        break;
+    case CMD_ENARMOR: {
+        pgp_armored_msg_t msgt = PGP_ARMORED_UNKNOWN;
+        *cmd = val;
+
+        if (arg) {
+            if (!strncmp("msg", arg, strlen(arg))) {
```
flowher: Code duplication, can you reuse ``armor_str_to_data_type``?

src/librepgp/stream-armor.c
```diff
@@ -500,21 +501,21 @@ armor_str_to_data_type(const char *str, size_t len)
         return PGP_ARMORED_UNKNOWN;
     }
 
-    if (!strncmp(str, "BEGIN PGP MESSAGE", len) || !strncmp("msg", str, len)) {
```
flowher: I actually prefer this, so that we don't duplicate the code

src/lib/rnp.c
```diff
@@ -1371,8 +1371,48 @@ rnp_armor_stream(
...
+    }
+
+    src_close(&src);
+    dst_close(&dst, result != RNP_SUCCESS);
+    return result;
+}
+
+rnp_result_t
+rnp_dearmor_stream(rnp_ctx_t *ctx, const char *in, const char *out)
+{
```
flowher: Why do we need to duplicate, this?

src/librepgp/stream-armor.c
flowher: function should be static now

src/rnp/rnp.c
flowher: I don't think there is a reason to duplicate this code (nevertheless, I think I now understand your point).
So,
```
struct {
 uint8_t ID,
 char *msg_data_type,
 char *cli_data_type,
} name[] = {
  { PGP_ARMORED_PUBLIC_KEY, "pubkey", "BEGIN PGP PUBLIC KEY" },
  .....
};
```
and then ```armor_str_to_data_type``` gets additional argument and does ```for``` inside to find correct id. That's how we do do in some other places

src/lib/rnp.c
```diff
@@ -1371,8 +1371,48 @@ rnp_armor_stream(
...
+    pgp_source_t src;
+    pgp_dest_t   dst;
+    rnp_result_t result;
+    bool         is_stdin;
+    bool         is_stdout;
+
+    is_stdin = !in || (strlen(in) == 0) || (strcmp(in, "-") == 0);
```
flowher: that's simply code  duplication. let's see how it looks after rebase when the other PR is integrated

flowher:
Looks OK,  but I would prefer take a second look after the ``--overwrite`` stuff is integrated, ok?

ronaldtse:
Right now it fails here:
```
RUNNING : rnp_armor()


[pgp_mem_readfile() misc.c:603] can't open "/home/travis/.rnp/pubring.gpg"
cannot read pub keyring
fatal: failed to load keys


[init_file_src() stream-common.c:261] can't stat '/tmp/rnpctmpAcY7AX/mid.dst'
[rnp_armor_stream() rnp.c:1363] failed to initialize reading


  File "cli_tests.py", line 890, in <module>
    run_tests()
  File "cli_tests.py", line 864, in run_tests
    run_rnp_tests()
  File "cli_tests.py", line 806, in run_rnp_tests
    rnp_misc_operations()
  File "cli_tests.py", line 794, in rnp_misc_operations
    run_test(rnp_armor)
  File "cli_tests.py", line 167, in run_test
    func(*args)
  File "cli_tests.py", line 782, in rnp_armor
    compare_files(dst_beg, dst_fin, "RNP armor/dearmor test failed")
  File "cli_tests.py", line 129, in compare_files
    if file_text(src) != file_text(dst):
  File "/home/travis/build/riboseinc/rnp/src/tests/cli_common.py", line 59, in file_text
    with open(path, 'r') as f:
IOError: [Errno 2] No such file or directory: '/tmp/rnpctmpAcY7AX/fin.dst'
```

src/rnp/rnp.c
ni4: I don't like this because 'BEGIN PGP PUBLIC KEY' is the constant from data layer which implements RFC, which does not depend on user in any way. While 'pubkey' is a CLI parameter which should be present on the CLI layer only. And mixing those layers in code will bring us back to GnuPG :)

src/lib/rnp.c
ni4: After merging other PR the duplication will be only in `rnp_initialize_input`/`rnp_initialize_input` calls. Actually this will be duplicated in encrypt/sign/decrypt calls as well. Probably I'll wrap it to something better later on, within implementation of #456

ni4:
@flowher @ronaldtse Thanks for the review. After merging #519 I'll rebase this and introduce fixes.

src/lib/rnp.c
ni4: ...also the main reason for making two functions is the fact that logically armoring and dearmoring are different operations. First one generates data, second - parses. First one may have additional parameters (for instance, we may wish later on to add other line length, line ending, make it raw base64 without the headers, and so on). It's like mixing together encryption and decryption and writing inside `if (decrypt) {...do decryption...} else {...do encryption...}`

src/rnp/rnp.c
flowher: Nop, that's not a problem, because you can add an argument to ``armor_str_to_data_type`` which will indicate what too look for.

flowher:
Thnx

src/lib/rnp.c
flowher: I don't think that's good enough reason. In fact mixing decryption/encryption is good example - interfaces often use same method to encrypt and decrypt (CipherInit(), CipherUpdate(), CiperFinalize()).
I'm not saying two methods are not good, I'm saying code duplication must be avoided.

src/rnp/rnp.c
ni4: Why that's not a problem? This is a completely wrong approach when entities from different parts of the library are mixed together just to save few lines of code. And, even now, command line argument does not directly fit to `armor_str_to_data_type` since it must not allow cleartext message.

@dewyatt , @ronaldtse what do you think? This is actually minor issue but we should follow some architectural logic and separate entities based on their usage.

src/lib/rnp.c
ni4: While we are talking about raw symmetric encryption/decryption - then yeah, this may be done in single function call (however, it is not like this - many libraries have different calls for encrypt and decrypto). However raw encryption/decryption does not parse input data.

For PGP decryption/encryption (and, armoring/dearmoring) situation is opposite - a lot of efforts put into data parsing, so it's quite logical to separate those calls. Anyway, it has just few lines of code duplication.

src/rnp/rnp.c
flowher: Ok, that's minor issue indeed. It seems it's faster if I'll fix it later, then explain.

src/lib/rnp.c
```diff
@@ -1453,12 +1486,10 @@ rnp_armor_stream(
         return RNP_ERROR_WRITE;
     }
 
-    result = is_armor ? rnp_armor_source(&src, &dst, (pgp_armored_msg_t) data_type) :
```
flowher: Ok, so this one line was better, because it was one line. We don't need whole new function.

src/rnp/rnp.c
ronaldtse: @ni4 I agree we should isolate the architecture first and then attempt to DRY the code. Maybe there are ways we can share those functions without polluting the architecture.

ronaldtse:
Could we have this branch merged soon since it's open for a while? Thanks!

ni4:
@ronaldtse Ok, I'll get back to it later today or tomorrow, once finished with some other stuff.
@flowher So, do I correctly understand that the main showstopper for you is having two functions in lib/rnp.c - enarmor and dearmor, and I should merge them into one?

flowher:
as in comment. you can merge it if you want. I'll do fix later

ni4:
I made some fixes and rebased, so now it should be okay to merge, once having the second approval.
Any objections?

flowher:
Thank you!

PR #522, opened 2017-10-24
(core): CRC24-Radix64 performance improvement (32 bits in parallel)

### LINKED ISSUE #521: CRC - process bytes in parallel
## Description

Currently CRC is processing 8 bits in one shot. This can be improved to process at least 32 bytes in one shot.
Initial PoC shows that processing 32 bytes in parallel, gives 44% improvement of writing/reading armored data (as CRC calc is hot spot in armor data processing)

Current:
```
rnp: default key set to "0BB1950A9E83E139"
/home/flowher/repos/rnp/src/rnp/.libs/rnp --enarmor=msg /tmp/1gb.rnd --output  4.48s user 0.89s system 98% cpu 5.429 total
rnp: default key set to "0BB1950A9E83E139"
/home/flowher/repos/rnp/src/rnp/.libs/rnp --dearmor out.xxx --output out.d  5.58s user 0.65s system 98% cpu 6.338 total
```

32-bytes in parallel:
```
/home/flowher/repos/rnp/src/rnp/.libs/rnp --enarmor=msg /tmp/1gb.rnd --output  2.58s user 0.84s system 90% cpu 3.755 total
/home/flowher/repos/rnp/src/rnp/.libs/rnp --dearmor out.xxx --output out.d  3.37s user 0.77s system 93% cpu 4.429 total
```

flowher:
Algorithm uses 4 tables of precalculated CRC24 values, thanks to which
it can process in parallel 32 bits of data. This tric doubles performance
Further improvements are possible (and coming later).

Results - processing 1GB armor data

```
OLD: rnp --enarmor=msg /tmp/1gb.rnd --output  4.48s user 0.89s system 98% cpu 5.429 total
NEW: rnp --enarmor=msg /tmp/1gb.rnd --output  2.38s user 0.86s system 79% cpu 4.089 total

OLD: rnp --dearmor out.xxx --output out.d  5.58s user 0.65s system 98% cpu 6.338 total
NEW: rnp --dearmor out.xxx --output out.d  3.28s user 0.84s system 96% cpu 4.275 total
```

Results for CRC calculation only (not sure about those results, but ratio is correct):
```
OLD: 2303MB  / s
NEW: 8025MB / s
```

Closes #521


ronaldtse:
@flowher since this is so awesome and is not dependent on RNP, I wonder if we should at least move it to `libfastcrc`? 😉

flowher:
@ronaldtse okis, I'll do that. Now thinking about it, it may not work on big-endian processor. But that's quick fix.

dewyatt:
Why aren't we just using botan's CRC24 (via `pgp_hash_t` or some other way)?

ni4:
@dewyatt Good question. Botan appears to use the PGP values for CRC. When I was optimizing crc-24 I didn't even think that it may implement it, since there are a lot of CRCs with different polynoms/initial values.

ni4:
@flowher Did you check the performance for all-zero 1Gb files, or with the random data? For constant values performance may be affected by the cache hits/misses.

flowher:
@dewyatt it's slower than this

flowher:
@ni4 I'm using random data in the test. I havn't check all zeros, but that case isn't really interesting

ronaldtse:
@flowher happy to contribute this to Botan via PR as long as @randombit is happy with it 😉

flowher:
Good for merging?

src/lib/crc24_radix64.c
```diff
@@ -0,0 +1,237 @@
...
+crc24_update(uint32_t crc, const uint8_t *in, size_t length)
+{
+    while (((uintptr_t) in & 3) && length) {
+        crc = process8(crc, *in++);
+        length--;
+    }
+
+    while (length >= 16) {
+        length -= 16;
+        crc = process32(crc, ((uint32_t *) in)[0]);
```
flowher: This will fail on big-endian cpu

flowher:
Code tested on qemu with PowerPC. Works OK

ronaldtse:
@dewyatt @ni4 if it's OK let's merge it!

src/lib/crc24_radix64.c
```diff
@@ -0,0 +1,243 @@
...
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+ * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <stdint.h>
+#include <stddef.h>
+#include <stdio.h>
```
dewyatt: I think you can drop `stdio.h`

flowher: thnx! done

PR #525, opened 2017-10-28
(core) Fix CRC24: Was finlizing wrong CRC state

flowher:
Found small bug during work on moving CRC24 to botan.


flowher:
I think tests are OK. Actually, it so happens that this check is still correct (by chance) and we have tests which would catch it. But it looks weird

PR #528, opened 2017-10-31
Change symmetric password encryption to not use provider in lower layers

dewyatt:
The main thing here is just moving password retrieval out of the `init_encrypted_dst` code path and instead deriving the key for passwords immediately.

Also adds support for setting the s2k cipher that will wrap the session key (when applicable). (We'll probably want to add a `--s2k-cipher` CLI arg at some point)

The reason for this change is that with FFI I'm aiming to use password parameters directly when possible. It just makes for a very awkward API in some places otherwise.

I will admit that in some ways this could be viewed as less secure since previously I believe the pass+derived key were isolated to one function, and now the key ends up in `ctx->passwords` until used. I think it's practically no difference though and is a decent compromise.

### DIFF
--- a/src/rnp/rnp.c
+++ b/src/rnp/rnp.c
@@ -400,7 +400,9 @@ rnp_cmd(rnp_cfg_t *cfg, rnp_t *rnp, int cmd, char *f)
         ret = rnp_process_stream(&ctx, f, rnp_cfg_get(cfg, CFG_OUTFILE)) == RNP_SUCCESS;
         break;
     case CMD_SYM_ENCRYPT:
-        ctx.passwordc = 1;
+        ctx.ealg = pgp_str_to_cipher(rnp_cfg_get(cfg, CFG_CIPHER));
+        ctx.halg = pgp_str_to_hash_alg(rnp_cfg_get(cfg, CFG_HASH));
+        rnp_encrypt_add_password(&ctx);
     /* FALLTHROUGH */
     case CMD_ENCRYPT: {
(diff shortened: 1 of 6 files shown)


dewyatt:
Spotted a small bug and another unused var. I think it's good now.

src/rnp/rnp.c
```diff
@@ -400,7 +400,9 @@ rnp_cmd(rnp_cfg_t *cfg, rnp_t *rnp, int cmd, char *f)
         ret = rnp_process_stream(&ctx, f, rnp_cfg_get(cfg, CFG_OUTFILE)) == RNP_SUCCESS;
         break;
     case CMD_SYM_ENCRYPT:
-        ctx.passwordc = 1;
+        ctx.ealg = pgp_str_to_cipher(rnp_cfg_get(cfg, CFG_CIPHER));
+        ctx.halg = pgp_str_to_hash_alg(rnp_cfg_get(cfg, CFG_HASH));
+        rnp_encrypt_add_password(&ctx);
```
ni4: Should not we take some action here if `rnp_encrypt_add_password()` fails?

ni4:
Except single comment everything looks good.
I also added issue #531 for s2k CLI parameters.

src/rnp/rnp.c
dewyatt: Good point, I'll add a check in here.

ronaldtse:
Thank you @dewyatt and @ni4 !

PR #529, opened 2017-10-31
Make string passphrase provider a bit stricter.

dewyatt:
This is just to prevent the (unlikely) possibility of truncating a password with the strncpy below.


ni4:
LGTM!

PR #537, opened 2017-11-15
Misc changes for CLI tests and hash interface

flowher:
3 misc changes in 3  separated commits :

* Removes code duplication for hash interface

* Use 'unitests' python module for CLI tests
   
    Use unittest python module to do CLI tests
    This way we will be able to choose test to
    run
  
    This is done in order to be able to run single
    test during development process. Example:
    ```./src/tests/cli_tests.py -f  -v Keystore.test_generate_with_rnp_import_to_gpg```

    This patch removes '--debug' cmd line option.
    Instead '-v' and '-f' options should be used
    
    For more details on how to run the tests issue
    ./src/tests/cli_tests.py -h

* Refactors cli_performance tests in order to
  be able to run single test during development process
   
  For more info how to do it, run:
  ```./src/tests/cli_perf.py -h```

  Example: ```./src/tests/cli_perf.py large_file_armored_decryption```

### DIFF
--- a/ci/main.sh
+++ b/ci/main.sh
@@ -37,5 +37,5 @@ cd src/tests
 
 LD_LIBRARY_PATH="$LD_LIBRARY_PATH:${GPG21_INSTALL}/lib"
 export LD_LIBRARY_PATH
-env RNPC_GPG_PATH="${GPG21_INSTALL}/bin/gpg" python2 cli_tests.py all --debug
+env RNPC_GPG_PATH="${GPG21_INSTALL}/bin/gpg" python2 cli_tests.py -w -v
 
--- a/src/lib/hash.c
+++ b/src/lib/hash.c
@@ -79,18 +79,19 @@
 #include <rnp/rnp_sdk.h>
 #include <stdio.h>
 
-static pgp_map_t hash_alg_map[] = {
-  {PGP_HASH_MD5, "MD5"},
-  {PGP_HASH_SHA1, "SHA1"},
-  {PGP_HASH_RIPEMD, "RIPEMD160"},
-  {PGP_HASH_SHA256, "SHA256"},
-  {PGP_HASH_SHA384, "SHA384"},
-  {PGP_HASH_SHA512, "SHA512"},
-  {PGP_HASH_SHA224, "SHA224"},
-  {PGP_HASH_SM3, "SM3"},
-  {PGP_HASH_CRC24, "CRC24"},
-  {0x00, NULL}, /* this is the end-of-array marker */
-};
+static const struct hash_alg_map_t {
+    pgp_hash_alg_t  type;
+    const char      *name;
+    const char      *botan_name;
+} hash_alg_map[] = {{PGP_HASH_MD5, "MD5", "MD5"},
+                    {PGP_HASH_SHA1, "SHA1", "SHA-1"},
+                    {PGP_HASH_RIPEMD, "RIPEMD160", "RIPEMD-160"},
+                    {PGP_HASH_SHA256, "SHA256", "SHA-256"},
+                    {PGP_HASH_SHA384, "SHA384", "SHA-384"},
+                    {PGP_HASH_SHA512, "SHA512", "SHA-512"},
+                    {PGP_HASH_SHA224, "SHA224", "SHA-224"},
+                    {PGP_HASH_SM3, "SM3", "SM3"},
+                    {PGP_HASH_CRC24, "CRC24", "CRC24"}};
 
 /**
  * \ingroup Core_Print
@@ -102,7 +103,23 @@ static pgp_map_t hash_alg_map[] = {
 const char *
 pgp_show_hash_alg(uint8_t hash)
 {
-    return pgp_str_from_map(hash, hash_alg_map);
+    const char *ret = NULL;
+    ARRAY_LOOKUP_BY_ID(hash_alg_map, type, name, hash, ret);
+    return ret;
+}
+
+const char *
+pgp_hash_name_botan(pgp_hash_alg_t hash)
+{
+    const char *ret = NULL;
+    ARRAY_LOOKUP_BY_ID(hash_alg_map, type, botan_name, hash, ret);
+    return ret;
+}
+
+const char *
+pgp_hash_name(const pgp_hash_t *hash)
+{
+    return pgp_show_hash_alg(hash->_alg);
 }
 
 /**
@@ -117,60 +134,12 @@ pgp_str_to_hash_alg(const char *hash)
     if (hash == NULL) {
         return PGP_DEFAULT_HASH_ALGORITHM;
     }
-    for (int i = 0; hash_alg_map[i].string != NULL; ++i) {
-        if (rnp_strcasecmp(hash, hash_alg_map[i].string) == 0) {
+    for (int i = 0; i < ARRAY_SIZE(hash_alg_map); i++) {
+        if (!rnp_strcasecmp(hash, hash_alg_map[i].name)) {
             return hash_alg_map[i].type;
         }
     }
     return PGP_HASH_UNKNOWN;
-}
-
-const char *
-pgp_hash_name_botan(pgp_hash_alg_t hash)
-{
-    switch (hash) {
-#if defined(BOTAN_HAS_MD5)
-    case PGP_HASH_MD5:
-        return "MD5";
-#endif
-
-#if defined(BOTAN_HAS_SHA1)
-    case PGP_HASH_SHA1:
-        return "SHA-1";
-#endif
-
-#if defined(BOTAN_HAS_RIPEMD_160)
-    case PGP_HASH_RIPEMD:
-        return "RIPEMD-160";
-#endif
-
-#if defined(BOTAN_HAS_SHA2_32)
-    case PGP_HASH_SHA224:
-        return "SHA-224";
-    case PGP_HASH_SHA256:
-        return "SHA-256";
-#endif
-
-#if defined(BOTAN_HAS_SHA2_64)
-    case PGP_HASH_SHA384:
-        return "SHA-384";
-    case PGP_HASH_SHA512:
-        return "SHA-512";
-#endif
-
-#if defined(BOTAN_HAS_SM3)
-    case PGP_HASH_SM3:
-        return "SM3";
-#endif
-
-#if defined(BOTAN_HAS_CRC24)
-    case PGP_HASH_CRC24:
-        return "CRC24";
-#endif
-
-    default:
-        return NULL;
-    }
 }
 
 /**
@@ -262,18 +231,6 @@ pgp_hash_finish(pgp_hash_t *hash, uint8_t *out)
     hash->handle = NULL;
     hash->_output_len = 0;
     return outlen;
-}
-
-/**
-   \ingroup Core_Hashes
-   \brief Get Hash name
-   \param hash Hash struct
-   \return Hash name
-*/
-const char *
-pgp_hash_name(const pgp_hash_t *hash)
-{
-    return pgp_show_hash_alg(hash->_alg);
 }
 
 size_t
--- a/src/lib/utils.h
+++ b/src/lib/utils.h
@@ -67,14 +67,14 @@
  * lookup_value lookup value
  * ret          return value
  */
-#define ARRAY_LOOKUP_BY_ID(array, id_field, ret_field, lookup_value, ret)   \
-    do {                                                                    \
-        for (size_t i__ = 0; i__ < ARRAY_SIZE(array); i__++) {              \
-            if ((array)[i__].id_field == (lookup_value)) {                  \
-                (ret) = (array)[i__].ret_field;                             \
-                break;                                                      \
-            }                                                               \
-        }                                                                   \
+#define ARRAY_LOOKUP_BY_ID(array, id_field, ret_field, lookup_value, ret) \
+    do {                                                                  \
+        for (size_t i__ = 0; i__ < ARRAY_SIZE(array); i__++) {            \
+            if ((array)[i__].id_field == (lookup_value)) {                \
+                (ret) = (array)[i__].ret_field;                           \
+                break;                                                    \
+            }                                                             \
+        }                                                                 \
     } while (0)
 
 /* Formating helpers */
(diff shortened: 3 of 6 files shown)


flowher:
Test failures to be addressed in following commits

src/lib/hash.c
```diff
@@ -79,18 +79,37 @@
...
+
+/*
+ * @params
+ * array:       array of the structures to lookup
+ * id_field     name of the field to compare against
+ * ret_field    filed to return
+ * id           id to look for
+ * ret          return value
+ */
+#define LOOKUP_ARRAY_BY_ID(array, id_field, ret_field, id, ret) \
```
dewyatt: It seems like this doesn't really need to be a macro, and I think if we can easily avoid a macro, we ought to do so (unless it provides some good benefits).

What do you think about something like this?

```diff
diff --git a/src/lib/hash.c b/src/lib/hash.c
index 429aad4..f2fe324 100644
--- a/src/lib/hash.c
+++ b/src/lib/hash.c
@@ -80,7 +80,7 @@
 #include <stdio.h>
 
 static const struct hash_alg_map_t {
-    uint8_t     type;
+    pgp_hash_alg_t type;
     const char *name;
     const char *botan_name;
 } hash_alg_map[] = {{PGP_HASH_MD5, "MD5", "MD5"},
@@ -93,23 +93,14 @@ static const struct hash_alg_map_t {
                     {PGP_HASH_SM3, "SM3", "SM3"},
                     {PGP_HASH_CRC24, "CRC24", "CRC24"}};
 
-/*
- * @params
- * array:       array of the structures to lookup
- * id_field     name of the field to compare against
- * ret_field    filed to return
- * id           id to look for
- * ret          return value
- */
-#define LOOKUP_ARRAY_BY_ID(array, id_field, ret_field, id, ret) \
-    do {                                                        \
-        for (size_t __i = 0; __i < ARRAY_SIZE(array); __i++) {  \
-            if ((array)[__i].id_field == (id)) {                \
-                (ret) = (array)[__i].ret_field;                 \
-                break;                                          \
-            }                                                   \
-        }                                                       \
-    } while (0)
+static const struct hash_alg_map_t *lookup_hash_alg_by_id(pgp_hash_alg_t id) {
+    for (size_t i = 0; i < ARRAY_SIZE(hash_alg_map); i++) {
+        if (hash_alg_map[i].type == id) {
+            return &hash_alg_map[i];
+        }
+    }
+    return NULL;
+}
 
 /**
  * \ingroup Core_Print
@@ -121,17 +112,15 @@ static const struct hash_alg_map_t {
 const char *
 pgp_show_hash_alg(uint8_t hash)
 {
-    const char *ret = NULL;
-    LOOKUP_ARRAY_BY_ID(hash_alg_map, type, name, hash, ret);
-    return ret;
+    const struct hash_alg_map_t* entry = lookup_hash_alg_by_id(hash);
+    return entry ? entry->name : NULL;
 }
 
 const char *
 pgp_hash_name_botan(pgp_hash_alg_t hash)
 {
-    const char *ret = NULL;
-    LOOKUP_ARRAY_BY_ID(hash_alg_map, type, botan_name, hash, ret);
-    return ret;
+    const struct hash_alg_map_t* entry = lookup_hash_alg_by_id(hash);
+    return entry ? entry->botan_name : NULL;
 }
 
 const char *
```

ci/main.sh
```diff
@@ -37,5 +37,5 @@ cd src/tests
 
 LD_LIBRARY_PATH="$LD_LIBRARY_PATH:${GPG21_INSTALL}/lib"
 export LD_LIBRARY_PATH
-env RNPC_GPG_PATH="${GPG21_INSTALL}/bin/gpg" python2 cli_tests.py all --debug
+env RNPC_GPG_PATH="${GPG21_INSTALL}/bin/gpg" python2 cli_tests.py -w
```
dewyatt: I think we should have `-wvf` here for CI.

src/tests/cli_tests.py
```diff
@@ -673,219 +542,408 @@ def rnp_cleartext_signing_gpg_to_rnp(filesize):
...
+            rnp_signing_gpg_to_rnp(size)
+            rnp_detached_signing_gpg_to_rnp(size)
+            rnp_cleartext_signing_gpg_to_rnp(size)
 
-def cleanup():
-    if DEBUG:
-        return
+if __name__ == '__main__':
+    main = unittest.main
+    main.USAGE = "\nRNP test client specific flags:\n  -w,\t\t Don't remove working directory" + main.USAGE
```
dewyatt: Looks like this needs another newline before `main.USAGE`.

src/lib/hash.c
ni4: I agree on this - we should minimize macros usage for code readability. Except performance cases - where you need to duplicate bunch of code lines to trade memory for speed, and places where there is no other solution - like `RNP_LOG`.
In other cases it is more clear and simple to use a function instead of macros.

src/tests/cli_perf.py
```diff
@@ -9,6 +9,8 @@
 from timeit import default_timer as perf_timer
 from cli_common import find_utility, run_proc, run_proc_fast, pswd_pipe, rnp_file_path, random_text, file_text, size_to_readable, raise_err
 import cli_common
+import inspect
+from optparse import OptionParser
```
dewyatt: I think `optparse` was deprecated like 7 years ago, would prefer `argparse`, but not a big deal.

ci/main.sh
flowher: -v yes, but -f no because I would prefer to get all the errors not only first one

src/tests/cli_tests.py
```diff
@@ -673,219 +542,408 @@ def rnp_cleartext_signing_gpg_to_rnp(filesize):
...
-        sys.exit(2)
-
-    tests = []
-
-    for arg in args[1:]:
-        if arg in ['-h', '--help']:
-            print 'Usage:\ncli_tests [-h | --help] [rnp] [rnpkeys] [all] [-d | --debug]'
-            sys.exit(0)
-        elif arg in ['-d', '--debug']:
-            DEBUG = True
```
ni4: Looks like we lost ability to print out all running commands, at least it doesn't work in `-v` mode

ci/main.sh
dewyatt: I kind of prefer CI failing fast/early, but it's OK either way.

src/lib/hash.c
flowher: This macro will be used for not only the one related to ``hash_alg_map_t``. I don't like returning ``void*``

src/tests/cli_tests.py
dewyatt: Yeah I noticed this too, I think it was very useful. Maybe we should count the `-v` args and have a verbosity level? (Or bring the debug arg back for this?)

src/tests/cli_tests.py
flowher: Not sure why it would need it.

src/tests/cli_perf.py
flowher: agreed

src/tests/cli_tests.py
dewyatt: Right now it looks like this for me:
```
RNP test client specific flags:
  -w,		 Don't remove working directoryUsage: cli_tests.py [options] [test] [...]
```

With a couple newlines:
```
RNP test client specific flags:
  -w,		 Don't remove working directory

Usage: cli_tests.py [options] [test] [...]
```

src/tests/cli_tests.py
flowher: Yes, the output is different. ``-v`` shows all the tests being run. Unfortunately those tests needs to be further redesigned. Currently in one single test runs f.e. encryption for all algorithms, sizes, compressions... this is far from useful.

src/tests/cli_tests.py
flowher: ah, I see what's you mean. Agreed,

ci/main.sh
flowher: I would say ``-f`` is useful during development time. CI should show all regressions (so that developer doesn't have to wait again after fixing first one)

src/tests/cli_tests.py
flowher: hmmm... which interesting output you are talking about exactly?

src/lib/hash.c
ni4: Extensions are even worse :) But what is the problem in having another functions named `lookup_encryption_alg_by_id`, `lookup_pk_alg_by_id` which will do the same cycle? It is just another 20 bytes of code, and 4-5 code lines, in price of better readability and usability.
Nobody here loves code duplication but there should be some compromise between complicating things and saving code lines.

dewyatt: Yeah I would say code readability is generally more important than reducing duplication (/DRY), particularly when working in a language like C.

flowher: @ni4 X number of useless functions like that can be eliminated by simple, readable 5 line macro.

ni4: In price of this you'll need to keep the same format of maps. For instance, the next thing you'll need to add hash length to the hash map, key bit size to the symmetric key algos, and some flags to the pk algo. You'll need to rewrite all these or start thinking about the next universal solution. 
Anyway, it's question of the coding style and everybody has his own opinion based on experience.

src/tests/cli_tests.py
flowher: Done - now we are using ``logging`` module to get debug logs. ``-d`` option is back

flowher:
Anything more for anyone else?

ronaldtse:
@ni4 @dewyatt are we ready to go? Thanks!

dewyatt:
I think the only remaining issue was the macro, so I took a second look at how many places it would be beneficial.

I think I can see maybe 6 places I could use it in some of my recent (JSON-related) code, and probably more soon. Maybe it's one of the few cases where introducing a macro would actually be worth the trade-offs.

Either way, I have some remaining nits that I will add.

src/lib/utils.h
```diff
@@ -56,6 +59,24 @@
...
+ * @params
+ * array:       array of the structures to lookup
+ * id_field     name of the field to compare against
+ * ret_field    filed to return
+ * id           id to look for
+ * ret          return value
+ */
+#define LOOKUP_ARRAY_BY_ID(array, id_field, ret_field, id, ret) \
+    do {                                                        \
+        for (size_t __i = 0; __i < ARRAY_SIZE(array); __i++) {  \
```
dewyatt: The underscore prefixes on `__i` clash with the existing conventions and don't really add anything. Should just be `i`.

src/lib/utils.h
```diff
@@ -56,6 +59,24 @@
...
 
+/*
+ * @params
+ * array:       array of the structures to lookup
+ * id_field     name of the field to compare against
+ * ret_field    filed to return
+ * id           id to look for
+ * ret          return value
+ */
+#define LOOKUP_ARRAY_BY_ID(array, id_field, ret_field, id, ret) \
```
dewyatt: The naming here is awkward to me. Something like `ARRAY_SEARCH_BY_INT` or similar would make more sense.

The way it currently is, it sounds like we're looking up an array somewhere, but that's not really the case.

src/lib/hash.c
```diff
@@ -79,18 +79,19 @@
...
-  {PGP_HASH_SHA256, "SHA256"},
-  {PGP_HASH_SHA384, "SHA384"},
-  {PGP_HASH_SHA512, "SHA512"},
-  {PGP_HASH_SHA224, "SHA224"},
-  {PGP_HASH_SM3, "SM3"},
-  {PGP_HASH_CRC24, "CRC24"},
-  {0x00, NULL}, /* this is the end-of-array marker */
-};
+static const struct hash_alg_map_t {
+    uint8_t     type;
```
dewyatt: This should be `pgp_hash_alg_t` so that the compiler has a chance to catch issues (at least clang seems to be smart enough).

src/lib/utils.h
flowher: actually 'i' is really bad name as it's commonly used for counter and it may clash. let's call it 'i_'

src/lib/utils.h
flowher: ``ARRAY_SEARCH_BY_ID`` as it doesn't have to be an int

src/lib/utils.h
dewyatt: @flowher Actually `i_` might be a good idea here as you mentioned.

Otherwise, we're going to have some subtle macro bugs like:
```c
int i = 3;
const char *ret = NULL;
ARRAY_SEARCH_BY_ID(hash_alg_map, type, name, i, ret);
// ret is still NULL here if variable name is i
```

flowher: yes, this macro was OK initially

flowher:
Done, now it needs to be merged.

ni4:
@flowher Looks good, except line wraps. We have 95 column limit for C code, but in Python even 87 columns are wrapped. Is this some kind of standard? And can you please introduce the tool-script-config-whatever else used so later on we may stick to the same coding style?

flowher:
@ni4 I would say let's stick to standard PEP8. Currently we are not, but there is no need to invent new coding style. How about that?

https://www.python.org/dev/peps/pep-0008/#maximum-line-length

ronaldtse:
Looks fine with me. Let’s get this merged.

src/lib/utils.h
```diff
@@ -56,6 +59,24 @@
...
 
+/*
+ * @params
+ * array:       array of the structures to lookup
+ * id_field     name of the field to compare against
+ * ret_field    filed to return
+ * lookup_value lookup value
+ * ret          return value
+ */
+#define ARRAY_LOOKUP_BY_ID(array, id_field, ret_field, lookup_value, ret) \
```
dewyatt: I think you were going to change this to `ARRAY_SEARCH_BY_ID` (https://github.com/riboseinc/rnp/pull/537#discussion_r151877028).

There is some nuance here and search is the more accurate descriptor.

flowher: Thanks. I've decided that I prefer ``lookup`` as this is what this macro does - it retrieves a record from a table corresponding to lookup_value. it doesn't search whole array

dewyatt: >  it doesn't search whole array

Well, surely it does search the whole array when it doesn't find what it's looking for.

Anyways, one distinction is that `lookup` implies that you are retrieving something that *definitely* exists, whereas `search` implies you are trying to find something that *may* exist. In our case, it's the latter - we may not find what we are looking for.

http://www.thefreedictionary.com/look+up
> To search for and find, as in a reference book.

https://english.stackexchange.com/a/165859

But I can certainly see some points in favor of either terms, and I'm sure developers are going to read the macro contents anyways.

flowher: So my way of thinking is - when "calling" this macro you expect that value exist (otherwise it's an error condition), you have a method for locating candidates (lookup_value) and what's most important - you expect that there is only one result.

dewyatt: Seems logical to me, let's just leave the name.

dewyatt:
I think everything was addressed here, so going to merge.

ronaldtse:
Thank you guys!!

PR #538, opened 2017-11-15
(doc) Moves dev-guide to doc/ and adds some rules regarding commits

flowher:
How about that?


ronaldtse:
I agree with the rules, and placing in `doc/` seems fine to me.

ronaldtse:
We probably should rename to `DEVELOPMENT-GUIDE` in-line with the rest of the documents too.

flowher:
Ok, renamed

PR #541, opened 2017-11-25
Fix code to be compliant with -Wsign-compare

ni4:
As per discussion at https://github.com/riboseinc/rnp/pull/539#discussion_r152394006 I checked code with `-Wsign-compare`, and introduced required fixes.
This PR doesn't add changes to the streamed code to avoid merge problems since they were added in PR #539.

@dewyatt @ronaldtse @flowher  Probably we should add `-Wsign-compare` to default build options, or travis checks, what do you think?

### DIFF
--- a/src/lib/crypto/ecdsa.c
+++ b/src/lib/crypto/ecdsa.c
@@ -127,7 +127,7 @@ pgp_ecdsa_verify_hash(const pgp_ecc_sig_t *   sign,
         return RNP_ERROR_BAD_PARAMETERS;
     }
 
-    if ((BN_num_bytes(pubkey->point) > sizeof(point_bytes)) ||
+    if ((BN_num_bytes(pubkey->point) > (int) sizeof(point_bytes)) ||
         BN_bn2bin(pubkey->point, point_bytes) || (point_bytes[0] != 0x04)) {
         RNP_LOG("Failed to load public key");
         goto end;
@@ -155,8 +155,8 @@ pgp_ecdsa_verify_hash(const pgp_ecc_sig_t *   sign,
         goto end;
     }
 
-    if ((BN_num_bytes(sign->r) > curve_order) || (BN_num_bytes(sign->s) > curve_order) ||
-        (curve_order > MAX_CURVE_BYTELEN)) {
+    if ((BN_num_bytes(sign->r) > (int) curve_order) ||
+        (BN_num_bytes(sign->s) > (int) curve_order) || (curve_order > MAX_CURVE_BYTELEN)) {
         goto end;
     }
 
--- a/src/lib/crypto/sm2.c
+++ b/src/lib/crypto/sm2.c
@@ -127,7 +127,7 @@ pgp_sm2_verify_hash(const pgp_ecc_sig_t *   sign,
 
     const size_t sign_half_len = BITS_TO_BYTES(curve->bitlen);
 
-    if ((BN_num_bytes(pubkey->point) > sizeof(point_bytes)) ||
+    if ((BN_num_bytes(pubkey->point) > (int) sizeof(point_bytes)) ||
         BN_bn2bin(pubkey->point, point_bytes) || (point_bytes[0] != 0x04)) {
         RNP_LOG("Failed to load public key");
         goto end;
@@ -152,8 +152,8 @@ pgp_sm2_verify_hash(const pgp_ecc_sig_t *   sign,
         goto end;
     }
 
-    if ((BN_num_bytes(sign->r) > sign_half_len) || (BN_num_bytes(sign->s) > sign_half_len) ||
-        (sign_half_len > MAX_CURVE_BYTELEN)) {
+    if ((BN_num_bytes(sign->r) > (int) sign_half_len) ||
+        (BN_num_bytes(sign->s) > (int) sign_half_len) || (sign_half_len > MAX_CURVE_BYTELEN)) {
         goto end;
     }
 
@@ -213,7 +213,7 @@ pgp_sm2_encrypt(uint8_t *               out,
         goto done;
     }
 
-    if ((BN_num_bytes(pubkey->point) > sizeof(point_bytes)) ||
+    if ((BN_num_bytes(pubkey->point) > (int) sizeof(point_bytes)) ||
         BN_bn2bin(pubkey->point, point_bytes) || (point_bytes[0] != 0x04)) {
         RNP_LOG("Failed to load public key");
         goto done;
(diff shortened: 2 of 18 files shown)


src/lib/generate-key.c
```diff
@@ -254,7 +254,7 @@ static bool
 set_default_user_prefs(pgp_user_prefs_t *prefs)
 {
     if (!prefs->symm_algs) {
-        for (int i = 0; i < ARRAY_SIZE(DEFAULT_SYMMETRIC_ALGS); i++) {
+        for (unsigned i = 0; i < ARRAY_SIZE(DEFAULT_SYMMETRIC_ALGS); i++) {
```
flowher: Technically it should be ``size_t`` as (in most cases) you compare to size_t not to ``unsigned ``. But all casts will work

src/lib/hash.c
```diff
@@ -80,9 +80,9 @@
 #include <stdio.h>
 
 static const struct hash_alg_map_t {
-    pgp_hash_alg_t  type;
-    const char      *name;
-    const char      *botan_name;
+    pgp_hash_alg_t type;
+    const char *   name;
```
flowher: Previous code was correct

src/tests/cipher.c
```diff
@@ -612,9 +612,8 @@ sm2_roundtrip(void **state)
     uint8_t ctext_buf[1024];
     uint8_t decrypted[27];
 
-    const rnp_keygen_crypto_params_t key_desc = {.key_alg = PGP_PKA_SM2,
-                                                 .hash_alg = PGP_HASH_SM3,
-                                                 .ecc = {.curve = PGP_CURVE_SM2_P_256}};
+    const rnp_keygen_crypto_params_t key_desc = {
```
flowher: Why changing this? previous one was better

ronaldtse:
@ni4 I agree that we should enable `-Wsign-compare` to both build and Travis 👍

src/lib/hash.c
flowher: Does it actually comes from clang-format?

src/lib/generate-key.c
ni4: Agree, will change to size_t so things look more consistent.
Got to fun situation while playing with `-Wextra` - it doesn't allow `{0}` initializer for records while this is directly described in C standard.

dewyatt:
I feel like some parts of this are band-aids that don't address the underlying issues.

`BN_num_bytes`, for example, shouldn't truncate the `size_t` to an `int`, It should return `bool` and take a `size_t *` param. (Or maybe return `size_t` and use `SIZE_MAX` to signal an error)

At a glance, it looks like merging this would cover up some exploitable buffer overflows.

I do agree that we should have `-Wsign-compare`, but we may also want to add `-Wno-error=sign-compare` until we can properly fix these. (And we should not rush this)

src/lib/hash.c
ni4: Yeah, this and the following changes were introduced by clang-format, and I mislooked it. Will revert.

ni4: ...will revert, but not this one - it seems it wasn't clang-formatted at some point,  * should be one space after the type.

src/lib/generate-key.c
dewyatt: > Got to fun situation while playing with -Wextra - it doesn't allow {0} initializer for records while this is directly described in C standard.

Are you sure you don't just need `{{0}}` instead? IIRC the requirement is that you initialize the first field (and it cascades from there). If the first field is a struct, `{{0}}` or similar is needed.

ni4:
This is mostly a quick-fix so later on it would be easier to check code for sign-compare errors.
> BN_num_bytes, for example, shouldn't truncate the size_t to an int, It should return bool and take a size_t * param. (Or maybe return size_t and use SIZE_MAX to signal an error)

@dewyatt  When this may fail? Given that argument is a correctly loaded bignum. The only fail-point is NULL as parameter or some problems in internal structures, or do I miss something?

src/lib/generate-key.c
ni4: @dewyatt Hm, nope. For the following line:
`pgp_data_t OID = {0};` with the declaration:
```
typedef struct {
    size_t   len;
    uint8_t *contents;
    uint8_t  mmapped; /* contents need an munmap(2) */
} pgp_data_t;
```
Compiler tells:
```
error: missing field 'contents' initializer [-Werror,-Wmissing-field-initializers]
    pgp_data_t OID = {0};
```

dewyatt: Ah, are you using clang? I think clang still has a bug there: https://bugs.llvm.org/show_bug.cgi?id=21689

src/lib/hash.c
flowher: yeah, ok I see. I think clang-format should put the ``*`` before the variable really. anyways, that's find

src/lib/generate-key.c
ni4: Yeah, as clang is default on macOS. Probably they will resolve the issue in a few more years :-D

ni4:
@dewyatt @flowher I updated the code, addressing comments.
So, may this be merged now?

src/lib/crypto/ecdsa.c
```diff
@@ -155,8 +155,8 @@ pgp_ecdsa_verify_hash(const pgp_ecc_sig_t *   sign,
         goto end;
     }
 
-    if ((BN_num_bytes(sign->r) > curve_order) || (BN_num_bytes(sign->s) > curve_order) ||
-        (curve_order > MAX_CURVE_BYTELEN)) {
+    if ((BN_num_bytes(sign->r) > (int) curve_order) ||
```
flowher: FIX: BN_num_bytes must be > 0

src/lib/crypto/sm2.c
```diff
@@ -152,8 +152,8 @@ pgp_sm2_verify_hash(const pgp_ecc_sig_t *   sign,
         goto end;
     }
 
-    if ((BN_num_bytes(sign->r) > sign_half_len) || (BN_num_bytes(sign->s) > sign_half_len) ||
-        (sign_half_len > MAX_CURVE_BYTELEN)) {
+    if ((BN_num_bytes(sign->r) > (int) sign_half_len) ||
```
flowher: FIX: BN_num_bytes must be > 0

flowher:
The problem with ``BN_num_bytes`` is marginal and I'm not sure it will happen. I'll take care of that, as I"m adding some tests anyway.
Go ahead and merge if you feel like it's OK now

dewyatt:
Sounds good, will merge. I just think some of the casts here are only silencing the warning and not really a solution.

I'm thinking the overflows aren't exploitable, given PGP's restrictions on MPI sizes.

flowher:
I agree -that's definitely not a real solution.

PR #545, opened 2017-11-27
Streamed processing improvements

ni4:
This PR includes some improvements over streamed code which I noticed but didn't include into the previous PR:
- added common initialization functions, removed duplicate zero-initializations
- replaced default RNP_SUCCES for return code (thanks to @dewyatt for noticing some time ago)
- shortcuted iteration over lists of items
- replaced DYNARRAY with list for signature subpackets
- added pgp_hash_list_* functions to use them later in streamed signing

### DIFF
--- a/src/lib/hash.c
+++ b/src/lib/hash.c
@@ -162,14 +162,14 @@ pgp_hash_create(pgp_hash_t *hash, pgp_hash_alg_t alg)
 
     rc = botan_hash_init(&impl, hash_name, 0);
     if (rc != 0) {
-        (void) fprintf(stderr, "Error creating hash object for '%s'", hash_name);
+        RNP_LOG("Error creating hash object for '%s'", hash_name);
         return false;
     }
 
     rc = botan_hash_output_length(impl, &outlen);
     if (rc != 0) {
         botan_hash_destroy(hash->handle);
-        (void) fprintf(stderr, "In pgp_hash_create, botan_hash_output_length failed");
+        RNP_LOG("In pgp_hash_create, botan_hash_output_length failed");
         return false;
     }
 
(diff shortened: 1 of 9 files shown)


src/lib/hash.c
```diff
@@ -271,3 +271,50 @@ pgp_digest_length(pgp_hash_alg_t alg, size_t *output_length)
     botan_hash_destroy(handle);
     return ret;
 }
+
+bool
+pgp_hash_list_add(list *hashes, pgp_hash_alg_t alg)
```
flowher: comment here would be useful. Something that tells me what I should do if this function returns false. In more details - if ``pgp_hash_create`` fails - then memory is not allocated, but if ``list_append`` fails then memory is allocated and I need to free it with.

src/lib/hash.h
```diff
@@ -77,16 +78,42 @@ void pgp_calc_mdc_hash(
...
+ *         on other error
+ **/
+bool pgp_hash_list_add(list *hashes, pgp_hash_alg_t alg);
+
+/* @brief Get hash structure for the corresponding algorithm
+ * @param hashes List of pgp_hash_t structures
+ * @param alg Hash algorithm
+ * @return pointer to the pgp_hash_t structure or NULL if there is no one in the list
+ **/
+pgp_hash_t *pgp_hash_list_get(list hashes, pgp_hash_alg_t alg);
```
flowher: It would be better to return const pointer here

src/lib/hash.h
```diff
@@ -77,16 +78,42 @@ void pgp_calc_mdc_hash(
...
+ * @return true if hash was added successfully or already exists in the list.
+ *         false will be returned if memory allocation failed, or alg is not supported, or
+ *         on other error
+ **/
+bool pgp_hash_list_add(list *hashes, pgp_hash_alg_t alg);
+
+/* @brief Get hash structure for the corresponding algorithm
+ * @param hashes List of pgp_hash_t structures
+ * @param alg Hash algorithm
+ * @return pointer to the pgp_hash_t structure or NULL if there is no one in the list
```
flowher: "or NULL if list doesn't contain ``alg``"

src/lib/hash.c
ni4: Thanks for spotting this. Actually, I forgot to `pgp_hash_finish()` in case of `list_append()` fails. Will fix.

src/lib/hash.h
ni4: Agree.

src/lib/hash.h
ni4: Ok

ni4:
@flowher Please see the updated commits ( I squashed one with fixes with the old one)

flowher:
Thnx !

PR #555, opened 2017-12-02
Better checks for results from BN_num_bits/bytes

### LINKED ISSUE #544: SM2 and ECDSA: code doesn't check result of BN_num_bytes
## Description

Code doesn't check result from ``BN_num_bytes``. See comments in #541
This may cause buffer overflows

flowher:
* Changes interface of BN_num_bytes and BN_num_bits: function now returns ``bool`` and number of bits/bytes as an argument
* Some cleanup in bn.c
* Introduces ``BN_hash`` for hashing bingums and moves already existing implementation

Closes #544

### DIFF
--- a/src/lib/crypto/bn.c
+++ b/src/lib/crypto/bn.c
@@ -205,18 +205,23 @@ PGPV_BN_clear_free(PGPV_BIGNUM *a)
     PGPV_BN_free(a);
 }
 
-int
-BN_num_bits(const PGPV_BIGNUM *a)
+bool
+BN_num_bits(const PGPV_BIGNUM *a, size_t *bits)
 {
-    size_t num_bits;
-    if (a == NULL) {
-        return -1;
+    if (!a || botan_mp_num_bits(a->mp, bits)) {
+        return false;
     }
+    return true;
+}
 
-    if (botan_mp_num_bits(a->mp, &num_bits) < 0) {
-        return -1;
+bool
+BN_num_bytes(const PGPV_BIGNUM *a, size_t *bits)
+{
+    if (BN_num_bits(a, bits)) {
+        *bits = BITS_TO_BYTES(*bits);
+        return true;
     }
-    return num_bits;
+    return false;
 }
 
 void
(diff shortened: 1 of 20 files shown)


ronaldtse:
Thanks @flowher . Could we also replace `grip_hash_bignum` with `BN_hash` stuff? There are some other `fprintf` calls in the edited functions -- could you also clean them up? Appreciate it!

src/librepgp/packet-print.c
```diff
@@ -292,16 +294,19 @@ print_block(int indent, const char *name, const uint8_t *str, size_t length)
 }
 
 /* return the number of bits in the public key */
-static int
+static size_t
```
ni4: This function is used to print information about the key, and I think users would like to see '2048' bits instead of more correct 2045 or 2047.

flowher:
@ronaldtse Okis for cleanup. For ``grip_hash_bignum``, the difficulty is that ``BN_hash`` calculates hash in slidely different way (length of mpi is not hashed by ``grip_hash_bignum``). After change I would need to regenerate test data in ``src/test/data/keyrings/`` and change hardcoded keygrip values in the tests. I think, such change wouldn't be  backward compatible (keygrip value is stored with a key, isn't?)

src/librepgp/packet-print.c
flowher: argh... yes, your right

src/lib/crypto/bn.c
```diff
@@ -205,32 +206,23 @@ PGPV_BN_clear_free(PGPV_BIGNUM *a)
...
+BN_num_bytes(const PGPV_BIGNUM *a, size_t *bits)
 {
-    size_t num_bits;
-    if (a == NULL) {
-        return -1;
-    }
-
-    if (botan_mp_num_bits(a->mp, &num_bits) < 0) {
-        return -1;
+    if (BN_num_bits(a, bits)) {
```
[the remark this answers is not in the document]
flowher: I don't think there is big difference. One argument is - because then I can handle errors from botan in single place. I guess one is as good as the other one internally botan_mp_num_bits does same thing as bn_mp_num_bytes.

flowher:
Can we merge it ? I think it's ready to go. I'll need this code for another PR and I would like to avoid merging
@ronaldtse I've open #558 for your request. I would like to do it after DRBG work

ni4:
@flowher You may merge it, but probably @dewyatt would like to take a look. The same is with #552.

ronaldtse:
Oops @ni4 I've already merged it... @dewyatt please feel free to review again.

flowher:
Thnx!
Please add comments here and I'll align if needed.

PR #563, opened 2017-12-07
Fix annoying erroneous message about disabled core dumps

ni4:


dewyatt:
Good catch. I think it's also best to initialize with `bool coredumps = true` rather than `false`, so that the user is warned when compiled without `sys/resource.h`. (From #556)

Side note: I think the error message should be changed -- rather than "warning: core dumps enabled [...]", it's more accurate to say "warning: core dumps may be enabled [...]" (we don't really verify that they are enabled - they could be globally disabled already, and the user would be confused by this message in that case).

ronaldtse:
Thanks @ni4 for fixing this annoying issue. Agree with @dewyatt on both proposed changes.

Somehow this test is failing?

```
ERROR: test_generate_key_with_gpg_import_to_rnp (__main__.Keystore)
  File "cli_tests.py", line 684, in test_generate_key_with_gpg_import_to_rnp
    raise_err('gpg key generation failed', err)
    raise CLIError(msg, log)
CLIError: None

```

flowher:
I think the test randomly fails sometimes. I've seen it before.

flowher:
I've restarted CI - test is OK now.

ronaldtse:
Thanks @flowher !

ni4:
@dewyatt @ronaldtse and @flowher Thanks all! 
Please see the updated commit.

PR #564, opened 2017-12-07
Use HMAC_DRBG for random number generation

### LINKED ISSUE #251: Single instance of RNG

### LINKED ISSUE #554: RNP uses consant IV equal to 0 for symmetric encryption (CFB)
After quick look at symmetric ciphers, it seems to me that RNP doesn't use IV CFB encryption. Design of this mode strongly requires IV to be unpredictable in order to be secure. NIST specification 800-38A says "The  IV  need  not  be  secret,  but  it  must  be  unpredictable". It means - using constant IV is insecure.

This bug is mainly about ``init_encrypted_dst`` (``stream-write.c``), but this occures in many calls to ``pgp_cipher_start``.

flowher:
This patch changes random number generation. Instead of initializing random number generator each time request to PRNG is made, we do it only once when ``rng_t`` (or ``rnp_ffi_t``) is created. Instead of using system resources we use HMAC_DRBG, which makes implementation independent of entropy available in the operating system (resource like ``/dev/random`` is used only during initialization of DRBG). This makes random less error prone and may make it faster (especially in situation when multiple threads access same system PRNG. Getting random data should almost never fail or block (when correct parameters are passed to the function). If needed, reseeding of DRBG is done automatically by botan.

This patch may be bigger then expected (it's actually way bigger than I thought). That's mainly because ``rng_t`` context must be passed from most upper level (client API) to the core of rnp. Hence diff is unusually quite big as for such small feature. Reviewing commit after commit makes probably more sense than whole diff.

In some rare cases we don't use DRBG. That's because initializing DRBG may take some time. Places where we don't use DRBG is mainly code which will be deleted after stream processing is finished and some code in keystore (as there is no ``rnp_t`` context). 

Additionally some smaller were fixed.

Regarding performance - we don't have tests which could measure that. I've run all key generation tests from our unit tests. Results are variable - this implementation was always faster, but sometimes 10sec, sometimes 2sec. But I think we would need quite special set of tests to measure it correctly.

Closes #554 
Closes #251


src/lib/crypto.h
```diff
@@ -128,7 +130,8 @@ bool pgp_generate_subkey(rnp_keygen_subkey_desc_t *     desc,
  *  @param subkey_pub pointer to store the generated public key, must not be NULL
  *  @return true if successful, false otherwise.
  **/
-bool pgp_generate_keypair(rnp_keygen_primary_desc_t *primary_desc,
+bool pgp_generate_keypair(struct rng_t *             rnp,
```
dewyatt: `rnp` -> `rng`

src/tests/ffi.c
```diff
@@ -939,6 +944,7 @@ test_ffi_encrypt_pk(void **state)
     assert_int_equal(RNP_SUCCESS, rnp_output_to_file(&output, "decrypted"));
     assert_non_null(output);
     assert_int_equal(RNP_SUCCESS, rnp_ffi_set_pass_provider(ffi, getpasscb, "password"));
+#if 0
```
dewyatt: Accident I assume

src/lib/crypto/rng.h
```diff
@@ -0,0 +1,105 @@
...
+#define RNP_RANDOM_H_
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <botan/ffi.h>
+
+enum { RNG_DRBG, RNG_SYSTEM };
+typedef uint8_t rng_type_t;
+
+struct rng_t {
```
dewyatt: This should have a typedef and that should be used everywhere instead of `struct rng_t`.

flowher: yep

include/rnp/rnp_types.h
```diff
@@ -95,7 +98,11 @@ typedef struct rnp_symmetric_pass_info_t {
 
 /* rnp operation context : contains additional data about the currently ongoing operation */
 typedef struct rnp_ctx_t {
-    rnp_t *        rnp;           /* rnp structure */
+    union {
```
dewyatt: No anonymous unions in C99.

(Also probably easier to add at the end of this structure to avoid all the `{0}` -> `{{0}}` changes, but not a particularly big deal)

flowher: yeah, I know. I was wondering if this should be done in this PR as it will make this PR even bigger. But ... let's do it!

flowher: Done

dewyatt:
I can't see the need for the rnp2-specific handling here, it looks a bit hacky. The diff below works well for me and doesn't require exposing definitions of rnp2 internal structures to the rest of the codebase, doesn't need the `is_rnp2` stuff, etc.

https://gist.github.com/dewyatt/a5c34cef118d17602db0dd988c034887

flowher:
It works becouse you now have pointer to rng in rnp_ctx and rnp in rng_ctx. 
I think there is bigger problem - we have too many operation contexts.
I'll give it a thought later today but looks Ok.

ni4:
Actually, we have just 3 contexts:
- rnp_t, which is a global context with per-application information (like keyrings, io, and some other stuff. Global rng should go here)
- rnp_ctx_t, which is context for current operation. It has pointer to rnp_t, which was a temporary solution, and actually pointers to required fields, like keyrings, should be copied here instead of having pointer to rnp_t
- rnp_ffi_t, which should be used only in FFI and cannot go to other places like rnp_ctx_t and so on.

So, logically, it would be to add pointer to rng to rnp_ctx_t and copy it there before operations start - from the rnp_t, or from rnp_ffi_t, depending on caller.

flowher:
We also have ``rnp_op_encrypt_t`` which is same as ``rnp_ctx_t``. I think we only need FFI which will serve very well as "global context with per-application information" + opaque functions to get required information ("like keyrings, io, and some other stuff").

ni4:
@flowher `rnp_op_encrypt_t` is the ffi thing to keep data between calls to `rnp_op_encrypt_calls`, and it is kinda ffi wrapper around `rnp_ctx_t` and input/output streams. Anyway, FFI layer is built on top of lib layer so data should be transferred from upper layer to bottom, and bottom layer (`rnp_ctx_t`, `rnp_t`) must not have pointers to the upper layer.

dewyatt:
@flowher 
> We also have rnp_op_encrypt_t which is same as rnp_ctx_t.

I see these as quite different, not sure what you mean.

> I think we only need FFI (or rnp2.c) in future, which will serve very well as "global context with per-application information" + opaque functions to get required information ("like keyrings, io, and some other stuff").

> Except if there is some other reason why we need FFI interface and the other one?

Right, when Jack and I were discussing the initial approach for the FFI, I mentioned specifically "if we find a lot of inflexible things within rnp (likely), we may just want to treat FFI as the rnp2 API you wrote about previously, at least to some degree.". So yes, FFI has always had a sort of secondary purpose as a testbed for a v2 API.

flowher:
If "rnp_ctx_t, which is context for current operation" then `` rnp_op_encrypt_t`` is a context for current encryption operation, isn't?
I assume it is, but then ``rnp_ctx_t`` is also context for encryption operation (as it is used in ``rnp_encrypt_stream``). So as far as I can see, they are a bit duplicated.

Of course, I see obvious difference - one is used for FFI and encryption only and the other one is not. But my point then is -  if we have a way to access FFI structure (opaque functions) then we can use it internally. 
And having 1 structure instead of 2 has quite obvious benefits - less copying/initialization/code/bugs to cope with. So today RNP2 is for FFI stuff, but my point is - it doesn't have to be

flowher:
Done

flowher:
Done

src/lib/crypto/rng.h
```diff
@@ -0,0 +1,105 @@
...
+#include <botan/ffi.h>
+
+enum { RNG_DRBG, RNG_SYSTEM };
+typedef uint8_t rng_type_t;
+
+typedef struct rng_st_t {
+    bool        initialized;
+    rng_type_t  rng_type;
+    botan_rng_t botan_rng;
+} rng_t;
```
dewyatt: I think ideally this would be opaque, with the definition only in `rng.c`, and an `rnp_new` or similar function for allocation, etc. I think we ought to aim to do that on new additions like this.

It may not be top priority at the moment, so not a big deal.

src/lib/rnp2.h
```diff
@@ -0,0 +1,105 @@
+/*-
```
dewyatt: This file can be deleted.

src/lib/rnp2_internal.h
```diff
@@ -0,0 +1,105 @@
+/*-
```
dewyatt: I would prefer we drop this file and any other rnp2 reorganizing changes. It's not really neccessary in this PR and it's already going to be enough work to rebase.

(Plus I've already done some work towards a larger reorganization of the FFI.)

src/lib/rnp.c
```diff
@@ -104,6 +104,7 @@ __RCSID("$NetBSD: rnp.c,v 1.98 2016/06/28 16:34:40 christos Exp $");
 
 #include <json.h>
 #include <rnp.h>
+#include "rnp2_internal.h"
```
dewyatt: Can be dropped now

src/lib/rnp2.c
```diff
@@ -1118,6 +1070,11 @@ rnp_decrypt(rnp_ffi_t ffi, rnp_input_t input, rnp_output_t output)
         return RNP_ERROR_NULL_POINTER;
     }
 
+    // TODO: This is actually work around. Long term GH #557 needs to be implemented
```
dewyatt: I can't see how this TODO makes sense at the moment. Setting the rng pointer doesn't seem like any more of a work around than setting `.password_provider` and similar fields below.

@ni4 what do you think, am I missing something? Maybe the parse handler should have an explicit `rng` field or something? (because it looks like `pgp_parse_handler_t->ctx` isn't actually used in master?)

ni4: @dewyatt Thanks for pointing at this. Actually, being part of rnp_ctx_t is pretty enough. ctx is not used since it currently contains only options to produce data, not to parse. 
And, actually, rng logically is not needed for data decryption/signature verification - botan just use it for secert key validation. I don't think that key validation for every operation is correct, probably it would be better to move it to key provider, or other place where keys are loaded/decrypted.

ni4:
@flowher Please do not forget to squash/reorganize commits to few ones.

src/lib/rnp2_internal.h
flowher: Agreed. I've forgotten to do it

src/lib/crypto/rng.h
flowher: No. That's not needed. It's internal structure done in a way to be easily included

src/lib/rnp2.c
flowher: I plan to add rnp_ctx_init_ffi .

src/lib/crypto/rng.h
dewyatt: I guess I didn't look at the internal fields much, but I see now they're all simple types, so indeed I would say it doesn't matter.

flowher:
Don't merge yet

dewyatt:
@ni4 you can always squash merge it yourself in the GH web interface too ([like so](https://help.github.com/assets/images/help/pull_requests/select-squash-and-merge-from-drop-down-menu.png))

flowher:
Commits are OK, I've been working to have clear history. Please follow dev guide and don't squash anything

dewyatt:
The dev guide only says prefer merge over squash. In this case, not all commits are buildable+testable, so IMO it's up to the final reviewer.

flowher:
All commits are buildable. I've tested most of them so I think it's OK.

ronaldtse:
Feel free to merge it!

dewyatt:
@flowher Maybe buildable, not testable. Oh well.

PR #566, opened 2017-12-08
Update developers guide with some handy clang-format commands.

ni4:


dewyatt:
It may be good to make the regex a little stricter, otherwise this would currently match files like `.clang-format`, for example.

Something like `grep ".\(\.c\|\.h\)\$"`? Probably room for improvement there, I don't use the BRE syntax much.

ni4:
Thanks, @dewyatt and @ronaldtse 
Adding '\$' definitely makes sense, updated PR. And merging.

PR #567, opened 2017-12-10
Cleanup in bn.c - automatic changes to cleanup interface

### LINKED ISSUE #543: Remove madness related to redefining BN_ functions
## Description

BN_bin2bn() is defined as PGPV_BN_bin2bn which wraps calls to botan... That's a heritage related to usage OpenSSL in old times.

Cleanup needed. We don't need all those #defines

flowher:
* Changes PGPV_BN_xxxx to bn_xxxx
* Removes not needed defines
* Changes structure called BIGNUM to bignum_t
* All done with ``sed``
 
Closes #543

### DIFF
--- a/src/lib/crypto/bn.c
+++ b/src/lib/crypto/bn.c
@@ -265,7 +265,7 @@ bn_is_zero(const bignum_t *n)
 }
 
 int
-bn_set_word(bignum_t *a, PGPV_BN_ULONG w)
+bn_set_word(bignum_t *a, uint32_t w)
 {
     if (a == NULL) {
         return -1;
--- a/src/lib/crypto/bn.c
+++ b/src/lib/crypto/bn.c
@@ -44,13 +44,13 @@
 /* these wrappers also check the arguments passed for sanity */
 
 PGPV_BIGNUM *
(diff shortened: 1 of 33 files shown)


src/lib/crypto/bn.c
```diff
@@ -221,15 +221,15 @@ bn_bn2hex(const PGPV_BIGNUM *a)
     return NULL;
 }
 
-/* hash a bignum, possibly padded - first length, then string itself */
+/* hash a bignum_t, possibly padded - first length, then string itself */
 size_t
-bn_hash(const PGPV_BIGNUM *bignum, pgp_hash_t *hash)
+bn_hash(const bignum_t *bignum_t, pgp_hash_t *hash)
```
dewyatt: I don't think this was intentional.

dewyatt:
> All done with sed

This tends to cause problems, unfortunately. I only saw one so far. I use vim but I usually fire up Eclipse for this sort of thing, since it actually creates an AST and is smart about things.

(Clang 6 has `clang-rename` too, though I haven't used it and it's pretty new.)

src/lib/crypto/bn.c
flowher: Thnx. Wasn't. closing this as automatic sed wasn't succcessful

PR #569, opened 2017-12-12
Move character/string constants to the separate file

ni4:
These constants will be needed in cleartext streamed signing. Also there were some places for the same double-defined constants.
I didn't touch old code (like reader.c, writer.c) as it will be removed once streamed signing is finished.

### DIFF
--- a/src/librepgp/stream-armor.c
+++ b/src/librepgp/stream-armor.c
@@ -25,6 +25,7 @@
  */
 
 #include "config.h"
+#include "stream-def.h"
 #include "stream-armor.h"
 #include "stream-packet.h"
 #include <stdlib.h>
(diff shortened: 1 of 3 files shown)


src/librepgp/stream-armor.c
```diff
@@ -908,40 +904,37 @@ init_armored_dst(pgp_dest_t *dst, pgp_dest_t *writedst, pgp_armored_msg_t msgtyp
...
 
     read = src_peek(src, buf, sizeof(buf));
-    if (read < (ssize_t) sizeof(armor_start)) {
+    if (read < (ssize_t) strlen(ST_ARMOR_BEGIN) + 1) {
         return false;
     }
 
     buf[read - 1] = 0;
-    return !!strstr((char *) buf, armor_start);
+    return !!strstr((char *) buf, ST_ARMOR_BEGIN);
```
dewyatt: We don't need the double logical not (`!!`) when returning a C99 bool, but it's not a big deal.

PR #575, opened 2017-12-20
Remove obsolete code superseded with streaming processing

ni4:
This PR cleanups codebase from old code related to encryption/decryption/signing/verification.
On the rnp level it introduces the single `rnp_protect_file()`/`rnp_protect_mem()` call which later detects the operation by signers/recipients in `rnp_ctx_t` and calls corresponding streaming method. Actually, later on rnp_sign_src/rnp_encrypt_src could be merged to single rnp_protect_src call as well.

A lot of old code is still left because it is used in packet dumper and keyring reading/writing code.


src/tests/rnp_tests.c
```diff
@@ -171,7 +171,6 @@ main(int argc, char *argv[])
       cmocka_unit_test(test_load_keyring_and_count_pgp),
       cmocka_unit_test(test_load_check_bitfields_and_times),
       cmocka_unit_test(test_load_check_bitfields_and_times_v3),
-      cmocka_unit_test(pgp_compress_roundtrip),
```
flowher: Do we have some other test which is testing compression? Or not needed anymore?

flowher:
> A lot of old code is still left because it is used in packet dumper and keyring reading/writing code.
It could use "streaming" code instead, isn't ? I propose to open a ticket for that

src/tests/rnp_tests.c
ni4: Yeah, CLI tests use compression, so it is tested as well.

ni4:
> It could use "streaming" code instead, isn't ? I propose to open a ticket for that
Yeah, It would be quite logical to move keys/keyrings loading and saving code to streams as well.
And then add packet dumping functionality.
I added issues #576 and #577 and will work on them after implementation of AEAD.

Issue #576, opened 2017-12-22
Move keyring and keys loading/saving code to streams


ni4:
## Description

Currently we use streamed code for data processing, however old code is still used for keyrings and keys parsing and saving. It should be moved to streamed approach/interface as well, so new armored and packet parsing code will be reused.

PR #580, opened 2018-01-18
Header cleanups

randombit:
Move declarations of functions defined in X.c to X.h instead of some_random.h, and remove unneeded includes esp signature.h

I am concerned about signature.h/.c - it seems like the majority of signature code is now in src/librepgp/stream-write.c, but `signature.c` still has all the same functionality duplicated. This confused the heck out of me when I was trying to debug something in FFI, I could see my printfs in the RSA code firing but not in signature.c - turns out these functions are now only used for subkey creation. Should these operations be reimplemented using the functions in stream-write?


flowher:
@randombit Indeed duplication is bad. We have #576 which should address this issue. Code in signature.c is used by (for example) rnpkeys, which shouldn't be the case soon.

ronaldtse:
Thank you @randombit @ni4 @flowher -- very happy to see things rolling again 👍

PR #594, opened 2018-01-24
Random fixes

flowher:
Found during DSA development

### DIFF
--- a/src/lib/crypto/ec.c
+++ b/src/lib/crypto/ec.c
@@ -204,18 +204,11 @@ pgp_genkey_ec_uncompressed(rng_t *                rng,
     ret = RNP_SUCCESS;
 
 end:
-    if (pr_key != NULL) {
-        botan_privkey_destroy(pr_key);
-    }
-    if (pu_key != NULL) {
-        botan_pubkey_destroy(pu_key);
-    }
-    if (public_x != NULL) {
-        bn_free(public_x);
-    }
-    if (public_y != NULL) {
-        bn_free(public_y);
-    }
+    botan_privkey_destroy(pr_key);
+    botan_pubkey_destroy(pu_key);
+    bn_free(public_x);
+    bn_free(public_y);
+
     if (RNP_SUCCESS != ret) {
         RNP_LOG("EC key generation failed");
         pgp_seckey_free(seckey);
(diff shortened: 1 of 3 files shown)


src/lib/crypto/ecdsa.c
```diff
@@ -79,14 +79,9 @@ pgp_ecdsa_sign_hash(rng_t *                 rng,
         goto end;
     }
 
-    // Allocate memory and copy results
+    // Allocate memory and copy results. bn_bin2bn won't return NULL
     sign->r = bn_bin2bn(out_buf, curve_order, sign->r);
     sign->s = bn_bin2bn(out_buf + curve_order, curve_order, sign->s);
-    if (!sign->r || !sign->s) {
```
dewyatt: I don't get the point of removing these checks, it can certainly return NULL.

flowher: Only if sign->r is NULL. But it this is the case then botan_pk_op_sign_finish() will fail.

dewyatt: I mean that `bn_bin2bn` can fail if memory allocation fails.

flowher: I think in this case there is no memory allocation done (out_buf != NULL && sign->r!=NULL)

dewyatt: `botan_mp_from_bin` can throw `std::bad_alloc` here https://github.com/randombit/botan/blob/8f17d65056903ca4da9bb16a85db717e4bea4456/src/lib/math/bigint/bigint.cpp#L284, in which case `bn_bin2bn` would return NULL.

flowher: If exception is thrown then FFI returns BOTAN_FFI_ERROR_EXCEPTION_THROWN, so this check wouldn't work anyway.

flowher: In such case botan returns BOTAN_FFI_ERROR_EXCEPTION_THROWN, isn't?

dewyatt: Yes, and then `bn_bin2bn` would return NULL here: https://github.com/riboseinc/rnp/blob/33fc010a2e2ff4cf08a8eec92159ad9d004d7352/src/lib/crypto/bn.c#L60

dewyatt: On a side note, I think it's bad practice to ignore return values from botan's FFI. Just because the current implementation of a function can't fail (not the case here), doesn't mean it won't evolve and introduce some failure points later on.

flowher: argh.. I'm wrong. sign->r always == NULL, that's even checked above.

ronaldtse:
Thanks @flowher !

Issue #596, opened 2018-01-25
Broken build in master due to `cli_tests.py`


ronaldtse:
## Description

```
+env RNPC_GPG_PATH=/home/travis/local-builds/gpg21-install/bin/gpg python2 cli_tests.py -w -v
Running in /tmp/rnpctmpdTXfbM
test_rnp_compression (__main__.Compression) ... ok
test_file_encryption__gpg_to_rnp (__main__.Encryption) ... ok
test_file_encryption__rnp_to_gpg (__main__.Encryption) ... ok
test_sym_encryption__gpg_to_rnp (__main__.Encryption) ... ok
test_sym_encryption__rnp_to_gpg (__main__.Encryption) ... ok
test_generate_4096_rsa_key (__main__.Keystore) ... ok
test_generate_default_rsa_key (__main__.Keystore) ... ok
test_generate_key_with_gpg_import_to_rnp (__main__.Keystore) ... gpg key generation failed
gpg key generation failed
ERROR
test_generate_multiple_rsa_key__check_if_available (__main__.Keystore) ... ok
test_generate_with_rnp_import_to_gpg (__main__.Keystore) ... ok
test_armor (__main__.Misc) ... ok
test_encryption_no_mdc (__main__.Misc) ... ok
test_encryption_s2k (__main__.Misc) ... ok
test_gpg_to_rnp_default_key (__main__.SignDefault) ... ok
test_rnp_to_gpg_default_key (__main__.SignDefault) ... ok
test_sign_P256 (__main__.SignECDSA) ... ok
test_sign_P384 (__main__.SignECDSA) ... ok
test_sign_P521 (__main__.SignECDSA) ... ok
test_verify_P256 (__main__.SignECDSA) ... ok
test_verify_P384 (__main__.SignECDSA) ... ok
test_verify_P521 (__main__.SignECDSA) ... ok
======================================================================
ERROR: test_generate_key_with_gpg_import_to_rnp (__main__.Keystore)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "cli_tests.py", line 684, in test_generate_key_with_gpg_import_to_rnp
    raise_err('gpg key generation failed', err)
  File "/home/travis/build/riboseinc/rnp/src/tests/cli_common.py", line 22, in raise_err
    raise CLIError(msg, log)
CLIError: None
----------------------------------------------------------------------
Ran 21 tests in 343.512s
FAILED (errors=1)
```

ni4:
Please see #550 
[...]

PR #597, opened 2018-01-25
not-fix for broken CLI tests

flowher:
How about skipping broken tests for now?
Feel free to close this PR in issue can be fixed quickly  ( #596)


ni4:
@flowher I just added PR #599 which should fix the issue

flowher:
@ni4 feel free to reject if this is fixed

PR #602, opened 2018-01-26
Add key to JSON conversion in FFI

dewyatt:
Long-term I'm not sure json is a good way to handle this piece, but I added this initially because the alternative that I could think of (adding like 30 or so accessor functions w/tests) seemed like it would take too long. So it wouldn't be a bad idea to revisit this at some point to try to provide a more traditional way of getting all this extra information.

One of the flags here (not enabled by default) allows secret MPIs to be in the json results, so an `rnp_scrub_memory` or similar in the FFI would be good (or maybe have `rnp_buffer_free` do this automatically for all buffers?).

For an idea of what this json looks like: [primary](https://gist.github.com/dewyatt/2c778699899bed56843aa81c12ce54ee) and [sub](https://gist.github.com/dewyatt/b8ccf4dcaee85c7d59a12b5a7ecf5de8).


dewyatt:
Oh I forgot to add the test data here, one sec...

ronaldtse:
@dewyatt interesting! Have you seen [JWS](https://tools.ietf.org/html/rfc7515) and [JWE](https://tools.ietf.org/search/rfc7516)?

dewyatt:
@ronaldtse Those are interesting, I think I may have glanced at one of them before, and I've used some related stuff like JWT.

ronaldtse:
The reason I ask is that the JW* series defines a JSON public/private/symmetric key format that we might want to serialize into rather than maintaining our own JSON format. I wonder if using this format would be beneficial, in the future, if there is use for a JSON OpenPGP format rather than the binary OpenPGP data format.

dewyatt:
I follow you and I think it's definitely worth looking at, much better than maintaining our own random format.

dewyatt:
Rebased to resolve conflicts.

ronaldtse:
Would the second approver please merge this? Thanks!

randombit:
LGTM merging

PR #604, opened 2018-01-27
Add identifier iterator to FFI

dewyatt:
#602 should be merged first. Just the last commit is relevant here.

This is just a way of iterating through the unique identifiers. Iterating the grips would be one way of iterating the keys, for example. There's no support for duplicate identifiers - like the same userid in different keys - though you could iterate by grip and figure it out from there.

One oddity you may note is that this uses json-c's `json_object` as a way of maintaining a unique set. Originally I used linkhash from json-c, but they recently [clarified](https://github.com/json-c/json-c/commit/f2f103b98663d63ff8a9f60b2e1432f78463f486#diff-f2eed63a55689488f16a5a82015d558f) that linkhash is not meant for public use. In the future we may benefit from pulling in a hash table impl (I'd personally like to avoid the POSIX ones and the ones that are primarily preprocessor-based if possible), since there are probably other places where it would be useful.


dewyatt:
Ah I forgot json-c 0.13 changed `json_object_object_add`. I think we're targeting 0.12 so I'll make a small change here.

PR #606, opened 2018-01-29
Add signatures and verification to FFI

randombit:
(Originally was on `ffi` branch, I manually picked it back to master so it can be merged as an independent PR)


include/rnp/rnp2.h
```diff
@@ -260,6 +264,54 @@ rnp_result_t rnp_key_have_public(rnp_key_handle_t key, bool *result);
 
 /* TODO define functions for password-based encryption */
 
+rnp_result_t rnp_op_sign_create(rnp_op_sign_t *op,
+                                rnp_ffi_t      ffi,
+                                rnp_input_t    input,
+                                rnp_output_t   output);
+
+rnp_result_t rnp_op_sign_set_signing_key(rnp_op_sign_t op, rnp_key_handle_t key);
```
dewyatt: I think this should be named `rnp_op_sign_add_signer` to align with `rnp_op_encrypt_add_recipient`.

include/rnp/rnp2.h
```diff
@@ -260,6 +264,54 @@ rnp_result_t rnp_key_have_public(rnp_key_handle_t key, bool *result);
...
+
+rnp_result_t rnp_op_sign_set_compression(rnp_op_sign_t op, const char *compression, int level);
+rnp_result_t rnp_op_sign_set_armor(rnp_op_sign_t op, bool armored, bool cleansign);
```
dewyatt: typo: cleansign -> clearsign

include/rnp/rnp2.h
```diff
@@ -260,6 +264,54 @@ rnp_result_t rnp_key_have_public(rnp_key_handle_t key, bool *result);
...
+rnp_result_t rnp_op_sign_set_detached(rnp_op_sign_t op, bool detached);
+
+rnp_result_t rnp_op_sign_set_timestamps(rnp_op_sign_t op,
+                                        uint32_t      creation_time,
+                                        uint32_t      expiration_time);
+rnp_result_t rnp_op_sign_set_hash_fn(rnp_op_sign_t op, const char *hash);
```
dewyatt: I think naming this `rnp_op_sign_set_hash` would align better with the json key generation and key listing stuff.

include/rnp/rnp2.h
```diff
@@ -260,6 +264,54 @@ rnp_result_t rnp_key_have_public(rnp_key_handle_t key, bool *result);
...
+rnp_result_t rnp_op_sign_set_file_name(rnp_op_sign_t op, const char *filename);
+rnp_result_t rnp_op_sign_set_file_mtime(rnp_op_sign_t op, uint32_t mtime);
+
+rnp_result_t rnp_op_sign_execute(rnp_op_sign_t op);
+rnp_result_t rnp_op_sign_destroy(rnp_op_sign_t op);
+
+rnp_result_t rnp_op_verify(rnp_op_verify_result_t *result,
```
dewyatt: The verify API isn't really consistent with the create/execute/destroy convention that most of the other ops in here follow.

src/lib/rnp2.c
```diff
@@ -1207,11 +1264,365 @@ rnp_op_encrypt_destroy(rnp_op_encrypt_t op)
...
+    if (!op || !key)
+        return RNP_ERROR_NULL_POINTER;
+
+    char *keyid = NULL;
+
+    rnp_result_t res = rnp_key_get_keyid(key, &keyid);
+    if (res != RNP_SUCCESS)
+        return res;
+
+    list_append(&op->rnpctx.signers, keyid, strlen(keyid) + 1);
```
dewyatt: Ideally check for out-of-mem here.

include/rnp/rnp2.h
ni4: Agree, it will also go well with multiple signers.

include/rnp/rnp2.h
ni4: I would vote for separation between armoring and cleartext signing. First is applicable for both encryption and signing of binary/text data, while cleartext signing is applicable only for signing of text documents.

include/rnp/rnp2.h
```diff
@@ -260,6 +264,54 @@ rnp_result_t rnp_key_have_public(rnp_key_handle_t key, bool *result);
...
+
+rnp_result_t rnp_op_sign_set_signing_key(rnp_op_sign_t op, rnp_key_handle_t key);
+
+rnp_result_t rnp_op_sign_set_compression(rnp_op_sign_t op, const char *compression, int level);
+rnp_result_t rnp_op_sign_set_armor(rnp_op_sign_t op, bool armored, bool cleansign);
+rnp_result_t rnp_op_sign_set_detached(rnp_op_sign_t op, bool detached);
+
+rnp_result_t rnp_op_sign_set_timestamps(rnp_op_sign_t op,
+                                        uint32_t      creation_time,
+                                        uint32_t      expiration_time);
```
ni4: From these two only creation_time is timestamp, expiration time internally is counted in seconds. Or do you convert it from timestamp?

include/rnp/rnp2.h
ronaldtse: I agree with @ni4 here.

ronaldtse:
@randombit could we update the code here a bit and get this merged? Thanks!

include/rnp/rnp2.h
randombit: Fixed

include/rnp/rnp2.h
randombit: Fixed - now 2 distinct functions for these

include/rnp/rnp2.h
randombit: Included hash function as an arg to `rnp_op_sign_add_signer` instead since it's totally reasonable when multi singing to use different hashes for different keys (eg SHA-512 for RSA + SHA-256 for P-256 ECDSA).

src/lib/rnp2.c
randombit: Fixed

include/rnp/rnp2.h
randombit: That is just seconds until expiration, parameter name changed.

include/rnp/rnp2.h
randombit: True, but verify needs to return potentially a lot of different status information (verification info, hash fn used, etc - for each signature) so returning an array of verify struct seemed the best way to handle it. I am open to doing something else but don't see a clear approach.

randombit:
Rebased + addressed review comments except for @dewyatt concern about the verify API.

@dewyatt @ni4 ready for another look.

include/rnp/rnp2.h
```diff
@@ -261,6 +265,57 @@ rnp_result_t rnp_key_have_public(rnp_key_handle_t key, bool *result);
...
+rnp_result_t rnp_op_sign_add_signer(
+  rnp_op_sign_t    op,
+  rnp_key_handle_t key,
+  const char *     hash_fn,
+  uint32_t         creation_time, /* seconds since Jan 1 1970 UTC */
+  uint32_t         expiration_seconds);
+
+rnp_result_t rnp_op_sign_set_compression(rnp_op_sign_t op, const char *compression, int level);
+rnp_result_t rnp_op_sign_set_armor(rnp_op_sign_t op, bool armored);
+rnp_result_t rnp_op_sign_set_clearsign(rnp_op_sign_t op, bool clearsign);
```
dewyatt: @ni4 is this what you had in mind here, or were you thinking a separate op (like `rnp_op_clearsign_*`) is best?
I can see an argument for having a separate op since clearsign+not-armored isn't a valid combo IIRC. Thoughts?

ni4: Yeah, that's it - have clearsign/armor as separate functions. 
Creating another set of rnp_op_clearsign_* functions would be too redundant. 
Actually, once clearsign is set to true armoring flag doesn't make sense at all, so could be ignored (or asserted to report misuse).

include/rnp/rnp2.h
ronaldtse: @dewyatt do you have in mind an alternative what verify should return -- or do we want to break it down further?

include/rnp/rnp2.h
ronaldtse: I agree. @randombit what do you think?

include/rnp/rnp2.h
dewyatt: Mostly I was just thinking that if it has `rnp_op` in the name, it should follow the same create/execute/destroy pattern that all the other `rnp_op` groups do.

dewyatt:
@ni4 When you have time, would you care to take ownership of this PR and try to resolve any remaining items so we can keep this moving?

ronaldtse:
Agree with @dewyatt . Thanks @ni4 !

ni4:
@dewyatt @ronaldtse Ok, sure!

include/rnp/rnp2.h
ni4: I would also stick to: 
- `rnp_op_verify_create(input, output)`
- `rnp_op_verify_detached_create(input, signature)`
- `rnp_op_verify_execute()`
The reason for this is difference of purpose and main parameters of detached and normal verification.
@dewyatt @ronaldtse What do you think?

dewyatt: @ni4 Yep I think that makes sense. I guess similar changes would make sense on the sign op but that's up to you.

randombit:
@ni4 I have some time available to work on this this week, what issues remain?

include/rnp/rnp2.h
ni4: @randombit Here is the main issue in this PR. What do you think about splitting methods in this way?

ni4:
@randombit Will you have time to look into this? Please see this comment https://github.com/riboseinc/rnp/pull/606#discussion_r172279074

ni4:
@randombit Nevermind, I finally got to this task. Will update PR later today or tomorrow.

ronaldtse:
That's wonderful @ni4 -- look forward to completing this 👍

ni4:
@ronaldtse @dewyatt @randombit 
Please see the updated PR.
Changes: 
- separated `rnp_op_sign_create` into `rnp_op_sign_create`, `rnp_op_sign_detached_create`, `rnp_op_sign_cleartext_create`
- made verify call to follow the same logic: `rnp_op_verify_create`, `rnp_op_verify_detached_create` and `rnp_op_verify_execute`
- changed the way verify call return signatures (get_count/get_at instead of bulk pointer return)
- changed the way signer is added (removed extra parameters, since right now there is no way to set hash algorithm/creation/expiration on per-signature basis, added issue #651)

ronaldtse:
@ni4 the approach looks reasonable to me. Thoughts @dewyatt @henrydcase ?

src/lib/rnp2.c
```diff
@@ -97,6 +98,34 @@ struct rnp_output_st {
...
+struct rnp_op_verify_st {
+    rnp_ffi_t    ffi;
+    rnp_input_t  input;
+    rnp_input_t  detached_input; /* for detached signature will be source file/data */
+    rnp_output_t output;
+    rnp_ctx_t    rnpctx;
+    /* these fields are filled after operation execution */
+    struct rnp_op_verify_signature_st *signatures;
+    size_t                             signature_count;
+    char                               filename[128];
```
dewyatt: Shouldn't this be dynamically allocated and not arbitrarily limited?

dewyatt:
> changed the way signer is added (removed extra parameters, since right now there is no way to set hash algorithm/creation/expiration on per-signature basis, added issue #651)

I don't know if I agree with this - I added those parameters since these things should eventually be supported. We can always return a "not implemented" error when something isn't supported internally (and add a note to the docs).

I think it's important to remember that this API is supposed to be RNP2 -- I don't think we should design it around the limitations of the current code base.

I think it makes sense for the API to support multiple signers with varying expiration times (but maybe not creation times), and multiple signers with varying hash functions, etc.

include/rnp/rnp2.h
```diff
@@ -277,10 +282,71 @@ rnp_result_t rnp_key_have_public(rnp_key_handle_t key, bool *result);
...
+
+rnp_result_t rnp_op_sign_detached_create(rnp_op_sign_t *op,
+                                         rnp_ffi_t      ffi,
+                                         rnp_input_t    input,
+                                         rnp_output_t   signature);
+
+rnp_result_t rnp_op_sign_add_signer(rnp_op_sign_t op, rnp_key_handle_t key);
+rnp_result_t rnp_op_sign_set_compression(rnp_op_sign_t op, const char *compression, int level);
+rnp_result_t rnp_op_sign_set_armor(rnp_op_sign_t op, bool armored);
+rnp_result_t rnp_op_sign_set_hash_fn(rnp_op_sign_t op, const char *hash);
```
dewyatt: Kind of a nitpick but we don't refer to this as `hash_fn` anywhere else in the code base, so I think we should be consistent and use `hash`.

(and if we ever need to disambiguate between a hash function and its output, we can probably use "digest" or "hash output" or something)

src/lib/rnp2.c
```diff
@@ -1223,11 +1288,451 @@ rnp_op_encrypt_destroy(rnp_op_encrypt_t op)
...
+{
+    if (!sig) {
+        return RNP_ERROR_NULL_POINTER;
+    }
+
+    return sig->verify_status;
+}
+
+rnp_result_t
+rnp_op_verify_signature_get_hash_fn(rnp_op_verify_signature_t sig,
```
dewyatt: This is a bit inconsistent on memory handling. Most of the other functions will return allocated memory which must be freed with `rnp_buffer_free`

We have also talked about avoiding in/out params before. I know botan uses this convention but @ronaldtse has [mentioned](https://github.com/riboseinc/rnp/pull/319#issuecomment-316705575) that it might be good to avoid this (I think I dropped the ball on adding it to the dev guide though).

src/lib/rnp2.c
```diff
@@ -97,6 +98,34 @@ struct rnp_output_st {
...
+struct rnp_op_sign_st {
+    rnp_ffi_t    ffi;
+    rnp_input_t  input;
+    rnp_output_t output;
+    rnp_ctx_t    rnpctx;
+};
+
+struct rnp_op_verify_signature_st {
+    uint8_t      keyid[PGP_KEY_ID_SIZE];
+    char         hash_fn[16];
```
dewyatt: Can you not use a `pgp_hash_alg_t` here? (mostly don't like seeing the arbitrary size)

include/rnp/rnp2.h
ni4: Agree, thanks. Just copied the notation from previous edits.

src/lib/rnp2.c
ni4: Ok, agree.

src/lib/rnp2.c
ni4: Sure, done.

ni4:
@dewyatt I addressed your review comments, please see the updated PR.
Also added some documentation.

> I don't know if I agree with this - I added those parameters since these things should eventually be supported. We can always return a "not implemented" error when something isn't supported internally (and add a note to the docs).

What about renaming this to add_signature, and adding some opaque signature type?
Pointer to it will be returned in _add_signer(operation, key), and then have a possibility to call _set_hash(), _set_expiration_time() and so on on it? Now those will return _NOT_IMPLEMENTED, but later we may even more extend it with setting the signature type, adding subpackets, and so on so on.

ni4:
@dewyatt 
> Any chance you can do a quick mockup of a header so we can get a clearer picture?
It should not take much time, will update PR soon

ni4:
@dewyatt Please see the updated PR. If it is good to go then I will remove rnp_op_add_signer and add implementations for other functions. 
Within current implementation (single hash alg/create/expire for all signatures) we can set corresponding 'global' parameters if signature parameter for `rnp_op_sign_signature_set_something` is NULL.

dewyatt:
@ni4 It looks good I think, more flexible than before since it will allow easier expansion in the future with different subpackets, etc.

ni4:
@dewyatt @ronaldtse @henrydcase Now this should be finnally done. Do you have any further suggestions/comments?

PR #607, opened 2018-01-29
Fixes for ECDH encryption and decryption

### LINKED ISSUE #598: ECDH - regression when testing against GnuPG
## Description

I'm adding tests for ECDH against GnuPG and found out that we actually have regression.

```
/usr/local/bin/gpg --homedir /tmp/rnpctmpcbwAqZ/.gpg --yes --passphrase password --pinentry-mode loopback -o /tmp/rnpctmpcbwAqZ/encrypt1.dec --decrypt /tmp/rnpctmpcbwAqZ/encrypt1.enc
gpg: NOTE: THIS IS A DEVELOPMENT VERSION!
gpg: It is only intended for test purposes and should NOT be
gpg: used in a production environment or with production keys!
gpg: ecdh failed in gcry_cipher_decrypt: Checksum error
gpg: encrypted with 256-bit ECDH key, ID DCAD9A0766F02CD6, created 2018-01-25
      "Test Testovich <encrypttest@secure.gov>"
**gpg: public key decryption failed: Checksum error**
gpg: decryption failed: No secret key
```
~Not sure yet why, but I'm on it (something related to aes-wrap)~

henrydcase:
* According to point RFC 4880bis04 (13.5) size of symmetric encryption key
      can be padded to 40 bytes (obfuscated) or padded to 8 nearest byte granularity.
      When encrypting, our implementation was padding the key to 40 bytes,
      but this is incompatible with GnuPG implementation (2.3.0-beta190). Also
      it is questionable how much security such solution bring. Also there is
      a bug in the RFC which states that PKCS5 padding must be used when
      padding symmetric key before it gets wrapped with AES-wrap. This makes
      it impossible to obfuscate size of the key to 40 bytes.
      Hence, I've change our implementation to pad symmetric key to nearest
      8 byte granularity.
      For more details see: https://dev.gnupg.org/T3763
    
    * Some cleanup and removal of unnecessary checking against NULL was introduced

    * When computing KEK, the KDF parameters (other_info) were wrongly supplied
      as salt to key derivation function `botan_pk_op_key_agreement'.
      It needs to be passed to KDF as a label.

Tests in #607
Closes #598
Requires #609

### DIFF
--- a/src/lib/crypto/ecdh.c
+++ b/src/lib/crypto/ecdh.c
@@ -146,37 +146,30 @@ compute_kek(uint8_t *              kek,
 {
     botan_pk_op_ka_t op_key_agreement = NULL;
     uint8_t          point_bytes[MAX_CURVE_BYTELEN * 2 + 1] = {0};
-    bool             ret = true;
+    bool             ret = false;
+    char             kdf_name[32] = {0};
+    size_t           num_bytes = 0;
+    uint8_t          s[MAX_CURVE_BYTELEN * 2 + 1] = {0};
+    size_t           s_len = sizeof(s);
 
-    // Name of Botan's KDF and backing hash algorithm
-    const char *hash_botan_name = pgp_hash_name_botan(hash_alg);
-    if (!hash_botan_name) {
+    if (botan_mp_to_bin(ec_pubkey, point_bytes) ||
+        botan_mp_num_bytes(ec_pubkey, &num_bytes)) {
         return false;
     }
 
-    char botan_kdf_name[64] = {0};
-    snprintf(botan_kdf_name, 64, "SP800-56A(%s)", hash_botan_name);
-
-    if (botan_mp_to_bin(ec_pubkey, point_bytes) < 0) {
-        return false;
-    }
-
-    size_t num_bytes = 0;
-    if (botan_mp_num_bytes(ec_pubkey, &num_bytes) < 0) {
-        return false;
-    }
-
-    if (botan_pk_op_key_agreement_create(&op_key_agreement, ec_prvkey, botan_kdf_name, 0) ||
+    if (botan_pk_op_key_agreement_create(&op_key_agreement, ec_prvkey, "Raw", 0) ||
         botan_pk_op_key_agreement(op_key_agreement,
-                                  kek,
-                                  &kek_len,
+                                  s,
+                                  &s_len,
                                   point_bytes,
                                   num_bytes,
-                                  other_info,
-                                  other_info_size)) {
-        ret = false;
+                                  NULL, //other_info,
+                                  0)) { //other_info_size)) {
         goto end;
     }
+
+    snprintf(kdf_name, sizeof(kdf_name), "SP800-56A(%s)",  pgp_hash_name_botan(hash_alg));
+    ret = !botan_kdf(kdf_name, kek, kek_len, s, s_len, NULL, 0, other_info, other_info_size);
 
 end:
     ret &= !botan_pk_op_key_agreement_destroy(op_key_agreement);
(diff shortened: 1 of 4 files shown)


ronaldtse:
The failure in Travis is again GnuPG key generation...

```

ERROR: test_generate_key_with_gpg_import_to_rnp (__main__.Keystore)
  File "cli_tests.py", line 705, in test_generate_key_with_gpg_import_to_rnp
    raise_err('gpg key generation failed, error ' + str(ret) , err)
    raise CLIError(msg, log)
CLIError: None
```

flowher:
@ronaldtse It's probably because travis has run before PR #605 was integrated. I'll this build few times today and see if it breaks

flowher:
RUN1: OK
RUN2: FAILS in following way:
```

test_verify_P2048_Q256 (__main__.SignDSA) ... /home/travis/local-builds/gpg21-install/bin/gpg --gen-key --expert --batch --pinentry-mode loopback --homedir /tmp/rnpctmpyJgPTV/.gpg --yes --passphrase password
gpg: keybox '/tmp/rnpctmpyJgPTV/.gpg/pubring.kbx' created
gpg: WARNING: some OpenPGP programs can't handle a DSA key with this digest size
gpg: can't connect to the agent: IPC connect call failed
gpg: agent_genkey failed: No agent running
gpg: key generation failed: No agent running

FAIL
kill -9 gpg-agent
kill: failed to parse argument: 'gpg-agent'

gpg-connect-agent reloadagent /bye
gpg-connect-agent: can't connect to the agent: IPC connect call failed
```

dewyatt:
I guess we missed this in review of 4170eb9940e1d57b65ffb9dbe8b79e39af8ab9b3, maybe killall was meant there, obviously kill wants a PID. 

The workaround suggested with gpgconf does a wait on the process, so that sounds good to me.

flowher:
GPG update testing here #27

ni4:
@dewyatt @flowher Yeah, that should be my fault, was a bit in hurry and mislooked the problem, thought it was fixed.

flowher:
@ni4 no worries - it's good stuff anyway

dewyatt:
Rebased to get CI passing. I'll try to take a closer look tomorrow but ideally would prefer to see an approval from @randombit before merge.

flowher:
Last CI run was on rebased version

ronaldtse:
@randombit @dewyatt @ni4 could we have an approval to merge this? Thanks!

src/lib/crypto/ecdh.c
```diff
@@ -225,14 +221,15 @@ pgp_ecdh_encrypt_pkcs5(rng_t *                  rng,
...
     }
 
-    if (*wrapped_key_len < ECDH_WRAPPED_KEY_SIZE) {
+    // +8 because of AES-wrap adds 8 bytes
+    if (*wrapped_key_len < (m_padded_len + 8)) {
         return RNP_ERROR_SHORT_BUFFER;
     }
 
     // See 13.5 of RFC 4880 for definition of other_info_size
     const size_t other_info_size = (pubkey->ec.curve == PGP_CURVE_NIST_P_256) ? 54 : 51;
```
ni4: This was not changed, however RFC 4880bis defines this size as `The size of the KDF parameters sequence, defined above, is either 54 for the NIST curve P-256, 51 for the curves P-384 and P-521, or 56 for Curve25519.`.
Or we are not supposed to support ECDH with 25519 yet?

henrydcase: Nop, that's part of RFC. RFC references RFC3394 

``   Compute C = AESKeyWrap( Z, m ) as per [RFC3394] ``
And that's what aes-3394 does
``
2.2.1 Key Wrap
...
   Outputs:     Ciphertext, (n+1) 64-bit values {C0, C1, ..., Cn}.
``

henrydcase: We don't support djb's curve yet

dewyatt:
@henrydcase Per the dev guide, PRs are to be merged by a reviewer, not the submitter/author.

PR #609, opened 2018-01-30
CI stabilization

flowher:
*    (tests) Clear keyrings only when finishing tests suite
*    (tests) Use gpgconf instead of kill -9
*    (tests) Update GPG to 2.2.4

### DIFF
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -89,7 +89,7 @@ if [ ! -e "${GPG21_INSTALL}/bin/gpg" ]; then
     cd ..
   done
 
-  for archive in libgcrypt:1.8.0 libassuan:2.4.3 libksba:1.3.5; do
+  for archive in libgcrypt:1.8.2 libassuan:2.5.1 libksba:1.3.5; do
     pkgname="${archive%:*}"
     version="${archive#*:}"
 
@@ -104,11 +104,11 @@ if [ ! -e "${GPG21_INSTALL}/bin/gpg" ]; then
     cd ..
   done
 
-  wget -c https://www.gnupg.org/ftp/gcrypt/pinentry/pinentry-1.0.0.tar.bz2
-  wget -c https://www.gnupg.org/ftp/gcrypt/pinentry/pinentry-1.0.0.tar.bz2.sig
-  gpg --verify pinentry-1.0.0.tar.bz2.sig
-  tar -xjf pinentry-1.0.0.tar.bz2
-  cd pinentry-1.0.0
+  wget -c https://www.gnupg.org/ftp/gcrypt/pinentry/pinentry-1.1.0.tar.bz2
+  wget -c https://www.gnupg.org/ftp/gcrypt/pinentry/pinentry-1.1.0.tar.bz2.sig
+  gpg --verify pinentry-1.1.0.tar.bz2.sig
+  tar -xjf pinentry-1.1.0.tar.bz2
+  cd pinentry-1.1.0
   ./configure --prefix="${GPG21_INSTALL}" \
     --with-libgpg-error-prefix="${GPG21_INSTALL}" \
     --with-libassuan-prefix="${GPG21_INSTALL}" \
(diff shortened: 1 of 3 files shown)


src/tests/cli_tests.py
```diff
@@ -109,14 +110,15 @@ def clear_keyrings():
...
-    run_proc('kill', ['-9', 'gpg-agent'])
+    run_proc(GPGCONF, ['--homedir', GPGDIR, '--kill', 'gpg-agent'])
     while os.path.isdir(GPGDIR):
         try:
             shutil.rmtree(GPGDIR)
         except:
             time.sleep(0.1)
     os.mkdir(GPGDIR, 0700)
     run_proc('gpg-connect-agent', ['reloadagent', '/bye'])
+    time.sleep(1)
```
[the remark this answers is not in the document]
flowher: ok, let see

ci/install.sh
```diff
@@ -89,7 +89,7 @@ if [ ! -e "${GPG21_INSTALL}/bin/gpg" ]; then
     cd ..
   done
 
-  for archive in libgcrypt:1.8.0 libassuan:2.4.3 libksba:1.3.5; do
+  for archive in libgcrypt:1.8.2 libassuan:2.5.1 libksba:1.3.5; do
```
ronaldtse: Could we find some way in not hardcoding the package versions but dynamically determine them based on GnuPG version? I find this very difficult to manage...

ci/install.sh
```diff
@@ -104,11 +104,11 @@ if [ ! -e "${GPG21_INSTALL}/bin/gpg" ]; then
     cd ..
   done
 
-  wget -c https://www.gnupg.org/ftp/gcrypt/pinentry/pinentry-1.0.0.tar.bz2
-  wget -c https://www.gnupg.org/ftp/gcrypt/pinentry/pinentry-1.0.0.tar.bz2.sig
-  gpg --verify pinentry-1.0.0.tar.bz2.sig
-  tar -xjf pinentry-1.0.0.tar.bz2
-  cd pinentry-1.0.0
+  wget -c https://www.gnupg.org/ftp/gcrypt/pinentry/pinentry-1.1.0.tar.bz2
```
ronaldtse: Same here.

ci/install.sh
```diff
@@ -117,11 +117,11 @@ if [ ! -e "${GPG21_INSTALL}/bin/gpg" ]; then
   ${MAKE} -j${CORES} install
   cd ..
 
-  wget -c https://www.gnupg.org/ftp/gcrypt/gnupg/gnupg-2.1.23.tar.bz2
-  wget -c https://www.gnupg.org/ftp/gcrypt/gnupg/gnupg-2.1.23.tar.bz2.sig
-  gpg --verify gnupg-2.1.23.tar.bz2.sig
-  tar -xjf gnupg-2.1.23.tar.bz2
-  cd gnupg-2.1.23
+  wget -c https://www.gnupg.org/ftp/gcrypt/gnupg/gnupg-2.2.4.tar.bz2
```
ronaldtse: Or we could fetch the latest one?

src/tests/cli_tests.py
```diff
@@ -1065,9 +1070,10 @@ def setUp(self):
         self.rnp = Rnp(RNPDIR, RNP, RNPK)
         self.gpg = GnuPG(GPGDIR, GPG)
         self.rnp.password = self.gpg.password = PASSWORD
-        self.rnp.userid = self.gpg.userid = 'singingtest@secure.gov'
+        self.rnp.userid = self.gpg.userid = 'encrypttest'+self.test_id+'@secure.gov'
```
ronaldtse: Would `@example.com` work here? I know it's been here for a while...

ci/install.sh
flowher: I agree, but this PR is just quick fix rather than big improvement of those scripts (which probably is needed anyway).

ci/install.sh
flowher: Unfortunately they don't have link  to "latest"

404 Not Found
The requested URL /ftp/gcrypt/gnupg/gnupg-latest.tar.bz2 was not found on this server.

ronaldtse: Maybe we need a script to parse the [site](https://www.gnupg.org/ftp/gcrypt/gnupg/) and extract the highest version number... maybe @erikbor has an easy solution?

flowher: I had this thought - I would rather parse this https://gnupg.org/download/index.html.
But then we depend on format of the site and build may break if format changes (small chances). Why not...

dewyatt: I think @flowher's idea seems like the simplest if they don't have any nice/official way of doing this. Some python w/html5lib and some xpath or something could work.

We could also use the most recent tags from the git repos or something like that, but the download page feels like a safer bet to me.

ci/install.sh
ronaldtse: Agree, let's just merge this first and leave generalizing to a new issue.

PR #612, opened 2018-01-31
AEAD OCB mode and GnuPG compatibility

ni4:
This PR adds support for AEAD-OCB encryption mode and ensures compatibility with latest GnuPG development build.
It updates CLI tests to check AEAD implementation against GnuPG (if it has AEAD implementation).

To select AEAD algorithm CLI parameter `--aead[=EAX|=OCB]` should be used.
Also (mostly for tests) you may control AEAD chunk size as per RFC4880bis via `--aead-chunk-bits` parameter.

### DIFF
--- a/include/repgp/repgp_def.h
+++ b/include/repgp/repgp_def.h
@@ -119,6 +119,9 @@
 /* Maximum authenticated data length for AEAD */
 #define PGP_AEAD_MAX_AD_LEN 32
 
+/* Default chunk bits, equals to 1mb chunks */
+#define PGP_AEAD_DEF_CHUNK_BITS 14
+
 /** Old Packet Format Lengths.
  * Defines the meanings of the 2 bits for length type in the
  * old packet format.
--- a/include/repgp/repgp_def.h
+++ b/include/repgp/repgp_def.h
@@ -98,33 +98,6 @@
 /* Size of the fingerprint */
 #define PGP_FINGERPRINT_SIZE 20
 
-/* Maximum length of the packet header */
-#define PGP_MAX_HEADER_SIZE 6
-
-/* Nonce len for AEAD/EAX */
-#define PGP_AEAD_EAX_NONCE_LEN 16
-
-/* Nonce len for AEAD/OCB */
-#define PGP_AEAD_OCB_NONCE_LEN 15
-
-/* Maximum AEAD nonce length */
-#define PGP_AEAD_MAX_NONCE_LEN 16
-
-/* Authentication tag len for AEAD/EAX and AEAD/OCB */
-#define PGP_AEAD_EAX_OCB_TAG_LEN 16
-
-/* Maximum AEAD tag length */
-#define PGP_AEAD_MAX_TAG_LEN 16
-
-/* Maximum authenticated data length for AEAD */
-#define PGP_AEAD_MAX_AD_LEN 32
-
-/* Default chunk bits, equals to 1mb chunks */
-#define PGP_AEAD_DEF_CHUNK_BITS 14
-
-/* Preallocated cache length for AEAD encryption/decryption */
-#define PGP_AEAD_CACHE_LEN (PGP_INPUT_CACHE_SIZE + PGP_AEAD_MAX_TAG_LEN)
-
 /** Old Packet Format Lengths.
  * Defines the meanings of the 2 bits for length type in the
  * old packet format.
--- a/include/repgp/repgp_def.h
+++ b/include/repgp/repgp_def.h
@@ -101,14 +101,17 @@
 /* Maximum length of the packet header */
 #define PGP_MAX_HEADER_SIZE 6
 
-/* Nonce (IV) len for AEAD/EAX */
+/* Nonce len for AEAD/EAX */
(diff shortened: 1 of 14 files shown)


ronaldtse:
Travis is failing in "sanitize" with this:

```
ERROR: test_sym_encryption__rnp_aead (__main__.Encryption)
  File "cli_tests.py", line 947, in test_sym_encryption__rnp_aead
    rnp_sym_encryption_rnp_aead(cipher, size, 6, 'zip', mode, bits, usegpg)
  File "cli_tests.py", line 484, in rnp_sym_encryption_rnp_aead
    rnp_decrypt_file(enc, dst)
  File "cli_tests.py", line 202, in rnp_decrypt_file
    raise_err('rnp decryption failed', out + err)
    raise CLIError(msg, log)
CLIError: None

Ran 39 tests in 841.114s

```

ni4:
@ronaldtse Thanks. Weird, before the rebase all tests were passing. Will re-check again.

ni4:
@ronaldtse I found the source of problem, it's kind of rare edge-case condition. So, lucky to hit it :) Will work on fixing

include/repgp/repgp_def.h
```diff
@@ -101,21 +101,30 @@
...
+/* Nonce len for AEAD/OCB */
+#define PGP_AEAD_OCB_NONCE_LEN 15
+
 /* Maximum AEAD nonce length */
 #define PGP_AEAD_MAX_NONCE_LEN 16
 
-/* Authentication tag len for AEAD/EAX */
-#define PGP_AEAD_EAX_TAG_LEN 16
+/* Authentication tag len for AEAD/EAX and AEAD/OCB */
+#define PGP_AEAD_EAX_OCB_TAG_LEN 16
```
henrydcase: Do we need to expose those defines in public header files? (in /src folder)

src/tests/cli_tests.py
```diff
@@ -903,14 +928,25 @@ def test_sym_encryption__rnp_to_gpg(self):
             rnp_sym_encryption_rnp_to_gpg(cipher, size, 6, 'bzip2')
 
     def test_sym_encryption__rnp_aead(self):
+        AEAD_CIPHERS = ['AES', 'AES192', 'AES256', 'TWOFISH', 'CAMELLIA128', 'CAMELLIA192', 'CAMELLIA256']
+        AEAD_MODES = ['eax', 'ocb']
+        AEAD_BITS = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12, 14, 18, 24, 30, 40, 50, 56]
+
+        usegpg = gpg_supports_aead()
```
henrydcase: I wonder if we need this. Shouldn't we just always have version which supports OCB and EAX?
If we use version which doesn't support those modes then test will simply pass without testing them? Doesn't sound good.

ni4:
@henrydcase  Looks like we got some problem in ElGamal:
```
in botan_pk_op_decrypt exception Invalid argument Decoding error: Invalid public key ciphertext, cannot decrypt
[pgp_elgamal_private_decrypt_pkcs1() crypto/elgamal.c:217] Decryption failed

Warning: cipher 0 not supported[encrypted_try_key() stream-parse.c:1444] unsupported symmetric algorithm 0
```

src/tests/cli_tests.py
ni4: If there is no latest gpg beta available then it will just do self-test. We should expand Travis to run with gpg beta, but this I think should be made as separate PR.

include/repgp/repgp_def.h
ni4: Yeah, probably it should be moved to better place. Will think about it.

henrydcase:
Hmm, that's interesting. Seems like something is corrupting memory. I'll take a look. Thnx

ni4:
Travis passes AEAD tests fine, but fails again on #613

dewyatt:
@ni4 Is [this](https://gist.github.com/dewyatt/c1ae77f945c69ccaec2f0b024e628d22) failure already known?

ni4:
@dewyatt Yeah, sorry that I forgot to mention it. I reported it in https://dev.gnupg.org/T3774 and waiting for resolution.

dewyatt:
@ni4 Can we add a temporary workaround to `cli_tests.py` to avoid that T3774 bug? We want to keep CI passing as consistently as possible.

ni4:
@dewyatt I don't see the easy workaround except ignoring GnuPG return code at all. Which would make the test against GnuPG pointless :)

dewyatt:
@ni4 I re-read the bug and I see what you mean, I can't see any obvious way to work around it either.

ni4:
@dewyatt @ronaldtse Any action on this? Or let's wait till GnuPG issue is solved?

ronaldtse:
@ni4 Can you help temporarily disable the failing test and merge it first -- there is value in having a merged implementation 😉

ronaldtse:
Going to merge this first. @ni4 could you re-enable the tests in a separate PR? Thanks!

ni4:
@ronaldtse Thanks! Added PR #618

Issue #610, opened 2018-01-31
Clang build with sanitizers is broken


henrydcase:
## Description

Seems our build with clang sanitizers was always red so nobody looks at it, but now it seems it doesn't even run:

https://travis-ci.org/riboseinc/rnp/jobs/334984424 (build for commit d3cb665 - current head)

Not sure when it was broken

ni4:
This should be because of leak sanitizer, I got (the same?) problem on macOS:
`clang: error: unsupported option '-fsanitize=leak' for target 'x86_64-apple-darwin16.7.0`

dewyatt:
This is only the sanitize-leaks build, which we have always explicitly allowed to fail (until we have fixed every leak in our test runs).

It looks like LSan requires ptrace, which isn't currently permitted in travis containers: https://github.com/travis-ci/travis-ci/issues/9033

We could use a VM for that particular build, but since we don't really use this build I think it probably makes more sense to just wait and see how travis responds. You can always run it locally with something like:

`docker run --cap-add SYS_PTRACE -ti --rm travisci/ci-garnet:packer-1512502276-986baf0 bash -l`.

@ni4 Your issue appears to be different. LSan should work on macOS [AFAIK](https://clang.llvm.org/docs/LeakSanitizer.html).

dewyatt:
Fixed in the PR referenced, closing.

PR #614, opened 2018-02-03
ci: Add GPG beta/master tests

dewyatt:
There's room for improvement here, but I didn't want to spend much more time on this and it has been working well for me so far.

* Makes it a bit easier to modify stable GPG and related package versions.
* Should enable AEAD testing I believe.
* Adds a workaround for #610.


ci/install.sh
```diff
@@ -99,35 +102,127 @@ if [ ! -e "${GPG21_INSTALL}/bin/gpg" ]; then
...
+  cat << 'END' | git apply -
+diff --git a/Makefile.am b/Makefile.am
+index 8c8b8e5..412244c 100644
+--- a/Makefile.am
++++ b/Makefile.am
+@@ -85,7 +85,7 @@ endif
+ SUBDIRS = m4 secmem pinentry ${pinentry_curses} ${pinentry_tty} \
+ 	${pinentry_emacs} ${pinentry_gtk_2} ${pinentry_gnome_3} \
+ 	${pinentry_qt} ${pinentry_tqt} ${pinentry_w32} \
+-	${pinentry_fltk} doc
```
dewyatt: Just a note: I realize this is not ideal and will likely require modification at some point, but I couldn't see an easy alternative. pinentry doesn't have a `--disable-doc` configure option like some of the others do, and it requires `makeinfo`, and the version of `makeinfo` in the ubuntu:trusty `texinfo` package seems to be too old.

ci/install.sh
```diff
@@ -99,35 +102,127 @@ if [ ! -e "${GPG21_INSTALL}/bin/gpg" ]; then
...
+  gpg --verify gettext-latest.tar.xz.sig
+  tar -xJf gettext-latest.tar.xz --strip 1
+  ./configure --prefix="${GPG_INSTALL}"
+  ${MAKE} -j${CORES} install
+  cd ..
+  export GETTEXT_PREFIX="${GPG_INSTALL}/bin/"
+
+  # workaround https://github.com/travis-ci/travis-ci/issues/8613
+  # alternatively, forcing the libgpg-error build to use gcc should work
+  export LD_LIBRARY_PATH="/usr/local/clang-5.0.0/lib"
```
dewyatt: I realize this is also not ideal and may require an update at soem point. Hopefully travis will fix the issue here. (I also tried this with globbing, rather than hardcoding the version, but it ended up set to the literal value, maybe travis is doing something weird there for security.)

ronaldtse:
Thank you @dewyatt!

Issue #487, opened 2017-10-07
Implement 'keyfile' mode of operation


ni4:
## Description

This is mostly call for discussion. Often user doesn't need to use large keyrings and deal with trust flags, key ids and so on, and only single key exported to the file is used for operations. Like in automated environment where multiple files are processed in batch and encrypted for the specified receiver.

So it probably worth to add `--keyfile` parameter, which will disable loading of keyrings and just use key(s) loaded from the specific file.

dewyatt:
I think we used to have this as `--keyring`, though I can't recall whether it was ever fully functional.

Anyways, GnuPG 2.1.14 added similar options for quick encryption without a keyring: `--recipient-file`/`-f`, so it would be good to have.

ronaldtse:
I want to bump this since the `keyfile` concept is very useful in the case where we we don't need a keyring. Thoughts @ni4 @dewyatt ?

ni4:
I can do this.

ronaldtse:
Thanks @ni4 !

PR #619, opened 2018-02-07
Allow multiple-recipient encryption

### LINKED ISSUE #236: Implement multiple recipient encryption
## Description

Currently single-recipient-only encryption is implemented, however OpenPGP standard allows multiple public key and/or password encryption.

ni4:
This PR adds support for multiple-key encryption via `-r` or `--recipient` CLI parameter, which replaces previously used `--userid`. `--userid` will be used for signer's private key selection.
Also it adds `--passwords` CLI option which allows symmetric encryption to multiple passwords.

I didn't add CLI tests yet since we lack a functionality on keyring manipulation (or do I miss somehow the option to export private key, delete key from keyring, and so on?)

Fixes #236

### DIFF
--- a/src/rnp/rnp.c
+++ b/src/rnp/rnp.c
@@ -76,6 +76,7 @@ static const char *usage = "--help OR\n"
                            "\t\t[--output=file] file OR\n"
                            "\t--version\n"
                            "where options are:\n"
+                           "\t[-r, --recipient] AND/OR\n"
                            "\t[--armor] AND/OR\n"
                            "\t[--cipher=<ciphername>] AND/OR\n"
                            "\t[--zip, --zlib, --bzip, -z 0..9] AND/OR\n"
@@ -111,6 +112,7 @@ enum optdefs {
     OPT_KEYRING,
     OPT_KEY_STORE_FORMAT,
     OPT_USERID,
+    OPT_RECIPIENT,
     OPT_ARMOR,
     OPT_HOMEDIR,
     OPT_DETACHED,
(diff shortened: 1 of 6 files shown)


henrydcase:
I would prefer this PR has a test for this new functionality

ni4:
@henrydcase Ok. I'll think how to better workaround current import/export/key deletion limitations. Probably I'll implement #487 first.

ni4:
@henrydcase Please see the updated PR with CLI testcase
@ronaldtse @dewyatt I added test to the PR so it may be finally reviewed

ni4:
@dewyatt Looks like PR #625 didn't solve the problem with dsa roundtrip timeout, it still fails.

dewyatt:
@ni4 Ok, filed an issue.

ni4:
@dewyatt Thanks. Even if that would be fixed and this PR is approved please do not merge - looks like I found some issue which wasn't addressed in tests.

ni4:
Update: this may be reviewed/merged now. Is seems there is an issue in GnuPG (https://dev.gnupg.org/T3795) which fails to decrypt file which is encrypted to number of passwords (if AEAD is not used)
Also we do not support encryption with keys and passwords together - this will be implemented within signing + encryption, since requires changes in the way we parse commands.

ronaldtse:
@ni4 are we ready to merge this given the tests pass? Thanks!

ni4:
@ronaldtse Yeah, it's ready. I just need to check why Travis fails with sanitize-leaks

dewyatt:
> Yeah, it's ready. I just need to check why Travis fails with sanitize-leaks

I ran it again without any issues and haven't been able to reproduce any issues locally. It might be fine, but we'll want to keep an eye on it I guess.

ni4:
@dewyatt Thanks. So, we may merge it once have second approval.

PR #621, opened 2018-02-08
Add rnp_key_add_uid function

randombit:


src/lib/rnp2.c
```diff
@@ -1999,6 +2000,44 @@ key_get_uid_at(pgp_key_t *key, size_t idx, char **uid)
...
+                bool             primary)
+{
+    rnp_selfsig_cert_info info;
+    pgp_hash_alg_t        hash_alg = PGP_HASH_UNKNOWN;
+    pgp_key_t *           key = get_key_prefer_public(handle);
+    pgp_key_t *           seckey = get_key_require_secret(handle);
+
+    memset(&info, 0, sizeof(info));
+
+    if (!key || !seckey)
```
henrydcase: I would add !uid as you call strlen() later on it

include/rnp/rnp2.h
```diff
@@ -237,6 +237,13 @@ rnp_result_t rnp_key_get_primary_uid(rnp_key_handle_t key, char **uid);
 rnp_result_t rnp_key_get_uid_count(rnp_key_handle_t key, size_t *count);
 rnp_result_t rnp_key_get_uid_at(rnp_key_handle_t key, size_t idx, char **uid);
 
+rnp_result_t rnp_key_add_uid(rnp_key_handle_t key,
```
henrydcase: As that's public API some comments could be helpful

src/lib/rnp2.c
```diff
@@ -1999,6 +2000,44 @@ key_get_uid_at(pgp_key_t *key, size_t idx, char **uid)
...
+        return RNP_ERROR_BAD_PARAMETERS;
+
+    ARRAY_LOOKUP_BY_STRCASE(hash_alg_map, string, type, hash, hash_alg);
+    if (hash_alg == PGP_HASH_UNKNOWN) {
+        return RNP_ERROR_BAD_FORMAT;
```
henrydcase: shouldn't the error code be RNP_ERROR_BAD_PARAMETERS?

randombit: BAD_FORMAT is used in other similar cases (eg `rnp_op_encrypt_add_password`) and I probably just copied it from there. I agree BAD_PARAM makes more sense here though.

src/lib/rnp2.c
randombit: Fixed

include/rnp/rnp2.h
randombit: Done

PR #622, opened 2018-02-10
Add defaults.h and change default symmetric cipher to AES256

ni4:
This PR moves all default algorithms and other constants related to default parameters selection to single defaults.h header.

We use cast5 as default symmetric algorithm, and, at least, AEAD without --cipher parameter doesn't work (since cast5 has 8-byte block).
RFC4880bis has AES-128 as a single MUST-implement algorithm. However, since ffi already used AES256, and GnuPG uses it as default, I replaced cast5 with aes256.

### DIFF
--- a/src/rnp/rnpcfg.c
+++ b/src/rnp/rnpcfg.c
@@ -35,6 +35,7 @@
 
 #include "rnpcfg.h"
 #include "utils.h"
+#include "defaults.h"
 #include <rnp/rnp_sdk.h>
 #include <rekey/rnp_key_store.h>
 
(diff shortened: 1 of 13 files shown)


src/rnp/rnpcfg.c
```diff
@@ -75,7 +75,7 @@ rnp_cfg_load_defaults(rnp_cfg_t *cfg)
     rnp_cfg_setstr(cfg, CFG_HASH, DEFAULT_HASH_ALG);
     rnp_cfg_setint(cfg, CFG_ZALG, PGP_C_ZIP);
     rnp_cfg_setint(cfg, CFG_ZLEVEL, 6);
-    rnp_cfg_setstr(cfg, CFG_CIPHER, "cast5");
+    rnp_cfg_setstr(cfg, CFG_CIPHER, "aes128");
```
dewyatt: LGTM but ideally this would be a define somewhere. We could use a `defaults.h` or something to centralize these things. FFI actually defaults to AES-256 -- if we add a define then we can update FFI to use the same here:
https://github.com/riboseinc/rnp/blob/ce5f4bcd2df94ab2dd08c9153c31c20254989f43/src/lib/rnp2.c#L1098

(Though some of this stuff will change when we have an actual preferences framework in place anyways)

ni4: @dewyatt Agree, I had a bunch of other constants which I didn't know where to put. Will update PR soon with defaults.h

ni4:
@dewyatt Please see the updated PR. Actually now it changes a lot more things then just a single algorithm. I updated PR description accordingly.

dewyatt:
@ni4 Looks like a step in the right direction to me!

PR #623, opened 2018-02-11
Fix leaks in tests with ask_expert_details.

dewyatt:
It looks like `setup_rnp_common` duplicates some things that `rnpkeys_init` does, and a few tests were missing `rnp_end`.


ronaldtse:
Thanks @ni4! @dewyatt it's all yours to merge in the order you wanted 😉

PR #625, opened 2018-02-11
Add a printf to test_dsa_roundtrip.

dewyatt:
This is just to prevent travis-ci from thinking the process is hung.

This test is pretty thorough and takes 5x longer than all the other `rnp_tests` combined, so travis-ci will sometimes kill it after 10 minutes of no output.

Obviously there's a few other ways to handle this, but this is simple and should work.

### DIFF
--- a/src/tests/cipher.c
+++ b/src/tests/cipher.c
@@ -718,6 +718,11 @@ test_dsa_roundtrip(void **state)
 
 
         assert_true(pgp_generate_seckey(&key_desc, &sec_key1));
+        // try to prevent timeouts in travis-ci
+        printf("p: %zu q: %zu h: %s\n",
+               key_desc.dsa.p_bitlen,
+               key_desc.dsa.q_bitlen,
+               pgp_show_hash_alg(key_desc.hash_alg));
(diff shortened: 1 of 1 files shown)


henrydcase:
Printing this information is useful anyway

src/tests/cipher.c
```diff
@@ -718,6 +718,11 @@ test_dsa_roundtrip(void **state)
 
 
         assert_true(pgp_generate_seckey(&key_desc, &sec_key1));
+        // try to prevent timeouts in travis-ci
+        printf("p: %lu q: %lu h: %s\n",
```
henrydcase: %lu ideally should be %zu

henrydcase:
Some of those tests are also done in ``cli_tests.py``. The purpose is a bit different but  if running time is  problem, then it is OK to disable some of those cases

src/tests/cipher.c
dewyatt: Good catch, fixed

randombit:
I recently spent some time trying to make group generation in botan faster but no luck so far

PR #626, opened 2018-02-11
Fix up rnp_key_add_uid to handle locked keys and add uid to both keys.

dewyatt:
I didn't get a chance to review #621 originally but I looked back and fixed some issues I saw:
* Updated to handle locked keys.
* Adds uid to both pub and sec keys (depends on keyring format a bit).
* Fixed a small mem leak.

### DIFF
--- a/include/repgp/repgp_def.h
+++ b/include/repgp/repgp_def.h
@@ -509,7 +509,8 @@ typedef enum pgp_op_t {
     PGP_OP_UNPROTECT = 6,   /* removing protection from a (locked) key */
     PGP_OP_DECRYPT_SYM = 7, /* symmetric decryption */
     PGP_OP_ENCRYPT_SYM = 8, /* symmetric encryption */
-    PGP_OP_VERIFY = 9       /* signature verification */
+    PGP_OP_VERIFY = 9,      /* signature verification */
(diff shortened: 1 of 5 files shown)


dewyatt:
> did not occur to me that pgp_key_add_userid would only modify one of the keys

Yeah it might make sense to modify that function to take both the pub and sec keys. I think that would allow us to generate the signature *once* and then just duplicate the relevant packets to the other key, more efficient. Maybe a task for another time, haven't thought about it much.

ni4:
@dewyatt @randombit  Storing userids/self signatures together with secret key doesn't make sense since you may always generate any of those? In my understanding secret keyring needs to store only secret key(s), while signatures/userids are things hanging on public key.

dewyatt:
> @dewyatt @randombit Storing userids/self signatures together with secret key doesn't make sense since you may always generate any of those? In my understanding secret keyring needs to store only secret key(s), while signatures/userids are things hanging on public key.

I get what you mean, particularly for GPG21 w/G10, but GPG2 keeps most self-sigs synced between the rings. I can't recall all the reasons for this right now since I'm not quite awake yet but if you think about the `--keyring` option on a copied `secring.gpg` or something, that might make some sense.

PR #627, opened 2018-02-11
Fix leaks in rnp_tests and re-enforce sanitize-leaks CI run.

dewyatt:
**Note**: The other PRs I have open should be merged before this one so that we can see the sanitize-leaks CI run pass. (I expect it to fail here initially)

Right now in master `rnp_tests` leaks quite a bit for me:
```SUMMARY: AddressSanitizer: 265317 byte(s) leaked in 1587 allocation(s).```

This, combined with the other PRs, should get us back to 0 so that we can re-enforce the sanitize-leaks CI run. (Hoping to add a valgrind run soon too)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
       sudo: true
 
     # codecov
-    - env: BUILD_MODE=coverage GPG_VERSION=stable
+    - env: BUILD_MODE=coverage GPG_VERSION=beta
       compiler: gcc
 
     # clang-format style checks
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,9 +60,6 @@ matrix:
             - clang-format-4.0
 
   allow_failures:
-    # allowed to fail until we have eliminated all memory leaks from our test runs
-    - env: BUILD_MODE=sanitize-leaks GPG_VERSION=stable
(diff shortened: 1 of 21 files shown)


dewyatt:
I see one thing I'll need to fix here, I'll take a look soon but this is pending the other PRs anyways.

dewyatt:
Ok, fixed the previous issue. CI will fail on the sanitize-leaks runs until the other PRs are merged and I rebase, so we're just waiting on those.

dewyatt:
It looks like there's a few leaks that I somehow missed locally, so I'll continue adding commits to get all them fixed.

@ni4 Can you double-check me on 601c78a0acafb29ddf4c406b952bb3b43e6c4252? The cache is leaked for partial packet sources, I can see multiple ways to fix it, not sure what is most appropriate though.

ni4:
@dewyatt Thanks for noticing this. Fix looks pretty correct to me.

dewyatt:
Ok it sounds like we're good to go then, sanitize-leaks is passing now, permission to merge is granted so I'll go ahead and merge.

Ideally we should rebase all open PRs on master after this to ensure we're not introducing any new leaks.

PR #630, opened 2018-02-14
Split DSA roundtrip test into 2 cases

### LINKED ISSUE #629: test_dsa_roundtrip test timeouts
## Description

Continuing from https://github.com/riboseinc/rnp/pull/619#issuecomment-365655632.

I was hoping the simple solution in #625 would be OK but it looks like it can take > 10 minutes on the first key generation within that test on travis.

Anyways, I guess the simplest solutions would be:
1. Increase the timeout, or
2. Speed up/trim the test

I think @henrydcase talked about (2) before so that might be simplest. Thoughts?

## Steps to Reproduce

1. https://travis-ci.org/riboseinc/rnp/jobs/341324652#L1894

henrydcase:
Splits DSA test into 2. One is testing all key sizes required by RFC, the other one does some negative tests for selected key.
This should make dsa roundtrip test a bit less than 2x faster as key generation is done only once

More improvements are possible, but let see first if that's fast enough...

Closes #629

### DIFF
--- a/src/tests/cipher.c
+++ b/src/tests/cipher.c
@@ -675,7 +675,6 @@ test_dsa_roundtrip(void **state)
     rnp_test_state_t *rstate = *state;
     uint8_t           message[PGP_MAX_HASH_SIZE];
     pgp_seckey_t      sec_key1 = {{0}};
-    pgp_seckey_t      sec_key2 = {{0}};
     pgp_dsa_sig_t     sig = {0};
 
     struct key_params {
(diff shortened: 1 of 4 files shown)


ronaldtse:
Thanks @henrydcase , obviously we'd prefer DSA key generation to be faster but perhaps @randombit should look at it. The approach is fine for now with me.

henrydcase:
@ronaldtse I'll look at DSA key generation to make it at least as fast as in gnupg (after doing same thing for elgamal). This PR is just quick solution to stabilize test

ronaldtse:
@henrydcase that would be great! In any case, let's get this merged first for now...

PR #632, opened 2018-02-20
Fix leak when encrypted_decrypt_cfb_header fails to decrypt.

dewyatt:
This is a rare case because we have to pass checks on the algorithm id earlier in the code in order to reach this point.
Nevertheless, this did leak randomly on occasion (less than 2% chance in some of my tests).

### DIFF
--- a/src/librepgp/stream-parse.c
+++ b/src/librepgp/stream-parse.c
@@ -1332,6 +1332,7 @@ encrypted_decrypt_cfb_header(pgp_source_encrypted_param_t *param,
 
         return true;
     } else {
+        pgp_cipher_cfb_finish(&crypt);
         return false;
     }
 }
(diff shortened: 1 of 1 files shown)


dewyatt:
@ni4 Please feel free to push!
By the way, I think this is the issue you saw in #619. I revisited it and was able to reproduce locally (rarely). The stack trace was useless because I need to change some build flags in CI, which I'll do a PR for today.

ronaldtse:
Got it -- thanks @ni4 and @dewyatt !!

ni4:
@dewyatt Yeah, probably. Please see the updated PR. If it's ok to you then it may be merged.
I want to add encrypt-decrypt-sign and other operations on sources to cmocka tests, including negative cases with modified input, so more leaks will be covered and codecov will be happier as well.

ni4:
@dewyatt Ahh, I see what you mean. Updated the PR. Is this approach good enough? It uses more code lines but less nesting level/number of ifs.

ronaldtse:
Yes! @dewyatt is exactly right, and thanks @n4 for updating it. Looks much cleaner now, feel free merge when ready.

dewyatt:
Ok everyone appears to approve overall, merging.

PR #633, opened 2018-02-20
Multiple-key signing

### LINKED ISSUE #620: Allow multiple signers
## Description

Currently only single signer's key/userid may be specified. 
CLI should be extended to allow multiple `--userid` parameters.

ni4:
This PR allows to stack `-u/--user` CLI parameters to sign/clearsign/detached sign file with multiple keys. Includes CLI test case.

Next PR, built on top of this and #619 , will allow encrypt-and-sign.

Fixes #620

### DIFF
--- a/src/librepgp/stream-write.c
+++ b/src/librepgp/stream-write.c
@@ -1321,10 +1321,11 @@ signed_dst_finish(pgp_dest_t *dst)
 {
     rnp_result_t             ret;
     pgp_dest_signed_param_t *param = dst->param;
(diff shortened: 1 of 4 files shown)


ronaldtse:
Looking at the code involved perhaps we should start considering some refactoring; some of the code seem too deeply branched under loops and if/else.

ni4:
@ronaldtse Agree. Added #635, will get to it at some point.

ronaldtse:
@dewyatt @henrydcase could we get one more approval and merge? Thanks!

PR #636, opened 2018-02-21
ci: Make local CI runs simpler.

dewyatt:
This just makes it easier to run things locally.
You can now do something like:
```sh
./travis.sh
env GPG_VERSION=beta BUILD_MODE=sanitize-leaks ci/local.sh
```

Other related changes in `travis.sh`:
* Changed to all verbose options, for readability.
* Added the ptrace capability, since it's needed for sanitize-leaks.
* Run as the travis user (uid 2000).
* Switch to working directory `/usr/local/rnp` immediately.
* Default to `bash -l` as the command, but allow others.

There's some duplication here, but we already had that duplication in the docs, so it's not too bad.

Maybe the macOS devs can let me know if any of this actually works for you guys...

### DIFF
--- a/ci/local.sh
+++ b/ci/local.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+set -eux
+
+rsync -a /usr/local/rnp /tmp
+sudo -iu travis bash <<EOF
+cd /tmp/rnp
+env GPG_VERSION=$GPG_VERSION BUILD_MODE=$BUILD_MODE ci/run-local.sh
+EOF
+
--- a/ci/run-local.sh
+++ b/ci/run-local.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+set -eux
+
+export LOCAL_BUILDS="$HOME/local-builds"
+export BOTAN_INSTALL="${LOCAL_BUILDS}/botan-install"
+export CMOCKA_INSTALL="${LOCAL_BUILDS}/cmocka-install"
+export JSONC_INSTALL="${LOCAL_BUILDS}/jsonc-install"
+export GPG_INSTALL="${LOCAL_BUILDS}/gpg-install"
+export CC=clang
+export CORES=$(grep -c '^$' /proc/cpuinfo)
+ci/install.sh
+ci/main.sh
+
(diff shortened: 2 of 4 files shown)


ronaldtse:
Very interesting... LGTM since it's not related -- maybe we need to do something for the multiple recipient thing?

ni4:
@dewyatt @ronaldtse  It fails due to memory leak during decryption:
```
Direct leak of 24 byte(s) in 1 object(s) allocated from:
    #0 0x4bbbc3 in __interceptor_malloc /local/mnt/workspace/tmp/ubuntu_rel/llvm/utils/release/final/llvm.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:67:3
    #1 0x7fb0e0619dac in operator new(unsigned long) (/usr/lib/x86_64-linux-gnu/libstdc++.so.6+0x5edac)

Indirect leak of 480 byte(s) in 2 object(s) allocated from:
    #0 0x4bbdba in calloc /local/mnt/workspace/tmp/ubuntu_rel/llvm/utils/release/final/llvm.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:76:3
    #1 0x7fb0e0e2d13a in Botan::allocate_memory(unsigned long, unsigned long) (/home/travis/local-builds/botan-install/lib/libbotan-2.so.4+0x36213a)

Indirect leak of 104 byte(s) in 1 object(s) allocated from:
    #0 0x4bbbc3 in __interceptor_malloc /local/mnt/workspace/tmp/ubuntu_rel/llvm/utils/release/final/llvm.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:67:3
    #1 0x7fb0e0619dac in operator new(unsigned long) (/usr/lib/x86_64-linux-gnu/libstdc++.so.6+0x5edac)
```
And it should be solved in #632 (it is exactly the case there - we try to decrypt file encrypt to 2 passwords, and passing first wrong one).

dewyatt:
@ni4 Ah, I forgot we hadn't merged that one yet. Probably need to clear the cache to get better stack traces now too. I'l get that merged and rebase here to make sure.

dewyatt:
OK travis is happy now after the rebase.

PR #638, opened 2018-02-25
Encrypt-and-sign support

### LINKED ISSUE #582: Implement encrypt-and-sign
## Description

Currently data can be signed or encrypted, however we should also allow encrypt and sign data.

ni4:
This PR adds encrypt-and-sign support to CLI and streams.
Also it adds some improvements/refactoring to CLI tests and encrypt-and-sign test as well.

Fixes #582

### DIFF
--- a/src/rnp/rnp.c
+++ b/src/rnp/rnp.c
@@ -652,6 +652,9 @@ setoption(rnp_cfg_t *cfg, int val, char *arg)
         }
 
         rnp_cfg_setint(cfg, CFG_PASSWORDC, count);
+        if (count > 0) {
+            rnp_cfg_setbool(cfg, CFG_ENCRYPT_SK, true);
+        }
         break;
     }
     case OPT_OUTPUT:
--- a/src/rnp/rnp.c
+++ b/src/rnp/rnp.c
@@ -93,7 +93,7 @@ static const char *usage = "--help OR\n"
                            "\t[--verbose]\n";
 
 enum optdefs {
-    /* commands */
+    /* Commands as they are get via CLI */
     CMD_ENCRYPT = 260,
     CMD_DECRYPT,
     CMD_SIGN,
@@ -108,7 +108,11 @@ enum optdefs {
     CMD_VERSION,
     CMD_HELP,
 
-    /* options */
+    /* OpenPGP data processing commands. Sign/Encrypt/Decrypt mapped to these */
+    CMD_PROTECT,
+    CMD_PROCESS,
+
+    /* Options */
     OPT_SSHKEYS,
     OPT_KEYRING,
     OPT_KEY_STORE_FORMAT,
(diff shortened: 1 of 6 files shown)


src/rnp/rnp.c
```diff
@@ -304,181 +308,209 @@ rnp_on_signatures(pgp_parse_handler_t *handler, pgp_signature_info_t *sigs, int
...
 
-            if (!list_append(&ctx.signers, rnp->defkey, strlen(rnp->defkey) + 1)) {
-                RNP_LOG("allocation failed");
-                ret = false;
-                break;
+            rnp_cfg_copylist_str(cfg, &ctx->signers, CFG_SIGNERS);
+
+            if (!list_length(ctx->signers)) {
+                if (!rnp->defkey) {
+                    fprintf(stderr, "No userid or default key for signing\n");
```
henrydcase: RNP_LOG

src/rnp/rnp.c
```diff
@@ -304,181 +308,209 @@ rnp_on_signatures(pgp_parse_handler_t *handler, pgp_signature_info_t *sigs, int
...
+            if (rnp_cfg_getbool(cfg, CFG_ENCRYPT_PK)) {
+                rnp_cfg_copylist_str(cfg, &ctx->recipients, CFG_RECIPIENTS);
 
-        for (int i = 0; i < passwordc; i++) {
-            if ((ret = rnp_encrypt_add_password(&ctx))) {
-                RNP_LOG("Failed to add password");
-                goto done;
+                if (!list_length(ctx->recipients)) {
+                    if (!rnp->defkey) {
+                        fprintf(stderr, "No userid or default key for encryption\n");
```
henrydcase: RNP_LOG

henrydcase:
LGTM - if travis OK, then good for me

src/rnp/rnp.c
ni4: This is actually not 'technical' logging but message to user. Later on we should separate it with some specific function, responsible for interaction with user.

src/rnp/rnp.c
ni4: The same as previous

ni4:
Looks like timeout happened:
```
test_sign_P3072_Q256 (__main__.SignDSA) ... /home/travis/build/riboseinc/rnp/src/rnpkeys/rnpkeys --homedir /tmp/rnpctmpW5qC1n/.rnp --generate-key --expert --pass-fd 3 --userid SignDSAtest_sign_P3072_Q256@example.com
The job exceeded the maximum time limit for jobs, and has been terminated.
```

src/rnp/rnp.c
henrydcase: ah, ok. I see

henrydcase:
Ok, let's remove tests for 3072 key. Seems that will cause problems

ronaldtse:
In case @dewyatt wishes a further look, I'll hold off merging for now.

henrydcase:
And it doesn't timeout now

src/tests/cli_tests.py
```diff
@@ -178,34 +178,62 @@ def rnp_params_insert_z(params, pos, z):
             params[pos:pos] = ['-z', str(z[1])]
 
 def rnp_params_insert_aead(params, pos, aead):
-    if aead:
+    if aead != None:
```
dewyatt: Minor thing but the PEP8 version of this would be `if aead is not None:`. There's a lot of other things that don't align with PEP8 here though, so it's a longer-term thing, no big deal.

ni4: Thanks, @dewyatt. Added #640 to fix this at some point.

PR #641, opened 2018-02-28
Fix AEAD corner cases

ni4:
This PR fixes found corner cases in rnp which didn't work properly.
Now it is finally 100% compatible with latest GnuPG master.
Enables AEAD CLI tests as well.

### DIFF
--- a/src/lib/symmetric.c
+++ b/src/lib/symmetric.c
@@ -617,11 +617,13 @@ void
 pgp_cipher_aead_reset(pgp_crypt_t *crypt)
 {
     uint32_t flags = BOTAN_CIPHER_UPDATE_FLAG_FINAL;
-    uint8_t tag[PGP_AEAD_EAX_OCB_TAG_LEN] = {0};
-    size_t  in = 0, out = 0;
-    size_t  len = crypt->aead.decrypt ? crypt->aead.taglen : 0;
-    size_t  dlen = crypt->aead.decrypt ? 0 : crypt->aead.taglen;
+    uint8_t  tag[PGP_AEAD_EAX_OCB_TAG_LEN] = {0};
+    size_t   in = 0, out = 0;
+    size_t   len = crypt->aead.decrypt ? crypt->aead.taglen : 0;
+    size_t   dlen = crypt->aead.decrypt ? 0 : crypt->aead.taglen;
 
+    /* botan_cipher_clear will not work since it resets scheduled key as well
+       so we just finishing operation and ignoring the result */
     botan_cipher_update(crypt->aead.obj, flags, tag, dlen, &out, tag, len, &in);
 }
 
--- a/src/lib/symmetric.c
+++ b/src/lib/symmetric.c
@@ -613,6 +613,18 @@ pgp_cipher_aead_update(pgp_crypt_t *crypt, uint8_t *out, const uint8_t *in, size
     return true;
 }
 
+void
+pgp_cipher_aead_reset(pgp_crypt_t *crypt)
+{
+    uint32_t flags = BOTAN_CIPHER_UPDATE_FLAG_FINAL;
+    uint8_t tag[PGP_AEAD_EAX_OCB_TAG_LEN] = {0};
+    size_t  in = 0, out = 0;
+    size_t  len = crypt->aead.decrypt ? crypt->aead.taglen : 0;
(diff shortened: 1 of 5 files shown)


src/librepgp/stream-parse.c
```diff
@@ -603,7 +616,8 @@ encrypted_src_read_aead_part(pgp_source_encrypted_param_t *param)
         }
     }
 
-    res = encrypted_start_aead_chunk(param, chunkend ? param->chunkidx + 1 : param->chunkidx, lastchunk);
+    res = encrypted_start_aead_chunk(
+      param, chunkend && param->chunkin ? param->chunkidx + 1 : param->chunkidx, lastchunk);
```
dewyatt: I think some parentheses around the ternary conditional expr. would make this a bit clearer.

src/lib/symmetric.c
```diff
@@ -613,6 +613,18 @@ pgp_cipher_aead_update(pgp_crypt_t *crypt, uint8_t *out, const uint8_t *in, size
...
+void
+pgp_cipher_aead_reset(pgp_crypt_t *crypt)
+{
+    uint32_t flags = BOTAN_CIPHER_UPDATE_FLAG_FINAL;
+    uint8_t  tag[PGP_AEAD_EAX_OCB_TAG_LEN] = {0};
+    size_t   in = 0, out = 0;
+    size_t   len = crypt->aead.decrypt ? crypt->aead.taglen : 0;
+    size_t   dlen = crypt->aead.decrypt ? 0 : crypt->aead.taglen;
+
+    botan_cipher_update(crypt->aead.obj, flags, tag, dlen, &out, tag, len, &in);
```
henrydcase: Why not to simply use "botan_cipher_clear"

src/librepgp/stream-write.c
```diff
@@ -331,8 +331,16 @@ encrypted_start_aead_chunk(pgp_dest_encrypted_param_t *param, size_t idx, bool l
     STORE64BE(param->ad + param->adlen - 8, idx);
 
     if (last) {
-        total = idx ? (idx - 1) * param->chunklen : 0;
-        total += param->cachelen + param->chunkout;
+        if (param->chunkout + param->cachelen == 0) {
```
henrydcase: parenthesis also here, please

src/librepgp/stream-write.c
```diff
@@ -331,8 +331,16 @@ encrypted_start_aead_chunk(pgp_dest_encrypted_param_t *param, size_t idx, bool l
...
+            /* we need to clearly reset it since cipher was initialized but not finished */
+            pgp_cipher_aead_reset(&param->encrypt);
+        }
+
+        total = idx * param->chunklen;
+        if (param->cachelen + param->chunkout) {
+            total -= param->chunklen - param->cachelen - param->chunkout;
```
henrydcase: Code is a bit complicated, but I think in theory it is possible that ``chunklen < chachelen``. Which will result in bug here

ronaldtse:
Thanks @ni4 , this looks good once the concerns raised by @henrydcase are resolves!

src/librepgp/stream-parse.c
ni4: Thanks. Replaced it with additional variable and if, seems to look a bit better.

src/lib/symmetric.c
ni4: It will also clear the key, while it is reused with each next chunk.

src/librepgp/stream-write.c
ni4: Okay, thanks

src/librepgp/stream-write.c
ni4: It is safe to use since in the calling function it is checked.
However I also added check here, 'just in case'.

ni4:
@henrydcase @dewyatt Please see the updated PR.

src/lib/symmetric.c
henrydcase: ok, I wouldn't guess that. Can you please add comment why we do that, as it's not obvious?
I'll check if there is a way to improve botan's FFI api

ni4: Updated code with comment. Created issue #645 to sum up what we need to improve in botan

henrydcase: perfect

ronaldtse:
Thank you @ni4 @dewyatt @henrydcase !

PR #644, opened 2018-03-02
Fix for ElGamal encryption/decryption

### LINKED ISSUE #613: Suspected memory corruption causes some tests to fail
### Problem  1 (reproduced ``rnpctmpN4D_3l``)
In test ``EncryptElgamal.test_decrypt_P1234`` we got

```

/home/travis/build/riboseinc/rnp/src/rnp/rnp --homedir /tmp/rnpctmpxWeSbm/.rnp --pass-fd 3 --userid encrypttestEncryptElgamaltest_decrypt_P1234@example.com --decrypt /tmp/rnpctmpxWeSbm/encryptEncryptElgamaltest_decrypt_P1234.enc --output /tmp/rnpctmpxWeSbm/encryptEncryptElgamaltest_decrypt_P1234.dec
in botan_pk_op_decrypt exception Invalid argument Decoding error: Invalid public key ciphertext, cannot decrypt
[pgp_elgamal_private_decrypt_pkcs1() crypto/elgamal.c:217] Decryption failed

Warning: cipher 0 not supported[encrypted_try_key() stream-parse.c:1444] unsupported symmetric algorithm 0
[init_encrypted_src() stream-parse.c:2035] failed to obtain decrypting key or password
[rnp_process_file() rnp.c:1393] error 0x12000006
```

Happens rarely and randomly. Memory corruption suspected. Seems like old bug as it's in elgamal encryption which was modified long time ago.

Root cause is that RNP uses elgamal decryption wrongly. As a description of the function says 
``- * @param length length of g2k or in (must be equal to byte size of prime `p')`` and this is not enforced by the code which causes memory overwrite inside a function.
Right solution is to make the function misuse resistant.

### Problem 2 (reproduced - `` `` )
This check doesn't seem correct (leading zeros)
```

    if (length != p_len) {
        RNP_LOG("Wrong size of modulus in public key %u %u", p_len, length);
        goto end;
    }
```

### Problem 3
elgamal encryption and decryption functions should return ``rnp_error_t``, then error handling won't be that weird

henrydcase:
(elgamal) Fixes memory corruption
    
    * Reimplements ElGamal encryption and decryption
    * Introduces structure which describes buffer as a
      pointer to content and size
    * Enables ElGamal CLI decryption tests

Fixes #613


src/lib/crypto/elgamal.c
```diff
@@ -84,167 +84,117 @@
...
         RNP_LOG("Failed to load private key");
         goto end;
     }
 
-    if (botan_privkey_check_key(key, rng_handle(rng), 1)) {
-        RNP_LOG("Wrong private key");
-        goto end;
-    }
-
-    memcpy(bt_plaintext, g2k, p_len);
```
henrydcase: bug was here. g2k not alwayswas p_len

src/lib/types.h
```diff
@@ -792,4 +792,13 @@ typedef struct rnp_key_protection_params_t {
     pgp_hash_alg_t    hash_alg;
 } rnp_key_protection_params_t;
 
+/*
+ * Data structure for storing pointer to the buffer together with
+ * it's length.
+ */
+typedef struct buf_t {
+  uint8_t *bx;
```
dewyatt: I don't really get the field name `bx`, I can't think of what that is meant to stand for. We have `buf` in `pgp_memory_t` that makes a bit more sense to me (or `ptr`, or something). What does it mean here?

src/lib/crypto/elgamal.c
```diff
@@ -84,167 +84,117 @@
...
-    uint8_t *             bt_ciphertext = NULL;
+    rnp_result_t          ret = RNP_ERROR_BAD_PARAMETERS;
+    uint8_t               enc_buf[ELGAMAL_MAX_P_BYTELEN*2] = {0};   /* Max size of an output
+                                                                    len is twice an order of
+                                                                    underlying group (twice
+                                                                    byte-size of p) */
 
-    if (botan_mp_num_bytes(pubkey->p->mp, &p_len)) {
-        RNP_LOG("Wrong public key");
+    size_t tmp;
```
dewyatt: It would be good to have a better name than `tmp` here. I get that it's reused later but I think it's more readable to just have separate variables with more descriptive names.

henrydcase: it's used for different temporary things which are not worth having it's own variable defined

src/lib/types.h
henrydcase: pointer to buffer. I prefer short names, I'll rename it to pbuf

henrydcase:
Crap, tests will take too long (which is obvious when I now think about it - elgamal needs both DSA+Elgamal keys and elgamal keygen uses strong primes, so it's even longer than DSA).
But I'll create another PR tomorrow with some workaround. So this one is not to be merge just yet, but also there will no more changes here

ronaldtse:
@henrydcase can we merge this now?

henrydcase:
@ronaldtse first of all - sorry for delay. Can we wait day or two  - I'll try to finish my test improvement (which is here: https://github.com/henrydcase/rnp/pull/34/files), then elgamal stuff should be faster. But need day or two.
It will also add possibility to export private key

henrydcase:
Needs, little re-work to align with changes introduced in  #652

ronaldtse:
@henrydcase sounds good -- thanks!

henrydcase:
This is done now

src/lib/types.h
```diff
@@ -792,4 +792,13 @@ typedef struct rnp_key_protection_params_t {
     pgp_hash_alg_t    hash_alg;
 } rnp_key_protection_params_t;
 
+/*
+ * Data structure for storing pointer to the buffer together with
+ * it's length. Value of 'len' depends on context in which it is used.
```
dewyatt: > Value of 'len' depends on context in which it is used.

That sounds a bit error-prone to me.

ronaldtse: Indeed, @henrydcase could you fix this a bit and merge? Thanks!

henrydcase: I think it is OK as it's just utility.

PR #647, opened 2018-03-04
ci: Split CI test jobs up and increase some timeouts

dewyatt:
We still get some occasional timeouts in CI.

There are two main limits we're running in to:
1. 10 minutes with no output. This is configurable so we can extend it if needed.
2. 50 minutes total job run time. This is a hard limit and includes the install time (when cache isn't available).

So this PR:
1. Extends the default 10-minute no-output limit to 20 minutes. The downside is that the output of `ci/main.sh` is suppressed until the entire script has finished. There are workarounds for this but I didn't spend the time testing any so I'll skip it for now.
2. Further splits up the matrix to separate the cmocka tests and cli tests. This is a pretty standard thing to do anyways.

I expect we may still hit the 50 minute timeout when rebuilding cache, but that's a rare thing anyways (and it will just require restarting the relevant jobs, since it will timeout after creating the new cache).


dewyatt:
I'll make some improvements on this since there's still issues. I probably need to add a stage to warm the cache.

henrydcase:
good stuff, LGTM.

dewyatt:
I added a commit to use a separate stage for the cache.

I think travis is having issues so we may have timeouts anyways:
https://github.com/travis-ci/travis-ci/issues/9294#issuecomment-370335966

I did also see failues due to network timeouts for git clones and wgets from gnupg.org.

Anyways, let's see how this build goes.

dewyatt:
Ok, merging with permission. I should add that technically travis stages are a beta feature. But...it mostly seems to work OK.

PR #649, opened 2018-03-08
Fix possible memory leak during signature writing

ni4:
It could happen only during quite negative scenario, however still worth fixing.


ronaldtse:
Thanks @ni4 , the fix seems good except that the `res = res && xxx` calls don't seem very clean. I do understand that it is the convention of this file, so maybe we can take some time later to clean them up.

ni4:
@ronaldtse Thanks. I don't like such notation as well, but doing if (!res) { goto error; } after each call IMHO will look uglier.

ronaldtse:
Agree, that’s why I didn’t suggest that :wink:

PR #652, opened 2018-03-10
Improves CI test execution 

### LINKED ISSUE #141: No way to export a private key
Wanted the generate a key with rnp and read it with gpg to test #140, but it doesn't seem `rnpkeys` or any API exposed in `rnp.h` supports this operation - `rnp_export_key` only handles public keys.

henrydcase:
I propose not to generate ElGamal, DSA and RSA keys during test execution but rather import pre-generated keys. Key generation for those algorithms is time consuming (by design of an algorithm) and even if we improve speed of key generation it might not be enough to fix all problems related to broken CI builds. From the other hand I would very much like to test algorithms with different key sizes (encryption, signing but also import and export of the keys itself).
Importing pre-generated keys will also ensure backward compatibility. Also .... who on earth would generate DSA or ElGamal key in 21st century :)

This PR:
* adds pre-generated keys and uses them in CLI tests. 
* adds possibility to export secure keys which Closes  #141
* adds some tests for RSA
* blocks all the tests for ElGamal: this is done in order to facilitate integration of #644 for which this PR is a pre-requisit. #644 will re-enable those tests (it needs a bit of rework, because API used in tests has changed)


ronaldtse:
Agree. We should separate testing of key generation, key importing/exporting and key usage.

src/tests/cli_tests.py
```diff
@@ -678,11 +678,44 @@ def setup(loglvl):
...
+        self.__op_key_location = None
+        self.__op_key_gen_cmd = None
+
+    @property
+    def operation_key_location(self):
+        return self.__op_key_location
+
+    @operation_key_location.setter
+    def operation_key_location(self, key):
+        # OZAPTF key must be tuple
```
dewyatt: Can we take care of the TODO?

src/tests/cli_tests.py
```diff
@@ -1425,51 +1482,89 @@ class SignDSA(Sign):
...
-        cmd = SignDSA.RNP_GENERATE_DSA_PATTERN.format(1024)
         rnp = self.rnp.copy()
         rnp.hash = 'SHA512'
-        self._rnp_sign_verify(rnp, self.gpg, cmd)
+        self.operation_key_gencmd = SignDSA.RNP_GENERATE_DSA_PATTERN.format(1024)
+        self._sign_verify(rnp, self.gpg)
+
+class EncryptSignRSA(Encrypt, Sign):
+
+    GPG_GENERATE_RSA_PATERN = """
```
dewyatt: "pattern" is the word you want

src/tests/data/regenerate_keys
```diff
@@ -0,0 +1,117 @@
...
+
+sys.path.append("../")
+from gnupg import GnuPG as GnuPG
+from rnp import Rnp as Rnp
+
+PASSWORD = "password"
+USERID_PFX = '@example.com'
+
+
+class KeyFormater(object):
```
dewyatt: "formatter" is probably what you meant

dewyatt:
Great, I was thinking we should do something like this at some point. I'm thinking as long as we still have tests that use dynamically generated keys for each type (maybe at least the smallest sizes) then we should be pretty well covered on any key generation regressions.

src/tests/cli_tests.py
henrydcase: thnx, I've forgotten about it

henrydcase:
It seems master branch is broken:
https://travis-ci.org/riboseinc/rnp/builds/352507158?utm_source=github_status&utm_medium=notification

This is causing broken build for this PR after it was rebased on most recent trunk
@dewyatt Can you take a look on CI please?

dewyatt:
@henrydcase It's from what I mentioned here: https://github.com/riboseinc/rnp/issues/611#issuecomment-372447367
Basically I made the assumption that GnuPG and related items are kept somewhat stable in their master branches (buildable and testable), but I see some mistakes do sometimes sneak in there. 
Anyways, it looks like it was fixed and tests should be completing soon.

I'll look into pinning our GPG "beta" to specific known-good commits, sort of like we do for stable.

henrydcase:
Perfect, all good now. Please merge

PR #653, opened 2018-03-13
ci: Allow setting specific commits for GPG beta and deps.

dewyatt:
Right now they're all the same as before, but we have the option of picking known-good commits in case we need to in the future.

This is just something minor I did earlier, ref: https://github.com/riboseinc/rnp/pull/652#issuecomment-372684295

### DIFF
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -137,13 +137,22 @@ build_gpg_stable() {
 }
 
 build_gpg_beta() {
+  local GETTEXT_VERSION=$1
+  local NPTH_VERSION=$2
+  local LIBGPG_ERROR_VERSION=$3
+  local LIBGCRYPT_VERSION=$4
+  local LIBASSUAN_VERSION=$5
+  local LIBKSBA_VERSION=$6
+  local PINENTRY_VERSION=$7
+  local GNUPG_VERSION=$8
+
   mkdir gettext
   pushd gettext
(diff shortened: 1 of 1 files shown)


ronaldtse:
LGTM -- I had this thought cross my head sometime ago -- how about we extract the script that installs GnuPG somewhere else so others could also use it?

dewyatt:
> LGTM -- I had this thought cross my head sometime ago -- how about we extract the script that installs GnuPG somewhere else so others could also use it?

I'm not sure it would really be very useful outside of CI. I mean most users will be installing via a package anyways (or maybe building their own package for gpg 2.3) - only reason we aren't is because we don't have root in travis. Plus we have specific configure options we use for CI, some slightly hacky patches, etc.

ronaldtse:
@dewyatt we actually have a CI container build job that installs GnuPG 2.2/2.3 (since RHEL still provides 2.0), and we have run into installation problems so many times due to dependency issues. So if there is a script that we can just give the major version to install, it would be fantastic.

dewyatt:
@ronaldtse I think I see what you're saying, though you're usually better off building some custom RPMs in my experience. I guess it's not that unusual to do things with a plain script for container infrastructure though.

Anyways, it might be possible to come up with something better, maybe using the info provided in https://github.com/riboseinc/rnp/issues/611#issuecomment-368871937. Probably a task for another week though.

PR #654, opened 2018-03-16
FFI improvements: memory input/output, documentation, tests

ni4:
This PR just adds memory input/output streams to FFI and extends FFI tests to check those streams during signing/verification.

@dewyatt I changed my rnp_input_from_memory to match your implementation, however it still may have a merge conflict. So if you have branch(es) with only those changes feel free to make a PR and I'll rebase on it, fixing the conflicts.


ni4:
@dewyatt Thanks. I don't like that approach very much too, but 4 almost identical large methods I don't like as well. Will see what can be reorganized/reused when implement encrypt-and-sign and implement more tests for FFI.

PR #664, opened 2018-04-08
Streamed PGP keys parsing and writing

ni4:
This PR introduces new code for PGP keys parsing and writing, and corresponding tests.
However it doesn't replace any functionality for keyrings yet - it's quite huge and will require separate PRs in order to not break anything.

Benefits of new data formats/approach are:
- separation between 'data' and 'key' entities (pgp_key_pkt_t will contain only parsed raw data)
- easier and more logical key encryption/decryption (no need to reparse key's raw packets)
- simpler approach for signatures (fill data - sign - write)
- storage of key data in structure which is described in RFC (transferable OpenPGP keys).


dewyatt:
This is looking very promising! I'll need some time to review.

ni4:
Thanks, @dewyatt! All comments are welcome.

ni4:
@ronaldtse I doubt that there will be some performance difference since not complicated processing is done during the parsing. I'll test it anyway once finish with things I currently working on (move packet dump code to streams, so large amount of old code could be finally removed).
Also, there is at least one major advantage - current key loading code dumps the whole file into the memory before processing, while within current approach we may filter out unneeded keys later on.

dewyatt:
Yeah I wouldn't necessarily expect much of an improvement on performance for loading/saving keys. The only good thing about the old way of reading/writing keys is that it really was a simple fread/fwrite, so you could always round-trip things without any problem. I'm not sure if this maintains that, still reviewing.

I bet this new stuff should be much more efficient for encrypting/decrypting keys though.

So I'm still reviewing but I wanted to check on a use case, so imagine something like this:
* An application stores a primary key in some super secure database.
* The application stores multiple subkeys in various other secure databases.

So basically it seems like an interesting use-case for an application to be able to load/save a lone subkey (and the key provider will use the binding sig to request the primary). I did some work to make sure this is possible with the old code in #665 so I just wanted to check if it will be possible to load/save a subkey separately from a primary in this new code.

src/lib/types.h
```diff
@@ -106,6 +106,12 @@ typedef struct {
     const char *string;
 } pgp_map_t;
 
+/** multi-precision integer, used in signatures and public/secret keys */
+typedef struct pgp_mpi_t {
```
dewyatt: Why are we introducing a new type for this? Shouldn't we be using `bignum_t`, which is backed by secure memory in botan?

ni4: The idea is that all data is in 'raw' format once it is parsed from the source/written out. And conversion to bignum_t is done only during cryptography operation. Say, we don't need convert to bignum_t all 20 keys we read from the keyring, only one which is used for verification or signing.

src/librepgp/stream-key.c
```diff
@@ -0,0 +1,725 @@
...
+    pgp_forget(body.data, body.len);
+    free_packet_body(&body);
+    return ret;
+}
+
+void
+forget_secret_key_fields(pgp_key_pkt_t *key)
+{
+    if (!is_secret_key_pkt(key->tag) || !key->sec_avail) {
+        return;
```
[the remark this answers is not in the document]
ni4: As we don't use assert() almost anywhere probably it would be better to add RNP_LOG. Ok.

src/lib/types.h
dewyatt: Ok that sounds reasonable. I think once we start using C++ in here we ought to use `Botan::secure_vector` for this, just to make sure things are wiped on free (I see it's manually done here) and mlock'd.

ni4:
Thanks, @ronaldtse and @dewyatt !

PR #665, opened 2018-04-11
Rework the ffi key provider and tweak key loading

dewyatt:
The main thing here is changing how the key provider works in the FFI. This changes it to just be a signal to inform the application that it should load the key via the ffi methods.

I also changed key loading a bit to use the key provider where it seemed to make sense.
This allows, for example, loading a subkey for which the primary wasn't loaded yet. The G10 store also now uses the key provider to load the pubkey, instead of having that confusing pubring parameter in there.

(I did some last minute changes here so I'll probably want to review this myself tomorrow too.)

### DIFF
--- a/src/lib/key-provider.c
+++ b/src/lib/key-provider.c
@@ -63,28 +63,26 @@ rnp_key_matches_search(const pgp_key_t *key, const pgp_key_search_t *search)
     return false;
 }
 
-bool
-pgp_request_key(const pgp_key_provider_t *   provider,
-                const pgp_key_request_ctx_t *ctx,
-                pgp_key_t **                 key)
+pgp_key_t *
+pgp_request_key(const pgp_key_provider_t *provider, const pgp_key_request_ctx_t *ctx)
 {
-    if (!provider || !provider->callback || !ctx || !key) {
-        return false;
+    pgp_key_t *key = NULL;
+    if (!provider || !provider->callback || !ctx) {
+        return NULL;
     }
-    *key = NULL;
-    if (!provider->callback(ctx, key, provider->userdata)) {
-        return false;
+    if (!(key = provider->callback(ctx, provider->userdata))) {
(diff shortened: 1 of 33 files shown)


dewyatt:
Just noticed I forgot to commit a test I added so I'll get that in shortly too.

ronaldtse:
@ni4 could you help review / approve / merge this? Thanks!

src/lib/key-provider.c
```diff
@@ -24,55 +24,114 @@
...
+    case PGP_KEY_SEARCH_FINGERPRINT:
+        return (key->fingerprint.length == search->by.fingerprint.length) &&
+               !memcmp(key->fingerprint.fingerprint,
+                       search->by.fingerprint.fingerprint,
+                       key->fingerprint.length);
+    case PGP_KEY_SEARCH_GRIP:
+        return !memcmp(key->grip, search->by.grip, PGP_FINGERPRINT_SIZE);
+    case PGP_KEY_SEARCH_USERID: {
+        for (unsigned i = 0; i < key->uidc; i++) {
+            if (!strcmp((char *) key->uids[i], search->by.userid)) {
```
ni4: Maybe use strstr/strcasestr here? Or do we need exact match here?

dewyatt: My feeling is it's best to go for an exact match within the key provider itself. An application using the FFI can always use the id iterator to do fancier matces like case-insensitive/regex userid, or partial keyid, and so on. (Or we could implement an FFI search function with a callback to check custom criteria, etc.)

dewyatt:
It looks like I left some debugging stuff in here, so let me get rid of that...

src/lib/key-provider.c
ronaldtse: @ni4 if it’s OK feel free to merge it. Thanks!

PR #672, opened 2018-04-22
Fix G10 protected keys and also fix some static analysis issues

dewyatt:
* Fixed a regression in G10 protected keys.
* (Hopefully) Fixed a bunch of issues reported by Coverity (which is working again).
* Fixed some other issues reported by clang's static analyzer.
* Fixed some issues with `pgp_keyid` and `pgp_fingerprint` (note that these can no longer be treated as if they return bool).

### DIFF
--- a/src/fuzzing/fuzz_keys.c
+++ b/src/fuzzing/fuzz_keys.c
@@ -5,7 +5,7 @@ int
 main(int argc, char *argv[])
 {
     rnp_t        rnp;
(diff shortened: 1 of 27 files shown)


dewyatt:
It looks like I got 38 steps forward and 5 steps back on Coverity.
@ni4 When you get a chance would you be able to look at the `src_peek_line`-related issues in Coverity? I think there's a handful of high issues there.

PR #684, opened 2018-06-01
Use `pgp_signature_t` for signature operations

ni4:
This PR is followup to #683 and improves rnp with `rnp_signature_t` usage for signature-related information.
So, within this PR:
- `pgp_signature_t` and streamed approach is used to read/write/store signatures
- removed old signature/subpacket parsing code
- removed old armoring/dearmoring code
- use only stream-sig methods for signing/verification

However, there are some old-style methods are left (including parse_packet, and callback-based key signatures validation). Next PRs will resolve this.


dewyatt:
@ni4 I pushed a fix to master for the CI issue here, so if you rebase and push it should be OK.

ni4:
@dewyatt Thanks, rebased.

PR #685, opened 2018-06-04
CMake

dewyatt:
#### Overview
* Removed autotools items, switched to CMake.
* Collapsed the CI matrix to run `rnp_tests` and `cli_tests.py` in parallel (within the same job).
* Fixed overflow in `init_armored_dst` (really long version strings).

#### Usage
* Generally CMake encourages out-of-source builds, so it's good to do something like:
```sh
mkdir /tmp/build && cd /tmp/build
cmake -DBUILD_TESTING=on /usr/local/rnp && make
```
* For running tests, it's easiest to use `ctest`:
```sh
# run all tests (two at a time), showing output
ctest -j2 -V
# run rnp_tests only
ctest -R rnp_tests
```

#### Notes
* Requires CMake 3.7+. We can probably reduce this if needed.
* Packaging is still a bit limited. I focused on RPM for the most part. It's just a big monolith so at some point we'll probably want to split it out into separate packages (devel, etc).
* There's currently a bug that makes building FreeBSD packages fail ([here](https://gitlab.kitware.com/cmake/cmake/issues/18031)), but it's not difficult to call `pkg create` manually on the generated manifest.
* I'm still working on more docs for this. Some of the most useful docs on modern cmake I found were [here](https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1) (especially the videos).


dewyatt:
(resolved conflict)

ni4:
@dewyatt I'm on a poor mobile connection for the next day, will review it as soon as get back to better conditions.

dewyatt:
@ni4 No worries!

src/lib/CMakeLists.txt
```diff
@@ -0,0 +1,231 @@
...
+    "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>"
+    "$<INSTALL_INTERFACE:include>"
+  PRIVATE
+    "${CMAKE_CURRENT_SOURCE_DIR}"
+    "${PROJECT_SOURCE_DIR}/src"
+)
+
+# This has precedence and can cause some confusion when the binary
+# dir one isn't actually being used. To be improved.
+file(REMOVE "${CMAKE_CURRENT_SOURCE_DIR}/config.h")
```
ni4: @dewyatt Got hit by this line on Mac (or, probably, because of the way I used to build). While it worked on Travis container It does not work if run as `cmake .`.
Do you have any ideas on improvement?

dewyatt: I added it at the last minute honestly. Yes, I can add a workaround - one minute.

dewyatt:
Argh, I committed a fix but accidentally pushed the branch to master so it merged.

Anyways, @ni4 give `master` a try, it should work for in-source and out-of-source builds.

ni4:
@dewyatt Thanks, now it works perfectly.

PR #688, opened 2018-06-12
Initial C++ conversion

randombit:
This is the most primitive possible conversion. I basically just renamed all the files to `.cpp` and fixed the bugs. In addition right now it relies on `-fpermissive` as there are a number of missing casts I have not yet fixed. However it does compile and tests pass.

C++ does not support C99's designated initializers, but GCC accepts them with a warning. In addition a bug? in GCC meant I had to change anywhere a designated initializer was used with a non-literal because "sorry, unimplemented: non-trivial designated initializers not supported". I think while the C99 syntax works pretty well for creating a descriptive request we can do better in C++ anyway.

I don't want to go too crazy in this PR. I think it's better to get this merged to master asap, which would allow multiple people to work independently on converting different areas of the source.

There is a doc on C++ usage which is incomplete and doesn't really justify its decisions but it's something to start a discussion.

I probably will not be able to make further changes here until Friday, so feel free to push to this branch.


randombit:
Oh it appears the GCC in Ubuntu 14.04 doesn't accept the missing casts even with `-fpermissive`, so CI is failing. I will try to fix those in the morning.

ni4:
@ronaldtse @randombit If there are no objections I'll push update with fixes a bit later. There are more then these two warnings.

randombit:
Using `{}` seems good to me for now. Eventually (IMO) these structs can all be classes with constructors so a default constructor can zero-init.

dewyatt:
I'm just going to push a few minor things in here (and clear the cache to be thorough).

dewyatt:
Huh I kind of expected git clone --branch would accept a commit hash since it accepts a branch name or tag. Oh well, I'll fix that and also rebase this on master.

ronaldtse:
@ni4 @dewyatt could we have one more approval to get this merged? Thanks!

ni4:
@ronaldtse Done. But maybe @dewyatt wanted to add some more commits?

ronaldtse:
Thanks @ni4 and @dewyatt it's all yours!

dewyatt:
I'd like to take a closer look - I'm seeing a possible issue, just trying to see if it was introduced here or elsewhere.

dewyatt:
I think there's a regression introduced here, this breaks ruby-rnp on a few tests. (With and without my additional changes here)

I think we should hold off on merging this until we figure out what is going on there. I can create a PR to add ruby-rnp tests in CI (been working on that this week), which should help.

dewyatt:
Ok I have a fix for the ruby-rnp issue (reading from some uninitialized mem in here). I would personally like to see #691 merged first so we can rebase, confirm the issue, commit the fix, etc.

dewyatt:
Rebased on master. CI will fail because of the uninitialized read, I'll commit a quick fix and make sure CI passes and then merge.

ni4:
Thanks, @dewyatt. I'll rebase and update #689 with C++ afterwards.

ronaldtse:
Thank you @dewyatt, @randombit and @ni4  !!

PR #689, opened 2018-06-12
Use new key signatures validation method

ni4:
This PR replaces old parser-based key signature validation with stream-based.
This is next step in removal of old code.


ni4:
@ronaldtse Thanks. Yeah, depending on which one is merged first I'll update this one or Jack's PR.

ni4:
@dewyatt Do you know a good way to restart the Travis build? It timed out on download on some stuff, other then this everything should be working with the latest commit.

Also it would be a good idea to squash commits before merge since only the last one builds well with C++ changes. Tried to do rebases/single commit fixes but that didn't work well.

dewyatt:
@ni4 I went ahead and restarted it. For reference: if you go to the build page (Details link above), you can hit "Restart build" to restart the entire build or use the little button to the right of each job.

I see you're going to have some failures like @randombit mentioned here.

dewyatt:
I think Jack introduced `pgp_io_from_fp` to fix some of these, so I guess you'll want to use that instead of the `pgp_io_t io = {...}`.

ni4:
@dewyatt Thanks, I believe I looked over the whole travis screen in search for refresh/like buttons while they are just in front of me.

Corrected order of pgp_io_t fields should fix the build.

ronaldtse:
Thanks @ni4 , look forward getting this merged 👍

dewyatt:
> P.S. @dewyatt Looks like codecov now checks ruby-rnp coverage instead of rnp?

codecov does indeed appear confused ,I'll take care of that soon

PR #691, opened 2018-06-16
ci: Test against ruby-rnp.

dewyatt:
This can probably be improved but I already spent more time on it than I planned due to various issues with ASan, some travis cache timeouts, etc.

On a side note: I also experimented with adding a valgrind job to CI  (cc5a82d40cd767aaf4ae331c787e7fdb47b7f037) but found it to be too slow to complete within the 50 minute limit.


ronaldtse:
This would definitely help improve the quality of checks against RNP now, but there will be occasions that ruby-rnp gets out of sync with RNP.

Do you plan to have this as a temporary attempt to fix RNP, or will we have some mechanism on synchronizing changes later on?

dewyatt:
@ronaldtse I do view this as temporary. Once I get a release branch going I can probably improve this.

dewyatt:
@ni4 Good catch, I must have made some last-minute changes. Should be good now (I think I might work on reducing this duplication now too)

dewyatt:
Actually I think I might push to this one more time...

dewyatt:
I guess I'll go ahead and merge this, seems like concerns were addressed and such.

ronaldtse:
Thank you @dewyatt !

PR #692, opened 2018-06-18
ci: Ensure ruby-rnp test does not upload to codecov.

dewyatt:
Quick fix to prevent a codecov upload during the ruby-rnp test.


dewyatt:
Actually I broke coverage with the CMake switchover anyways, but I'll do a separate PR to get that fixed up.

PR #693, opened 2018-06-20
Make CMake actually read the version tag

ronaldtse:


ronaldtse:
Apparently this is because the tag didn't have a commit message (https://stackoverflow.com/questions/6445148/git-fatalno-tags-can-describe-sha1-number)

So this is not necessary. I'll close it.

PR #695, opened 2018-06-22
docs: Add release workflow document.

dewyatt:
This is a bit light on details in some places, but can be expanded as needed.

I created the `release/v0.x` branch and tagged the 0.9.0 release. I also did some minor supporting commits directly to master.


ronaldtse:
We just need to automate release packaging and distribution now. At the moment it would be the easiest to add a `rnp/prepare.sh` file in https://github.com/riboseinc/rpm-specs to make this happen.

PR #696, opened 2018-06-22
docs: Add missing step for release workflow.

dewyatt:
This step is important to ensure github/`git archive` tarballs are buildable.

I failed to do this for the initial release so we should discuss how to fix that. My thought is to simply amend the commit and move the tag in the release branch, since it's the release itself that is the problem and the code is not changing (so a version bump would be a bit misleading).

### DIFF
--- a/doc/RELEASE-WORKFLOW.md
+++ b/doc/RELEASE-WORKFLOW.md
@@ -14,11 +14,15 @@ version (and x is a literal -- not a placeholder).
 git checkout -b release/v1.x master
 ```
 
-## Create a CHANGELOG.md
+## Create a CHANGELOG.md and version.txt
 
 ```
 vim CHANGELOG.md
 git add CHANGELOG.md
+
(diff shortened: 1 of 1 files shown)


ronaldtse:
Feel free to go ahead to make the commit, it is indeed one issue I encountered in https://github.com/riboseinc/rpm-specs/pull/30

dewyatt:
> Feel free to go ahead to make the commit, it is indeed one issue I encountered in riboseinc/rpm-specs#30

Done (and yes I've been looking at rpm-specs too)

dewyatt:
I went ahead and added the fixed branch naming in here too, as discussed earlier.

dewyatt:
I think I see an ordering issue in these docs, will fix that here or in master later.

ronaldtse:
FYI The RNP yum package is now available on our yum repo for CentOS/RHEL: https://github.com/riboseinc/yum

PR #697, opened 2018-06-23
CMake: Fix library versioning

dewyatt:
Apparently I made some last-minute changes and broke this.
This is broken in our tagged release too :( so currently an RPM of 0.9.0 will have libs like:
```
/usr/lib64/librnp.so
/usr/lib64/librnp.so.
/usr/lib64/librnp.so.0
```

After this change, we are back to normal:
```
/usr/lib64/librnp.so
/usr/lib64/librnp.so.0
/usr/lib64/librnp.so.0.9.0
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,9 @@
 # 3.7+ for the BZip2 target in the cmake-bundled finder
 cmake_minimum_required(VERSION 3.7)
 
+# contact email, other info
+include(cmake/info.cmake)
+
 # determine version
 include(cmake/version.cmake)
 determine_version("${CMAKE_CURRENT_SOURCE_DIR}" RNP)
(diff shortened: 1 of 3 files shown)


dewyatt:
I also just realized I slightly deviated from my original naming for release branches (did this in the doc too).
I originally intended to do `release/0.x` rather than `release/v0.x`. (I meant to use the 'v' prefix just for version tags.) @ronaldtse Do we want to correct this or just go with it?

ronaldtse:
Oh, let's correct this since we're still early on. At least we won't have any regrets 👍

dewyatt:
> Oh, let's correct this since we're still early on. At least we won't have any regrets +1

Ok, once this is merged I'll update those docs, drop the old branch, and recreate the properly-named one (which will include this fix we need too).

PR #699, opened 2018-06-23
ci: Fail with uninitialized and unused CMake vars.

dewyatt:
This only tests the cmake code paths used in travis, so it's far from foolproof but it's better than nothing.


ronaldtse:
Looks good but some TEST builds are failing due to it.

dewyatt:
Ok, should be good now.

PR #700, opened 2018-06-24
Use new code for keyring loading/saving

ni4:
This PR migrates keyring operations to the new streamed code. It currently just replace old functionality without adding new, but now it would be much easier to make operations like key export, certification, generation of new subkeys and so on so on.
There are some leftovers, like raw subpackets in keys, pgp_memory_t usage in g10 keys and so on. Those will be updated via separate PRs.


ni4:
Thanks, @dewyatt . I just removed some more old code/updated test to make Codecov happier, so probably you'd like to take another look.

PR #702, opened 2018-07-02
More flexible key by name search

ni4:
While initially it was a fix for AV in key-by-name search (function str2keyid) I decided to move ahead and refactor it slightly.
Now it will allow search by fingerprint, and allow to use spaces/tabs inside.

### DIFF
--- a/src/lib/misc.cpp
+++ b/src/lib/misc.cpp
@@ -350,10 +350,6 @@ pgp_mem_writefile(pgp_memory_t *mem, const char *f)
     return true;
 }
 
-typedef struct {
-    uint16_t sum;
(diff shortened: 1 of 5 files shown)


ronaldtse:
@dewyatt quick question — is the functionality already exposed to ruby-rnp? Thanks.

dewyatt:
@ronaldtse No, this is a CLI-level thing. Plus I've actually tried to avoid any kind of fuzzy (/multi-field) searching in the FFI because I figure the application can always implement whatever filtering it wants.

ronaldtse:
That's true. Thanks!

PR #705, opened 2018-07-03
Remove RFCs from sources and add links to them to development guide

### LINKED ISSUE #704: Remove `doc/rfcs`
RFCs are all viewable on the IETF site and there's no reason why we duplicate them here.

ni4:
If you have any suggestions on other protocol information/updates to development guide feel free to add.

Fixes #704


ronaldtse:
This is good, but it would be even better if we keep a specification on what RFCs RNP implement. We will definitely need that on our upcoming public pages 👍

PR #707, opened 2018-07-03
Fix signing/encryption with default key

### LINKED ISSUE #682: rnp --sign doesn't work without the username parameter
## Description

While rnp should be able to use default key for signing, it doesn't.

## Steps to Reproduce

`rnp --sign somefile`

ni4:
Recently signing/encryption with not specified signer/recipient was broken so this PR fixes it.
Also, in order to cover updated code, I added cmocka CLI test and some overhead to allow call rnp.cpp:main from the rnp_tests.
It uses define `RNP_TESTS`, which is set in `CMakeLists.txt` for `rnp_tests` target.
@dewyatt , @ronaldtse what do you think about such approach?
It would allow to add more tests later on and improve coverage and static analysis on rnp/rnpkeys

Fixes #682

### DIFF
--- a/src/rnp/rnp.cpp
+++ b/src/rnp/rnp.cpp
@@ -828,7 +828,7 @@ parse_option(rnp_cfg_t *cfg, const char *s)
 }
 
 int
-#ifndef RNP_TESTS
+#ifndef RNP_RUN_TESTS
 main(int argc, char **argv)
 #else
 rnp_main(int argc, char **argv)
--- a/src/rnp/rnp.cpp
+++ b/src/rnp/rnp.cpp
@@ -503,11 +503,11 @@ rnp_cmd(rnp_cfg_t *cfg, rnp_t *rnp)
         break;
     case CMD_VERSION:
         print_praise();
-        ret = RNP_SUCCESS;
+        ret = true;
         break;
     default:
         print_usage(usage);
-        ret = RNP_SUCCESS;
+        ret = true;
     }
 
 done:
@@ -609,7 +609,9 @@ setoption(rnp_cfg_t *cfg, int val, char *arg)
     case CMD_ENARMOR:
     case CMD_HELP:
     case CMD_VERSION:
-        setcmd(cfg, val, arg);
+        if (!setcmd(cfg, val, arg)) {
+            return false;
+        }
         break;
     /* options */
     case OPT_COREDUMPS:
@@ -779,7 +781,9 @@ setoption(rnp_cfg_t *cfg, int val, char *arg)
         rnp_set_debug(arg);
         break;
     default:
-        setcmd(cfg, CMD_HELP, arg);
+        if (!setcmd(cfg, CMD_HELP, arg)) {
+            return false;
+        }
         break;
     }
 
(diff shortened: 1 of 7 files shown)


ni4:
@dewyatt Agree, will update now to RNP_RUN_TESTS.
Btw, I believed that codecov does coverage for `rnp_tests` run but according to diff rnp/rnpkeys was also covered. I.e. does it calculate coverage for runs via python cli_tests, or I just mislooked something?

dewyatt:
> Btw, I believed that codecov does coverage for rnp_tests run but according to diff rnp/rnpkeys was also covered. I.e. does it calculate coverage for runs via python cli_tests, or I just mislooked something?

It does both `rnp_tests` and `cli_tests`. They are done in two separate jobs in CI and then codecov merges that info.

src/rnp/rnp.cpp
```diff
@@ -856,8 +866,8 @@ main(int argc, char **argv)
                 rnp_cfg_setstr(&cfg, CFG_SSHKEYFILE, optarg);
                 break;
             case 'V':
-                print_praise();
-                exit(EXIT_SUCCESS);
+                setcmd(&cfg, CMD_VERSION, optarg);
```
dewyatt: Shouldn't this check the return value now?

ni4:
@dewyatt Thanks for explanation. I updated the latest commit with new define name

src/rnp/rnp.cpp
```diff
@@ -906,6 +919,15 @@ main(int argc, char **argv)
         }
     }
 
+    switch (rnp_cfg_getint(&cfg, CFG_COMMAND)) {
+    case CMD_HELP:
+    case CMD_VERSION:
+        rnp_cmd(&cfg, &rnp);
+        ret = EXIT_SUCCESS;
```
dewyatt: Should this be conditional based on the return of `rnp_cmd`?

src/rnp/rnp.cpp
ni4: Generally - yes. I allowed it here because CMD_VERSION/CMD_HELP always succeed. But you are right, it is better to add checks.

src/rnp/rnp.cpp
ni4: The same as previous. Will update the PR shortly.

src/rnp/rnp.cpp
ni4: Argh, sorry, misunderstood the question. Here it will just set the command, return value will be checked below (the next your comment).

ni4: Omg, no. I will get back to this tomorrow. Too late at the evening to think well.
Only now got what did your comment actually mean :-)

ni4:
@dewyatt Please see the latest commit. It should fix the issues you described.

src/tests/cli.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "utils.h"
+
+int rnp_main(int argc, char **argv);
+
+int
+call_rnp(const char *cmd, ...)
+{
+    int     argc = 0;
+    int     res;
+    char ** argv = (char **) calloc(32, sizeof(char *));
```
dewyatt: Might be a good place to start using C++ (just a simple `std::string` on the stack).

PR #714, opened 2018-07-13
Fix rnp_op_add_signature.

dewyatt:
This fixes the issue in https://github.com/riboseinc/enmail/issues/53#issuecomment-403798923 and is basically what I meant in https://github.com/riboseinc/rnp/issues/496.

The reason we never encountered this issue before is because it only affects G10 secrings and we don't have any tests using `rnp_op_add_signature` in that case (I'll work on adding that too, but wanted to get this fix in).


dewyatt:
Looks like I'll have to update the ruby-rnp specs, it makes an assumption on the key request ordering that isn't always true after this change.

ronaldtse:
@dewyatt I suppose the Ruby specs should have been build that allowed failures, and this repo should trigger a downstream build of ruby-rnp specs rather than being directly linked?

PR #720, opened 2018-07-23
Fix typo: Allistair -> Alistair

strogonoff:
After a Google search, it seems that Alistair is the right spelling
(and not Allistair) for NetBSD Foundation president Alistair G. Crooks.


strogonoff:
Found the typo when looking up the name out of curiosity

strogonoff:
Looks like a glitch in codecov/project since the change is just two words in README

PR #723, opened 2018-08-07
Validate key signatures

ni4:
This PR forces key signature validation for all keys added to the keyring(s).
Also it validates key material, using botan's key_check calls. Previously those calls were embedded into sign/encrypt calls, slowing them down.

So, after applying this PR:
- all keys, added to the keyring, will be checked for validity (certifications, subkeys bindings, expiration times, and so on).
- keys which did not pass check will be denied to be used for encryption/verification, erroring the operation
- signing/encryption will work faster

Most likely some more checks should be added, but this one will address most of 'wrong key' usecases.


ronaldtse:
@dewyatt any thoughts? Hoping to merge this soon. Thanks!

src/librepgp/stream-sig.cpp
```diff
@@ -1150,6 +1151,194 @@ signature_validate_direct(const pgp_signature_t *   sig,
...
+    /* Check signature's expiration time */
+    now = time(NULL);
+    create = signature_get_creation(sinfo->sig);
+    expiry = signature_get_expiration(sinfo->sig);
+    if (create > now) {
+        /* signature created later then now */
+        RNP_LOG("signature created %d seconds in future", (int) (create - now));
+        sinfo->expired = true;
+    }
+    if (create && expiry && (create + expiry < now)) {
```
dewyatt: Wouldn't this be a regression of 109760747450b7300c2015531d25de7b3f83d5de?

```
create = 1533813318
expiry = 1 (sec)
now = create + expiry
(create + expiry) < now == true (expired)
(create + expiry) <= now == false (not expired)
```

In the example above, the signature has been alive for 1s (the expiry), so I believe it should be considered expired.

ronaldtse: True (@dewyatt you meant "should not be considered expired", right?).

@ni4 could we also add a test against this? Thanks!

ni4: @dewyatt @ronaldtse Thanks with this. Got a bit messed up with those `<`/`<=`.
Will fix it.

ronaldtse: But at 1001 it is not expired, right? `(1000 + 1) < 1001 == true (expired)`

ni4: Yes, at 1001 it is not expired. However `(1000 + 1) < 1001` will be false :)

dewyatt: I mixed up the example a little so I fixed that, but I actually meant what I wrote: that it *should* be considered expired.

If I create a signature with a lifetime of 1 second, it should be considered expired 1 second after the creation time.
If it is created at tick 0, and the time is now 1, it has been alive for a full second and should be considered expired.

I played around with some other implementations, and fed them the following inputs:

```
Created: 1533917179
Signature Duration: 1
Now: 1533917180
```

| Implementation Name | Signature Validity |
|---------------------|--------------------------------------------------|
| [GnuPG](https://github.com/gpg/gnupg/blob/e88f56f1937ac92f6a3b94e50b6db2649ec0be41/g10/parse-packet.c#L2047-L2049)               | Signature Expired                                |
| [OpenPGP.js](https://github.com/openpgpjs/openpgpjs/blob/49153db63b5a11f805a19b7187842b466067947b/src/packet/signature.js#L715)          | Signature Expired                                |
| [PGPy](https://github.com/SecurityInnovation/PGPy/blob/51fe8f8864ae46a3f7fed6060c0cc5a546f62458/pgpy/pgp.py#L178)                | Signature Ok                                     |
| [kbpgp](https://github.com/keybase/kbpgp/blob/055d8bd1d96450a53b09794209f3ee5355d0fbb9/lib/openpgp/packet/signature.js#L740) | Signature Ok |

Some of the js/py ones do use higher-res clock sources, so if the time was something like `1533917180001` (created+duration+1ms), then it would be considered expired.

Anyways, I'm OK as far as merging this. I'm just not totally sure what is most correct as far as that particular line.

ronaldtse: @dewyatt should this be clarified in RFC4880-bis?

dewyatt: @ronaldtse It does look like implementations are inconsistent, but I'm not sure it's worth it. Some err on the side of potentially expiring early like I suggested, some err on the side of potentially expiring late like ni4 does here. 

I see the JWT RFC (which uses a date for expiration, rather than a seconds count) has language like:
>  The "exp" (expiration time) claim identifies the expiration time on
   or after which the JWT MUST NOT be accepted for processing.

So it leaves it up to the implementation (plus it allows implementations to account for a few minutes of clock skew anyways).

I'm pretty much neutral on whether the RFC should add clarification.

ronaldtse: Then let's stay with the convention of GnuPG, seems safe enough for us.

dewyatt: > @ronaldtse Then let's stay with the convention of GnuPG, seems safe enough for us.

This PR switched us away from GnuPG's behavior though, so if we want to switch back we'll want to use `<=` again like `(create + expiry <= now)`.

PR #724, opened 2018-08-07
Adjust library and package naming

dewyatt:
This is mostly name adjustments that ought to be useful if we ever want to support installing multiple major versions of rnp, for example.

| Old               | New                     |
|----------------------------|-----------------------------|
| librnp.so.0.9.1            | librnp-0.so.0.9.1           |
| rnp-0.9.1-1.el7.x86_64.rpm | rnp0-0.9.1-1.el7.x86_64.rpm |
| yum install rnp            | yum install rnp0            |


ronaldtse:
Thanks @dewyatt this is a good approach.

1. The ruby-rnp specs are failing as expected due to the change in library name, once that is fixed it should be good to go.

2. Could you also update https://github.com/riboseinc/rpm-spec-rnp/ so that the RNP RPM package gets built properly (automatically)?

dewyatt:
@ronaldtse 
> Could you also update https://github.com/riboseinc/rpm-spec-rnp/ so that the RNP RPM package gets built properly (automatically)?

I'll get that updated and work on automating things.

ronaldtse:
@dewyatt is this good to merge? Thanks!

ronaldtse:
Not at all, it was all approved but just waiting for your merge 😉

PR #725, opened 2018-08-13
Release 0.9.2

### LINKED ISSUE #721: Bump version to 0.9.2
As described here: 
https://github.com/riboseinc/enmail/issues/53

0.9.1 contains some issues that are now resolved in master, which are necessary for the correct functioning of signing emails via Ruby-RNP.

@ni4 could you help us make a release 0.9.2? Thanks!

ni4:
This should follow the rules from `RELEASE-WORKFLOW.md`. 
The only doubt I have is it correct that GitHub show two commiters for some commits below, or it doesn't matter?

Fixes #721


dewyatt:
> The only doubt I have is it correct that GitHub show two commiters for some commits below, or it doesn't matter?

It's normal, it's because on some of them you are the committer but not the author (rebasing and cherry-picking will do this).
You can do something like `git log --format=fuller` to see what I mean.

dewyatt:
Also this reminded me I had some local doc updates so I pushed fcd33da. Nothing different, just expanded a bit.

ronaldtse:
@dewyatt now I see that master has also been updated. Somehow it was still showing last month's commits as HEAD in master few hours ago...

Issue #231, opened 2017-06-19
Generated secret keys always have 65536 as iteration count for s2k


ni4:
GnuPG uses much higher iteration count: 13107200 or 7864320, depending on installation.
We should also pick up  better default encryption algorithm (currently CAST5 is used).

randombit:
This could be an additional parameter to the keygen structure added in #241

ni4:
Hm, s2k parameters are not part of key generation but part of secret key saving process - i.e. user logically is able to take already generated key, change security parameters (add/change/remove password, strengthen s2k) and save the key again.

randombit:
Yeah true - I think the code would need to be refactored to split up these two operations.

dewyatt:
Just adding this note for when someone works on this: The way gpg appears to handle this is it calculates an iteration count that takes ~100ms on the current machine, and uses that (see [here](https://github.com/gpg/gnupg/blob/5cf95157c5db88dd599ac4d48f619782179b1438/agent/protect.c#L163)).

IIRC another piece to keep in mind is that there is the RFC encoding of the iteration count, and then there's another encoding that gnupg sometimes uses (can't really recall).

ni4:
Their approach seems to be a bit weird. Stuff which is encrypted on slow computer should not be less secure. I think better and simpler approach is to just declare default constant which is secure for nowadays and update it every year or so. And of course add configurable parameter for this.

dewyatt:
> Their approach seems to be a bit weird. Stuff which is encrypted on slow computer should not be less secure.

I think the idea is probably to keep up with advances in hardware, without requiring occasional software updates just to change the iteration count? They do enforce a lower limit (65536) at least.

> I think better and simpler approach is to just declare default constant which is secure for nowadays and update it every year or so. And of course add configurable parameter for this.

I actually think the calibration approach may be better, if combined with a growing lower limit every year. Think about the case where you're using the newest high-end machine with all kinds of specific hardware acceleration. Such a machine may have the *potential* to use a much higher iteration count (without noticeable delays) than other machines, and so it may as well default to doing that.

So I think a machine-specific calibration combined with a growing lower limit might be a good compromise.

ni4:
We also forgot another small thing - default iteration count should also depend on hash algorithm. It must be higher for weaker hashes and vice versa.

ni4:
See the issue #531

dewyatt:
[...]

PR #728, opened 2018-08-17
Add option to set S2K iterations for rnpkeys

randombit:
Increase default iteration count to match GPG (33554432)

GH #231


dewyatt:
This is a good improvement but I also wanted to ask @randombit do you have any thoughts on this being dynamic? GnuPG does it like this: [agent/protect.c](https://github.com/gpg/gnupg/blob/5cf95157c5db88dd599ac4d48f619782179b1438/agent/protect.c#L163)
So the value will actually vary based on system load, etc.

@ni4 mentions in https://github.com/riboseinc/rnp/issues/231#issuecomment-329514180, that is kind of weird in some ways.
On the other hand, doesn't it kind of act as a small source of extra randomness (sort of extending the salt a bit)?

Anyways, it's a step forward, so I'd say ship it. Hopefully we can travis-ci to pass (pretty slow today).

dewyatt:
It looks like the increased iteration count is going to be enough to hit the 50-minute limit in CI. We need to trim some fat.

randombit:
Ouch! I will switch to something self-tuning, and also probably add use of `--s2k-iterations` when generating keys to test so we don't hit the runtime limits.

randombit:
e6babb5 adds a runtime estimator. For the (bad) OpenPGP S2K it turns out to be easy because the PGP "iterations" is exactly equal to the number of bytes hashed by the S2K. The function runs the hash computation over a buffer for a short while to get an estimate of the hash functions bytes per second speed, then uses that to answer how many bytes would have to be hashed to consume the requested CPU time.

I considered adding a cache of the hash speed results, but it did not seem worthwhile since most of the time this sort of autotuning would occur just once per process lifetime (`rnpkeys` execution or such).

I reduced the S2K iterations used in the key generation tests to 1, which actually ends up being 1024 bytes hashed (the minimum possible amount in the OpenPGP S2K encoding). That may help the test runtimes.

dewyatt:
(Rebased to resolve conflict)

src/lib/crypto/s2k.cpp
```diff
@@ -128,9 +129,60 @@ pgp_s2k_encode_iterations(size_t iterations)
...
+
+   if(duration == 0)
+      return MIN_ITERS;
+
+   const double bytes_per_usec = static_cast<double>(bytes) / duration;
+   const double desired_usec = desired_msec * 1000.0;
+   const double bytes_for_target = bytes_per_usec * desired_usec;
+   const uint8_t iters = pgp_s2k_encode_iterations(bytes_for_target);
+
+#if 0
```
dewyatt: I'm thinking this should use `rnp_get_debug` -- best to avoid commented-out chunks of code.

randombit: Much better that way, fixed now.

randombit:
@ronaldtse @dewyatt Review comment addressed is there anything else that needs to change before merge?

dewyatt:
Also, wouldn't it make sense to get rid of `DEFAULT_S2K_ITERATIONS` and always use the dynamic value?
```diff
diff --git a/src/lib/defaults.h b/src/lib/defaults.h
index 771f528..5b840c4 100644
--- a/src/lib/defaults.h
+++ b/src/lib/defaults.h
@@ -43,9 +43,6 @@
 /* Default symmetric algorithm as PGP constant */
 #define DEFAULT_PGP_SYMM_ALG PGP_SA_AES_256
 
-/* Default number of S2K iterations */
-#define DEFAULT_S2K_ITERATIONS 33554432
-
 /* Default number of msec to run S2K derivation */
 #define DEFAULT_S2K_MSEC 200
 
diff --git a/src/rnp/rnpcfg.cpp b/src/rnp/rnpcfg.cpp
index 93640c8..81d5426 100644
--- a/src/rnp/rnpcfg.cpp
+++ b/src/rnp/rnpcfg.cpp
@@ -80,7 +80,7 @@ rnp_cfg_load_defaults(rnp_cfg_t *cfg)
     rnp_cfg_setstr(cfg, CFG_CIPHER, DEFAULT_SYMM_ALG);
     rnp_cfg_setstr(cfg, CFG_SUBDIRGPG, SUBDIRECTORY_RNP);
     rnp_cfg_setint(cfg, CFG_NUMTRIES, MAX_PASSWORD_ATTEMPTS);
-    rnp_cfg_setint(cfg, CFG_S2K_ITER, DEFAULT_S2K_ITERATIONS);
+    rnp_cfg_setint(cfg, CFG_S2K_ITER, 0);
 }
 
 bool
diff --git a/src/rnpkeys/rnpkeys.cpp b/src/rnpkeys/rnpkeys.cpp
index 81630e8..2366966 100644
--- a/src/rnpkeys/rnpkeys.cpp
+++ b/src/rnpkeys/rnpkeys.cpp
@@ -216,7 +216,6 @@ rnp_cmd(rnp_cfg_t *cfg, rnp_t *rnp, optdefs_t cmd, char *f)
         primary_desc->crypto.rng = &rnp->rng;
         protection->hash_alg = primary_desc->crypto.hash_alg;
         protection->symm_alg = pgp_str_to_cipher(rnp_cfg_getstr(cfg, CFG_CIPHER));
-        protection->iterations = rnp_cfg_getint(cfg, CFG_S2K_ITER);
         action->subkey.keygen.crypto.rng = &rnp->rng;
 
         if (!rnp_cfg_getbool(cfg, CFG_EXPERT)) {
```

Or is there a reason to not do that? I'm not fully awake yet so I'm a bit uncertain myself.

randombit:
You are right the hard-coded iteration count should be removed entirely. I will check into what is going on with rnpkeys.

randombit:
@dewyatt Ready for another look 0f2d4ae should have addressed last issues.

src/lib/crypto/s2k.cpp
```diff
@@ -128,9 +133,64 @@ pgp_s2k_encode_iterations(size_t iterations)
...
+   {
+   // Consider clock_gettime
+   struct timeval tv;
+   ::gettimeofday(&tv, NULL);
+   return (static_cast<uint64_t>(tv.tv_sec) * 1000000) + static_cast<uint64_t>(tv.tv_usec);
+   }
+
+size_t pgp_s2k_compute_iters(pgp_hash_alg_t alg, size_t desired_msec, size_t trial_msec)
+{
+   const uint64_t TRIAL_USEC = trial_msec * 1000;
```
dewyatt: I think this should be set after the `DEFAULT_S2K_TUNE_MSEC` override below.

src/lib/crypto/s2k.cpp
```diff
@@ -128,9 +133,64 @@ pgp_s2k_encode_iterations(size_t iterations)
...
+      bytes += sizeof(buf);
+      end = get_timestamp_usec();
+      }
+
+   pgp_hash_finish(&hash, buf);
+
+   const uint64_t duration = end - start;
+
+   if(duration == 0)
+      return MIN_ITERS;
```
dewyatt: Should be wrapped in `pgp_s2k_decode_iterations`?

src/rnp/rnpcfg.h
```diff
@@ -61,6 +61,8 @@
 #define CFG_CREATION "creation"         /* signature validity start */
 #define CFG_CIPHER "cipher"             /* symmetric encryption algorithm as string */
 #define CFG_HASH "hash"                 /* hash algorithm used, string like 'SHA1'*/
+#define CFG_S2K_ITER "s2k-iter"         /* number of S2K hash iterations to perform */
+#define CFG_S2K_MSEC "s2k-msec"         /* number of milliseconds S2K should target */
```
dewyatt: I'm not seeing this option being read in the code.

src/lib/crypto/s2k.cpp
```diff
@@ -128,9 +133,64 @@ pgp_s2k_encode_iterations(size_t iterations)
...
+size_t pgp_s2k_compute_iters(pgp_hash_alg_t alg, size_t desired_msec, size_t trial_msec)
+{
+   const uint64_t TRIAL_USEC = trial_msec * 1000;
+   const uint8_t MIN_ITERS = 96;
+   uint8_t buf[8192] = { 0 };
+   size_t bytes = 0;
+
+   if(desired_msec == 0)
+      desired_msec = DEFAULT_S2K_MSEC;
+   if(trial_msec == 0)
```
dewyatt: Also, on the style side of things, [conditionals should always have braces](https://github.com/riboseinc/rnp/blob/master/doc/DEVELOPMENT-GUIDE.md#general-guidelines).

src/lib/crypto/s2k.cpp
randombit: Fixed, thanks

src/lib/crypto/s2k.cpp
randombit: Fixed

src/rnp/rnpcfg.h
randombit: Fixed

src/lib/crypto/s2k.cpp
randombit: Fixed, thanks

dewyatt:
@randombit 
> On a fast-ish machine it tops at 300 msec or so since that's as hard as the PGP S2K can get.

On my machine (w/a Ryzen 7 1800X), it looks like anything above 55ms (so including the default 150ms) ends up using the max iteration count (255/65011712). Does that seem right? I just want to be sure. It seems like an improvement regardless at this point though, so approving.

dewyatt:
Maybe we can pester @ronaldtse for a review too since this has been open for so long.

ronaldtse:
Can someone help merge once the build is done? Thanks 👍

dewyatt:
CI has been done for like 4 hours, github UI just isn't updating for some reason. Merging!

randombit:
> On my machine (w/a Ryzen 7 1800X), it looks like anything above 55ms (so including the default 150ms) ends up using the max iteration count (255/65011712). Does that seem right? 

@dewyatt  Expected I think, because Ryzen has SHA-1 and SHA-256 hardware which is very fast. I am on an Intel chip that doesn't have these instructions. On a Ryzen 2700, I see SHA-256 at about 1600 MB/s, at that speed it can hash the max S2K iterations in about 40 ms or so.

randombit:
I suspect if you use `--hash SHA384` (which doesn't have hardware accel on Ryzen) you'll see it top at more like 150-200 ms.

dewyatt:
@randombit Ok that makes sense, thanks! I just checked and I do get like 1600 MiB/s SHA-256 (`botan speed SHA-256`) and only 315 MiB/s w/SHA-384. And indeed using `--hash SHA384` has the desired effect. :thumbsup:

PR #745, opened 2018-09-17
Correct return codes from validate_pgp_key_signature

### LINKED ISSUE #726: validate_pgp_key_signature doesn't fail when it should
## Description

Lots of assigning to `res` in various places:
https://github.com/riboseinc/rnp/blob/dff813481ad9925749ec57faa7a8752c47aa8e5b/src/librepgp/stream-key.cpp#L1314

But it's never used:
https://github.com/riboseinc/rnp/blob/dff813481ad9925749ec57faa7a8752c47aa8e5b/src/librepgp/stream-key.cpp#L1335

@ni4 can you take a look?

ni4:
Fixes and documents logic of the return codes from key signature validation code.
Fixes #726

### DIFF
--- a/src/librepgp/stream-key.cpp
+++ b/src/librepgp/stream-key.cpp
@@ -1462,10 +1462,19 @@ typedef struct validate_info_t {
     rng_t *                 rng;
 } validate_info_t;
 
+/**
+ * @brief validate single key's signatures
+ *
+ * @param sig signature to validate
+ * @param info additional validation info
+ * @return RNP_SUCCESS if signature was successfully validated (however, this doesn't
+ *         guarantee that it is valid). Other error code means that there were problems
+ *         during the validation.
+ */
 static rnp_result_t
 validate_pgp_key_signature(const pgp_signature_t *sig, validate_info_t *info)
 {
-    rnp_result_t         res = RNP_ERROR_SIGNATURE_INVALID;
+    rnp_result_t         res = RNP_ERROR_GENERIC;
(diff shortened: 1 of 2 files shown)


randombit:
The two doxygen comments seem to disagree - one (in `.c`) says it will return `RNP_SUCCESS` if the validation was processed (vs some error occurring) and in the `.h` file "RNP_SUCCESS if all checks were passed, RNP_ERROR_SIGNATURE_INVALID for invalid sig, RNP_ERROR_SIGNATURE_EXPIRED for expired signature."

It looks like in fact `RNP_ERROR_SIGNATURE_INVALID` and `RNP_ERROR_SIGNATURE_EXPIRED` are squashed into `RNP_SUCCESS` and the comment in the C file is correct. Is the idea the caller look at the `validate_info_t` to find the actual status? At the very least the comment in the header should be corrected. Maybe also hint there that the caller must examine the validation result in order to see if the signature is actually correct?

Otherwise ok.

ni4:
@randombit It's a bit messy, but actually two functions (`signature_check` and `validate_pgp_key_signature`) serves a bit different purposes:
- first one checks signature when all required data is available (and is used for document signature checks as well), so return the validation status itself
- second one used to to check one of the key signatures and then populate validate_info_t structure (increment valid/invalid counts and so on). So it returns failure only in case validation was broken (out of memory and other conditions), so later on we cannot be sure on validation status of one or more signatures.

randombit:
@ni4 Sorry I had completely missed those are actually different functions.

ronaldtse:
Thanks @ni4 @dewyatt @randombit !

Issue #748, opened 2018-09-19
Building on (my) debian


cfcs:
## Description

The build instructions for debian are missing

## Steps to Reproduce

Here's what I did to build the tools:
```
sudo apt install g++-8 cmake libbz2-dev libjson-c-dev libbotan-2-dev
cmake .
make
```

ni4:
@cfcs Do you mean the primary key binding signature  for signing subkey? Then it looks like my bug, will provide a fix soon.

ni4:
Hm, it doesn't seem to be an error, at least it is compatible with GnuPG: it generates primary key binding only for sign-able key, and not for the certify one (see the packet listing below).
Looks like 'certify' means that this subkey may be used to issue certification signatures only.

```
:off 657: packet header 0xb852 (tag 14, len 82)
Public subkey packet
    version: 4
    creation time: 1537520054 (Fri Sep 21 11:54:14 2018)
    public key algorithm: 19 (ECDSA)
    public key material:
        ecc p: 515 bits
        ecc curve: NIST P-256
:off 741: packet header 0x88ef (tag 2, len 239)
Signature packet
    version: 4
    type: 24 (Subkey Binding Signature)
    public key algorithm: 19 (ECDSA)
    hash algorithm: 8 (SHA256)
    hashed subpackets:
        :type 33, len 21
        issuer fingerprint: 0x2c536e2d839b03fc1b61dc134042507083ec2bc6 (20 bytes)
        :type 2, len 4
        signature creation time: 1537520054 (Fri Sep 21 11:54:14 2018)
        :type 27, len 1
        key flags: 0x02 ( sign )
    unhashed subpackets:
        :type 16, len 8
        issuer key ID: 0x4042507083ec2bc6
        :type 32, len 117
        embedded signature:
            version: 4
            type: 25 (Primary Key Binding Signature)
            public key algorithm: 19 (ECDSA)
            hash algorithm: 8 (SHA256)
            hashed subpackets:
                :type 33, len 21
                issuer fingerprint: 0x1a38cf93396f82f6aa6392a34076a626a0c24b1e (20 bytes)
                :type 2, len 4
                signature creation time: 1537520054 (Fri Sep 21 11:54:14 2018)
            unhashed subpackets:
                :type 16, len 8
                issuer key ID: 0x4076a626a0c24b1e
            lbits: 0x7981
            signature material:
                ecc r: 256 bits
                ecc s: 255 bits
    lbits: 0xa7a0
    signature material:
[...]

PR #752, opened 2018-09-24
Update readme with instructions for debian

ni4:
I'm not quite familiar with debian but these instructions at least allows to build on clean debian/stretch Docker container.

Based on #748

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,8 +207,21 @@ yum install -y rnp
 
 ## On Debian
 
-(WIP)
-
+``` sh
+# Clone the repository by version tag (or omit it to get the latest sources)
+sudo apt install git
+git clone https://github.com/riboseinc/rnp.git -b v0.11.0
+# Enable access to `testing` packages by editing /etc/apt/sources.list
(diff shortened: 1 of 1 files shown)


dewyatt:
I think it should also install `git` (before the clone), and `zlib1g-dev`.

ni4:
@dewyatt Thank you. Actually it is a good idea to try those instructions before issuing the PR :) Now it seems to work fine.

dewyatt:
Let's go ahead and merge this since it's just docs. :)

ronaldtse:
Thanks @ni4 @dewyatt ! We really should go ahead with the automated CI build of all sorts of packages...

dewyatt:
> We really should go ahead with the automated CI build of all sorts of packages...

@ronaldtse I'm working on the gitlab ci automation for this sort of stuff but it will be a few more weeks. I'm trying to improve things and take advantage of gitlab ci features and tune things up, etc., rather than do a 1:1 conversion. (I already spent most of last week at the NYC maker faire for my birthday today)

Issue #572, opened 2017-12-17
rnp prompts for input and reads from stdin when it should prefer /dev/tty


dewyatt:
## Description

I believe things like this should try to read from `/dev/tty` if avail, and `stdin` otherwise:
https://github.com/riboseinc/rnp/blob/7bafcf181962c12075e92d64b12ca377e91c37ff/src/lib/rnp.c#L1225

See `rnp_getpass` for an example or `ttyio.c` from GnuPG for a more complete solution.

On the output side (https://github.com/riboseinc/rnp/issues/542), things like this should output to `stderr` or `/dev/tty`:
https://github.com/riboseinc/rnp/blob/7bafcf181962c12075e92d64b12ca377e91c37ff/src/lib/rnp.c#L1224

## Steps to Reproduce

1. touch out.gpg
2. echo test | src/rnp/rnp --sign --output=out.gpg

## Expected Behavior

It should warn about the overwrite and prompt for a filename, then proceed.

## Actual Behavior

It prompts for a filename but the user does not get a chance to input it since it tries to read from the pipe.

ni4:
@dewyatt Agree, actually we also should move all such code parts to CLI level, so input from user will be handled by CLI not library.

Issue #542, opened 2017-11-25
Redirected output can result in extra invalid bytes being appended


dewyatt:
## Description

I think it's common to do something like `echo message | rnp --sign > output.gpg` and expect it to work.
We do have a `--output` option as well, but *nix users are definitely going to expect that they can pipe/redir stdout in this case and it will work.

Unfortunately this doesn't work 100% correctly in rnp because of things like:
https://github.com/riboseinc/rnp/blob/82fb956244d721fc6fa002922b36b9f3fe1887e6/src/lib/pass-provider.c#L124
https://github.com/riboseinc/rnp/blob/82fb956244d721fc6fa002922b36b9f3fe1887e6/src/lib/pass-provider.c#L132

A quick solution here might be to output to stderr (or use `pgp_io_t`).

## Steps to Reproduce

1. `echo test | src/rnp/rnp --sign | pgpdump` or `echo test | src/rnp/rnp --sign | gpg --list-packets`

## Expected Behavior

There shouldn't be any issues with the PGP data.

## Actual Behavior

At least one newline (0x0a) is appended to the data (more if password is incorrect). pgpdump and `gpg --list-packets` both spot this and complain in different ways (`gpg --list-packets` exits with a non-zero code, though `--verify` is good).

ni4:
I think that `puts("\nPasswords do not match!");` should be considered as communication with user on the CLI layer, and we should have some callback/function/whatever else like `show_user_message()` which will check for output redirection and so on.

ronaldtse:
Agree with @ni4 . We shouldn't mix lib and CLI responsibilities.

dewyatt:
The `rnp_password_provider_stdin` provider is only meant to be used by the CLI utils, that's why it has print statements. It would be good to move it somewhere else though.

ni4:
@dewyatt I think we should create some 'clicommon' source file where it should be put in, among with other sources which may be shared between `rnp`,  `rnpkeys` and `rnp_tests`

PR #753, opened 2018-09-25
Remove pgp_io_t

### LINKED ISSUE #157: Remove pgp_io_t usage
The usage of `pgp_io_t` is deemed unnecessary (as discussed with @catap and @dewyatt ) and should be replaced with normal `stdout`/`stderr`.

We should replace all instances with the normal approach.

ni4:
Since we moved to RNP_LOG this structure is no longer needed (especially as input parameter to all key/keyring processing functions).
Also in CLI we used it only for res stream, and in FFI - for err stream, which were replaced with single stream variable.

Fixes #157

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -339,69 +339,10 @@ set_pass_fd(rnp_t *rnp, int passfd)
 {
     rnp->passfp = fdopen(passfd, "r");
     if (!rnp->passfp) {
-        fprintf(rnp->io->errs, "cannot open fd %d for reading\n", passfd);
+        RNP_LOG("cannot open fd %d for reading", passfd);
         return false;
     }
     return true;
-}
-
-/* Initialize a RNP context's io stream handles with a user-supplied
- * io struct. Returns 1 on success and 0 on failure. It is the caller's
- * responsibility to de-allocate a dynamically allocated io struct
- * upon failure.
- */
-static bool
-init_io(rnp_t *rnp, pgp_io_t *io, const char *outs, const char *errs, const char *ress)
-{
-    /* TODO: I think refactoring can go even further here. */
-
-    /* Configure the output stream. */
-    io->outs = stdout;
-    if (outs && strcmp(outs, "<stderr>") == 0) {
-        io->outs = stderr;
-    }
-
-    /* Configure the error stream. */
-    io->errs = stderr;
-    if (errs && strcmp(errs, "<stdout>") == 0) {
-        io->errs = stdout;
-    }
-
-    /* Configure the results stream. */
-    if (ress == NULL) {
-        io->res = io->errs;
-    } else if (strcmp(ress, "<stdout>") == 0) {
-        io->res = stdout;
-    } else if (strcmp(ress, "<stderr>") == 0) {
-        io->res = stderr;
-    } else {
-        if ((io->res = fopen(ress, "w")) == NULL) {
-            fprintf(stderr, "cannot open results %s for writing\n", ress);
-            return false;
-        }
-    }
-
-    rnp->io = io;
-    return true;
-}
-
-/* Allocate a new io struct and initialize a rnp context with it.
- * Returns 1 on success and 0 on failure.
- *
- * TODO: Set errno with a suitable error code.
- */
-static bool
-init_new_io(rnp_t *rnp, const char *outs, const char *errs, const char *ress)
-{
-    pgp_io_t *io = (pgp_io_t *) calloc(1, sizeof(*io));
-
-    if (io != NULL) {
-        if (init_io(rnp, io, outs, errs, ress))
-            return true;
-        free((void *) io);
-    }
-
-    return false;
 }
 
 /*************************************************************************/
@@ -412,8 +353,7 @@ init_new_io(rnp_t *rnp, const char *outs, const char *errs, const char *ress)
 rnp_result_t
 rnp_init(rnp_t *rnp, const rnp_params_t *params)
 {
-    bool      coredumps = true;
-    pgp_io_t *io;
+    bool coredumps = true;
 
     /* If system resource constraints are in effect then attempt to
      * disable core dumps.
@@ -430,11 +370,15 @@ rnp_init(rnp_t *rnp, const rnp_params_t *params)
           stderr);
     }
 
-    /* Initialize the context's io streams apparatus. */
-    if (!init_new_io(rnp, params->outs, params->errs, params->ress)) {
-        return RNP_ERROR_GENERIC;
+    /* Configure the results stream. */
+    if (!params->ress || !strcmp(params->ress, "<stderr>")) {
+        rnp->resfp = stderr;
+    } else if (strcmp(params->ress, "<stdout>") == 0) {
+        rnp->resfp = stdout;
+    } else if (!(rnp->resfp = fopen(params->ress, "w"))) {
+        fprintf(stderr, "cannot open results %s for writing\n", params->ress);
+        return RNP_ERROR_BAD_PARAMETERS;
     }
-    io = rnp->io;
 
     // set the key provider
     rnp->key_provider.callback = rnp_key_provider_keyring;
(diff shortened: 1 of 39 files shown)


src/lib/rnp.cpp
```diff
@@ -472,14 +414,14 @@ rnp_init(rnp_t *rnp, const rnp_params_t *params)
     if (params->pubpath) {
         rnp->pubring = rnp_key_store_new(params->ks_pub_format, params->pubpath);
         if (rnp->pubring == NULL) {
-            fputs("rnp: can't create empty pubring keystore\n", io->errs);
+            fprintf(stderr, "rnp: can't create empty pubring keystore\n");
```
randombit: Is there a reason this isn't `RNP_LOG`?

src/lib/rnp.cpp
```diff
@@ -472,14 +414,14 @@ rnp_init(rnp_t *rnp, const rnp_params_t *params)
...
             return RNP_ERROR_BAD_PARAMETERS;
         }
     }
     if (params->secpath) {
         rnp->secring = rnp_key_store_new(params->ks_sec_format, params->secpath);
         if (rnp->secring == NULL) {
-            fputs("rnp: can't create empty secring keystore\n", io->errs);
+            fprintf(stderr, "rnp: can't create empty secring keystore\n");
```
randombit: Same

src/librekey/rnp_key_store.cpp
```diff
@@ -598,15 +581,15 @@ rnp_key_store_merge_key(pgp_key_t *dst, const pgp_key_t *src)
...
 pgp_key_t *
-rnp_key_store_add_key(pgp_io_t *io, rnp_key_store_t *keyring, pgp_key_t *srckey)
+rnp_key_store_add_key(rnp_key_store_t *keyring, pgp_key_t *srckey)
 {
     pgp_key_t *added_key = NULL;
 
-    if (io && rnp_get_debug(__FILE__)) {
-        fprintf(io->errs, "rnp_key_store_add_key\n");
+    if (rnp_get_debug(__FILE__)) {
+        fprintf(stderr, "rnp_key_store_add_key\n");
```
randombit: `RNP_LOG`?

src/librekey/rnp_key_store.cpp
```diff
@@ -643,8 +625,8 @@ rnp_key_store_add_key(pgp_io_t *io, rnp_key_store_t *keyring, pgp_key_t *srckey)
         return NULL;
     }
 
-    if (io && rnp_get_debug(__FILE__)) {
-        fprintf(io->errs, "rnp_key_store_add_key: keyc %lu\n", list_length(keyring->keys));
+    if (rnp_get_debug(__FILE__)) {
+        fprintf(stderr, "rnp_key_store_add_key: keyc %lu\n", list_length(keyring->keys));
```
randombit: `RNP_LOG`?

src/librekey/rnp_key_store.cpp
```diff
@@ -438,11 +424,11 @@ rnp_key_store_free(rnp_key_store_t *keyring)
    \return none
 */
 bool
-rnp_key_store_list(pgp_io_t *io, const rnp_key_store_t *keyring, const int psigs)
+rnp_key_store_list(const rnp_key_store_t *keyring, const int psigs)
 {
     unsigned keyc = (keyring != NULL) ? list_length(keyring->keys) : 0;
 
-    (void) fprintf(io->res, "%u key%s\n", keyc, (keyc == 1) ? "" : "s");
+    (void) fprintf(stdout, "%u key%s\n", keyc, (keyc == 1) ? "" : "s");
```
randombit: Maybe this function should take a `FILE` (or `std::ostream`?) which accepts the output, instead of it always being printed to stdout?

src/librekey/rnp_key_store.cpp
```diff
@@ -695,13 +677,12 @@ rnp_key_store_remove_key_by_id(pgp_io_t *io, rnp_key_store_t *keyring, const uin
...
 pgp_key_t *
-rnp_key_store_get_key_by_id(pgp_io_t *             io,
-                            const rnp_key_store_t *keyring,
+rnp_key_store_get_key_by_id(const rnp_key_store_t *keyring,
                             const uint8_t *        keyid,
                             pgp_key_t *            after)
 {
     if (rnp_get_debug(__FILE__)) {
-        fprintf(io->errs, "searching keyring %p\n", keyring);
+        fprintf(stderr, "searching keyring %p\n", keyring);
```
randombit: `RNP_LOG`?

src/librekey/rnp_key_store.cpp
```diff
@@ -717,8 +698,8 @@ rnp_key_store_get_key_by_id(pgp_io_t *             io,
          key_item = list_next(key_item)) {
         pgp_key_t *key = (pgp_key_t *) key_item;
         if (rnp_get_debug(__FILE__)) {
-            hexdump(io->errs, "keyring keyid", key->keyid, PGP_KEY_ID_SIZE);
-            hexdump(io->errs, "keyid", keyid, PGP_KEY_ID_SIZE);
+            hexdump(stderr, "keyring keyid", key->keyid, PGP_KEY_ID_SIZE);
+            hexdump(stderr, "keyid", keyid, PGP_KEY_ID_SIZE);
```
randombit: I feel like we want a `RNP_LOG_BIN` for things like this.

src/rnp/rnp.cpp
```diff
@@ -270,37 +270,37 @@ rnp_on_signatures(pgp_parse_handler_t *handler, pgp_signature_info_t *sigs, int
...
-            key = rnp_key_store_get_key_by_id(io, handler->ctx->rnp->pubring, keyid, NULL);
-            repgp_print_key(io, handler->ctx->rnp->pubring, key, "signature ", 0);
+            key = rnp_key_store_get_key_by_id(handler->ctx->rnp->pubring, keyid, NULL);
+            repgp_print_key(handler->ctx->rnp->pubring, key, "signature ", 0);
         }
     }
 
     if (count == 0) {
-        fprintf(io->res, "No signature(s) found - is this a signed file?\n");
+        fprintf(stderr, "No signature(s) found - is this a signed file?\n");
```
randombit: Should these just go to `resfp` instead?

ronaldtse:
> not a fan of hardcoding use of stdout anywhere within the library - it seems like this should be some kind of FILE* or std::ostream& arg instead so the application can always control where the output goes.
> - @randombit 

I agree and in fact it is a general rule we should apply across the whole application. Other than changes made here, perhaps we need a separate effort in cleaning up these references?

src/lib/rnp.cpp
ni4: Here and in places below I just made changes to use stderr directly instead of io->errs. 
Some of the places can be definitely changed to RNP_LOG, but I was in doubt which ones - i.e. where we are 'showing the message to the CLI user' and should use separate function, and where we have just an error message. Will update the PR anyway.

ni4:
@randombit @ronaldtse I also agree that writing directly to stdout is not a good idea. However, I left it as it is to address later on in separate issues - like #572 and #542. We need some separate function/method for input/output via CLI.

src/librekey/rnp_key_store.cpp
ni4: Yeah, makes sense. Added argument and will pass rnp->resfp there.

src/librekey/rnp_key_store.cpp
ni4: I added RNP_DHEX(), which also wraps if (rnp_get_debug(...))

src/rnp/rnp.cpp
ni4: Not sure here - resfp should get results, but this is more like error message/message to the user (i.e. to the terminal). I'll leave this one to be resolved in another PR (see comment below).

ronaldtse: Agreed. Merging now.

PR #756, opened 2018-09-28
Display key id/fingerprint/grip in packet listing

ni4:
Just needed this functionality within other task (correcting key grip calculation).
Also now times/s2k iterations are printed as unsigned.


ronaldtse:
This happened so fast 😆  -- thank you guys!

PR #760, opened 2018-10-03
Fix key grip calculation

### LINKED ISSUE #540: Some key grips are calculated incorrectly
## Description

Key grips are not calculated correctly for all PK types.
With Elgamal keys, for example, the calculated grip does not match up with GnuPG.

I had to dig into the libgcrypt source a bit and I can see that there is a "generic" keygrip calc method that is used for certain PK types:
https://github.com/gpg/libgcrypt/blob/ce1cbe16992a7340edcf8e6576973e3508267640/cipher/pubkey.c#L669
We have not implemented the generic method, which appears to be used for elg, as can be seen here:
https://github.com/gpg/libgcrypt/blob/ce1cbe16992a7340edcf8e6576973e3508267640/cipher/elgamal.c#L1136
Note the missing keygrip computation func, compared to RSA:
https://github.com/gpg/libgcrypt/blob/ce1cbe16992a7340edcf8e6576973e3508267640/cipher/rsa.c#L1624

## Steps to Reproduce

1.` gpg --homedir=$PWD/src/tests/data/keyrings/1/ --list-keys --with-keygrip`
2. ` src/rnpkeys/rnpkeys --homedir=$PWD/src/tests/data/keyrings/1 --list-keys` (actually we don't list grips right now, so just add a hexdump call in to `rnp_key_store_add_keydata` maybe)

### LINKED ISSUE #558: Remove ``grip_hash_bignum`` and use ``BN_hash`` instead
## Description

See #555 

We could replace ``grip_hash_bignum`` with ``BN_hash stuff``. Implementation of both functions does same thing, but in slidely different way.
This task requires regeneration of test data as current key rings store value of keygrip precalculated with ``grip_hash_bignum`` function

BTW: This may be related to #540  (indeed, I confirm GPG keygrips are different than RNP, but I havn't found any specification which would describe algorithm for keygrip calculation)

ni4:
This PR fixes key grip calculation so key grips are equal to ones calculated with GnuPG.

Fixes #540 
Fixes #558


src/lib/misc.cpp
```diff
@@ -838,3 +838,80 @@ userid_to_id(const uint8_t *userid, char *id)
...
+        if ((hexid[i] == ' ') || (hexid[i] == '\t')) {
+            continue;
+        }
+        return false;
+    }
+    return true;
+}
+/* convert hex string, probably prefixes with 0x, to binary form */
+bool
+hex2bin(const char *hex, size_t hexlen, uint8_t *bin, size_t len, size_t *out)
```
dewyatt: Can you use `rnp_hex_decode` instead?

src/lib/misc.cpp
```diff
@@ -838,3 +838,80 @@ userid_to_id(const uint8_t *userid, char *id)
     id[8 * 2] = 0x0;
     return id;
 }
+
+/* check whether string is hex */
+bool
+ishex(const char *hexid, size_t hexlen)
```
dewyatt: Similar to the below, I think we have existing functionality for this.

dewyatt: Ok but in the future I would say we should be moving towards a more C++-y approach.

dewyatt:
I added a small commit to resolve some TODOs in the tests that were waiting on this exact fix.

Issue #436, opened 2017-09-03
Support selectable hash functions in SM2


ronaldtse:
## Description

Refer to https://github.com/riboseinc/rfc-openpgp-oscca/pull/2

## Steps to Reproduce

1.  SM2DSA, SM2PKE should allow selection of internal hash function

## Expected Behavior

* Generation of SM2 public key
* Public-key encryption using SM2
* Signatures using SM2DSA / SM2PKE

## Actual Behavior

SM2 is hardcoded to use SM3 right now.

ronaldtse:
Thanks @randombit !

randombit:
Looking into resolving this issue, I find there is a problem with SM2 signatures as implemented in RNP. The digest is provided pre-hashed within RNP (by SM3, SHA-256, SHA-512, ...) and it is then signed using a botan pk_sign operator initialized with padding string "". Which causes SM2 to use the default userid field, and hash the message with SM3. As a result the ZA string is computed with the wrong digest (for non-SM3 hash), and a secondary SM3 is applied on the already hashed input.

Botan's current C API could be made to work, if RNP were rearranged to hash the message incrementally (through `botan_pk_op_verify_update`) instead of giving just the digest to the signature algorithm. However incremental processing for signatures requires a significant restructuring in RNP. It is probably better to change Botan SM2 to support a Raw(H) digest, which accepts a pre-hashed input. This already existed for RSA/DSA/ECDSA, but not for SM2. Fixing this bug means SM2 in RNP will behave incompatibly (existing signatures, including self-signatures, will no longer verify).

@ronaldtse thoughts?

ronaldtse:
Oh no -- thank you for discovering this issue! I can make a guess that no one is using SM2 yet; let's fix this before anyone does. Whatever you think is the better way is fine.

PR #761, opened 2018-10-05
Correct SM2 signatures + misc refactorings

randombit:
The main goal of this PR is to address the bug described in #436, where SM2 signatures were actually generated by hashing the hash of the message. Fixing this requires Botan 2.8 or higher which added support for "Raw" SM2 signatures in https://github.com/randombit/botan/pull/1679. In this mode the signature scheme expects to be provided a digest directly.

Plus various refactorings and cleanups done along the way:

 * move the functions involved with directly creating/verifying the signature to `crypto/signature.cpp`
 * convert some macros in `utils.h` into inline functions
 * remove some unneeded functions from bn (some were there just to support a test for elgamal in `cipher.cpp`, this was a holdover from when RNP implemented ElGamal directly instead of calling Botan's version).
 * signature verification functions all took an `rng_t` argument, this was in the end solely needed because `rsa_verify_pkcs1` took an RNG - which it ignored. I removed them all.

This adds tests for creating and verifying SM2 signatures using SM3 and SHA-256. The signatures it verifies were created with GmSSL.


ronaldtse:
Thank you guys!!

PR #764, opened 2018-10-06
Support SM2 with arbitrary hashes

### LINKED ISSUE #436: Support selectable hash functions in SM2

randombit:
Mostly done in #761, this adds support for setting the hash in `rnpkeys`

Required some changes to rnpcfg as there was no way to distinguish between the user having set the hash to SHA-256 on the command line, vs that just having been the default. Now hash defaults to SHA256 as before, except SM2 which still uses SM3 by default. But in all cases an alternative can be set with `--hash`.

Closes #436

### DIFF
--- a/src/rnp/rnpcfg.cpp
+++ b/src/rnp/rnpcfg.cpp
@@ -74,7 +74,6 @@ rnp_cfg_load_defaults(rnp_cfg_t *cfg)
 {
     rnp_cfg_setbool(cfg, CFG_OVERWRITE, false);
     rnp_cfg_setstr(cfg, CFG_OUTFILE, NULL);
-    rnp_cfg_setstr(cfg, CFG_HASH, DEFAULT_HASH_ALG);
     rnp_cfg_setint(cfg, CFG_ZALG, DEFAULT_Z_ALG);
     rnp_cfg_setint(cfg, CFG_ZLEVEL, DEFAULT_Z_LEVEL);
     rnp_cfg_setstr(cfg, CFG_CIPHER, DEFAULT_SYMM_ALG);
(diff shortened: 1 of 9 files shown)


src/rnp/rnpcfg.cpp
```diff
@@ -319,6 +324,10 @@ rnp_cfg_getstr(const rnp_cfg_t *cfg, const char *key)
         return it->val.val._string;
     }
 
+    if (strcmp(key, CFG_HASH) == 0) {
```
dewyatt: This seems weird, @ni4 what do you think?

ni4: @dewyatt Agree, to hacky. I think this check should be moved to place where hash algorithm is checked (or moved to function like `rnp_cfg_gethashalg()`)

ronaldtse:
@randombit will this be updated soon for the merge? Thanks!

randombit:
@ronaldtse Yes I have been away this week but will update this PR later today.

randombit:
@ni4 @dewyatt Updated per comments, also I fixed #766 while in there.

randombit:
It appears the failing test is directly testing the behavior considered a bug in #766 - it sets an invalid hash and tests if it is being ignored. I would vote for changing the test. @dewyatt @ronaldtse

ronaldtse:
@randombit fully support correcting the test!

dewyatt:
Agreed, seems like that test can be removed.

randombit:
Test updated

PR #768, opened 2018-10-29
Support additional ECC curves

### LINKED ISSUE #615: Support for additional curves
## Description

New draft of the standards introduces some changes with which code base should be aligned
* Curve25519
    * it is added as SHOULD but in reality it's a must have as that's a curve which probably will be used the most with ECDH
    * it was always our plan to add it anyway
* Brainpool curves
    * they are added as MAY but as those curves use short Weierstrass form, so adding them should be easy even if not supported by botan (which I believe is not a case)
* compressed point format for curves expressed in Montgomery form - we only support uncompressed format (except for EdDSA)
* test for a zero point

ni4:
This PR adds support for the following curves:
- Brainpool p256, p384, p512 ECDSA/ECDH
- secp256k1 ECDSA/ECDH (it is not described in rfc4880bis but support by GnuPG and was easy to add)
- x25519 (i.e. ECDH with curve25519)

Also extends tests, adds some fixes and CLI support for these curves.

Fixes #615


src/lib/crypto/ecdh.cpp
```diff
@@ -261,9 +276,20 @@ ecdh_encrypt_pkcs5(rng_t *                  rng,
...
     }
 
-    out->p.len = sizeof(out->p.mpi);
-    if (botan_pk_op_key_agreement_export_public(eph_prv_key, out->p.mpi, &out->p.len)) {
-        goto end;
+    /* we need to prepend 0x40 for the x25519 */
+    if (key->curve == PGP_CURVE_25519) {
+        out->p.len = sizeof(out->p.mpi) - 1;
+        if (botan_pk_op_key_agreement_export_public(
+              eph_prv_key, out->p.mpi + 1, &out->p.len)) {
```
randombit: For x25519 does there need to be a byte reversal here?

src/lib/crypto/ec.cpp
```diff
@@ -158,6 +235,46 @@ get_curve_desc(const pgp_curve_t curve_id)
...
 
+rnp_result_t
+x25519_generate(rng_t *rng, pgp_ec_key_t *key)
+{
+    botan_privkey_t pr_key = NULL;
+    botan_pubkey_t  pu_key = NULL;
+    rnp_result_t    ret = RNP_ERROR_KEY_GENERATION;
+    uint8_t         keyle[32] = {0};
+
+    // at this point it must succeed
```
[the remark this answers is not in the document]
ni4: Yeah, will remove it.

src/lib/crypto/ecdh.cpp
ni4: No, it doesn't seem so, at least it is interoperable with GnuPG. Looks like 0x40 is considered to tell 'treat data as it is'. Will write to mailing list to ensure this since there is no any information in the 4880-bis.

ni4: ...ok, it was already discussed in the following thread: https://lists.gnupg.org/pipermail/gnupg-devel/2018-February/033437.html

ronaldtse:
Any further updates here? Would love to merge this soon. Thanks!

ni4:
@randombit Any further comments?
@dewyatt Please take a look on this once you have some time.

src/lib/crypto/ecdh.cpp
dewyatt: @randombit Was this point satisfied?

ronaldtse:
@randombit are we all good here? Can we merge this soon? Thanks!

ronaldtse:
I believe @dewyatt is busy fixing cars 😉

PR #770, opened 2018-11-14
Fix AEAD functionality

### LINKED ISSUE #769: AEAD cipher doesn't work with the latest botan
## Description

AEAD support just broken right now.

## Steps to Reproduce

`rnp --passwords 1 --aead -z 0 -c 100k.dat --output 100k.dat.encrypted`

ni4:
AEAD context was not correctly reseted, however it worked with previous botan versions.
Fixes #769


dewyatt:
@ni4 Should we have added a test here? I didn't think about it before.

ni4:
@dewyatt Existing CLI tests catched this issue, so I didn't add any specific test for this case

dewyatt:
I see now, thanks!

PR #771, opened 2018-11-14
Correctly copy all parsed pgp_key_t fields during load

### LINKED ISSUE #763: g10 secret keys cannot be used by userid in CLI.
## Description

It looks like regression, however I'm not sure whether this worked before.
List of user ids field is not filled for g10 secret keys.

## Steps to Reproduce

1. `rnp --homedir .gpg -u "userid" --sign somefile.txt`

## Expected Behavior

rnp finds key by userid and signs.

## Actual Behavior

Fails since secret key is not found.

ni4:
This PR was aimed to fix #763.
As the side effect it changed the way keys are copied - now all fields are copied instead of reparsing to/from transferable key.

Fixes #763 

Now travis will fail - need to wait till #770 is merged.

### DIFF
--- a/src/lib/defs.h
+++ b/src/lib/defs.h
@@ -136,24 +136,24 @@
         }                                                                            \
     } while (/*CONSTCOND*/ 0)
 
-#define EXPAND_ARRAY_EX(str, arr, num)                                                \
-    do {                                                                              \
-        if (str->arr##c == str->arr##vsize) {                                         \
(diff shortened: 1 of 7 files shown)


ni4:
@ronaldtse @dewyatt @randombit Any comments on this? Thanks!
Codecov doesn't seem happy but it looks like mostly because of code lines which checks memory allocation failure.

PR #772, opened 2018-11-21
Remove DYNARRAY and refactor for easier later C++ conversion

ni4:
This PR removes DYNARRAY and other #define-based array allocations, removing defs.h header.
Also it replaces direct field usage with accessor methods for easier later transition to C++ classes.


src/lib/rnp2.cpp
```diff
@@ -3152,15 +3155,16 @@ key_get_uid_at(pgp_key_t *key, size_t idx, char **uid)
     if (!key || !uid) {
         return RNP_ERROR_NULL_POINTER;
     }
-    if (idx >= key->uidc || DYNARRAY_IS_EMPTY(key, uid)) {
+    if (idx >= pgp_get_userid_count(key)) {
```
randombit: This removes the check for UIDs being empty, does it matter?

ronaldtse: @ni4 could you have a look here? Thanks!

ni4: Empty uid list equals to `pgp_get_userid_count() == 0`, which will satisfy the `idx >=` check  since idx is unsigned.

ni4:
@randombit Please see the last commit, I described all list elements there.

src/lib/rnp2.cpp
dewyatt: Agree with @ni4, this should be good. It looks like the `DYNARRAY_IS_EMPTY` check here before was bogus (added by yours truly :P).

ronaldtse: Thanks guys!

PR #775, opened 2018-11-29
Add some API usage examples

### LINKED ISSUE #773: Usage of RNP in key-generation and to encrypt/decrypt memory
I got this message from @r10s about using RNP in https://github.com/deltachat/deltachat-core  .

(Thanks @r10s! )

Could someone help him with this? Thanks!

ping @ni4 @dewyatt 

> hi all,
> 
> this is bjoern from delta chat writing to you.
> 
> delta chat is a messenger that uses smtp/imap to send and received messages.
> 
> for encryption, we currently use netpgp, that we've modified/fixed to
> fit our needs.
> 
> you can imagine, that we're very happy to see netpgp maintained again as
> ribose/rnp.
> 
> apart that we would be very happy to do pr on ribose/rnp, i am a little
> bit lost in the new api.
> 
> what we need is key-generation and to encrypt/decrypt memory, not files.
> i've found rnp_process_mem(), however, i've not found a function that
> takes rnp_ctx_t and allows adding a key to encrypt to.
> 
> also it's not clear how decryption works.
> 
> i would be very happy if you can point me to the correct functions :)
> 
> thanks in advance :)
> 
> oh, i am unsure if this is the correct channel, it would be great if you
> forward the question otherwise :)
> 
> ahh, and, of course you're welcome to join #deltachat on irc :)
> 
> best -
> bjoern

ni4:
These examples include all common functionality (key generation-load-import-export, encrypt/decrypt, sign/verify) and use lower-level API which most likely is subject to change.

While FFI is available and better defined now I think we should have raw C API as well for easier adoption in other C/C++ projects. Just need to stabilize it a bit.

Also development of these examples revealed some weak sides of the base code, so I updated it as well, making handier to use.
Fixes #773

cc @r10s

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,11 @@ rnp --decrypt --homedir=${keyringdir} ${filename}.gpg
 Creates: `${filename}`
 
 
+# For developers
+
+You may wish to use librnp library in your projects as well.
+See the [`src/examples/README.md`](./src/examples/README.md) for the usage details and sample code.
+
 
 # Install
 


dewyatt:
IMO it really doesn't seem like a good idea to encourage end-users to utilize the internal API. It's just going to lead to a lot of upset users.

randombit:
The library changes look good (hiding use of list etc) but agree with @dewyatt that providing example code with internal APIs is going to lead to sadness, people will copy that code and then later changes to rnp will break their applications. What is missing from `rnp2.h` that prevents writing the examples using that?

Also, the demos are compiled as C++ but it looks like they are basically C already, maybe better to compile them as C so it is clear that RNP is exposing a C API despite being written in C++.

ni4:
@dewyatt @randombit Thanks for the input. So, I think as first step I would make another PR with library internals changes, and as the second step - move examples to rnp2.h usage, updating it if needed.

Then, should we stick just to rnp2.h as a single interface? And (at some point) move CLI to this interface as well? What do you think?
cc @ronaldtse

ronaldtse:
I agree with the comments here, we should encourage users to use the external API only.

We should take the opportunity to improve `rnp2.h` (or make the necessary changes to simplify it as `rnp.h`) for external consumption.

@ni4 could you translate the examples into the external API so that people can copy/paste them but not access internal API?

On the other hand, I agree with @ni4 that we should move the CLI to this interface too. Thoughts @dewyatt @randombit ?

PR #780, opened 2018-12-20
Replace pgp_memory_t usage with streams

ni4:
This PR removes old pgp_memory_t, replacing it with pgp_source_t/pgp_dest_t.
Also it adds tests for file streams, and extends streams with some new methods (`init_tmpfile_dst`, `file_to_mem_src`).


ni4:
@dewyatt  Hm, looks like something changed in Travis configuration, do you have any idea? It fails to download public key for rvm signature check.

ni4:
@dewyatt Thank you! Now it passess all tests fine.

PR #781, opened 2018-12-23
Allow to set per-signature parameters

### LINKED ISSUE #651: Improve signing hash algorithm selection for multiple signatures
## Description

Currently there is only one field in `rnp_ctx_t` for signing hash algorithm.
However, this could not work for different key algorithms, so we should have a way to:
- set hash algorithm for each signer;
- pick correct hash algorithm for each signer (if it is not set explicitly). Key flags, algorithm and bit length could be used here.
- update FFI to allow settings of hash algorithm for each signer.

### LINKED ISSUE #774: Add possibility to pass parameter for on_signatures in rnp_process_mem/file.
## Description

Currently there is no way to pass some param to the on_signatures handler when `rnp_process_mem` function is used. It is possible to use own parse handler, however easier solution should be provided.

ni4:
This PR improves control over signatures:
- allows to send custom param to on_signatures handler
- adds structure to set per-signer info (hash algorithm, signature expiration, creation, later on may be extended with other parameters like additional subpackets and so on)
- implements corresponding functions for FFI
- extends FFI tests

Fixes #651
Fixes #774

### DIFF
--- a/include/rnp/rnp2.h
+++ b/include/rnp/rnp2.h
@@ -517,14 +517,16 @@ rnp_result_t rnp_op_sign_add_signature(rnp_op_sign_t            op,
                                        rnp_key_handle_t         key,
                                        rnp_op_sign_signature_t *sig);
 
-/** @brief Set hash algorithm used during signature calculation. Not implemented yet.
+/** @brief Set hash algorithm used during signature calculation instead of default one, or one
+ *         set by rnp_op_encrypt_set_hash/rnp_op_sign_set_hash
  *  @param sig opaque signature context, returned via rnp_op_sign_add_signature
  *  @param hash hash algorithm to be used
  *  @return RNP_SUCCESS or error code if failed
  */
 rnp_result_t rnp_op_sign_signature_set_hash(rnp_op_sign_signature_t sig, const char *hash);
 
-/** @brief Set signature creation time. Not implemented yet.
+/** @brief Set signature creation time. By default current time is used or value set by
+ *         rnp_op_encrypt_set_creation_time/rnp_op_sign_set_creation_time
  *  @param sig opaque signature context, returned via rnp_op_sign_add_signature
  *  @param create creation time in seconds since Jan, 1 1970 UTC
  *  @return RNP_SUCCESS or error code if failed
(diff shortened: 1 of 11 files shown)


dewyatt:
It looks like a ruby-rnp test is going to fail, can we avoid breaking compatibility here or do we need to update ruby-rnp? (I haven't looked closely)

ni4:
@dewyatt Thanks for noticing, I'll take a look later today.

dewyatt:
I feel like this introduces at least one regression. Here's a gist:
https://gist.github.com/dewyatt/eeeee2319f1937129358a7cfa073eae0
`g++ --std=c++11 -I/usr/include/rnp-0 -lrnp-0 rnp-pr-781.cpp -ornp-pr-781`

This will pass in current master but fail in this branch. I think the ruby-rnp changes proposed would have  covered up the issue.

ni4:
@dewyatt This is the same problem that arises in the ruby-rnp - expiration/creation time is set after the signature is added. Within new changes it is copied from 'default' operation value to the signature, so if the value is set after signature addition then it will not affect it. 
Is it good to go for you? I can add some flags like 'creation_set, expiration_set' and then check those flags, but it seems to overcomplicate things - it is easier to add lines 'This method must be called before any signature is added otherwise it will not affect already added signatures'.

dewyatt:
@ni4 I'll take a closer look tomorrow but I think I get what you mean. My initial thought is that it may be helpful to maintain the old behavior, because guidelines like "call this method before X" can be overlooked and lead to subtle bugs (obviously we can just say RTFM, but it may be nice to avoid that situation). `std::optional` is made for exactly these cases but would require switching the standard to c++17.

Anyways, I'd like to take a closer look tomorrow.

dewyatt:
@ni4 So it seems like we have at least 3 options (using expiration time as an example here):
1. Document that `rnp_op_sign_set_expiration_time` will now have different behavior.
2. Maintain the old behavior: `rnp_op_sign_set_expiration_time` will affect all sigs (added before or after the call). Individual signer sigs can be modified with the now-functional `rnp_op_sign_signature_set_expiration_time`.
3. Remove `rnp_op_sign_set_expiration_time` entirely. The option must be set for each signature individually, using `rnp_op_sign_signature_set_expiration_time`.

I'm not liking option 1 so much because it breaks previously written code. While that may generally be OK for an 0.x release, I don't like that it would break in a subtle/hidden way.

@ronaldtse Any thoughts? To be clear, what we're discussing essentially boils down to:
Current behavior:
```ruby
    sign = rnp.start_sign(input: Rnp::Input.from_string('data'),
                          output: output)
    sign.expiration_time = 60 # can be set before adding signers
    sign.add_signer(rnp.find_key(userid: 'key0-uid1'))
    sign.expiration_time = 60 # or after
    sign.execute
```

This branch:
```ruby
    sign = rnp.start_sign(input: Rnp::Input.from_string('data'),
                          output: output)
    sign.expiration_time = 60 # must come before adding signers
    sign.add_signer(rnp.find_key(userid: 'key0-uid1'))
    sign.expiration_time = 9090 # this will have no effect
    sign.execute
```

ronaldtse:
@dewyatt @ni4 I agree with option 2. Appreciate the detailed explanation! 😉

ni4:
@dewyatt @ronaldtse Please see the updated PR.

PR #789, opened 2019-01-08
ci: Fix FreeBSD build.

dewyatt:
* Add some missing pkg installs
* Set `GPG_VERSION`. This is set to `stable` because `beta` failed to build, which IIRC means some AEAD tests will not be run. Maybe someone else can look into that issue.
* Fix some shebang lines to play nice.
* Install bundler as root. There are other ways to solve this, but this is quick and easy.

### DIFF
--- a/ci/before_install.sh
+++ b/ci/before_install.sh
@@ -35,10 +35,23 @@ macos_install() {
 }
 
 freebsd_install() {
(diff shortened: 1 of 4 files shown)


ronaldtse:
Thanks @dewyatt, looks great!

PR #795, opened 2019-01-21
Separate CLI code from library code

ni4:
This PR is the next step on moving to FFI-only library interface.
It moves rnp_t-related CLI code to rnp/rnpcli.h(.cpp) and deleted unused code from repgp.h/repgp.cpp
Once this PR get merged librnp will include only FFI interface and underlying code.


ni4:
@dewyatt @ronaldtse Would you have time to take a look on this? Thanks!

src/lib/crypto/s2k.cpp
```diff
@@ -157,6 +157,10 @@ pgp_s2k_compute_iters(pgp_hash_alg_t alg, size_t desired_msec, size_t trial_msec
     const uint8_t MIN_ITERS = 96;
     uint8_t       buf[8192] = {0};
     size_t        bytes = 0;
+    pgp_hash_t    hash = {};
+    uint64_t      start;
+    uint64_t      end;
+    uint64_t      duration;
```
dewyatt: We don't really have anything official in our C++ guide on this, but I think it's standard practice to make variable declarations as late as possible, so I don't like seeing all the variables moved up top like this (and they should all probably be initialized).

Refs:
* C++ Coding Standards: 101 Rules, Guidelines, and Best Practices
`Declare variables as locally as possible.`
* [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Local_Variables)
`Place a function's variables in the narrowest scope possible, and initialize variables in the declaration.`
* [Mozilla Coding Style](https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style#General_C.2FC.2B.2B_Practices)
`Declare local variables as near to their use as possible.`

src/lib/crypto/s2k.cpp
```diff
@@ -165,11 +169,13 @@ pgp_s2k_compute_iters(pgp_hash_alg_t alg, size_t desired_msec, size_t trial_msec
...
 
-    pgp_hash_t hash;
-    pgp_hash_create(&hash, alg);
+    if (!pgp_hash_create(&hash, alg)) {
+        RNP_LOG("failed to create hash object");
+        return 0;
+    }
 
-    const uint64_t start = get_timestamp_usec();
-    uint64_t       end = start;
```
dewyatt: As above, it feels like a step backwards a bit, I think it's widely accepted that the best practice is to declare and initialize here.

src/lib/crypto/s2k.cpp
ni4: Thanks, @dewyatt ! Honestly I believed (without any reference :) ) that declaration of variables on very top of the function is best practice, but that was wrong. At least for the C++.
Will fix and update PR soon.

src/lib/crypto/s2k.cpp
```diff
@@ -179,10 +179,11 @@ pgp_s2k_compute_iters(pgp_hash_alg_t alg, size_t desired_msec, size_t trial_msec
 
     pgp_hash_finish(&hash, buf);
 
-    const uint64_t duration = end - start;
-
-    if (duration == 0)
+    uint64_t      duration = end - start;
+    const uint8_t MIN_ITERS = 96;
```
dewyatt: Just a note that I forgot to mention but typically the "don't declare variables until needed" guideline doesn't apply to constants, so this one could just as easily be at the top.

C++ Coding Standards: 101 Rules, Guidelines, and Best Practices
```
Declare variables as locally as possible.
...
Because constants don't add state, this Item does not apply to them.
```

(I personally don't mind either way, as we have no official guideline on it)

PR #801, opened 2019-02-15
FFI: non-json key generation interface

### LINKED ISSUE #796: FFI: Implement non-JSON key generation interface
## Description

While we have interface to generate keys, using the JSON format, it is not well usable from low-level languages which doesn't handle JSON natively (like Ruby does).
So it would be good to have function-based interface as well.

## Suggested interface

Following functions are  proposed to implement the same level of functionality which JSON interface has right now (should be subject for discussion):

``` C
rnp_result_t rnp_key_generate_create(rnp_ffi_t ffi, rnp_key_generate_handle_t *handle, const char *alg);

rnp_result_t rnp_subkey_generate_create(rnp_key_generate_handle_t primary, rnp_key_generate_handle_t *handle, const char *alg);

rnp_result_t rnp_key_generate_set_bits(rnp_key_generate_handle_t key, size_t bits);

rnp_result_t rnp_key_generate_set_hash(rnp_key_generate_handle_t key, const char *hash);

rnp_result_t rnp_key_generate_set_dsa_qbits(rnp_key_generate_handle_t key, size_t qbits);

rnp_result_t rnp_key_generate_set_curve(rnp_key_generate_handle_t key, const char *curve);

rnp_result_t rnp_key_generate_set_protection_cipher(rnp_key_generate_handle_t key, const char *cipher);

rnp_result_t rnp_key_generate_set_protection_hash(rnp_key_generate_handle_t key, const char *hash);

rnp_result_t rnp_key_generate_set_protection_mode(rnp_key_generate_handle_t key, const char *mode);

rnp_result_t rnp_key_generate_set_protection_iterations(rnp_key_generate_handle_t key, size_t iterations);

rnp_result_t rnp_key_generate_add_usage(rnp_key_generate_handle_t key, const char *usage);

rnp_result_t rnp_key_generate_remove_usage(rnp_key_generate_handle_t key, const char *usage);

rnp_result_t rnp_key_generate_set_userid(rnp_key_generate_handle_t key, const char *userid);

rnp_result_t rnp_key_generate_set_expiration(rnp_key_generate_handle_t key, uint32_t expiration);

rnp_result_t rnp_key_generate_clear_pref_hashes(rnp_key_generate_handle_t key);

rnp_result_t rnp_key_generate_add_pref_hash(rnp_key_generate_handle_t key, const char *hash);

rnp_result_t rnp_key_generate_clear_pref_compression(rnp_key_generate_handle_t key);

rnp_result_t rnp_key_generate_add_pref_compression(rnp_key_generate_handle_t key, const char *compression);

rnp_result_t rnp_key_generate_clear_pref_ciphers(rnp_key_generate_handle_t key);

rnp_result_t rnp_key_generate_add_pref_cipher(rnp_key_generate_handle_t key, const char *cipher);

rnp_result_t rnp_key_generate_set_pref_keyserver(rnp_key_generate_handle_t key, const char *keyserver);

rnp_result_t rnp_generate_key(rnp_key_generate_handle_t key, rnp_key_generate_handle_t subkey);

rnp_result_t rnp_key_generate_get_key(rnp_key_generate_handle_t genkey, rnp_key_handle_t *key);

rnp_result_t rnp_key_generate_handle_destroy(rnp_key_generate_handle_t *key);

```

ni4:
This PR extends FFI functionality with rnp_op_generate_* functions for key/subkey generation.
Also it adds some key properties getters and 'shortcut functions' for quick key/subkey generation.
For discussion see #796.

Fixes #796


ni4:
@dewyatt Could you please take a look? Most likely you'll have some suggestions on parameters/missing functions and so on.

src/lib/rnp2.cpp
```diff
@@ -3226,180 +3294,775 @@ rnp_generate_key_json(rnp_ffi_t ffi, const char *json, char **results)
...
+    }
+    if (scurve && (ret = rnp_op_generate_set_curve(subop, scurve))) {
+        goto done;
+    }
+    if ((ret = rnp_op_generate_execute(subop))) {
+        goto done;
+    }
+done:
+    if (!ret && key) {
+        *key = primary;
```
dewyatt: * It looks like `key` is an optional arg (which is fine IMO) but will leak if NULL is provided and generation succeeds. I guess you would probably want to do `if (!key) { rnp_key_handle_destroy(primary); }`.
* Adding a line to a test for this would be good too, something like:
`assert_rnp_success(rnp_generate_key_rsa(ffi, 1024, 1024, "test_key_null", NULL));`
* Some clarification in the docs on who frees what would be ideal.

include/rnp/rnp2.h
```diff
@@ -326,6 +330,251 @@ rnp_result_t rnp_generate_key_json(rnp_ffi_t ffi, const char *json, char **resul
...
+ *
+ * @param op pointer to opaque key generation context.
+ * @param keyserver NULL-terminated string with key server's URL, or NULL to delete it from
+ *                  user preferences.
+ * @return RNP_SUCCESS or error code if failed.
+ */
+rnp_result_t rnp_op_generate_set_pref_keyserver(rnp_op_generate_t op, const char *keyserver);
+
+/** Execute the prepared key or subkey generation operation.
+ *  Note: if you set protection algorithm, then you need to specify ffi password provider to
```
dewyatt: I don't think we should use the password provider for this, we should allow setting it directly with something like `rnp_op_generate_set_protection_password`. That would be more consistent with `rnp_key_protect`, `rnp_key_unprotect`, `rnp_op_encrypt_add_password`, etc.

I really intended for the pass provider to be used in cases where a password *may* be required, but it is not certain. In this case, if the user is specifying a protection cipher/etc, a password is *definitely* required.

dewyatt:
I don't think any of the `rnp_op_generate_add_pref_*` functions are working...whenever I use one of them, the corresponding preference is missing from the generated packet.

src/lib/rnp2.cpp
```diff
@@ -3226,180 +3294,775 @@ rnp_generate_key_json(rnp_ffi_t ffi, const char *json, char **results)
...
+rnp_op_generate_add_usage(rnp_op_generate_t op, const char *usage)
+{
+    if (!op || !usage) {
+        return RNP_ERROR_NULL_POINTER;
+    }
+    uint8_t flag = 0;
+    if (!str_to_key_flag(usage, &flag)) {
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+    if (op->primary) {
```
dewyatt: I'm wondering if we ought to do a check with `pgp_pk_alg_capabilities` here. What do you think?

src/lib/rnp2.cpp
ni4: Thanks for noticing this. Will fix. What do you think about making this function public? Make this any sense, or will give more questions from the users?

include/rnp/rnp2.h
ni4: Agree, at least we could add this method but fallback to password provider if required. Will fix this.

src/lib/rnp2.cpp
dewyatt: @ni4 I could see it being useful as a public function. We can just clarify in the docs that it is an alternative to other methods of key generation.
(I'd prefer we make the param names more verbose/descriptive if public though)

ni4: @dewyatt Yeah, good. What about adding password parameter as well, so key/subkey will go out encrypted after the single call?

dewyatt: @ni4 That sounds like a good move, I guess `NULL` would be "no protection"?

ni4: @dewyatt Yeah, exactly.

ni4: @dewyatt I pushed some commits which address this thread and thread below with `rnp_op_generate_set_protection_password` instead of password provider.
Will address other comments a bit later.

ni4:
> I don't think any of the `rnp_op_generate_add_pref_*` functions are working...
Yeah, you are right. PR #804 aims to fix this. Will update this PR as well with some tests.

dewyatt:
I think for the shortcut functions, if we're generating both a primary and sub, we should be calling `pgp_generate_keypair` so that it defaults to "sign+cert" usage for the primary and "encrypt" for the sub.
Right now it defaults to "everything the algorithm supports" for both, which is different from what `rnpkeys` and `gpg` do.

src/lib/rnp2.cpp
ni4: @dewyatt Yeah, correct. Just pushed a commit with some tests.

ni4:
> I think for the shortcut functions, if we're generating both a primary and sub, we should be calling `pgp_generate_keypair` so that it defaults to "sign+cert" usage for the primary and "encrypt" for the sub.

@dewyatt  As I checked `pgp_generate_keypair`, it only differs that it calls the following code:
```C
    if (!primary_desc->cert.key_flags && !subkey_desc->binding.key_flags) {
        primary_desc->cert.key_flags = PGP_KF_SIGN | PGP_KF_CERTIFY;
        subkey_desc->binding.key_flags = PGP_KF_ENCRYPT;
    }
```
What about setting default flags in shortcut function as well? One of the ideas was that shortcut functions as much as possible rely on rnp_op_generate_* functions.

dewyatt:
@ni4 That sounds reasonable to me.

ni4:
@dewyatt Please see the latest commit. Are there any other comments?

PR #802, opened 2019-02-18
ci: Add MacOS

dewyatt:
I ran into a few issues but in the end I've reduced it down to this.
I ran into some linker errors w/botan master, so I set it to use 2.9.0, which seems reasonable to me.


ronaldtse:
Merging since this is only CI.

PR #804, opened 2019-02-28
Fix pgp_user_prefs_add_val function: add items correctly

ni4:
This PR fixes error in aforementioned function (used to add preferred sym algs/hashes/etc), improves it to not allow duplicates, and adds tests.


ronaldtse:
Agree with @dewyatt , we can start taking advantage of the newer primitives 👍

PR #815, opened 2019-03-17
Fix double-free with invalid armored headers.

dewyatt:
Current master will crash on this test, so this fixes that.


dewyatt:
CI is going to fail because of https://github.com/riboseinc/rnp/pull/810 here, so maybe we should merge that first and rebase this.

Issue #816, opened 2019-03-18
Using tests under doc/


strogonoff:
A question regarding https://github.com/riboseinc/rnp/blob/master/doc/tests/rnpkeys-generate-key.md:

* Am I correct that it’s intended to be a manual testing aid?
* Is it for a subset of tests (perhaps integration/e2e) that can’t be easily automated?
* How does it go together with https://github.com/riboseinc/rnp/tree/master/src/tests?

This question arose while updating RNP site and integrating development documentation.

strogonoff:
From https://github.com/riboseinc/rnp/tree/master/doc/tests it seems that it’s used for acceptance tests. If that’s correct and these tests are up-to-date and should be kept, feel free to close!

strogonoff:
(Note: I might move these tests under docs/develop/tests/, but other than that won’t change. Unless you let me know that this is obsolete, in which case I’ll drop it.)

dkg:
without any clearer rationale for the split, i do support moving `doc/tests` somewhere under `docs/` (`docs/develop/tests` sounds resaonable to me) -- it's confusing to see both folders.

PR #817, opened 2019-03-18
Migrate develop docs to AsciiDoc and under docs/

strogonoff:


dewyatt:
It seems odd to have a top-level [`README.adoc`](https://github.com/strogonoff/rnp/blob/master/README.adoc) and also a [`docs/README.adoc`](https://github.com/strogonoff/rnp/blob/master/docs/README.adoc), especially when they start out very similarly.

docs/installation.adoc
```diff
@@ -0,0 +1,55 @@
...
+yum install -y rnp
+----
+
+== On Debian
+
+[source,console]
+----
+# Clone the repository by version tag (or omit it to get the latest sources)
+sudo apt install git
+git clone https://github.com/riboseinc/rnp.git -b v0.11.0
```
dewyatt: We have an 0.12.0 release -- I wonder if we should try to keep this line up-to-date?

ronaldtse: We should definitely keep installation steps up to date, but I wonder if this one in particular we can use some sort of variable...?

strogonoff: How about adding this item to release checklist @dewyatt 
Alongside any changelog updates and such

strogonoff: @dewyatt if no objections, I’ll add this to the release workflow doc (around where it tells the maintainer to update CHANGELOG)

dewyatt: @ronaldtse I'm not aware of any at the moment. @strogonoff That sounds like a good idea to me.

ronaldtse:
@strogonoff as @dewyatt mentioned, is the `docs/README.adoc` supposed to take a new life of its own?

strogonoff:
> It seems odd to have a top-level [`README.adoc`](https://github.com/strogonoff/rnp/blob/master/README.adoc) and also a [`docs/README.adoc`](https://github.com/strogonoff/rnp/blob/master/docs/README.adoc), especially when they start out very similarly.

@dewyatt that README shouldn’t be there, apparently I moved it and then accidentally recreated by writing from previously open Vim buffer so it didn’t make it in the commit. Deleted now.

strogonoff:
To clarity, now the README was moved back into the root. Parts of it were moved in respective files under `docs/` where they will be seen by the project site, and README now links to those.

Sometimes things move back and forth as I’m trying to sort out the best way to handle this, so please don’t mind that.

Thanks for noticing the stale README (I wouldn’t have noticed that until later)

ronaldtse:
@strogonoff are we ready to merge?

strogonoff:
@ronaldtse Yes, ready as of now.

Sorry, forgetting to add the no-ci tag at commit time

skalee:
Perhaps symlinking `README.adoc` to `docs/README.adoc` or _vice versa_ would solve the issue with maintaining two files?

dewyatt:
I have some other comments (might be outside the scope of this PR) after looking again:
1. I see we have both a [`doc/tests/`](https://github.com/riboseinc/rnp/tree/63e69a4908eec5b3c41ef25b24464559225dc607/doc/tests) and [`docs/`](https://github.com/riboseinc/rnp/tree/63e69a4908eec5b3c41ef25b24464559225dc607/docs). I know you have an open issue for this (#816) but I still thought it might be good to move it into `docs/` (until we have guidance on whether to remove it entirely). I think it will confuse people.
2. It looks like [`docs/develop.adoc`](https://github.com/riboseinc/rnp/tree/63e69a4908eec5b3c41ef25b24464559225dc607/docs/develop.adoc) would fit nicely in [`docs/develop/`](https://github.com/riboseinc/rnp/tree/63e69a4908eec5b3c41ef25b24464559225dc607/docs/develop). If it were a top-level doc that linked to the items in the `docs/develop` directory, then it would make sense, but it doesn't appear to be that. What do you think?
3. I assume [`docs/navigation.adoc`](https://github.com/riboseinc/rnp/tree/63e69a4908eec5b3c41ef25b24464559225dc607/docs/navigation.adoc) is maybe something for rnpgp.com and not intended to be viewed on GitHub? In that case, I think ideally we would make it appear like less of an end-user file. Mainly I think the name sounds too user-friendly and people will click on it thinking it will be helpful to navigate the docs, but it won't be. Could it be moved to a top-level `.ribose/` or something else?

Also, generally we need to distinguish between "internal development docs" and "development docs" (API docs for end-users). I think having a `docs/internal` might make sense at some point.

strogonoff:
> Perhaps symlinking `README.adoc` to `docs/README.adoc` or vice versa would solve the issue with maintaining two files?

@skalee the brief simultaneous existence of README.adoc and docs/README.adoc was an accident, the one under docs/ has been deleted as it should’ve been in the first place.

strogonoff:
@dewyatt 

1. If there’re no objections I’ll move test docs under docs/ as well in future

2. https://www.rnpgpg.com/software/rnp/docs/develop/ is intended to be the entry point to further development docs nested under the /develop/ path

3. It’s a valid concern, currently the navigation doc is required by open project site during Jekyll build, in future the navigation may be moved elsewhere or changed so that document contents contain user-clickable navigation.

4. I don’t think there’s any confusion. You have docs about _using_ RNP through its CLI tool and/or C APIs, and about _developing_ RNP itself, and that’s how they’re phrased. Unless you’re closed-source or closed for contributions, I believe calling contribution docs “internal” would be a mistake. It’s great that you have contribution guides, this is not common across Ribose Open projects currently!

I believe the situation with RNP docs will be improved with these changes, if that is so then we can file improvement issues later rather than delay the merge, on the other hand if there’re any show-stoppers then I’d like to hear them!

dewyatt:
I think I forgot to merge this!

PR #822, opened 2019-03-27
Add rnp_unload_keys function to the FFI.

ni4:
We didn't have such function before, and it is useful sometimes (at least need this for CLI to FFI transition).

### DIFF
--- a/include/rnp/rnp2.h
+++ b/include/rnp/rnp2.h
@@ -47,6 +47,9 @@ typedef uint32_t rnp_result_t;
 #define RNP_KEY_REMOVE_PUBLIC (1U << 0)
 #define RNP_KEY_REMOVE_SECRET (1U << 1)
 
+#define RNP_KEY_UNLOAD_PUBLIC (1U << 0)
+#define RNP_KEY_UNLOAD_SECRET (1U << 1)
+
 /**
  * Flags for optional details to include in JSON.
  */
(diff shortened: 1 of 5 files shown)


dewyatt:
Any thoughts on calling it `rnp_unload_keys`? `rnp_clear_keys` might be slightly ambiguous -- when I first read the name I thought it was zeroing key material or something.

ni4:
@dewyatt Yeah, thanks. Don't really like _clear_keys but was not able to find better alternative.

What do you think about the flags?
As I just remembered, we have `RNP_KEY_REMOVE_PUBLIC`/`RNP_KEY_REMOVE_SECRET` constants as well. 
Or would it be better to add `RNP_KEY_UNLOAD_*` ones?

dewyatt:
> @dewyatt Yeah, thanks. Don't really like _clear_keys but was not able to find better alternative.
> 
> What do you think about the flags?
> As I just remembered, we have `RNP_KEY_REMOVE_PUBLIC`/`RNP_KEY_REMOVE_SECRET` constants as well.
> Or would it be better to add `RNP_KEY_UNLOAD_*` ones?

Oh yeah I would probably create new `RNP_KEY_UNLOAD_*` ones, that's what users will likely search for after seeing the rest of the API.

(We probably should have created some generic ones but at this point we would break backwards-compat.)

ni4:
@dewyatt Ok, will update the PR a bit later.

ni4:
@dewyatt Please see the updated PR.

PR #823, opened 2019-03-31
Change all references from rnp2 to rnp.

dewyatt:
rnp2 was an internal transitional name that was useful for a time but is no longer necessary and may cause confusion.


ronaldtse:
@dewyatt could you help rebase and merge? Thanks!

PR #829, opened 2019-04-10
Fix redumper crash

### LINKED ISSUE #813: redumper crashes
## Description

I noticed `redumper` crashes here:
https://github.com/riboseinc/rnp/blob/786529d18ff5a6ae94640a3d0e9266147a7ac088/src/apps/packet-dumper/redumper.cpp#L77-L78

I'm not familiar with the history of this workaround, but it looks like it would always crash:
https://github.com/riboseinc/rnp/blob/786529d18ff5a6ae94640a3d0e9266147a7ac088/src/librepgp/stream-ctx.cpp#L34-L36

(We don't do any testing of `redumper` so it's easy for this to go unnoticed)

## Steps to Reproduce

1. `src/apps/packet-dumper/redumper somefile.gpg`

ni4:
This PR fixes #813 and adds corresponding test.

### DIFF
--- a/src/tests/cli.cpp
+++ b/src/tests/cli.cpp
@@ -24,6 +24,7 @@
  * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <sys/wait.h>
 #include "rnp_tests.h"
 #include "support.h"
 #include "utils.h"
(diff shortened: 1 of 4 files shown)


src/tests/cli.cpp
```diff
@@ -471,3 +471,27 @@ test_cli_rnpkeys(void **state)
...
+    char *            redumper_path =
+      rnp_compose_path(rstate->original_dir, "../apps/packet-dumper/redumper", NULL);
+    char cmd[512] = {0};
+    int  chnum;
+    int  status;
+    /* call redumper's help */
+    chnum = snprintf(cmd, sizeof(cmd), "%s -h", redumper_path);
+    assert_true(chnum < (int) sizeof(cmd));
+    status = system(cmd);
+    assert_true(WIFEXITED(status));
```
dewyatt: It looks like you'll need to add `#include <sys/wait.h>` for `WIFEXITED`etc on FreeBSD. That should be OK for all 3 supported platforms.

ni4: @dewyatt Thanks, now it should be ok. Worked on macOS without include.

ronaldtse:
Thanks!

PR #832, opened 2019-04-12
Correctly set armor header for detached signatures.

### LINKED ISSUE #811: Incorrect armor header line for detached signatures
## Description

According to the RFC, these should use "BEGIN PGP SIGNATURE", but we currently use "BEGIN PGP MESSAGE".

## Steps to Reproduce

1. `rnp --sign --detach --armor somefile`
2. `cat somefile.asc`

ni4:
This fixes #811 and update tests to not let this happen in the future.


ronaldtse:
Thank you @ni4!

PR #840, opened 2019-04-27
Remove the use of deprecated Botan functions.

dewyatt:
There's a handful of functions that were recently deprecated in Botan, which we currently use in rnp.
Some links for more info:
* https://github.com/randombit/botan/commit/d17dbed547765739e1885bde33b0165795bcbd72
* https://github.com/randombit/botan/blob/a82d746a51c786049e855098934349a7821e1a7e/src/lib/ffi/ffi.h#L1112
* https://github.com/randombit/botan/blob/a82d746a51c786049e855098934349a7821e1a7e/src/lib/ffi/ffi.h#L942

Note that I opened https://github.com/riboseinc/rnp/issues/839 since I know this introduces something that can throw (`std::to_string`). `std::to_chars` would be better but it's c++17.

### DIFF
--- a/src/lib/crypto/rsa.cpp
+++ b/src/lib/crypto/rsa.cpp
@@ -75,8 +75,8 @@
 
 /** \file
  */
-#include <string.h>
-#include <stdbool.h>
+#include <string>
+#include <cstring>
 #include <botan/ffi.h>
 #include "crypto/rsa.h"
 #include "hash.h"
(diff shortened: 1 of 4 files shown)


src/lib/crypto/rsa.cpp
```diff
@@ -369,6 +369,10 @@ rsa_generate(rng_t *rng, pgp_rsa_key_t *key, size_t numbits)
         goto end;
     }
 
+    if (botan_privkey_create(
+          &rsa_key, "RSA", std::to_string(numbits).c_str(), rng_handle(rng))) {
+        goto end;
+    }
     if (botan_privkey_create_rsa(&rsa_key, rng_handle(rng), numbits) != 0) {
```
ni4: Should not this be removed since replaced with previous `botan_privkey_create()` call?

dewyatt: Oops, good catch!

PR #844, opened 2019-05-02
FFI functions to check uid/key revocation status

ni4:
This PR updates lower layer code with revocation signature checks, and adds functions to check key or user id revocation status.
However, it still lacks revocation check logic behind which should be implemented on lower layer (I'll file separate issue for this).
For instance, it just checks whether key/userid signature has revocation info subpacket, without checking the signature validity and signature's issuer.

To retrieve data about the user id I added `rnp_uid_handle_t` opaque structure. Currently it allows only to check revocation status, however later on we'll need to retrieve list of certifications/photo data/whatever else.


src/lib/rnp.cpp
```diff
@@ -4391,6 +4461,54 @@ rnp_key_get_expiration(rnp_key_handle_t handle, uint32_t *result)
...
+{
+    if (!handle || !result) {
+        return RNP_ERROR_NULL_POINTER;
+    }
+    pgp_key_t *key = get_key_prefer_public(handle);
+    if (!key || !key->revoked) {
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+
+    *result = strdup(key->revocation.reason ? key->revocation.reason : "");
```
dewyatt: We should probably check for `NULL` so we can return `RNP_ERROR_OUT_OF_MEMORY`.

ni4: @dewyatt Yeah, you are right. Will fix it. Any other comments?

dewyatt: It might make sense to add a note that an `rnp_uid_handle_t` can be invalidated if a userid is removed (though we may not have a function to do this *yet*). But LGTM anyways.

ni4: @dewyatt Done, please see the latest commit.

PR #847, opened 2019-05-09
FFI: add functions to access key/uid signature(s).

ni4:
This PR adds some basic signature access functions which I needed for CLI.
Also it fixes `rnp_locate_key()` function to set result to NULL if key was not found (basically, it should be in a separate PR, needed this in `rnp_signature_get_signer()`).

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -2567,6 +2567,8 @@ rnp_locate_key(rnp_ffi_t         ffi,
         (*handle)->pub = pub;
         (*handle)->sec = sec;
         (*handle)->locator = locator;
+    } else {
+        *handle = NULL;
     }
     return RNP_SUCCESS;
 }
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -66,6 +66,12 @@ struct rnp_uid_handle_st {
     size_t     idx;
 };
 
+struct rnp_signature_handle_st {
+    rnp_ffi_t     ffi;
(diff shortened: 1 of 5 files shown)


src/lib/rnp.cpp
```diff
@@ -4212,6 +4220,207 @@ rnp_key_get_uid_handle_at(rnp_key_handle_t key, size_t idx, rnp_uid_handle_t *ui
...
+    }
+    uint8_t keyid[PGP_KEY_ID_SIZE] = {0};
+    if (!signature_get_keyid(&handle->sig->sig, keyid)) {
+        *result = NULL;
+        return RNP_SUCCESS;
+    }
+
+    size_t hex_len = PGP_KEY_ID_SIZE * 2 + 1;
+    *result = (char *) malloc(hex_len);
+    if (*result == NULL) {
```
dewyatt: Style-wise I would argue to use `if (!*result)` though I'm not sure we have an official guideline on it.

ronaldtse: I don’t think we have this specified in the guide, but I agree that your suggestion is more consistent (just a few lines above).

ronaldtse:
@ni4 please feel free to merge after perhaps a tweak of the line mentioned above. Thanks!

src/lib/rnp.cpp
ni4: Thanks, you are right. Will fix and merge this soon.

ronaldtse:
Thanks @ni4 !

PR #849, opened 2019-05-11
FFI: add function rnp_output_to_file()

ni4:
Main goal of this PR was to add rnp_output_t creation functions which would allow to control file overwrite/random temporary file creation. 
Probably it should have another name, like rnp_output_to_path_ex() or so on.
However then we'll probably need to control temporary dir creation as well.

### DIFF
--- a/src/tests/support.cpp
+++ b/src/tests/support.cpp
@@ -103,6 +103,19 @@ file_contents(const char *path, ssize_t *size)
     return mem;
 }
 
+off_t
+file_size(const char *path)
+{
+    struct stat path_stat;
+    if (stat(path, &path_stat) != -1) {
+        if (S_ISDIR(path_stat.st_mode)) {
+            return -1;
+        }
+        return path_stat.st_size;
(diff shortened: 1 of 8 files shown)


src/tests/support.cpp
```diff
@@ -103,6 +103,19 @@ file_contents(const char *path, ssize_t *size)
     return mem;
 }
 
+int
```
dewyatt: I know you're just moving this code but I think this should be `off_t`. I know on my system `st_size` is 8 bytes and int is 4.

ni4: @dewyatt  Agree on this. Just updated the PR.

ronaldtse:
@dewyatt as you mentioned about release scheduling, please feel free to merge at the right time. Thanks!

dewyatt:
Should be fine to merge now, I'm just trying to consider 0.13 as mostly frozen so I probably won't cherry-pick this into that branch.

PR #857, opened 2019-05-28
FFI: add rnp_import_keys() function.

ni4:
While we have `rnp_load_keys()` this one is different since it works only with OpenPGP transferable keys, and lets caller to know the results of the import (which keys were imported/updated).
Need the last part to use in CLI.

### DIFF
--- a/include/rekey/rnp_key_store.h
+++ b/include/rekey/rnp_key_store.h
@@ -114,6 +114,14 @@ typedef enum key_store_format_t {
     G10_KEY_STORE,
 } key_store_format_t;
 
+/* Key import status. Order of elements is important. */
+typedef enum pgp_key_import_status_t {
+    PGP_KEY_IMPORT_STATUS_UNKNOWN = 0,
+    PGP_KEY_IMPORT_STATUS_UNCHANGED,
+    PGP_KEY_IMPORT_STATUS_UPDATED,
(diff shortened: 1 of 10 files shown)


dewyatt:
The JSON seems a little odd:
```json
{
  "keys": [
    {
      "public": "new",
      "secret": "none",
      "fingerprint": "e95a3cbf583aa80a2ccc53aa7bc6709b15c23a4a"
    },
    {
      "public": "new",
      "secret": "none",
      "fingerprint": "e332b27caf4742a11baa677f1ed63ee56fadc34d"
    }
  ]
}
```

Any thoughts on just doing an array?
```json
[
  {
    "public": "new",
    "secret": "none",
    "fingerprint": "e95a3cbf583aa80a2ccc53aa7bc6709b15c23a4a"
  },
  {
    "public": "new",
    "secret": "none",
    "fingerprint": "e332b27caf4742a11baa677f1ed63ee56fadc34d"
  }
]
```

ni4:
@dewyatt
>  The JSON seems a little odd:

I did in this since to be able to expand this information in future. Say, add fields like total number of processed keys and so on. Or is this redundant?

ronaldtse:
Thank you @ni4 ! Might be useful to have a flexible structure.

PR #859, opened 2019-06-02
FFI: add rnp_calculate_iterations() function.

ni4:
Needed it in CLI, so here it comes.
Any comments whether we could need second parameter (sample time) or can always use the default one?

### DIFF
--- a/src/tests/ffi.cpp
+++ b/src/tests/ffi.cpp
@@ -4868,3 +4868,11 @@ test_ffi_keys_import(void **state)
     // cleanup
     rnp_ffi_destroy(ffi);
 }
+
+void
+test_ffi_calculate_iterations(void **state)
+{
+    size_t iterations = 0;
+    assert_rnp_success(rnp_calculate_iterations("sha256", 500, &iterations));
+    assert_true(iterations > 65536);
+}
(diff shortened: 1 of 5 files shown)


src/tests/ffi.cpp
```diff
@@ -1753,7 +1753,10 @@ test_ffi_key_generate_ex(void **state)
     /* key protection */
     assert_rnp_success(rnp_op_generate_set_protection_cipher(keygen, "aes256"));
     assert_rnp_success(rnp_op_generate_set_protection_hash(keygen, "sha256"));
-    assert_rnp_success(rnp_op_generate_set_protection_iterations(keygen, 65536));
+    size_t iterations = 0;
+    assert_rnp_success(rnp_calculate_iterations("sha256", 500, &iterations));
+    assert_true(iterations > 65536);
+    assert_rnp_success(rnp_op_generate_set_protection_iterations(keygen, iterations));
```
dewyatt: I'd rather see this in its own tiny test. The more granular we can keep the tests, the better we can parallellize test running soon.

dewyatt:
> Any comments whether we could need second parameter (sample time) or can always use the default one?

I'm thinking it should be fine to keep it like you did here, using the default `trial_msec`.

src/tests/ffi.cpp
ni4: @dewyatt Agree. Updated and rebased.

PR #864, opened 2019-06-19
Fix RSA algorithm name and dump fingerprint for v3 keys.

ni4:
Some minor dumping fixes and test for them.

### DIFF
--- a/src/librepgp/stream-dump.cpp
+++ b/src/librepgp/stream-dump.cpp
@@ -130,7 +130,7 @@ static pgp_map_t key_type_map[] = {
 };
 
 static pgp_map_t pubkey_alg_map[] = {
-  {PGP_PKA_RSA, "RSA"},
+  {PGP_PKA_RSA, "RSA (Encrypt or Sign)"},
   {PGP_PKA_RSA_ENCRYPT_ONLY, "RSA (Encrypt-Only)"},
   {PGP_PKA_RSA_SIGN_ONLY, "RSA (Sign-Only)"},
   {PGP_PKA_ELGAMAL, "Elgamal (Encrypt-Only)"},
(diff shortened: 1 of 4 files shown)


src/librepgp/stream-dump.cpp
```diff
@@ -130,7 +130,7 @@ static pgp_map_t key_type_map[] = {
 };
 
 static pgp_map_t pubkey_alg_map[] = {
-  {PGP_PKA_RSA, "RSA"},
+  {PGP_PKA_RSA, "RSA (Encrypt or Sign)"},
```
dewyatt: I just want to point out that we should keep in mind that when we change the JSON packet dumper we are effectively breaking our public interface. If I had an application using `rnp_dump_packets_to_json` that assumed this value was stable, it would be pretty easy to imagine it crashing/misbehaving/failing when upgrading to a newer version of rnp with the change in this PR.

Having said that, it seems like this change would make things more consistent, right? Maybe we should just document that the output of `rnp_dump_packets_to_json` is subject to change?

ni4: @dewyatt agree, we should not make breaking changes into the JSON output. But somehow we should handle the case when we need to fix some older error (like this PR does).
So in this particular case I think we may let it go, but later on try to avoid it as much as possible - anyway we'll need to update it with new signature subpackets, algorithms, v5 fingerprint and so on so on.

dewyatt: Well apparently this does break ruby-rnp's tests, so we'll need to figure out what to do about that.

dewyatt: I'll go ahead and put in a workaround for ruby-rnp's tests and then hopefully we'll be all green here.

ronaldtse: Thank you @dewyatt !

ni4: Thanks, @dewyatt !

dewyatt:
Ok we're caching the ruby-rnp clone in Cirrus and it doesn't have a manual method of clearing the cache. I'll get that taken care of this morning.

ronaldtse:
Thank you @ni4 and @dewyatt !

PR #867, opened 2019-06-20
ci: Various fixes.

dewyatt:
I encountered a few different issues when looking into the CI failures in https://github.com/riboseinc/rnp/pull/865.

I can't say that these changes will fix that particular issue but I think they are a step in the right direction.
If/when these are merged we can restart that build and go from there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -117,6 +117,7 @@ cache:
 before_install:
   - . ci/env.inc.sh
   - ./ci/before_install.sh
+  - rvm use 2.4.6 --install --binary
 
 install:
   - travis_wait 50 ./ci/install.sh
--- a/Brewfile
+++ b/Brewfile
@@ -9,5 +9,4 @@ brew "cmocka"
 brew "gnupg"
 brew "wget"
 brew "python@2"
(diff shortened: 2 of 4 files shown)


dewyatt:
It looks like this introduces a different failure on macOS. For some reason the ruby tests are not able to find the rnp dylib, though Linux and FreeBSD have no issues...I'll see if I can figure that out.

dewyatt:
Well I certainly spent more time on this than expected! So as far as the issue in this PR (`Could not open library`), I think macOS's SIP is to blame. The simplest workaround I've found is to copy the dylib's to `/usr/local/lib`. I'll add a workaround commit here.

I don't know why this is suddenly an issue when it was not a problem in the past.

The missing `bundle` issue in the other PR I think is likely some stale cache problem.

dewyatt:
It's strange, I *was* seeing consistent build failures for ruby-ffi 1.11.1 on macOS xcode10.1 in travis, but that issue has seemingly resolved itself (I was working around it by installed 1.9.25 before bundle install). Oh well, I guess.

PR #869, opened 2019-06-29
FFI: set default algorithm-dependent key/subkey flags before the generation.

ni4:
In previous PRs I missed default key flags setup, so keys/subkeys were by default generated with all available by the algorithm flags.
This PR fixes changes defaults to sign/certify for primary key, and encrypting (or signing for DSA/EdDSA/ECDSA) subkey.

I updated already existing tests instead of writing new one since new test would require to generate new keys, slowing things down.


ronaldtse:
@ni4 thanks for this!

PR #873, opened 2019-07-10
Use FFI in rnpkeys

ni4:
This PR makes rnpkeys to use FFI interface.
rnp will be moved later on so we'll finally be able to have single interface to rnp functions.
Also it removes no longer used code, which is replaced with FFI calls.
There are some minor usages of direct interface (see `fficli.h`), which will be adressed during `rnp` transition.

P.S. Unfortunately I don't see the good way to make compilable incremental commits for `rnpkeys` instead of single large one.

### DIFF
--- a/src/lib/librnp.3
+++ b/src/lib/librnp.3
@@ -93,10 +93,6 @@ The following functions are used for key management:
 .Fo rnp_match_list_keys
 .Fa "rnp_t *rnp" "char *pattern"
 .Fc
-.Ft int
-.Fo rnp_find_key
-.Fa "rnp_t *rnp" "char *userid"
-.Fc
 .Ft char *
 .Fo rnp_get_key
 .Fa "rnp_t *rnp" "const char *id"
--- a/src/lib/librnp.3
+++ b/src/lib/librnp.3
@@ -105,10 +105,6 @@ The following functions are used for key management:
 .Fo rnp_export_key
 .Fa "rnp_t *rnp" "char *userid"
 .Fc
-.Ft int
-.Fo rnp_generate_key
-.Fa "rnp_t *rnp" "char *userid" "int numbits"
-.Fc
 .Pp
 The following functions are used for file management:
 .Ft rnp_result_t
--- a/src/lib/librnp.3
+++ b/src/lib/librnp.3
@@ -97,10 +97,6 @@ The following functions are used for key management:
 .Fo rnp_get_key
 .Fa "rnp_t *rnp" "const char *id"
 .Fc
-.Ft int
-.Fo rnp_export_key
-.Fa "rnp_t *rnp" "char *userid"
-.Fc
 .Pp
 The following functions are used for file management:
 .Ft rnp_result_t
@@ -182,7 +178,7 @@ is appended to the home directory in order to search for
 the keyrings.
 .Pp
 To export a key, the
-.Fn rnp_export_key
+.Fn rnp_key_export
 function is used.
 Output is sent to the standard output.
 .Pp
--- a/src/lib/librnp.3
+++ b/src/lib/librnp.3
@@ -106,10 +106,6 @@ The following functions are used for key management:
 .Fa "rnp_t *rnp" "char *userid"
 .Fc
 .Ft int
-.Fo rnp_import_key
-.Fa "rnp_t *rnp" "char *file"
-.Fc
-.Ft int
 .Fo rnp_generate_key
 .Fa "rnp_t *rnp" "char *userid" "int numbits"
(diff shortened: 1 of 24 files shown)


ni4:
Okay, it seems to be a good time to investigate where and how tests can be optimized to not take that much time... I believe most of time is spent in temporary key generation.

ronaldtse:
And presumably due to the lack of entropy in containers?

ni4:
> And presumably due to the lack of entropy in containers?

This may be a part of the issue, but anyway we should make less generate key calls.

dewyatt:
We shouldn't jump to conclusions on what is causing it of course, we should investigate and verify.
I am still working on the Google Test conversion that would help here (not because of Google Test really, but because of the restructuring that would come with it).

Anyways, I think I can come up with a temporary solution for this in a bit.

ni4:
Looks like I was wrong with first assumption. More time is spent in CLI tests in exhaustive single operation runs (like for encryption - it is run for [each algo, multiplied by different file size, multiplied by compression used, etc]. 
I'll finish investigations and reduce number of runs in the separate PR.

dewyatt:
> Looks like I was wrong with first assumption. More time is spent in CLI tests in exhaustive single operation runs (like for encryption - it is run for [each algo, multiplied by different file size, multiplied by compression used, etc].

Oh yeah I see that now, the cmocka tests ran for ~35m, cli tests timed out @ 50m. :(
I have an idea on how to improve that, but your quick fix you mentioned is probably a good idea now.

ni4:
Update: found other issues, including slowed down in ~2 times CLI tests run. Will continue to work on them.

dewyatt:
I was able to get a ~45% speedup for cli_tests *locally* with some restructuring (no changes to actual tests). I assume that will translate to some gains in CI as well, though I expect it won't be as significant.

I'll continue doing some experimenting too and hopefully we can combine our efforts and see some good results here.

ni4:
@dewyatt @ronaldtse As I find out, the 95% of time is spent in key material validity checks. While signature check is quite fast, single RSA and ElGamal key check takes up to 0.5 second on my machine. After disabling RSA key material check cli_tests run 290 seconds instead of 3600.
And, btw, after transition to FFI tests run slower since there are more key checks in keyring copying/loading/whatever else.

So my idea is to disable key material check on every load and do it only when key is generated, and when key is imported from other file. What do you think?

dewyatt:
> @dewyatt @ronaldtse As I find out, the 95% of time is spent in key material validity checks. While signature check is quite fast, single RSA and ElGamal key check takes up to 0.5 second on my machine. After disabling RSA key material check cli_tests run 600 seconds instead of 3600.

Yeah I did some profiling today and saw that too (large amount of execution time in `validate_pgp_key_material`). I did a little proof-of-concept w/Boost.Interprocess to cache the key material validity between all processes. On my machine it reduced execution time by 32% for rnp_tests, 6%(?) for cli_tests, and 56% for ruby-rnp's tests. I don't intend it as an actual solution, it was more for getting a better idea on what a good solution may be.

> So my idea is to disable key material check on every load and do it only when key is generated, and when key is imported from other file. What do you think?

I think this sounds good but I need sleep and will have to check back tomorrow.

ronaldtse:
> So my idea is to disable key material check on every load and do it only when key is generated, and when key is imported from other file. What do you think?

Agree! Or caching. It's up to you guys. :smile:

ni4:
> On my machine it reduced execution time by 32% for rnp_tests, 6%(?) for cli_tests, and 56% for ruby-rnp's tests.
@dewyatt cli_tests do keyring load on each run, which includes key material validation. Probably that's why there is no major speedup. And, btw, the most slow for RSA key is a secret key part validation.

ni4:
@dewyatt I went ahead and did #875, please take a look on it.

ronaldtse:
@dewyatt are we waiting for the test speedup here or can we merge this?

ni4:
@ronaldtse Let me rebase this and update with changes from a3d5a828627042f3b0728de13f92dbce8bc4826c

dewyatt:
I'm planning to review it today.

ronaldtse:
Sounds like we can now split `rnpkeys` into a new repository!

ni4:
@ronaldtse I think it would be better to wait for rnp's FFI transition as well - while it should not make problems to transfer rnpkeys CLI code to separate repository, it would be harder to transfer tests (like cli_tests.py) which depends on both rnpkeys and rnp utilities.

ronaldtse:
Agree!

PR #874, opened 2019-07-12
FFI: add function rnp_op_generate_set_request_password().

ni4:
This PR adds a way to request key encryption password via FFI's password provider and then use it to encrypt generated key.
While it is already possible to encrypt key via rnp_op_generate_set_password(), this method could be handy in some situations (in rnpkeys for instance :)).

I'm not sure about the function name, so if you have better suggestions - you are welcome to share :)

### DIFF
--- a/src/tests/rnp_tests.cpp
+++ b/src/tests/rnp_tests.cpp
@@ -214,6 +214,7 @@ main(int argc, char *argv[])
       cmocka_unit_test(test_ffi_key_generate_eddsa),
       cmocka_unit_test(test_ffi_key_generate_sm2),
       cmocka_unit_test(test_ffi_key_generate_ex),
+      cmocka_unit_test(test_ffi_key_generate_protection),
       cmocka_unit_test(test_ffi_add_userid),
       cmocka_unit_test(test_ffi_detect_key_format),
       cmocka_unit_test(test_ffi_encrypt_pass),
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -574,6 +574,19 @@ rnp_result_t rnp_op_generate_set_curve(rnp_op_generate_t op, const char *curve);
 rnp_result_t rnp_op_generate_set_protection_password(rnp_op_generate_t op,
                                                      const char *      password);
 
+/**
+ * @brief Enable or disable password requesting via ffi's password provider. This password
+ *        then will be used for key encryption.
+ *        Note: this will be ignored if password was set via
+ *        rnp_op_generate_set_protection_password().
+ *
+ * @param op pointer to opaque key generation context.
+ * @param request true to enable password requesting or false otherwise. Default value is false
+ *                (i.e. key will be generated unencrypted).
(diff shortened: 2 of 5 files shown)


ni4:
```
bundler: command not found: rspec
Install missing gem executables with `bundle install`
```
Looks like the issue we had previously?

dewyatt:
> ```
> bundler: command not found: rspec
> Install missing gem executables with `bundle install`
> ```
> 
> Looks like the issue we had previously?

Yes, it does...it looks like maybe the `bundle install` was skipped so rspec is missing. I wish Cirrus had a way to manually clear cache. I'll take a closer look soon.

src/tests/rnp_tests.cpp
```diff
@@ -212,9 +212,10 @@ main(int argc, char *argv[])
       cmocka_unit_test(test_ffi_key_generate_dsa),
       cmocka_unit_test(test_ffi_key_generate_ecdsa),
       cmocka_unit_test(test_ffi_key_generate_eddsa),
-      cmocka_unit_test(test_ffi_key_generate_sm2),
+      cmocka_unit_test(test_ffi_key_generate_sm2),*/
       cmocka_unit_test(test_ffi_key_generate_ex),
-      cmocka_unit_test(test_ffi_add_userid),
+      cmocka_unit_test(test_ffi_key_generate_protection),
+      /*cmocka_unit_test(test_ffi_add_userid),
```
dewyatt: Let's make sure we don't merge this with tests commented out.

ni4: @dewyatt Wow, thanks, my fault. Commenting those out to make tests go faster, but this time missed to fix things back.

ni4: @dewyatt Fixed commented out tests. Any other comments? Anyway waiting for CI.

dewyatt:
It looks like there are a couple of test failures:
```
[  FAILED  ] s2k_iteration_tuning
[  FAILED  ] test_ffi_keygen_json_pair_dsa_elg
```

I've seen `s2k_iteration_tuning` fail on occasion, I think it's probably because `pgp_s2k_compute_iters` uses wall-clock time and the process may end up suspended longer than the test allows for, etc. Maybe we can improve it later.

The `test_ffi_keygen_json_pair_dsa_elg` is more interesting, it looks like it expects the key bits to be 1536 but it gets 1528. We may want to look more closely at that.

I'll try to restart CI and see if these are unrelated to the PR, file issues, etc.

ni4:
@dewyatt Looks like it is a rare case where exponentiation leads to 1-byte smaller Y then X. Will try to reproduce this case.

ni4:
@dewyatt Reproducing ElGamal problem takes time (it runs quite slowly), so I filed #881.
Any other comments or this can be merged?

ni4:
@ronaldtse Do you have any comments, or we can merge this (problems discussed in comments above were filed as separate issues)?

ronaldtse:
@ni4 sorry I missed this one. Looks good to go!

PR #877, opened 2019-07-16
Fix missing rspec in Cirrus

dewyatt:
It looks like the issue here was pretty much what I suspected, the cache had ruby gems installed for 2.4, but FreeBSD had some updates and a newer ruby was installed, etc.

So the main fix is in 43001e4. I also did some rearranging mostly to improve readability of logs.


ronaldtse:
Thanks for going through the grueling testing @dewyatt !

Merging since it’s only CI.

PR #879, opened 2019-07-17
ci: Don't call before_install twice in Cirrus.

dewyatt:
I accidentally left this duplicate call to before_install in here.


ronaldtse:
Thanks @dewyatt ! Merging since it's a CI change.

PR #883, opened 2019-07-19
Changes that allow the library to build on Windows (MinGW)

kaie:
Related to https://github.com/riboseinc/rnp/issues/878

Building on Windows succeeds. I haven't yet executed it on Windows. Note the TODO for password entry hiding, I suggest to do that in a follow-up. Removes the use of the GNU C regex code, and uses std:.regex from C++11 instead. Removes the use of strsep, replaced with C++ string code. The UNCONST macro seemed really dangerous, and didn't build on my system, a size mismatch was reported. The new code using C++ const operaters should be safer.

With these changes, the tests on Linux still pass for me.


kaie:
Hmm, the second commit wasn't supposed to be included with this pull request...

Anyway...

Porting the tests will be more difficult, because RNP uses several additional non-portable constructs.
The tests build with the second commit.
I added comments with TODO for the remaining work.

kaie:
I don't understand why the travis tests failed. I don't get failures when executing "make test" locally on Linux.

To trigger another travis run, I'll back out the second commit, which was supposed to be handled separately anyway

dewyatt:
> I don't understand why the travis tests failed. I don't get failures when executing "make test" locally on Linux.

Probably the same underlying issue as https://github.com/riboseinc/rnp/issues/881 (but on the ruby side). Re-running should be OK.

kaie:
@dewyatt The new build shows failure, too. Could you please have a look? Does that look like known CI failures?

Do you need me to produce a successful travis run, before you'll review the pull request?

kaie:
Is it possible to trigger a re-run of the tests without doing a commit?

src/lib/pass-provider.cpp
```diff
@@ -27,16 +27,21 @@
 
 #include <stdio.h>
 #include <string.h>
+
+#if !defined(_WIN32) && !defined(_WIN64)
```
dewyatt: `_WIN32` should be sufficient for both

dewyatt:
> @dewyatt The new build shows failure, too. Could you please have a look? Does that look like known CI failures?

The new build is showing different test failures, and no they are not known bugs.

> Do you need me to produce a successful travis run, before you'll review the pull request?

Generally, yes.

dewyatt:
> Is it possible to trigger a re-run of the tests without doing a commit?

Yes, but you currently lack the permissions. I'll re-run it.

kaie:
I cannot reproduce the travis test failures locally.

Could the travis failure be caused by an old, buggy version of clang, or a buggy version of a stdlibc++, which has bugs in the regular expression library code? I'm currently out of other ideas.

The tests pass on Mac, possibly because it has a newer clang version?

The Linux builder uses clang 5.0.0, which means it hasn't even been upgraded to the most stable branch release 5.0.x. (I tested locally with 5.0.1+ from debian 9 which passed tests.)

The sanitize builder fails with an undefined reference to __ubsan_vptr_type_cache. https://github.com/google/oss-fuzz/issues/713 suggests that an additional link option might be necessary.

dewyatt:
I can repro, I'll try to take a look.

The compiler is not the issue, some of our jobs in travis use clang, some use gcc.

kaie:
Do you build botan with threads enabled or disabled?

kaie:
> Do you build botan with threads enabled or disabled?

If botan is build with os threads enabled (default for me), then botan spawns a worker thread for key generation. I build botan with os threads disabled. I no longer see threads spawned. In both scenarios I don't see a test failure.

dewyatt:
>  a buggy version of a stdlibc++, which has bugs in the regular expression library code?

I think you're spot on here. I'm able to reproduce this in `centos:7` and `ubuntu:trusty` w/`libstdc++`. If I switch to using clang's `libc++`, the tests pass.

kaie:
What are the oldest distributions that you need to (or want to) support?

Ubuntu Trusty (14.04) is no longer maintained, apparently, only extended security updates. The most recent version of cmake available for it is 3.5.1, but RNP requires at least version 3.7.

What's the oldest version of Centos/RHEL 7.x that you want to support?

dewyatt:
I would say we're likely to drop trusty in CI soon (pending a bug fix), but CentOS 7 should remain supported.

We may have to keep the existing regex code alongside the new:
```cpp
#ifdef _WIN32
  // use std::regex
#else
  // use regcomp
#endif
```

kaie:
Ok. I'm able to reproduce using a Centos 7.6 VM with EPEL for cmake3, fully updated. I'll try to have a look tomorrow what's going on. If I cannot figure it out, I agree that we might need #ifdef.

dewyatt:
> Ok. I'm able to reproduce using a Centos 7.6 VM with EPEL for cmake3, fully updated. I'll try to have a look tomorrow what's going on. If I cannot figure it out, I agree that we might need #ifdef.

Sounds good, hopefully it's just a bug in the PR here and libstdc++ happens to be more forgiving.

kaie:
It's confirmed, C++ regexp code is broken in CentOS 7 using the default gcc 4.8 compiler.

This code below on centos 7 prints:
```
  return value: 0
  number of found matches: 0
```

On a modern Debian system it prints:
```
  return value: 1
  number of found matches: 1
```

(both compiled using: g++ -std=c++11)

```
#include <regex>
#include <string>
#include <stdio.h>

int main(int argc, char *argv[])
{
  std::string haystack("hello");
  std::string needle("h");
  std::regex re(needle, std::regex_constants::extended | std::regex_constants::icase);

  std::smatch found;

  bool result = std::regex_search(haystack, found, re);

  printf("return value: %d\n", (int)result);
  printf("number of found matches: %d\n", found.size());

  return 0;
}
```

Then I found this article, which also confirms it:
http://www.michaelbrich.com/no-working-around-broken-c11-regex-in-gcc-4-8/

kaie:
RHEL and Centos provide a "Developer Toolset" that provides a newer compiler toolchain.

For reference, see https://lists.centos.org/pipermail/centos-announce/2017-November/022641.html

You can install it using:
```
$ sudo yum install centos-release-scl
$ sudo yum install devtoolset-7
```

This doesn't change the default. If you want to use the modern tools, you need to start a bash shell with the tools enabled:
```
$ scl enable devtoolset-7 bash
$ g++ --version
g++ (GCC) 7.3.1 20180303 (Red Hat 7.3.1-5)
```

If you don't want to maintain the #ifdef and two different code paths for regular expressions, you could decide that Centos 7 is only supported if the modern C++ toolchain from the Developer Toolset are used.

I'll test if all of the RNP test suite passes with my patch and the newer toolchain.

kaie:
Based on information found here: 
https://stackoverflow.com/questions/12530406/is-gcc-4-8-or-earlier-buggy-about-regular-expressions

The regex support was added in gcc 4.9, and 4.8 simply contained experimental code. There's a macro suggested above to detect if the regex code can be expected to work correctly. We could use that macro to use the C++ code on all modern platforms, and only fallback to GNU C regexp on systems that don't have a working C++ regex implementation available.

dewyatt:
I'm splitting hairs a bit but it looks like libstdc++ specifically is to blame, not the compiler. Your example works on trusty when built with clang++ w/libc++, but fails when built with clang++ w/libstdc++ (default).

Anyways, I do already use the SCL and devtoolset-7/llvm-toolset-7 in the (in progress) rnp GitLab CI deployment. I believe these link the system libstdc++ but will use the newer regex headers. So I don't think it would introduce any new package dependencies (except for the toolset during build time, but that's no issue).

So that might be OK, I'll check more thoroughly later.

We do have a blocker either way and my current plan would be:
1. Wait on https://github.com/riboseinc/rnp/issues/882 (I suspect this is causing test failures in xenial+bionic).
2. Update Travis CI to use xenial.
3. Make sure CI is happy, then merge this.

src/librekey/rnp_key_store.cpp
```diff
@@ -851,28 +850,25 @@ get_key_by_name(const rnp_key_store_t *keyring,
...
+
     for (list_item *key_item = after ? list_next((list_item *) after) :
                                        list_front(keyring->keys);
          key_item;
          key_item = list_next(key_item)) {
         keyp = (pgp_key_t *) key_item;
 
         for (i = 0; i < pgp_key_get_userid_count(keyp); i++) {
-            if (regexec(&r, (char *) pgp_key_get_userid(keyp, i), 0, NULL, 0) == 0) {
+            std::string userid = pgp_key_get_userid(keyp, i);
```
dewyatt: Can probably be const.

kaie:
FYI, I've primarily used hg/mercurial in the past, and are not an expert with git. Please let me know if I did something unusual with this pull request. I had to submit a merge commit. If you think this pull request is confusing, I'd be happy to start over a fresh pull request on another branch. Just let me know.

kaie:
More changes are required because of the recent commits on master.

dewyatt:
@kaie We usually don't use merge commits much, we rebase. I can take care of it if you want.

kaie:
I made the commit before I noticed that the master had moved. If it's OK with you, I'd close this pull request, and do a fresh one. Will allow me to excercise with the proper procedures.

dewyatt:
> I made the commit before I noticed that the master had moved. If it's OK with you, I'd close this pull request, and do a fresh one. Will allow me to excercise with the proper procedures.

Sounds good to me!

ronaldtse:
Thank you @kaie for the contribution!

Since it now builds on MinGW, it's probably time we add AppVeyor testing... 👏

PR #890, opened 2019-07-24
dummy change

kaie:
Let's see if travis produces a correct build, with this no-op commit.


kaie:
This worked fine... Confirms that travis testing is working correctly, and it's really the code from https://github.com/riboseinc/rnp/pull/883 that is introducing failures.

PR #893, opened 2019-07-25
Allow cross-compiling for Windows with MinGW

kaie:
Replacement pull request for https://github.com/riboseinc/rnp/pull/883

### DIFF
--- a/src/lib/utils.h
+++ b/src/lib/utils.h
@@ -178,7 +178,7 @@ STORE64BE(uint8_t x[8], uint64_t y)
 /* Macro _GLIBCXX_USE_CXX11_ABI was first introduced with GCC 5.0, which
  * we assume to be bundled with a sane implementation of std::regex. */
 #if !defined(__GNUC__) || defined(_GLIBCXX_USE_CXX11_ABI)
-#define RNP_ASSUME_SANE_LIBSTDCPLUSPLUS_REGEX 1
+#define RNP_USE_STD_REGEX 1
 #endif
 
 #endif
--- a/src/lib/utils.h
+++ b/src/lib/utils.h
@@ -113,8 +113,14 @@
 /* for silencing unused parameter warnings */
 #define RNP_USED(x) /*LINTED*/ (void) &(x)
 
-#ifndef RNP_UNCONST
-#define RNP_UNCONST(a) ((void *) (unsigned long) (const void *) (a))
+#ifndef RNP_CONST_TO_VOID_PTR
+#define RNP_CONST_TO_VOID_PTR(a) (reinterpret_cast<void *>(const_cast<char *>(a)))
+#endif
+
+#ifdef _WIN32
+#define RNP_MKDIR(pathname, mode) mkdir(pathname)
+#else
+#define RNP_MKDIR(pathname, mode) mkdir(pathname, mode)
 #endif
 
 /* debugging helpers*/
@@ -167,6 +173,12 @@ STORE64BE(uint8_t x[8], uint64_t y)
 
 #ifndef MAX
 #define MAX(a, b) (((a) > (b)) ? (a) : (b))
+#endif
+
+/* Macro _GLIBCXX_USE_CXX11_ABI was first introduced with GCC 5.0, which
+ * we assume to be bundled with a sane implementation of std::regex. */
+#if !defined(__GNUC__) || defined(_GLIBCXX_USE_CXX11_ABI)
+#define RNP_ASSUME_SANE_LIBSTDCPLUSPLUS_REGEX 1
 #endif
 
 #endif
--- a/src/librepgp/stream-parse.cpp
+++ b/src/librepgp/stream-parse.cpp
@@ -30,7 +30,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <string>
-#include <deque>
+#include <vector>
 #include <time.h>
 #include <rnp/rnp_def.h>
 #include "stream-ctx.h"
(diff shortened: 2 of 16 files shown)


kaie:
FYI, the symbols I'm removing from fficli.h are duplicates, they are defined elsewhere as enums.

kaie:
Note that I implemented the fallback to the POSIX regex code, which means, this isn't blocked by any upgrade to your build infrastructure.

Because this isn't a small patch, it might be helpful to merge, to avoid further conflicts.

kaie:
> I'm splitting hairs a bit but it looks like libstdc++ specifically is to blame, not the compiler. Your example works on trusty when built with clang++ w/libc++, but fails when built with clang++ w/libstdc++ (default).

You're right, but I assumed that usually compilers use the libstdc++ that was bundled with it. So, when someone says GCC 4.9 fixed it, you're right, it's the newer libstdc++ that's bundled with the newer compiler that fixed it.

I used a simple check to decide if we use the old POSIX C code or the new C++ code. It's based on a macro that was first used with GCC version 5. I think if someone uses GCC 5, they are probably using a libstdc++ that's at least as new as the one bundled with GCC 5, and should work fine.

src/lib/utils.h
```diff
@@ -169,4 +175,10 @@ STORE64BE(uint8_t x[8], uint64_t y)
 #define MAX(a, b) (((a) > (b)) ? (a) : (b))
 #endif
 
+/* Macro _GLIBCXX_USE_CXX11_ABI was first introduced with GCC 5.0, which
+ * we assume to be bundled with a sane implementation of std::regex. */
+#if !defined(__GNUC__) || defined(_GLIBCXX_USE_CXX11_ABI)
+#define RNP_ASSUME_SANE_LIBSTDCPLUSPLUS_REGEX 1
```
dewyatt: Can we shorten this to something like `RNP_USE_STD_REGEX`?

src/librepgp/stream-parse.cpp
```diff
@@ -967,13 +969,55 @@ signed_src_finish(pgp_source_t *src)
     return ret;
 }
 
+static void tokenize(const std::string &str, const std::string &delims,
```
dewyatt: I think we can shorten this (and may as well templatize):
```cpp
template <typename T>
void
tokenize(const typename T::value_type &str, const typename T::value_type &delims, T &result)
{
    result.clear();
    size_t current;
    size_t next = -1;
    do {
        next = str.find_first_not_of(delims, next + 1);
        if (next == T::value_type::npos) {
            break;
        }
        next -= 1;
        current = next + 1;
        next = str.find_first_of(delims, current);
        result.push_back(str.substr(current, next - current));
    } while (next != T::value_type::npos);
}
```

Also a `deque` seems an odd choice here, I think a `vector` would be more typical.

src/librepgp/stream-parse.cpp
```diff
@@ -990,15 +1034,18 @@ cleartext_parse_headers(pgp_source_t *src)
...
 
-                if ((halg = pgp_str_to_hash_alg(hname)) == PGP_HASH_UNKNOWN) {
-                    RNP_LOG("unknown halg: %s", hname);
-                }
+            const std::string delimiters = ", \t";
+            std::deque<std::string> tokens;
+
+            tokenize(remainder, delimiters, tokens);
 
+            for (auto iter = tokens.begin(); iter != tokens.end(); ++iter) {
```
dewyatt: Since we're using C++11, can you please use the shorter form?:
```cpp
for (auto &token: tokens) {
```

src/librepgp/stream-parse.cpp
kaie: You simplifed the tokenizer function in a nice way, and used a smart way to use it as a template.

However, I'd like to suggest changes:
* I'd prefer to avoid assigning negative numbers to an unsigned type size_t. You're assuming that you'll know exactly how it will overflow. I'd like to avoid that.
* For the substring calculation, you might get an end position that is npos, and you're subtracting from it. I'd prefer to avoid that, and stick with the defined value npos for "give me all there is left".
* We don't know if size_t matches the size used internally by the string type. Using the value type's internal size type is safer.

Here's a new version, which I've tested with a variety of inputs:

```c++
/*
 * str is a string to tokenize.
 * delims is a string containing a list of delimiter characters.
 * result is a container<string_type> that supports push_back.
 */
template <typename T>
static void
tokenize(const typename T::value_type &str, const typename T::value_type &delims, T &result)
{
    typedef typename T::value_type::size_type string_size_t;
    const string_size_t npos = T::value_type::npos;

    result.clear();
    string_size_t current;
    string_size_t next = 0;
    do {
        next = str.find_first_not_of(delims, next);
        if (next == npos) {
            break;
        }
        current = next;
        next = str.find_first_of(delims, current);
        string_size_t count = (next == npos) ? npos : (next - current);
        result.push_back(str.substr(current, count));
    } while (next != T::value_type::npos);
}
```

dewyatt: I probably pulled some hacky play version of my snippet there, I don't typically rely on overflows :P.

Realistically, templatizing the `std::string` container is probably useless overkill for us and could be explicit too. I was more interested in the result container.

 Anyways, this looks good to me at a glance!

dewyatt:
Also if you'r familiar with `clang-format`, now would be a good time to run it on the particular sections of code changed here. Otherwise, we can take care of it before merge.

kaie:
If it's easy for you to run clang-format, I'd prefer that. Otherwise I'll have to setup the matching environment.

dewyatt:
Ok, clang-format made some minor changes. Be sure to `git pull` on this branch before making any further changes (if any).

kaie:
Thanks for committing clang-format to my branch.

I have more changes to files below src/tests to make tests build on Windows.
Do you prefer me to add those to this pull request, or start another pull request on a new branch?

kaie:
(hmm, separate branch would work only after you've merged this one I guess)

dewyatt:
> I have more changes to files below src/tests to make tests build on Windows.
> Do you prefer me to add those to this pull request, or start another pull request on a new branch?

Let's just do a separate PR for that after this one.

ronaldtse:
Thank you @kaie @dewyatt !!

PR #894, opened 2019-07-26
Allow cross-compiling of tests with MinGW (and add TODOs)

kaie:
I haven't yet tried to run the tests on Windows. As seen in TODO comments (see also #ifdef) there are several places that use code that isn't available on Windows, and need some new platform specific code.
This is just a starting point.

### DIFF
--- a/src/tests/cli.cpp
+++ b/src/tests/cli.cpp
@@ -24,7 +24,10 @@
  * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#ifndef _WIN32
 #include <sys/wait.h>
+#endif
+
 #include "rnp_tests.h"
 #include "support.h"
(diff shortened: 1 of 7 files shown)


dewyatt:
I wouldn't be comfortable approving this with the current approach, I would *really* like to avoid preprocessor activity in the body of tests (even if temporary).

Instead, I'd prefer we start to build out more cross-platform support functions.

kaie:
That's fine. Feel free to close. My primary intention was to investigate how much is missing, and point out those places to you.

dewyatt:
@kaie 
> That's fine. Feel free to close.

Ok, closing. I'm thinking I may do a quick PR today to start organizing our cross-platform support some, and I'm hoping I can even address some things you've pointed out in this PR.

> My primary intention was to investigate how much is missing, and point out those places to you.

You've definitely been a big help on that, much appreciated!

PR #909, opened 2019-09-05
Use FFI for rnp --enarmor/--dearmor commands.

ni4:
This PR continues transition to FFI-only library usage.
Also within it I started to write new code in C++ where applicable instead of C.


ronaldtse:
@ni4 build is failing here.

Screen Shot 2019-09-07 at 4 29 01 AM

ni4:
@ronaldtse Wow, thanks. Was so sure that everything was checked and it cannot fail so it failed as usual :-) Now it should pass Travis.

PR #912, opened 2019-09-14
Remove --show-keys command from rnp.

### LINKED ISSUE #911: Remove rnp --show-keys command.
## Description

This command existed since the netpgp and did a strange job - it did not display information about the keys but some basic information about the key signature expiration.
Since we have better functionality in `rnpkeys --list-keys/--list-sigs`, I think that duplicate one (which doesn't print any info now) may be removed.

ni4:
As title says.
Fixes #911


dewyatt:
Thanks! I thought I approved already. :/

PR #922, opened 2019-10-13
Update rnpcfg with some std::string based functions.

ni4:
This PR moves a bit forward to C++ code, and code will be used in CLI to FFI transition PR #920


dewyatt:
Looks like CI is failing. :(

ni4:
@dewyatt Argh, it's quite logical that this will not work until rebased on #921 :-)

PR #925, opened 2019-10-17
Allow to use password provider in rnp_op_encrypt_add_password

ni4:
Sometimes (in CLI for example) it is useful to be able to request password(s) for symmetric encryption via the configured pass provider instead of specifying it directly.

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -53,6 +53,7 @@
 #include "utils.h"
 #include "json_utils.h"
 #include "version.h"
+#include <botan/secmem.h>
 
 struct rnp_key_handle_st {
     rnp_ffi_t        ffi;
@@ -2152,13 +2153,11 @@ rnp_op_encrypt_add_password(rnp_op_encrypt_t op,
                             size_t           iterations,
                             const char *     s2k_cipher)
 {
-    rnp_result_t ret = RNP_ERROR_GENERIC;
-
     // checks
-    if (!op || !password) {
+    if (!op) {
         return RNP_ERROR_NULL_POINTER;
     }
-    if (!*password) {
+    if (password && !*password) {
         // no blank passwords
         FFI_LOG(op->ffi, "Blank password");
         return RNP_ERROR_BAD_PARAMETERS;
@@ -2182,9 +2181,20 @@ rnp_op_encrypt_add_password(rnp_op_encrypt_t op,
         FFI_LOG(op->ffi, "Invalid cipher: %s", s2k_hash);
         return RNP_ERROR_BAD_PARAMETERS;
     }
-    // derive key, etc
-    ret = rnp_ctx_add_encryption_password(&op->rnpctx, password, hash_alg, symm_alg, 0);
-    return ret;
+    try {
+        Botan::secure_vector<char> ask_pass(MAX_PASSWORD_LENGTH, '\0');
+        if (!password) {
+            pgp_password_ctx_t pswdctx = {.op = PGP_OP_ENCRYPT_SYM, .key = NULL};
+            if (!pgp_request_password(
+                &op->ffi->pass_provider, &pswdctx, &ask_pass[0], ask_pass.size())) {
+                return RNP_ERROR_BAD_PASSWORD;
+            }
+            password = ask_pass.data();
+        }
+        return rnp_ctx_add_encryption_password(&op->rnpctx, password, hash_alg, symm_alg, iterations);
+    } catch (...) {
+        return RNP_ERROR_OUT_OF_MEMORY;
+    }
 }
 
 rnp_result_t
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -1777,7 +1777,8 @@ rnp_result_t rnp_op_encrypt_set_expiration_time(rnp_op_encrypt_t op, uint32_t ex
  * @brief Add password which is used to encrypt data. Multiple passwords can be added.
  *
  * @param op opaque encrypting context. Must be allocated and initialized.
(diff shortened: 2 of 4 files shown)


src/lib/rnp.cpp
```diff
@@ -2182,8 +2180,18 @@ rnp_op_encrypt_add_password(rnp_op_encrypt_t op,
...
+        return rnp_ctx_add_encryption_password(&op->rnpctx, password, hash_alg, symm_alg, iterations);
+    }
+    // ask for password if it is NULL
+    char ask_pass[MAX_PASSWORD_LENGTH] = {0};
+    pgp_password_ctx_t pswdctx = {.op = PGP_OP_ENCRYPT_SYM, .key = NULL};
+    if (!pgp_request_password(&op->ffi->pass_provider, &pswdctx, ask_pass, sizeof(ask_pass))) {
+        return RNP_ERROR_BAD_PASSWORD;
+    }
+    rnp_result_t ret = rnp_ctx_add_encryption_password(&op->rnpctx, ask_pass, hash_alg, symm_alg, iterations);
+    pgp_forget(ask_pass, sizeof(ask_pass));
```
dewyatt: 1) It would be nice if we can avoid two separate calls to `rnp_ctx_add_encryption_password`.
2) Since we're more C++-y now, we ought to rely on destructors for zeroing memory. (Botan's `secure_vector` will do this)

Not well tested but something like this maybe:
```diff
diff --git a/src/lib/rnp.cpp b/src/lib/rnp.cpp
index cb68a914..1bdca7c3 100644
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -53,6 +53,7 @@
 #include "utils.h"
 #include "json_utils.h"
 #include "version.h"
+#include <botan/secmem.h>
 
 struct rnp_key_handle_st {
     rnp_ffi_t        ffi;
@@ -2180,19 +2181,16 @@ rnp_op_encrypt_add_password(rnp_op_encrypt_t op,
         FFI_LOG(op->ffi, "Invalid cipher: %s", s2k_hash);
         return RNP_ERROR_BAD_PARAMETERS;
     }
-    // just add password if is given
-    if (password) {
-        return rnp_ctx_add_encryption_password(&op->rnpctx, password, hash_alg, symm_alg, iterations);
-    }
-    // ask for password if it is NULL
-    char ask_pass[MAX_PASSWORD_LENGTH] = {0};
-    pgp_password_ctx_t pswdctx = {.op = PGP_OP_ENCRYPT_SYM, .key = NULL};
-    if (!pgp_request_password(&op->ffi->pass_provider, &pswdctx, ask_pass, sizeof(ask_pass))) {
-        return RNP_ERROR_BAD_PASSWORD;
+    Botan::secure_vector<char> ask_pass(MAX_PASSWORD_LENGTH, '\0');
+    if (!password) {
+        pgp_password_ctx_t pswdctx = {.op = PGP_OP_ENCRYPT_SYM, .key = NULL};
+        if (!pgp_request_password(
+              &op->ffi->pass_provider, &pswdctx, &ask_pass[0], ask_pass.size())) {
+            return RNP_ERROR_BAD_PASSWORD;
+        }
+        password = ask_pass.data();
     }
-    rnp_result_t ret = rnp_ctx_add_encryption_password(&op->rnpctx, ask_pass, hash_alg, symm_alg, iterations);
-    pgp_forget(ask_pass, sizeof(ask_pass));
-    return ret;
+    return rnp_ctx_add_encryption_password(&op->rnpctx, password, hash_alg, symm_alg, iterations);
 }
 
 rnp_result_t
```

Of course we do incur the constructor/destructor cost in both cases in the above. Open for discussion!

ronaldtse: I'm happy to delegate to destructors for cleanliness...

ni4: @dewyatt Thanks, it look much cleaner in this way.
And constructor/destructor cost may be called zero compared to hash function iterations.

The only thing is that we most likely need to handle exceptions in this case, since Botan's secure_allocator's method `allocate_memory` is written in the following way (while the standard allocator is defined as nothrow since C++11):
```C++
BOTAN_MALLOC_FN void* allocate_memory(size_t elems, size_t elem_size)
   {
#if defined(BOTAN_HAS_LOCKING_ALLOCATOR)
   if(void* p = mlock_allocator::instance().allocate(elems, elem_size))
      return p;
#endif

   void* ptr = std::calloc(elems, elem_size);
   if(!ptr)
      throw std::bad_alloc();
   return ptr;
   }
```
Am I correct, or just overcomplicating things?

dewyatt: @ni4 Good point! I did some thinking about whether we should start handling exceptions in rnp a while back but things are kind of a mess. There are a few proposals I've been watching like std::expected and zero-overhead exceptions that may be nice in the future.

For now, maybe we should avoid this. We can still get rid of the duplicate calls to `rnp_ctx_add_encryption_password` by doing something like `password = ask_pass` (and calling `pgp_forget` regardless).

ni4: @dewyatt Actually I added `try {  } catch(...) {...}` in few places in code, so maybe just use it here as well? Or try-catch in C/C++ mixed code has other problems which I'm not aware of?

dewyatt: It's not a problem per se, it's more that it increases indentation and adds some clutter. If you want to do that here, I say no problem!

ni4: @dewyatt Ah, okay then. Please see the updated code.
I don't like extra clutter as well, but it seems there is not good way to avoid it at the moment.

ronaldtse:
@dewyatt ?

PR #927, opened 2019-10-21
FFI: add function to dump signature packet to JSON

ni4:
This PR adds function which dumps signature packet information to JSON.
This would allow to get required info via JSON instead of having FFI function for each possible field.

As a side effect also function `rnp_op_verify_signature_get_handle` was added, allowing to have single interface to signature functions.

P.S. It is draft since it uses cherry-picked commits from yet unmerged PR #924 and is a part of #920.


ni4:
@ronaldtse @dewyatt could you please review this? Not sure whether GitHub notifies when status is changed from draft to ready for review.

ronaldtse:
Looks like the "draft => non-draft PR" action won't cause any notification. Best to ping after the change.

PR #928, opened 2019-10-21
Check for NULL in rnp_op_set_hash()

ni4:
It appeared that we were not checking for NULL in `rnp_op_set_hash()`, so NULL parameter value was causing AV which is not good.


ronaldtse:
Thanks @ni4 !

PR #931, opened 2019-10-25
FFI: add function rnp_op_verify_is_signed()

ni4:
At the moment it is impossible to distinguish via FFI whether input data was encrypted or encrypted-and-signed, so this function was added.
Using it we may process any encrypted and/or signed data in a single rnp_op_verify_*() call without trying to guess source message type.


dewyatt:
@ni4 Are you sure it's needed, can't the same check be done with `rnp_op_verify_get_signature_count`?

ni4:
@dewyatt Looks like I over-engineered things here, you are right. Actually was thinking about the cases when message is signed, but signature(s) is invalid, so sigcount is 0. However in this case rnp_op_verify_execute() will return an error which will be enough to understand that message is broken. So closing this and made another PR #935 to migrate test's code and update it with some more cases.

PR #945, opened 2019-11-01
Add test for large packet verification

### LINKED ISSUE #940: Ensure correct work with large packet sizes.
## Description

We need to check whether rnp correctly handles as large as possible packets.
This could be checked with zipped and signed 4Gb literal packet, built from single symbol, so it can be compressed to reasonable size to be added to tests suite.

Construct literal packet with largest possible size (4gb), using the 4-byte packet length, and compressed and then signed stream on top of it, and make sure it is processed well with rnp_op_verify() call.
Separate test file should be added to rnp_tests() suite.

rrrooommmaaa:
Added bzip2-zipped file containing a signed literal packet of 4G-1 bytes
test_large_packet() function test verification of this large packet.
Commented out code shows how to generate this file.
Closes #940


ni4:
@rrrooommmaaa Thanks for the PR. Please take a look at comment I left. 
Also no need to duplicate key files for the test - you may borrow those from `/data/keyrings/1`.
Other than this things looks good to me.

rrrooommmaaa:
@ni4 I have updated the test to use the keyring. Is there anything else for me to do?

ni4:
@rrrooommmaaa 

> So you'll need to manually construct the header via hex editor or code, so it will contain single, fixed-length packet (instead if chunked).

In the way you constructed that 4g file it uses partial lengths:
```
    :off 15: packet header 0xcbed (tag 11, ***partial len***)
    Literal data packet
        data format: 'b'
        filename:  (len 0)
        timestamp: 0 (Thu Jan  1 03:00:00 1970)
        data bytes: 4294967295
```

While the idea of this test is to be able to handle 4G fixed-length packet.
To construct it, you'll need to manually make packet header, then append data, then wrap this into compressed stream, and then wrap into signed stream.

You can read about the OpenPGP packet headers here: https://gitlab.com/openpgp-wg/rfc4880bis/blob/master/draft-ietf-openpgp-rfc4880bis-08.txt , section 4.2

ronaldtse:
@ni4 @dewyatt seems that this is done, could you help check? Thanks!

ni4:
@ronaldtse Nope, not at the moment - https://github.com/rnpgp/rnp/pull/945#issuecomment-549036929 is still not addressed.

ronaldtse:
Oh sorry for jumping the gun :wink:

rrrooommmaaa:
@ni4 There is an issue with the length that is encoded in the packet header.
Looks like this length includes the header (throughout the code). I'm looking at what exactly is specified in the protocol and there are interesting lines:
```
   A packet with length 100 may have its length encoded in one octet:
   0x64.  This is followed by 100 octets of data.

   A packet with length 1723 may have its length encoded in two octets:
   0xC5, 0xFB.  This header is followed by the 1723 octets of data.
```
According to this, the header isn't counted in `length' .  I generated a literal packet with length specified as 0xFFFFFFFF folllowed by 0xFFFFFFFF octets of data and it wasn't read correctly by the stream parser. I fixed the parser (see commit) and now keys can't be read.
So we have to either update the specification or implementation. Looks like they don't match.

ni4:
@rrrooommmaaa yeah, you should not change code in `stream-parse.cpp`, it had correct behavior.
Also it is good idea to run all tests locally when you change library internals to check for regression :-)

The confusion here is because of double usage of the word `header`: there is a header of packet itself, which is universal for all packets and contains packet type and length, and there is a header of literal data packet, which contains flags, file name, modification date, etc, before the actual data itself.

Also in such cases it is good to check not only with `rnp`, but with `gpg` as well:
```
rnp-build nickolay$ gpg --verify src/tests/data/test_large_packet/4g.bzip2.gpg 
gpg: [don't know]: invalid packet (ctb=58)
gpg: no signature found
gpg: uncompressing failed: No data
gpg: the signature could not be verified.
Please remember that the signature file (.sig or .asc)
should be the first file given on the command line.
```

So, once we hit this stone, please also add test case to `cli_tests.py`, which checks whether 4g.bzip2.gpg is successfully processed with gpg.

ni4:
@rrrooommmaaa Please also add test with GPG cross-verify (it verifies file ok now). Also it seems there are some problems with gpg key servers right now, let's wait for a while to restart Travis for successful build.

rrrooommmaaa:
@ni4 is it ok now?

ni4:
@rrrooommmaaa Not sure whether my comment on the commit 2d68562 triggered any notification.
> Thanks, but could you please move this call to cli_tests.py (to the Misc class)?

Basically, it would be better to keep cmake files as clean as possible and add test cases to cli_tests.py/rnp_tests instead.

rrrooommmaaa:
@ni4 I can neither see your comments nor receive notifications on them. Perhaps, they are set to visible to only certain users.

ronaldtse:
FWIW I can't see that particular comment too 😭

ni4:
@rrrooommmaaa @ronaldtse Arghh, maybe that's because of 'Pending' status:
Screen Shot 2019-11-05 at 23 04 03

Sorry then. Anyway, now PR looks good, let's just wait for Travis to finish.

ronaldtse:
@ni4 ah, you will have to act on the review, such as "Request changes" then the review will be posted.

PR #946, opened 2019-11-04
Hash uses Botan's C++ interface

joke325:
Converted pgp_hash_t to use Botan's C++ interface

### DIFF
--- a/src/lib/crypto/hash.cpp
+++ b/src/lib/crypto/hash.cpp
@@ -157,7 +157,6 @@ bool
 pgp_hash_create(pgp_hash_t *hash, pgp_hash_alg_t alg)
 {
     const char * hash_name = pgp_hash_name_botan(alg);
-    int          rc;
 
     if (hash_name == NULL) {
         return false;
@@ -266,23 +265,27 @@ pgp_hash_finish(pgp_hash_t *hash, uint8_t *out)
     }
 
     Botan::HashFunction *hash_fn = static_cast<Botan::HashFunction*>(hash->handle);
+    hash->handle = NULL;
+    if(hash_fn)
+    {
+        try
+        {
+            if(out)
+                hash_fn->final(out);
+            delete hash_fn;
+        }
+        catch(std::exception& ex)
+        {
+            RNP_LOG("Error finishing HashFunction ('%s')", ex.what());
+            outlen = 0;
+        }
+    }
+    hash->_output_len = 0;
+
     if(!out || !hash_fn) {
         RNP_LOG("Hash finalization failed");
         return 0;
     }
-
-    hash->handle = NULL;
-    try
-    {
-        hash_fn->final(out);
-        delete hash_fn;
-    }
-    catch(std::exception& ex)
-    {
-         RNP_LOG("Error finishing HashFunction ('%s')", ex.what());
-	 outlen = 0;
-    }
-    hash->_output_len = 0;
     return outlen;
 }
 
--- a/src/lib/crypto/hash.cpp
+++ b/src/lib/crypto/hash.cpp
@@ -157,26 +157,32 @@ bool
 pgp_hash_create(pgp_hash_t *hash, pgp_hash_alg_t alg)
 {
     const char * hash_name = pgp_hash_name_botan(alg);
-    size_t       outlen;
     int          rc;
 
     if (hash_name == NULL) {
         return false;
     }
 
-    std::unique_ptr<Botan::HashFunction> hash_fn = Botan::HashFunction::create(hash_name);
-    if(hash_fn == NULL) {
+    std::unique_ptr<Botan::HashFunction> hash_fn;
+    try
+    {
+        hash_fn = Botan::HashFunction::create(hash_name);
+    }
+    catch(std::exception& ex)
+    {
+        RNP_LOG("Error creating HashFunction ('%s')", ex.what());
+    }
+    if(!hash_fn) {
         RNP_LOG("Error creating hash object for '%s'", hash_name);
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


src/lib/crypto/hash.cpp
```diff
@@ -222,7 +222,8 @@ pgp_hash_add_int(pgp_hash_t *hash, unsigned n, size_t length)
 int
 pgp_hash_add(pgp_hash_t *hash, const void *buf, size_t len)
 {
-    return botan_hash_update((botan_hash_t) hash->handle, (const uint8_t *) buf, len);
+    static_cast<Botan::HashFunction*>(hash->handle)->update((const uint8_t*)buf, len);
```
dewyatt: May as well use `static_cast` for `buf` since we're doing C++-style casting here.

Also, what happens when `hash == nullptr`, `buf == nullptr`, `len == 0`?

src/lib/crypto/hash.cpp
```diff
@@ -233,12 +234,15 @@ pgp_hash_finish(pgp_hash_t *hash, uint8_t *out)
         return 0;
     }
 
-    if (out && botan_hash_final((botan_hash_t) hash->handle, out)) {
+    if(out == NULL) {
```
dewyatt: Style: `!out`

src/lib/crypto/hash.cpp
```diff
@@ -156,30 +157,28 @@ bool
...
     size_t       outlen;
     int          rc;
 
     if (hash_name == NULL) {
         return false;
     }
 
-    rc = botan_hash_init(&impl, hash_name, 0);
-    if (rc != 0) {
+    std::unique_ptr<Botan::HashFunction> hash_fn = Botan::HashFunction::create(hash_name);
```
dewyatt: I think this can throw since it calls `new`.

src/lib/crypto/hash.cpp
```diff
@@ -190,14 +189,15 @@ pgp_hash_copy(pgp_hash_t *dst, const pgp_hash_t *src)
         return false;
     }
 
-    botan_hash_t handle;
-    if (botan_hash_copy_state(&handle, (botan_hash_t) src->handle)) {
+    Botan::HashFunction *hash_fn = static_cast<Botan::HashFunction*>(src->handle);
```
dewyatt: What happens when `src->handle` is NULL? Do you think we should check for this case or just crash (which is sometimes fine)?

src/lib/crypto/hash.cpp
```diff
@@ -190,14 +189,15 @@ pgp_hash_copy(pgp_hash_t *dst, const pgp_hash_t *src)
...
+    std::unique_ptr<Botan::HashFunction> handle = hash_fn->copy_state();
```
dewyatt: This can throw.

src/lib/crypto/hash.cpp
```diff
@@ -233,12 +234,15 @@ pgp_hash_finish(pgp_hash_t *hash, uint8_t *out)
...
 
-    if (out && botan_hash_final((botan_hash_t) hash->handle, out)) {
+    if(out == NULL) {
         RNP_LOG("Hash finalization failed");
         return 0;
     }
-    botan_hash_destroy((botan_hash_t) hash->handle);
+
+    Botan::HashFunction *hash_fn = static_cast<Botan::HashFunction*>(hash->handle);
+    hash_fn->final(out);
```
dewyatt: I think this can throw.

dewyatt:
It looks like we're going to run into (probably unrelated) issues with the sanitize build in CI. Let me see.

src/lib/crypto/hash.cpp
```diff
@@ -222,7 +233,19 @@ pgp_hash_add_int(pgp_hash_t *hash, unsigned n, size_t length)
 int
 pgp_hash_add(pgp_hash_t *hash, const void *buf, size_t len)
 {
-    return botan_hash_update((botan_hash_t) hash->handle, (const uint8_t *) buf, len);
+    if (!hash->handle) {
+        return -1;
+    }
+
+    int error = 0;
```
dewyatt: This variable seems unnecessary, I'd rather we drop it and `return -2` directly in the catch body.

src/lib/crypto/hash.cpp
```diff
@@ -222,7 +233,19 @@ pgp_hash_add_int(pgp_hash_t *hash, unsigned n, size_t length)
...
+    try {
+        static_cast<Botan::HashFunction *>(hash->handle)
+          ->update(static_cast<const uint8_t *>(buf), len);
+    } catch (std::exception &ex) {
+        RNP_LOG("Error adding to HashFunction ('%s')", ex.what());
+        error = -2;
+    }
+    return error;
```
dewyatt: So just `return 0` here after the other change.

src/lib/crypto/hash.cpp
```diff
@@ -233,13 +255,24 @@ pgp_hash_finish(pgp_hash_t *hash, uint8_t *out)
         return 0;
     }
 
-    if (out && botan_hash_final((botan_hash_t) hash->handle, out)) {
+    Botan::HashFunction *hash_fn = static_cast<Botan::HashFunction *>(hash->handle);
+    hash->handle = NULL;
+    if (hash_fn) {
+        try {
+            if (out)
+                hash_fn->final(out);
```
dewyatt: By the way, style nit here, we always use braces even on one-line conditionals. See the [Style Guide](https://github.com/rnpgp/rnp/blob/master/docs/develop.adoc#style-guide).

Not something to hold up a merge though.

PR #950, opened 2019-11-05
(WIP) added test for checking key pkt length while importing

msdev428:


ni4:
@msdev428 as you see test is failing (most likely because RNP_SUCCESS == 0, instead of true).
Also please add 'positive' test cases - where key with 2-byte and 4-byte packet lengths are loaded correctly (and their self-signatures are verified correctly as well).
And, it's up to you - but I'd just edit key file in hex editor to modify header instead of doing that programmatically.

ronaldtse:
Ping @msdev428 could you help handle the change requests? Thanks.

msdev428:
@ni4 can you please check https://github.com/rnpgp/rnp/pull/950/checks?check_run_id=297779459?
It indicates some memory leaks but I think it's not from testing code.

msdev428:
btw, I'm using Ubuntu 18.04 for develop but I can't run tests properly on that system.
what platform do you recommended for development?

ni4:
@msdev428 Let's wait for tests to finish. I use macos + Visual Studio Code, however Linux/Ubuntu should work well too. @dewyatt uses some Linux AFAIK.
Since cmake is the same you should be able to run `ctest -V -R rnp_tests` without any problems.
Also can run tests in Docker container via `./docker.sh` script.

msdev428:
@ni4 I didn't free allocated memories for `ffi` and `input`. Now all tests were passed.
I will try to add test cases for normal key with 2bytes, 4bytes length.

msdev428:
@ni4 @dewyatt Can you try to review whether I used right commands for local testing?
```
mkdir build
cd build
cmake ..
ctest -V -R rnp_tests
```
When I used these commands for testing, I don't get proper log messages.
I attached it.
[LastTest.log](https://github.com/rnpgp/rnp/files/3832596/LastTest.log)

ni4:
@msdev428 
```
/bin/cp: cannot stat '(null)': No such file or directory
/home/msdev/work/ribose/GitWork/rnp/src/tests/support.cpp:239: Failure
```
This tells that copying of test data files fails, leading to line which calls it - `copy_recursively(getenv("RNP_TEST_DATA"), m_dir)`. 
So there could be 2 cases - RNP_TEST_DATA env variable is not (however it should be set by cmake), or m_dir, which is the result of `make_temp_dir()`, is NULL.
So please add debug output somewhere close by and check which one is failing. I just few days ago got hit by `make_temp_dir()` not working on MinGW/Windows, maybe you have the same issue.

msdev428:
@ni4 Thanks! It's working after set `RNP_TEST_DATA`.

src/tests/rnp_tests.h
```diff
@@ -62,6 +62,8 @@ void rnpkeys_generatekey_verifykeyHomeDirOption(void **state);
 
 void rnpkeys_generatekey_verifykeyKBXHomeDirOption(void **state);
 
+void rnpkeys_generatekey_verifykeyNonexistingHomeDir(void **state);
+
```
ni4: This looks like rebase-introduced artifact, as this test was removed some time ago.

src/tests/ffi.cpp
```diff
@@ -5279,6 +5279,124 @@ TEST_F(rnp_tests, test_ffi_keys_import)
...
+static bool
+fake_key_invalid_length(const char *valid_key_path,
+    const char *invalid_key_0_path, const char *invalid_key_1_path, const char *invalid_key_2_path)
+{
+    FILE *fp;
+    size_t read_len;
+    unsigned char pkt[2048];
+    unsigned char pkt_bit;
+
+    unsigned char fake_pkt[2048];
```
ni4: As we are moving to C++, could you please use std::string/std::vector here instead of static arrays/char * ?

ronaldtse:
@rrrooommmaaa I think this task has been abandoned. Do you have time to complete this PR? Thanks!

rrrooommmaaa:
@ronaldtse Sure, I'll take it

ronaldtse:
Thanks @rrrooommmaaa !

ronaldtse:
@rrrooommmaaa we will have to create a new dev branch and a new PR.

PR #954, opened 2019-11-10
Replaced and removed rnp_key_store_get_key_by_name() in tests.

### LINKED ISSUE #948: Replace rnp_key_store_get_key_by_name() in tests and remove it.
## Description

Function `rnp_key_store_get_key_by_name()` was used in non-FFI CLI, and now is deprecated.
So it and all underlying code should be removed and replaced by other functions in tests.

joke325:
Closes #948.

### DIFF
--- a/src/tests/support.cpp
+++ b/src/tests/support.cpp
@@ -672,3 +672,52 @@ check_json_field_bool(json_object *obj, const std::string &field, bool value)
     }
     return json_object_get_boolean(fld) == value;
 }
+
+/* return the next key which matches, starting searching after *after */
+pgp_key_t* 
+rnp_test_key_by_name(const rnp_key_store_t *keyring,
+                const char *           name,
+                pgp_key_t *            after)
+{
+    pgp_key_t *key = NULL;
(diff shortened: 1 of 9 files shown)


src/tests/support.cpp
```diff
@@ -672,3 +672,52 @@ check_json_field_bool(json_object *obj, const std::string &field, bool value)
...
+                if (regexec(&re, (char *) pgp_key_get_userid(keyp, i), 0, NULL, 0) == 0) {
+                    key = keyp;
+                }
+            }
+        }
+
+        regfree(&re);
+    }
+    return key;
+}
```
ni4: In this way tests will mostly test this function instead of direct library code.
Please change the behavior to the following:
- use functions `rnp_key_store_get_key_by_id()/rnp_key_store_get_key_by_fp()/rnp_key_store_get_key_by_grip()` where applicable. It would be handy to add some wrapper with `std::string` input parameter instead of uint8_t *
- where search is done by userid you could also add some minor wrapper which will call `rnp_key_store_search()`.
- only now I noticed that `rnp_keystore_get_key_by_userid` is also not used anywhere

PR #956, opened 2019-11-11
Fixes numerous 'Hash finalization failed' message while running rnp_t…

### LINKED ISSUE #955: Numerous 'Hash finalization failed' message while running rnp_tests
## Description

PR #946 introduced numerous error message output even on successfull tests run, which it didn't expose previously:
```
[pgp_hash_finish() /Users/nickolay/Ribose/rnp/src/lib/crypto/hash.cpp:274] Hash finalization failed
[pgp_hash_finish() /Users/nickolay/Ribose/rnp/src/lib/crypto/hash.cpp:274] Hash finalization failed
[pgp_hash_finish() /Users/nickolay/Ribose/rnp/src/lib/crypto/hash.cpp:274] Hash finalization failed
```

This is caused by incorrect handling of NULL `out` parameter, which is legitimate situation when you just want to cancel hashing and free allocated resources.

joke325:
Closes #955.

### DIFF
--- a/src/lib/crypto/hash.cpp
+++ b/src/lib/crypto/hash.cpp
@@ -256,24 +256,22 @@ pgp_hash_finish(pgp_hash_t *hash, uint8_t *out)
     }
 
     Botan::HashFunction *hash_fn = static_cast<Botan::HashFunction *>(hash->handle);
-    hash->handle = NULL;
-    if (hash_fn) {
-        try {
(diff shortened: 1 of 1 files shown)


ni4:
@joke325 it should be okay with codecov, since diff is small and most of it lines are error handling which aren't hit.

PR #957, opened 2019-11-13
Replaced and removed rnp_key_store_get_key_by_name() in tests.

### LINKED ISSUE #948: Replace rnp_key_store_get_key_by_name() in tests and remove it.

joke325:
Closes #948.

### DIFF
--- a/src/tests/support.cpp
+++ b/src/tests/support.cpp
@@ -677,40 +677,37 @@ pgp_key_t*
 rnp_tests_get_key_by_id(const rnp_key_store_t* keyring, const std::string& keyid, pgp_key_t* after)
 {
     pgp_key_t *key = NULL;
-    uint8_t keyid_bin[PGP_KEY_ID_SIZE];
+    std::vector<uint8_t> keyid_bin(PGP_KEY_ID_SIZE, 0);
     size_t binlen = 0;
 
-    if (!keyring || keyid.length() == 0) {
+    if (!keyring || keyid.empty()) {
         return NULL;
     }
     assert(!after || list_is_member(keyring->keys, (list_item *) after));
 
-    memset(keyid_bin, 0x0, sizeof(keyid_bin));
-    if (ishex(keyid.c_str(), keyid.size()) && hex2bin(keyid.c_str(), keyid.size(), keyid_bin, sizeof(keyid_bin), &binlen)) {
+    if (ishex(keyid.c_str(), keyid.size()) && hex2bin(keyid.c_str(), keyid.size(), keyid_bin.data(), keyid_bin.size(), &binlen)) {
         if (binlen <= PGP_KEY_ID_SIZE) {
-            key = rnp_key_store_get_key_by_id(keyring, keyid_bin, after);
+            key = rnp_key_store_get_key_by_id(keyring, keyid_bin.data(), after);
         }
     }
     return key;
 }
 
 pgp_key_t*
-rnp_tests_get_key_by_fpr(const rnp_key_store_t* keyring, const std::string& keyid, pgp_key_t* after)
+rnp_tests_get_key_by_fpr(const rnp_key_store_t* keyring, const std::string& keyid)
 {
     pgp_key_t *key = NULL;
-    uint8_t keyid_bin[PGP_FINGERPRINT_SIZE];
+    std::vector<uint8_t> keyid_bin(PGP_FINGERPRINT_SIZE, 0);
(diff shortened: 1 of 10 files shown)


src/tests/support.cpp
```diff
@@ -672,3 +672,60 @@ check_json_field_bool(json_object *obj, const std::string &field, bool value)
     }
     return json_object_get_boolean(fld) == value;
 }
+
+pgp_key_t*
+rnp_tests_get_key_by_id(const rnp_key_store_t* keyring, const std::string& keyid, pgp_key_t* after)
```
ni4: No need to add parameter after - every call uses NULL here. And if it will be needed in future then it can be easily added.

src/tests/support.cpp
```diff
@@ -672,3 +672,60 @@ check_json_field_bool(json_object *obj, const std::string &field, bool value)
...
+{
+    pgp_key_t *key = NULL;
+    uint8_t keyid_bin[PGP_KEY_ID_SIZE];
```
ni4: Since we are migrating to C++ then could you please use C++ vector here (and in other places where new code is added)?

src/tests/support.cpp
```diff
@@ -672,3 +672,60 @@ check_json_field_bool(json_object *obj, const std::string &field, bool value)
...
+    size_t binlen = 0;
+
+    if (!keyring || keyid.length() == 0) {
```
ni4: `keyid.empty()` would look better here.

src/tests/support.cpp
```diff
@@ -672,3 +672,60 @@ check_json_field_bool(json_object *obj, const std::string &field, bool value)
...
+    if (ishex(keyid.c_str(), keyid.size()) && hex2bin(keyid.c_str(), keyid.size(), keyid_bin, sizeof(keyid_bin), &binlen)) {
+        if (binlen <= PGP_KEY_ID_SIZE) {
+            key = rnp_key_store_get_key_by_id(keyring, keyid_bin, after);
+        }
+    }
+    return key;
+}
+
+pgp_key_t*
+rnp_tests_get_key_by_fpr(const rnp_key_store_t* keyring, const std::string& keyid, pgp_key_t* after)
```
ni4: The same comments as for `rnp_tests_get_key_by_id`.

src/tests/support.cpp
```diff
@@ -672,3 +672,60 @@ check_json_field_bool(json_object *obj, const std::string &field, bool value)
...
+            pgp_fingerprint_t fp = { {}, static_cast<unsigned>(binlen) };
+            memcpy(fp.fingerprint, keyid_bin, binlen);
+            key = rnp_key_store_get_key_by_fpr(keyring, &fp);
+        } 
+    }
+    return key;
+}
+
+pgp_key_t*
+rnp_tests_key_search(const rnp_key_store_t* keyring, const std::string& keyid, pgp_key_t* after)
```
ni4: The same comments as for the previous (where applicable).

src/tests/user-prefs.cpp
```diff
@@ -69,7 +69,7 @@ TEST_F(rnp_tests, test_load_user_prefs)
 
         // find the key
         pgp_key_t *key = NULL;
-        assert_non_null(key = rnp_key_store_get_key_by_name(pubring, userid, NULL));
+        assert_non_null(key = rnp_tests_key_search(pubring, userid, NULL));
         assert_non_null(key);
```
ni4: Once got here, we could also remove duplicate `assert_non_null(key)`, probably it was added by mistake during refactoring or other changes.

src/tests/user-prefs.cpp
```diff
@@ -114,7 +114,7 @@ TEST_F(rnp_tests, test_load_user_prefs)
...
```
ni4: The same - duplicate check for non-NULL.

src/tests/support.cpp
joke325: The 'after' is used, see 'key-store-search.cpp' lines 111, 140.

ni4: @joke325 Right. Sorry, mislooked that place.

ni4:
@ronaldtse @dewyatt Could you please review this to have to approvals?

PR #959, opened 2019-11-14
Check partial packet length edge cases.

### LINKED ISSUE #939: Check partial packet length edge cases.
## Description

rrrooommmaaa:
- Less than 512 bytes first chunk is not produced
- If the first chunk is less than 512 bytes, a warning is logged
- Largest possible chunk of 1G passed
- 0-size last chunk allowed
- partial lengths in `signature` and `public key` packets result in error.
Should we test all other packet types? There are plenty.

Closes #939


src/tests/partial-length.cpp
```diff
@@ -0,0 +1,262 @@
...
+pass_provider(rnp_ffi_t         ffi,
+            void *           app_ctx,
+            rnp_key_handle_t key,
+            const char *     pgp_context,
+            char             buf[],
+            size_t           buf_len)
+{
+    strncpy(buf, "password", buf_len);
+    return true;
+}
```
ni4: To avoid code duplication please use `ffi_string_password_provider` from the `support.h`.

src/tests/partial-length.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+static void
+test_partial_length_init(rnp_ffi_t *ffi, uint32_t key_flags)
+{
+    rnp_input_t     input = NULL;
+
+    /* init ffi */
+    assert_rnp_success(rnp_ffi_create(ffi, "GPG", "GPG"));
+
```
ni4: It would be better to not have extra empty lines :)

src/tests/partial-length.cpp
```diff
@@ -0,0 +1,262 @@
...
+    // init ffi
+    test_partial_length_init(&ffi, RNP_LOAD_SAVE_SECRET_KEYS);
+
+    // generate a sequence of octets with appropriate length using callback
+    dummy_reader_ctx_st reader_ctx;
+    reader_ctx.dummy = 'X';
+    reader_ctx.remaining = uncacheable_size;
+    assert_rnp_success(rnp_input_from_callback(&input, dummy_reader, NULL, &reader_ctx));
+
```
ni4: Extra empty lines, again.

src/tests/partial-length.cpp
```diff
@@ -0,0 +1,262 @@
...
+    assert_true(stream_read_partial_chunk_len(&src, &last) >= PGP_PARTIAL_PKT_FIRST_PART_MIN_SIZE);
+    assert_false(last);
+
+    // cleanup
+    src_close(&src);
+    assert_rnp_success(rnp_op_sign_destroy(sign));
+    assert_rnp_success(rnp_input_destroy(input));
+    assert_rnp_success(rnp_output_destroy(output));
+    assert_rnp_success(rnp_ffi_destroy(ffi));
+}
```
ni4: Empty line afterwards would save us from seeing 'No empty line at the end of file' warning.

ni4:
@rrrooommmaaa Please see comments. Also please rebase on master, so it can be merged (and doesn't contain 'merge pull request' commits), and squash small commits like 'type fixed'.

rrrooommmaaa:
@ni4 updated according to your request

src/tests/partial-length.cpp
ni4: @rrrooommmaaa  This wasn't resolved or do I miss something?

rrrooommmaaa: @ni4 yes, strange. I have applied the changes now.

ni4: @rrrooommmaaa yeah, now it looks good. Except two duplicate commit messages + merge commit, could you please rebawse/squash so there would be no extra commits in the history?

rrrooommmaaa:
@ni4 Can assist what kind of rebase should I do against what?
If I do simply `git rebase upstream/master` it shows some conflicts and after resolution adds `Merge` commit.

ni4:
@rrrooommmaaa From the GUI it looked as the following: 
 - update your fork's master to match original
 - interactively rebase your branch on your master
 - force-push your branch

rrrooommmaaa:
@ni4 I only have `master` branch in my repository. What do you mean by `my branch`?

ni4:
@rrrooommmaaa got it, this should be the source of problem. Then you can create new branch (see the development readme for branch naming  conventions), cherry-pick commits from master to that branch, and then you'll be able to do the previous. Also not sure whether GitHub would allow to change branch in pull request - possibly you'll need to re-create pull request.

PR #965, opened 2019-11-18
Test partial length packet edge cases

### LINKED ISSUE #939: Check partial packet length edge cases.

rrrooommmaaa:
Closes #939


ronaldtse:
Agree with @ni4 , @rrrooommmaaa can you help consolidate the commits and give it a good name? Thanks!

ronaldtse:
@rrrooommmaaa you don't need to close the PR, you can just "force push" to your PR branch, and the PR will be updated. Thanks.

rrrooommmaaa:
@ronaldtse I force-pushed it yesterday before your comment, but can't remove 'Change requested' flag

ronaldtse:
@rrrooommmaaa because the change requested flag can only be removed by the reviewer. Next time you can just ping us and we can resolve the review comment instead of opening a new PR.

PR #966, opened 2019-11-18
Generating keys with multiple userids is unsupported.

### LINKED ISSUE #929: Generating a key with multiple userids silently ignores all but last userid
## Description

Generating a key with multiple userids silently ignores all but last userid.

## Steps to Reproduce

```
% rnpkeys --homedir /tmp/rnphome --generate-key --userid="foo@example.org" --userid="bar@example.org"
[init_file_src() /home/teythoon/repos/pep/pgpzoo/src/rnp/src/librepgp/stream-common.cpp:378] can't stat '/tmp/rnphome/pubring.gpg'
wrong pubring path
Generating a new key...

sec   2048/RSA (Encrypt or Sign) e5505fc4d404a055 2019-10-24 [SC]
      c997b53e214cfd14ae2453f3e5505fc4d404a055
uid           bar@example.org
ssb   2048/RSA (Encrypt or Sign) 55d1444a84b39413 2019-10-24 [E]
      130b33cffbe9ee5b784ffc8255d1444a84b39413
% rnpkeys --homedir /tmp/rnphome --export-key c997b53e214cfd14ae2453f3e5505fc4d404a055 | sq inspect
-: Transferable Public Key.

    Fingerprint: C997 B53E 214C FD14 AE24  53F3 E550 5FC4 D404 A055
Public-key algo: RSA (Encrypt or Sign)
Public-key size: 2048 bits
  Creation time: 2019-10-24T10:18
       Keyflags: certification, signing

         Subkey: 130B 33CF FBE9 EE5B 784F  FC82 55D1 444A 84B3 9413
Public-key algo: RSA (Encrypt or Sign)
Public-key size: 2048 bits
  Creation time: 2019-10-24T10:18
       Keyflags: transport encryption, data-at-rest encryption

         UserID: bar@example.org
```

## Expected Behavior

Preferably generate a key with two userids, failing that, signal an error that giving multiple `--userid` arguments is not supported.

## Actual Behavior

All but the last `--userid` parameters are silently ignored.

joke325:
Closes #929

### DIFF
--- a/src/rnp/rnpcfg.cpp
+++ b/src/rnp/rnpcfg.cpp
@@ -66,6 +66,7 @@ typedef struct rnp_cfg_val_t {
 typedef struct rnp_cfg_item_t {
     char *        key;
     rnp_cfg_val_t val;
+    std::vector<rnp_cfg_val_t> multiVal;
 } rnp_cfg_item_t;
 
 void
@@ -110,6 +111,9 @@ rnp_cfg_val_free(rnp_cfg_val_t *val)
 static void
 rnp_cfg_item_free(rnp_cfg_item_t *item)
 {
+    for (auto i = item->multiVal.begin(); i != item->multiVal.end(); i++) {
+        rnp_cfg_val_free(&*i);
+    }
     rnp_cfg_val_free(&item->val);
     free(item->key);
     item->key = NULL;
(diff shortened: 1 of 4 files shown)


ni4:
@joke325 tests are failing here. Also there is no need to add new fields/values to rnp_cfg since it already has function to operate lists - see functionality on signers/recipient lists, and functions `rnp_cfg_addstr`/`rnp_cfg_copylist_string`. It should be good to add `rnp_cfg_addstring`, like it is done with `rnp_cfg_getstr`, but no need to change rnpcfg_t internals yet - those will be changed to C++  as a separate task, when all outer code will be more close to transition.

joke325:
@ni4 What this error means?

PR #967, opened 2019-11-20
Test partial length packet edge cases

### LINKED ISSUE #939: Check partial packet length edge cases.

rrrooommmaaa:
Closes #939


rrrooommmaaa:
@ni4 Is there anything I should do about codecov?

ni4:
@rrrooommmaaa Nope, it looks like something went wrong with codecov itself. Following the link brings correct coverage and icrease by 0.08%.

PR #968, opened 2019-11-20
Added an installed header containing RNP_SUCCESS

### LINKED ISSUE #897: No installed header containing RNP_SUCCESS
Using "make install", only the rnp.h header is installed.

That header mentions (in comments) the symbol RNP_SUCCESS, so you're encouraging applications to use that.

However, the exported header doesn't define it.

Is that intentional? Do you consider rnp_def.h a public head, and if yes, should it be installed by the "make install" target, too?

joke325:
Closes #897

### DIFF
--- a/include/rnp/rnp_err.h
+++ b/include/rnp/rnp_err.h
@@ -23,36 +23,54 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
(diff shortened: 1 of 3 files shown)


include/rnp/rnp_err.h
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (c) 2017, [Ribose Inc](https://www.ribose.com).
```
ni4: (c) 2017-2019

ronaldtse:
@dewyatt could you help review this? Thanks!

PR #970, opened 2019-11-22
Generating keys with multiple userids is unsupported.

### LINKED ISSUE #929: Generating a key with multiple userids silently ignores all but last userid

joke325:
Closes #929

### DIFF
--- a/src/rnp/fficli.cpp
+++ b/src/rnp/fficli.cpp
@@ -1467,9 +1467,9 @@ rnp_cfg_set_defkey(rnp_cfg_t *cfg)
 
     /* If a userid has been given, we'll use it. */
     userid = NULL;
-    list ids = NULL;
-    if (rnp_cfg_copylist_str(cfg, &ids, CFG_USERID)) {
-        userid = (char*)list_front(ids);
+    std::string uis = rnp_cfg_getlist_string(cfg, CFG_USERID, 0);
+    if (uis.length() > 0) {
+        userid = uis.c_str();
     }
     if (!userid) {
         /* also search in config file for default id */
@@ -1486,7 +1486,6 @@ rnp_cfg_set_defkey(rnp_cfg_t *cfg)
     } else {
         rnp_cfg_setstr(cfg, CFG_KR_DEF_KEY, userid);
     }
-    list_destroy(&ids);
 }
 
 bool
--- a/src/rnp/fficli.cpp
+++ b/src/rnp/fficli.cpp
@@ -1468,7 +1468,7 @@ rnp_cfg_set_defkey(rnp_cfg_t *cfg)
     /* If a userid has been given, we'll use it. */
     userid = NULL;
     std::string uis = rnp_cfg_getlist_string(cfg, CFG_USERID, 0);
-    if (uis.length() > 0) {
+    if (!uis.empty()) {
         userid = uis.c_str();
     }
     if (!userid) {
--- a/src/rnp/fficli.cpp
+++ b/src/rnp/fficli.cpp
@@ -1466,20 +1466,27 @@ rnp_cfg_set_defkey(rnp_cfg_t *cfg)
     }
 
     /* If a userid has been given, we'll use it. */
-    if (!(userid = rnp_cfg_getstr(cfg, CFG_USERID))) {
+    userid = NULL;
+    list ids = NULL;
+    if (rnp_cfg_copylist_str(cfg, &ids, CFG_USERID)) {
+        userid = (char*)list_front(ids);
+    }
+    if (!userid) {
         /* also search in config file for default id */
 
         if (defhomedir) {
             memset(id, 0, sizeof(id));
             conffile(homedir, id, sizeof(id));
             if (id[0] != 0x0) {
-                rnp_cfg_setstr(cfg, CFG_USERID, id);
+                rnp_cfg_unset(cfg, CFG_USERID);
+                rnp_cfg_addstr(cfg, CFG_USERID, id);
                 rnp_cfg_setstr(cfg, CFG_KR_DEF_KEY, id);
             }
         }
     } else {
         rnp_cfg_setstr(cfg, CFG_KR_DEF_KEY, userid);
     }
+    list_destroy(&ids);
 }
 
 bool
--- a/src/rnp/rnpcfg.cpp
+++ b/src/rnp/rnpcfg.cpp
@@ -375,15 +375,18 @@ rnp_cfg_getlist(rnp_cfg_t *cfg, const char *key)
 
 std::string rnp_cfg_getlist_string(const rnp_cfg_t *cfg, const std::string &key, size_t index) {
     list *lval = rnp_cfg_getlist(const_cast<rnp_cfg_t*>(cfg), key.c_str());
-    if (lval) {
-        if (index < list_length(*lval)) {
-            rnp_cfg_val_t *val = (rnp_cfg_val_t *)list_at(*lval, index);
-            return rnp_cfg_val_getstr(val);
-        } else {
-            RNP_LOG("wrong item index");
-        }
+    if (!lval) {
+        return "";
     }
-    return "";
+
+    if (!(index < list_length(*lval))) {
+        RNP_LOG("wrong item index");
+        return "";
+    }
+
+    rnp_cfg_val_t *val = (rnp_cfg_val_t*)list_at(*lval, index);
+    const char *sval = rnp_cfg_val_getstr(val);
+    return sval? sval : "";
 }
 
 bool
--- a/src/rnp/rnpcfg.cpp
+++ b/src/rnp/rnpcfg.cpp
@@ -247,10 +247,6 @@ rnp_cfg_setstr(rnp_cfg_t *cfg, const char *key, const char *val)
         return false;
     }
 
-    std::string mkey("\x01");
-    mkey.append(key);
-    rnp_cfg_addstr(cfg, mkey.c_str(), val? val : "");
-
     return true;
 }
 
@@ -288,26 +284,12 @@ rnp_cfg_addstr(rnp_cfg_t *cfg, const char *key, const char *str)
 }
 
 const char *
-rnp_cfg_getstr(const rnp_cfg_t *cfg, const char *key, const size_t item_idx)
+rnp_cfg_getstr(const rnp_cfg_t *cfg, const char *key)
 {
     rnp_cfg_item_t *it = rnp_cfg_find(cfg, key);
 
     if (it && (it->val.type == RNP_CFG_VAL_STRING)) {
-        if (item_idx == 0) {
-            return it->val.val._string;
-        } else {
-            std::string mkey("\x01");
-            mkey.append(key);
-            rnp_cfg_item_t *mit = rnp_cfg_find(cfg, mkey.c_str());
-            if (mit && (mit->val.type == RNP_CFG_VAL_LIST)) {
-                if (item_idx <= list_length(mit->val.val._list)) {
-                    rnp_cfg_val_t *val = (rnp_cfg_val_t*)list_at(mit->val.val._list, item_idx-1);
-                    if (val && (val->type == RNP_CFG_VAL_STRING)) {
-                        return val->val._string;
-                    }
-                }
-            }
-        }
+        return it->val.val._string;
     }
 
     return NULL;
--- a/src/rnp/rnpcfg.cpp
+++ b/src/rnp/rnpcfg.cpp
@@ -373,6 +373,19 @@ rnp_cfg_getlist(rnp_cfg_t *cfg, const char *key)
     return NULL;
 }
 
+std::string rnp_cfg_getlist_string(const rnp_cfg_t *cfg, const std::string &key, size_t index) {
+    list *lval = rnp_cfg_getlist(const_cast<rnp_cfg_t*>(cfg), key.c_str());
+    if (lval) {
+        if (index < list_length(*lval)) {
+            rnp_cfg_val_t *val = (rnp_cfg_val_t *)list_at(*lval, index);
+            return rnp_cfg_val_getstr(val);
+        } else {
+            RNP_LOG("wrong item index");
+        }
+    }
+    return "";
+}
+
 bool
 rnp_cfg_copylist_str(const rnp_cfg_t *cfg, list *dst, const char *key)
 {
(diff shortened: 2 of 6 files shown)


src/rnp/rnpcfg.cpp
```diff
@@ -247,6 +247,10 @@ rnp_cfg_setstr(rnp_cfg_t *cfg, const char *key, const char *val)
         return false;
     }
 
+    std::string mkey("\x01");
+    mkey.append(key);
+    rnp_cfg_addstr(cfg, mkey.c_str(), val? val : "");
+
```
ni4: You don't need to modify function `rnp_cfg_setstr()` as required functionality is already implemented in `rnp_cfg_addstr()`. Please see how handled `OPT_USERID` and `CFG_SIGNERS` in `rnp.cpp`.

src/rnp/rnpcfg.cpp
```diff
@@ -284,12 +288,26 @@ rnp_cfg_addstr(rnp_cfg_t *cfg, const char *key, const char *str)
...
 
 const char *
-rnp_cfg_getstr(const rnp_cfg_t *cfg, const char *key)
+rnp_cfg_getstr(const rnp_cfg_t *cfg, const char *key, const size_t item_idx)
 {
     rnp_cfg_item_t *it = rnp_cfg_find(cfg, key);
 
     if (it && (it->val.type == RNP_CFG_VAL_STRING)) {
-        return it->val.val._string;
+        if (item_idx == 0) {
```
ni4: The same as previous comment - you don't need to modify this function since all required functionality is already implemented.

src/rnpkeys/rnpkeys.cpp
```diff
@@ -254,6 +254,10 @@ rnp_cmd(rnp_cfg_t *cfg, cli_rnp_t *rnp, optdefs_t cmd, const char *f)
     case CMD_GENERATE_KEY: {
         if (f == NULL) {
             f = rnp_cfg_getstr(cfg, CFG_USERID);
+            if(rnp_cfg_getstr(cfg, CFG_USERID, 2) != NULL) {
```
ni4: You should use `rnp_cfg_getlist()` here, to quickly check it's length. Also here or later you may use more C++-ish wrapper `rnp_cfg_copylist_string()`.

src/rnpkeys/rnpkeys.cpp
```diff
@@ -254,6 +254,10 @@ rnp_cmd(rnp_cfg_t *cfg, cli_rnp_t *rnp, optdefs_t cmd, const char *f)
...
+                fprintf(stderr, "only userid is supported for generated keys\n");
```
ni4: The idea of this PR was to update code to be able to generate keys with multiple user id's, not just report an error. While key generation function supports single key id at the moment, you still may use `rnp_key_add_uid()` on the generated key. Also it worth adding `rnp_op_generate_add_uid()` method, so user id would be added in a single pass.

src/tests/generatekey.cpp
```diff
@@ -650,11 +650,14 @@ TEST_F(rnp_tests, rnpkeys_generatekey_testExpertMode)
...
 
-    assert_true(rnp_cfg_setbool(&ops, CFG_EXPERT, true));
-    assert_true(rnp_cfg_setint(&ops, CFG_S2K_ITER, 1));
+    assert_true(rnp_cfg_setbool(&tops, CFG_EXPERT, true));
+    assert_true(rnp_cfg_setint(&tops, CFG_S2K_ITER, 1));
 
     /* ecdsa/ecdh p256 keypair */
+    rnp_cfg_free(&ops);
+    rnp_cfg_copy(&ops, &tops);
     assert_true(rnp_cfg_setstr(&ops, CFG_USERID, "expert_ecdsa_p256"));
```
ni4: It would be better instead of modifying the whole test workflow just inset `rnp_cfg_unset()` calls. However, anyway, you should not modify inner code of `rnp_cfg_setstr()` and use `rnp_cfg_addstr()` or something similar.

joke325:
@ni4 There (the issue @929) is stated the Expected Behavior:
"Preferably generate a key with two userids, failing that, signal an error that giving multiple --userid arguments is not supported."

src/rnp/fficli.cpp
```diff
@@ -1466,20 +1466,27 @@ rnp_cfg_set_defkey(rnp_cfg_t *cfg)
     }
 
     /* If a userid has been given, we'll use it. */
-    if (!(userid = rnp_cfg_getstr(cfg, CFG_USERID))) {
+    userid = NULL;
+    list ids = NULL;
+    if (rnp_cfg_copylist_str(cfg, &ids, CFG_USERID)) {
+        userid = (char*)list_front(ids);
+    }
```
ni4: Instead of creating temporary list and copying data to it (and duplicate code below), it would be better to add helper function like `const std::string &rnp_cfg_getlist_string(const rnp_cfg_t *cfg, const std::string &key, size_t index)`. Inside you may use `rnp_cfg_getlist()` to check list's length and get value via `rnp_cfg_val_getstr()`.

src/rnpkeys/rnpkeys.cpp
```diff
@@ -230,15 +230,27 @@ bool
...
 
     switch (cmd) {
     case CMD_LIST_KEYS:
         if (!f) {
-            f = rnp_cfg_getstr(cfg, CFG_USERID);
+            list ids = NULL;
+            if (rnp_cfg_copylist_str(cfg, &ids, CFG_USERID) && list_length(ids) > 0) {
+                f = (fs = (char*)list_front(ids)).c_str();
+            }
+            list_destroy(&ids);
```
ni4: As the previous - no need to duplicate list, while you can get the element via some helper function.

src/rnpkeys/rnpkeys.cpp
```diff
@@ -230,15 +230,27 @@ bool
...
         return print_keys_info(cfg, rnp, stdout, f);
     case CMD_EXPORT_KEY: {
-        key = f ? f : rnp_cfg_getstr(cfg, CFG_USERID);
+        key = f;
+        if (!key) {
+            list ids = NULL;
+            if (rnp_cfg_copylist_str(cfg, &ids, CFG_USERID) && list_length(ids) > 0) {
+                key = (fs = (char*)list_front(ids)).c_str();
+            }
+            list_destroy(&ids);
```
ni4: As the previous.

src/rnpkeys/rnpkeys.cpp
```diff
@@ -253,7 +265,17 @@ rnp_cmd(rnp_cfg_t *cfg, cli_rnp_t *rnp, optdefs_t cmd, const char *f)
...
-            f = rnp_cfg_getstr(cfg, CFG_USERID);
+            list ids = NULL;
+            if (rnp_cfg_copylist_str(cfg, &ids, CFG_USERID) && list_length(ids) > 0) {
+                if (list_length(ids) == 1) {
+                    f = (fs = (char*)list_front(ids)).c_str();
+                } else {
+                    fprintf(stderr, "only userid is supported for generated keys\n");
+                    list_destroy(&ids);
+                    return false;
+                }
```
ni4: As the previous - helper function + `list_length(rnp_cfg_get_list())` to check list's length.

src/rnp/fficli.cpp
joke325: The rnp_cfg_val_getstr() doesn't support the RNP_CFG_VAL_LIST type.

ni4: @joke325 yeah, I mean something like `rnp_cfg_val_getstr((rnp_cfg_val_t*) list_front(rnp_cfg_getlist(cfg, name)))`

src/rnp/rnpcfg.cpp
```diff
@@ -373,6 +373,19 @@ rnp_cfg_getlist(rnp_cfg_t *cfg, const char *key)
     return NULL;
 }
 
+std::string rnp_cfg_getlist_string(const rnp_cfg_t *cfg, const std::string &key, size_t index) {
+    list *lval = rnp_cfg_getlist(const_cast<rnp_cfg_t*>(cfg), key.c_str());
+    if (lval) {
```
ni4: Here would be better to change to `if (!lval) {return ""}`, to have less nested code.

src/rnp/rnpcfg.cpp
```diff
@@ -373,6 +373,19 @@ rnp_cfg_getlist(rnp_cfg_t *cfg, const char *key)
...
+        if (index < list_length(*lval)) {
+            rnp_cfg_val_t *val = (rnp_cfg_val_t *)list_at(*lval, index);
+            return rnp_cfg_val_getstr(val);
```
ni4: `rnp_cfg_val_getstr()` may return NULL, which will raise an exception in std::string() constructor. Please check for NULL and return empty string in that case.

src/rnp/rnpcfg.cpp
```diff
@@ -373,6 +373,19 @@ rnp_cfg_getlist(rnp_cfg_t *cfg, const char *key)
     return NULL;
 }
 
+std::string rnp_cfg_getlist_string(const rnp_cfg_t *cfg, const std::string &key, size_t index) {
+    list *lval = rnp_cfg_getlist(const_cast<rnp_cfg_t*>(cfg), key.c_str());
+    if (lval) {
+        if (index < list_length(*lval)) {
```
ni4: Again, better return if index has wrong value, then nest further code.

src/rnpkeys/rnpkeys.cpp
```diff
@@ -253,7 +263,15 @@ rnp_cmd(rnp_cfg_t *cfg, cli_rnp_t *rnp, optdefs_t cmd, const char *f)
         return import_keys(cfg, rnp, f);
     case CMD_GENERATE_KEY: {
         if (f == NULL) {
-            f = rnp_cfg_getstr(cfg, CFG_USERID);
+            list *ids = NULL;
+            if ((ids = rnp_cfg_getlist(cfg, CFG_USERID)) && list_length(*ids) > 0) {
+                if (list_length(*ids) == 1) {
+                    f = (fs = rnp_cfg_getlist_string(cfg, CFG_USERID, 0)).c_str();
+                }  else {
+                    fprintf(stderr, "only userid is supported for generated keys\n");
```
ni4: I guess it should be something like 'Only single userid' or 'Only one userid'.

src/tests/generatekey.cpp
```diff
@@ -821,7 +832,7 @@ TEST_F(rnp_tests, generatekeyECDSA_explicitlySetBiggerThanNeededDigest_ShouldSuc
     assert_true(rnp_cfg_setstr(&ops, CFG_HASH, "SHA512"));
     assert_true(rnp_cfg_setint(&ops, CFG_S2K_ITER, 1));
 
-    assert_true(rnp_cfg_setstr(&ops, CFG_USERID, "expert_large_digest"));
+    assert_true(rnp_cfg_addstr(&ops, CFG_USERID, "expert_large_digest"));
```
ni4: We should also have negative test case - where two user ids are specified and key generation fails.

joke325:
@ni4 Could you please look at the failure in Cirrus CI / build. I don't understand how my last commit causes the error in the cli_tests.py / Misc::test_encryption_s2k test. I cannot reproduce the failure on my machine.

ni4:
@joke325 Looks like some problem on GnuPG side. I filed issue #984 to take a look later, and restarted test - let's see whether it helps.

src/rnpkeys/rnpkeys.cpp
```diff
@@ -230,15 +230,25 @@ bool
...
     const char *key;
+    std::string fs;
 
     switch (cmd) {
     case CMD_LIST_KEYS:
         if (!f) {
-            f = rnp_cfg_getstr(cfg, CFG_USERID);
+            list *ids = NULL;
+            if ((ids = rnp_cfg_getlist(cfg, CFG_USERID)) && list_length(*ids) > 0) {
+                f = (fs = rnp_cfg_getlist_string(cfg, CFG_USERID, 0)).c_str();
```
ni4: I missed it during the previous review - if you don't use fs later on then it would be enough to write `f = rnp_cfg_getlist_string(cfg, CFG_USERID, 0).c_str()`.

src/tests/generatekey.cpp
```diff
@@ -812,6 +823,22 @@ TEST_F(rnp_tests, generatekeyECDSA_explicitlySetSmallOutputDigest_DigestAlgAdjus
     rnp_cfg_free(&ops);
 }
 
+TEST_F(rnp_tests, generatekey_multipleUserIds_ShouldFail)
```
ni4: Please also add a line with test's declaration to rnp_tests.h.
While technically it is not needed it is sometimes handy to have all test names in one header. Or, if not - remove it completely in some separate PR.

src/rnpkeys/rnpkeys.cpp
joke325: I believe the fs used to keep values of strings, pointers from c_str() won't work without the fs.

ni4: Oh, agreed. Missed this side, sorry.

PR #971, opened 2019-11-24
gpg 2.1.15 complains about kbx format

### LINKED ISSUE #405: gpg 2.1.15 complains about our KBX format
## Description

This should be looked into at some point: https://github.com/riboseinc/rnp/pull/368#issuecomment-323347085

```
$ rm -rf ~/.rnp
$ src/rnpkeys/rnpkeys --gen-key --keystore-format=GPG21
$ docker run -ti --rm -v $HOME/.rnp:/root/.gnupg ubuntu:zesty gpg --list-keys
gpg: WARNING: unsafe ownership on homedir '/root/.gnupg'
gpg: parse_keyblock_image: more signatures than found in the meta data
gpg: keydb_get_keyblock failed: Invalid keyring
```

Newer gpg doesn't complain, but that may be because it's more lenient or something, so it should be more completely investigated.

rrrooommmaaa:
[NSIGS] "Number of signatures" field should match the total number of signatures in the blob including subkey signatures.

Though, this check was removed on Mar 30, 2017 update of GPG https://dev.gnupg.org/rGa6142dbdbc5783043deb847dc64998c421860941

Closes #405

### DIFF
--- a/src/librekey/key_store_kbx.cpp
+++ b/src/librekey/key_store_kbx.cpp
@@ -656,8 +656,7 @@ rnp_key_store_kbx_write_pgp(rnp_key_store_t *key_store, pgp_key_t *key, pgp_dest
 
     /* write keyblock and fix the offset/length */
     key_start = memdst.writeb;
-    /* absolute offset in output stream */
-    pt = key_start + dst->writeb;
+    pt = key_start;
     p = (uint8_t *) mem_dest_get_memory(&memdst) + 8;
     STORE32BE(p, pt);
 
--- a/src/librekey/key_store_kbx.cpp
+++ b/src/librekey/key_store_kbx.cpp
@@ -515,6 +515,8 @@ rnp_key_store_kbx_write_pgp(rnp_key_store_t *key_store, pgp_key_t *key, pgp_dest
     uint32_t   pt;
     pgp_hash_t hash = {0};
     bool       result = false;
+    list       subkey_sig_expirations = NULL; // expirations (uint32_t) of subkey signatures
+    uint32_t   expiration = 0;
 
     if (init_mem_dest(&memdst, NULL, BLOB_SIZE_LIMIT)) {
         RNP_LOG("alloc failed");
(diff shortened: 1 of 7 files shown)


rrrooommmaaa:
Should I add a unit test for this?

src/librekey/key_store_kbx.cpp
```diff
@@ -562,6 +564,11 @@ rnp_key_store_kbx_write_pgp(rnp_key_store_t *key_store, pgp_key_t *key, pgp_dest
             !pu16(&memdst, 0)) {                 // RFU
             goto finish;
         }
+        // load signature expirations while we're at it
+        for (i = 0; i < pgp_key_get_subsig_count(subkey); i++) {
+            expiration = signature_get_key_expiration(&pgp_key_get_subsig(subkey, i)->sig);
+            list_append(&subkey_sig_expirations, &expiration, sizeof(expiration));
```
ni4: Please check return result here since allocation failure may happen.

src/librekey/key_store_kbx.cpp
```diff
@@ -602,6 +607,12 @@ rnp_key_store_kbx_write_pgp(rnp_key_store_t *key_store, pgp_key_t *key, pgp_dest
             goto finish;
         }
     }
+    for (list_item *expiration_entry = list_front(subkey_sig_expirations); expiration_entry; expiration_entry = list_next(expiration_entry)) {
```
ni4: Probably this will not fit line length limitation - please check with clang-format (or use shorter variable name - personally I prefer this way since one-liner code looks better).

ni4:
@rrrooommmaaa
> Should I add a unit test for this?

Yeah, we usually add tests for all changes to make sure that later regression will not happen (and extend coverage).

rrrooommmaaa:
> 
> Yeah, we usually add tests for all changes to make sure that later regression will not happen (and extend coverage).

Done the test. And it turned out that writing keybox to memory output produced incorrect blob due to a bug. Fixed.

rrrooommmaaa:
@ni4 Is it ok now? What is codecov complaining about?

ni4:
@rrrooommmaaa It happens sometimes with codecov, it should be good to go. Let's just wait for the second approval.

ronaldtse:
Second approval!

PR #973, opened 2019-11-28
Defines unified names of algorithms

### LINKED ISSUE #852: Duplicate and inconsistent string maps between FFI and JSON dumping
## Description

We have defined multiple mappings from ints to strings for various types and this may present an inconsistent interface to the user.

For example:
https://github.com/riboseinc/rnp/blob/d562b479b12d930111025500d8c63ac442c1c4f3/src/librepgp/stream-dump.cpp#L154
https://github.com/riboseinc/rnp/blob/d562b479b12d930111025500d8c63ac442c1c4f3/src/lib/rnp.cpp#L309

There are differences in capitalization and in hyphenation. In the FFI this means that your inputs (say, "AES256") may not match the output from packet parsing to JSON ("AES-256").
This is confusing and seems like duplication we ought to be able to eliminate.

joke325:
Closes #852

### DIFF
--- a/src/tests/ffi.cpp
+++ b/src/tests/ffi.cpp
@@ -1390,7 +1390,7 @@ TEST_F(rnp_tests, test_ffi_key_generate_dsa)
     assert_false(boolres);
     /* algorithm */
     assert_rnp_success(rnp_key_get_alg(subkey, &alg));
-    assert_int_equal(strcasecmp(alg, "ElGamal"), 0);
+    assert_int_equal(strcasecmp(alg, "ELGAMAL"), 0);
     rnp_buffer_destroy(alg);
     /* key bits */
     assert_rnp_success(rnp_key_get_bits(subkey, &bits));
@@ -1735,10 +1735,10 @@ TEST_F(rnp_tests, test_ffi_key_generate_ex)
     /* preferred ciphers */
     assert_rnp_success(rnp_op_generate_clear_pref_ciphers(keygen));
     assert_rnp_failure(rnp_op_generate_add_pref_cipher(keygen, "unknown"));
-    assert_rnp_success(rnp_op_generate_add_pref_cipher(keygen, "blowfish"));
+    assert_rnp_success(rnp_op_generate_add_pref_cipher(keygen, "BLOWFISH"));
     assert_rnp_success(rnp_op_generate_clear_pref_ciphers(keygen));
-    assert_rnp_success(rnp_op_generate_add_pref_cipher(keygen, "camellia256"));
-    assert_rnp_success(rnp_op_generate_add_pref_cipher(keygen, "aes256"));
+    assert_rnp_success(rnp_op_generate_add_pref_cipher(keygen, "CAMELLIA256"));
+    assert_rnp_success(rnp_op_generate_add_pref_cipher(keygen, "AES256"));
     /* preferred compression algorithms */
     assert_rnp_success(rnp_op_generate_clear_pref_compression(keygen));
     assert_rnp_failure(rnp_op_generate_add_pref_compression(keygen, "unknown"));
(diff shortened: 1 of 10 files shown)


src/librepgp/packet-show.cpp
```diff
@@ -129,13 +129,13 @@ static pgp_map_t pubkey_alg_map[] = {
...
   {PGP_PKA_ELGAMAL, "Elgamal (Encrypt-Only)"},
-  {PGP_PKA_DSA, "DSA"},
-  {PGP_PKA_ECDH, "ECDH"},
-  {PGP_PKA_ECDSA, "ECDSA"},
+  {PGP_PKA_DSA, PGP_ALGNAME_DSA},
+  {PGP_PKA_ECDH, PGP_ALGNAME_ECDH},
+  {PGP_PKA_ECDSA, PGP_ALGNAME_ECDSA},
   {PGP_PKA_ELGAMAL_ENCRYPT_OR_SIGN, "Reserved (formerly Elgamal Encrypt or Sign"},
   {PGP_PKA_RESERVED_DH, "Reserved for Diffie-Hellman (X9.42)"},
-  {PGP_PKA_EDDSA, "EdDSA"},
```
ronaldtse: Looks good except there is a side effect that `EdDSA` is changed to `EDDSA`. @dewyatt @ni4 is this acceptable?

ni4: @ronaldtse Somehow missed your comment before writing my own :) I also don't think that we should change it this way for human-readable output, since common spelling is EdDSA, as well as AES-128, SHA-256.

ronaldtse:
> Generally it is good, however I have doubts in two places:
> 
> * packet dumping code: it's goal is to provide human-readable information, including algorithm names, and human-readable would be AES-128, SHA-256, EdDSA, RSA (Encrypt only) instead of AES128, SHA256, EDDSA, RSA (without notification whether it is Encrypt and Sign or Encrypt only).

Agree. Human-readable vs in code.

> * tests (some of them): those should check whether code works well with different case (aes128, rsa).

Agree that tests should test against variant inputs.

ronaldtse:
@joke325 for the algorithm names, I think @ni4 wanted to separate the "human-readable" forms from coded forms for display output:

e.g. AES-128, SHA-256, EdDSA, RSA (Encrypt only) vs `AES128`, `SHA256`, `EDDSA`, `RSA`

joke325:
@ronaldtse I've pulled down the proposed macros from from the "human-readable" strings, so they are separated now.

src/tests/ffi.cpp
```diff
@@ -2084,9 +2084,391 @@ TEST_F(rnp_tests, test_ffi_key_generate_ex)
...
+    assert_rnp_success(rnp_op_generate_get_key(keygen, &subkey));
+    assert_non_null(subkey);
+    assert_rnp_success(rnp_op_generate_destroy(keygen));
+    assert_rnp_success(rnp_key_handle_destroy(subkey));
+
+    assert_rnp_success(rnp_key_handle_destroy(key));
+    assert_rnp_success(rnp_ffi_destroy(ffi));
+}
+
+TEST_F(rnp_tests, test_ffi_key_generate_algnamecase)
```
ni4: While purpose of this test is to check whether code works well with misc cases it is not needed to do additional calls - rnp_op_generate_set_bits(), _add_usage(), etc.
Also it would be much more clear to not use pgp_algname_rsa and so on but use direct constants - "Rsa", "aes128", etc. So if something failed then you'll easy what was the cause.

src/tests/ffi.cpp
```diff
@@ -2106,7 +2488,7 @@ TEST_F(rnp_tests, test_ffi_key_generate_protection)
 
     /* Generate key and subkey without protection */
     rnp_op_generate_t keygen = NULL;
-    assert_rnp_success(rnp_op_generate_create(&keygen, ffi, "RSA"));
+    assert_rnp_success(rnp_op_generate_create(&keygen, ffi, PGP_ALGNAME_RSA));
```
ni4: I would suggest to leave string constants instead define usage so we'll have more coverage on misc case usage.

src/tests/rnp_tests.cpp
```diff
@@ -36,6 +36,41 @@ static char original_dir[PATH_MAX];
  */
 rng_t global_rng;
 
+/*
```
ni4: It is not a good idea to fill main test-runner file with those constants just to check some minor cases. As suggested before it would be much better to remove these string and just use constants directly in code with different spelling - "Rsa", "rsa", etc.

ronaldtse:
Looks like it's all good @joke325? If so, please feel free to merge!

joke325:
@ronaldtse (1) There is 1 pending reviewer. (2) The Merge PR option is unavailable to me, do I have rights for merging?

ni4:
@joke325 we have a rule of 2 approvals before merging. So merging this.

PR #976, opened 2019-11-28
If $LOGNAME is unavailable then $USER is used instead

### LINKED ISSUE #958: Variable LOGNAME may be not set under MSYS/MinGW
## Description

Probably it's installation-dependent, but at least in the case of MSYS2 + MinGW installation `$LOGNAME` is not set. This causes the generation of key `(null)@localhost` if userid was not specified.
So we should check it for availability and have fallback to `$USER`.

joke325:
Closes #958

### DIFF
--- a/src/lib/utils.h
+++ b/src/lib/utils.h
@@ -180,4 +180,14 @@ STORE64BE(uint8_t x[8], uint64_t y)
 #define RNP_USE_STD_REGEX 1
 #endif
 
+inline char *
+getenv_logname(void)
+{
+    char *name = getenv("LOGNAME");
+    if (!name) {
+        name = getenv("USER");
+    }
+    return name;
+}
+
 #endif
--- a/src/lib/generate-key.cpp
+++ b/src/lib/generate-key.cpp
@@ -297,7 +297,7 @@ keygen_primary_merge_defaults(rnp_keygen_primary_desc_t *desc)
                  "%s %d-bit key <%s@localhost>",
                  pgp_show_pka(desc->crypto.key_alg),
                  get_numbits(&desc->crypto),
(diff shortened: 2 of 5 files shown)


src/lib/utils.h
```diff
@@ -180,4 +180,14 @@ STORE64BE(uint8_t x[8], uint64_t y)
 #define RNP_USE_STD_REGEX 1
 #endif
 
+inline char *
+GETENV_LOGNAME(void)
```
ni4: Since it is not define it should be named in lowercase, i.e. getenv_logname().

src/lib/utils.h
```diff
@@ -180,4 +180,14 @@ STORE64BE(uint8_t x[8], uint64_t y)
...
+{
+	char *name = getenv("LOGNAME");
+    if (!name) {
+		name = getenv("USER");
+	}
```
ni4: There is some misalign here, or mixed space/tabs. Could you please apply clang-format so it is correctly spaced?

src/tests/rnp_tests.cpp
```diff
@@ -41,7 +41,7 @@ rnp_tests::rnp_tests() : m_dir(make_temp_dir())
     /* We use LOGNAME in a few places within the tests
      * and it isn't always set in every environment.
      */
-    if (!getenv("LOGNAME")) {
+    if (!getenv_logname()) {
```
ni4: This would slightly change previous behavior - since getenv_logname() may return non-NULL when `$USER` is set and $LOGNAME is unset. Most likely this will have no impact (and actually is on tests only) but it would be better to change it back.

ni4:
@joke325 And please squash so we'd have single commit instead of duplicate names.

PR #978, opened 2019-11-29
Split GitHub Actions into different platforms

ronaldtse:
And add badges for them in README.


ronaldtse:
Merging since this is purely about CI.

PR #990, opened 2019-12-03
Add missing -c, --symmetric command to rnp help message.

ni4:
As title says.


ronaldtse:
Thanks @jpo234 for discovering this and @ni4 for fixing it so fast 👍

PR #998, opened 2019-12-07
Fixes GCC compile warnings.

### LINKED ISSUE #979: Check and fix GCC compile warnings.
## Description

We have some GCC warnings during compilation, most of them because of GCC 8.0+ `-Wstringop-truncation`, as well as `-Wformat` (under windows) and some others.
Those should be fixed.

joke325:
Closes #979

### DIFF
--- a/src/librepgp/stream-armor.cpp
+++ b/src/librepgp/stream-armor.cpp
@@ -29,6 +29,7 @@
 #include <stdio.h>
 #include <unistd.h>
 #include <string.h>
+#include <algorithm>
 #include <rnp/rnp_def.h>
 #include "stream-def.h"
 #include "stream-armor.h"
@@ -193,10 +194,15 @@ armor_read_trailer(pgp_source_t *src)
     }
 
     stlen = strlen(param->armorhdr);
-    strncpy(st, ST_ARMOR_END, 8); /* 8 here is mandatory */
-    strncpy(st + 8, param->armorhdr + 5, stlen - 5);
-    strncpy(st + stlen + 3, ST_DASHES, 5);
-    stlen += 8;
+    if (stlen + 8 + 1 <= sizeof(st)) {
+        memcpy(st, ST_ARMOR_END, 8); /* 8 here is mandatory */
+        memcpy(st + 8, param->armorhdr + 5, stlen - 5);
+        memcpy(st + stlen + 3, ST_DASHES, 5);
+        stlen += 8;
+    } else {
+        RNP_LOG("Internal error");
+        return false;
+    }
     read = src_peek(param->readsrc, str, stlen);
     if ((read < (ssize_t) stlen) || strncmp(str, st, stlen)) {
         return false;
@@ -680,7 +686,7 @@ armor_message_header(pgp_armored_msg_t type, bool finish, char *buf)
 {
     const char *str;
     str = finish ? ST_ARMOR_END : ST_ARMOR_BEGIN;
-    strncpy(buf, str, strlen(str));
+    memcpy(buf, str, strlen(str));
     buf += strlen(str);
     switch (type) {
     case PGP_ARMORED_MESSAGE:
@@ -702,9 +708,9 @@ armor_message_header(pgp_armored_msg_t type, bool finish, char *buf)
         return false;
     }
 
-    strncpy(buf, str, strlen(str));
+    memcpy(buf, str, strlen(str));
     buf += strlen(str);
-    strncpy(buf, ST_DASHES, 5);
+    memcpy(buf, ST_DASHES, std::min<size_t>(sizeof(ST_DASHES) - 1, 5));
     buf[5] = '\0';
     return true;
 }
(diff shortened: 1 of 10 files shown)


src/lib/crypto.cpp
```diff
@@ -102,6 +102,20 @@ pgp_generate_seckey(const rnp_keygen_crypto_params_t *crypto,
...
+        }
+        if (crypto->ecc.curve == PGP_CURVE_25519) {
+            if (x25519_generate(rng, &seckey->material.ec)) {
+                RNP_LOG("failed to generate x25519 key");
+                goto end;
+            }
+            seckey->material.ec.curve = crypto->ecc.curve;
+            goto after_seckey_alg;
+        }
+    }
```
ni4: No need to move this outside from `switch()`. This makes code less readable and more complicated.
It is enough to add /* FALLTHROUGH */ or similar comment, please see the details here: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough

src/lib/crypto.cpp
```diff
@@ -158,6 +159,7 @@ pgp_generate_seckey(const rnp_keygen_crypto_params_t *crypto,
         goto end;
         break;
     }
+after_seckey_alg:
```
ni4: See the previous - it would be enough to add comment instead of code modifications.

src/librepgp/stream-armor.cpp
```diff
@@ -193,9 +194,9 @@ armor_read_trailer(pgp_source_t *src)
     }
 
     stlen = strlen(param->armorhdr);
-    strncpy(st, ST_ARMOR_END, 8); /* 8 here is mandatory */
+    memcpy(st, ST_ARMOR_END, std::min<size_t>(sizeof(ST_ARMOR_END)-1, 8)); /* 8 here is mandatory */
```
ni4: I think instead of std::min usage it would more readable/safe to check whether stlen + 8 + 1 less or equal to sizeof(st), and then use memcpy().

src/librepgp/stream-common.cpp
```diff
@@ -892,7 +892,8 @@ init_tmpfile_dest(pgp_dest_t *dst, const char *path, bool overwrite)
         RNP_LOG("failed to build file path");
         return RNP_ERROR_BAD_PARAMETERS;
     }
-    mktemp(tmp);
+    close(mkstemp(tmp));
+    unlink(tmp);
```
ni4: Why did you need to introduce these changes? This lines doesn't seem to produce any warning (GCC 9.2.0).

src/librepgp/stream-ctx.h
```diff
@@ -121,6 +121,12 @@ void             rnp_ctx_reset(rnp_ctx_t *);
 void             rnp_ctx_free(rnp_ctx_t *);
 struct rng_st_t *rnp_ctx_rng_handle(const rnp_ctx_t *ctx);
 
+rnp_result_t rnp_encrypt_set_pass_info(rnp_symmetric_pass_info_t *info,
+                                       const char *               password,
+                                       pgp_hash_alg_t             hash_alg,
+                                       size_t                     iterations,
+                                       pgp_symm_alg_t             s2k_cipher);
+
```
ni4: This function should be just removed, looks like was used previously and superseded by `rnp_ctx_add_encryption_password()`.

src/librepgp/stream-packet.cpp
```diff
@@ -1616,6 +1616,9 @@ stream_parse_signature_body(pgp_packet_body_t *pkt, pgp_signature_t *sig)
     }
 
     /* signature MPIs */
+    if (sig->palg == PGP_PKA_EDDSA && sig->version < PGP_V4) {
+        RNP_LOG("Warning! v3 EdDSA signature.");
+    }
```
ni4: The same as previous - no need to change code while it is enough to add a comment.

src/librepgp/stream-armor.cpp
```diff
@@ -193,10 +194,12 @@ armor_read_trailer(pgp_source_t *src)
     }
 
     stlen = strlen(param->armorhdr);
-    strncpy(st, ST_ARMOR_END, 8); /* 8 here is mandatory */
-    strncpy(st + 8, param->armorhdr + 5, stlen - 5);
-    strncpy(st + stlen + 3, ST_DASHES, 5);
-    stlen += 8;
+    if (stlen+8+1 <= sizeof(st)) {
```
ni4: 1. Missing spaces around '+' (please do not forget to run clang-format so code looks consistent)
2. Actually if stlen + 8 + 1 is larger than sizeof(st) then there is some problem with code so error message should be logged.

src/librepgp/stream-armor.cpp
```diff
@@ -702,9 +705,9 @@ armor_message_header(pgp_armored_msg_t type, bool finish, char *buf)
         return false;
     }
 
-    strncpy(buf, str, strlen(str));
+    memcpy(buf, str, strlen(str));
     buf += strlen(str);
-    strncpy(buf, ST_DASHES, 5);
+    memcpy(buf, ST_DASHES, std::min<size_t>(sizeof(ST_DASHES)-1, 5));
```
ni4: Clang-format, again.

ronaldtse:
Thanks @joke325 !

PR #1000, opened 2019-12-10
s2k iterations tests, documentation and packet dump

### LINKED ISSUE #982: rnp_op_encrypt_add_password ignores iterations argument
## Description

When specifying the number of iterations for the key derivation function the argument seems to be ignored

## Steps to Reproduce

1. specify the number of iterations to be used in rnp_op_encrypt_add_password: 
```
    if (rnp_op_encrypt_add_password(encrypt, password, "SHA256", 255, "AES256") != 0) {
        fprintf(stdout, "failed to set password\n");
        rc = 1;
        goto err_out_encrypt;
    }

```
2. encrypt some data to a file
3. verify the content of the encrypted file:
```
rnp --list-packets ciphertext.rnp
:off 0: packet header 0xc30d (tag 3, len 13)
Symmetric-key encrypted session key packet
    version: 4
    symmetric algorithm: 9 (AES-256)
    s2k specifier: 3
    s2k hash algorithm: 8 (SHA256)
    s2k salt: 0x2cd1000e45aa83ae
    s2k iterations: 0 (1024)
    encrypted key: 0x (0 bytes)
:off 15: packet header 0xd254 (tag 18, len 84)
Symmetrically-encrypted integrity protected data packet
```
## Expected Behavior

s2k iterations should match the argument

## Actual Behavior

s2k iterations always seems to be 0

[Full source](https://github.com/rnpgp/rnp/files/3911005/enc_main.zip)

### LINKED ISSUE #999: Update s2k iteration dumping and add tests for s2k iteration count check.
## Description

We need to be sure that correct iteration count is actually used when user calls `rnp_op_encrypt_add_password()`.
Check could be done via packet dumping and checking of output.

Also currently output is a bit misleading (see https://github.com/rnpgp/rnp/issues/982#issuecomment-562953955) so it should be corrected as well as `rnp_op_encrypt_ad_password()` function's documentation.

For the full discussion see #982

rrrooommmaaa:
- Added tests to ensure that correct iteration count is actually used when user calls rnp_op_encrypt_add_password(). Check is done via packet dumping to JSON and checking of output.
- Added tests for s2k iterations encode/decode functions
- Corrected packet dump output
- Corrected `rnp_op_encrypt_ad_password()` function's documentation.
Closes #982 
Closes #999

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -1807,10 +1807,13 @@ rnp_result_t rnp_op_encrypt_set_expiration_time(rnp_op_encrypt_t op, uint32_t ex
  *                 via password provider.
  * @param s2k_hash hash algorithm, used in key-from-password derivation. Pass NULL for default
  *        value. See rnp_op_encrypt_set_hash for possible values.
- * @param iterations number of iterations, used in key derivation function. Pass 0 for default
- *        value. Only 256 distinct values within the range [1024..0x3e00000] can be encoded.
- *        Thus, the number will be increased to the closest encodable value.
- *        In case it exceeds the maximum encodable value, it will be decreased to the maximum encodable value.
+ * @param iterations number of iterations, used in key derivation function.
+ *        According to RFC 4880, chapter 3.7.1.3, only 256 distinct values within the range
+ *        [1024..0x3e00000] can be encoded. Thus, the number will be increased to the closest
+ *        encodable value. In case it exceeds the maximum encodable value, it will be decreased
+ *        to the maximum encodable value.
+ *        If 0 is passed, an optimal number (greater or equal to 1024) will be calculated based
+ *        on performance measurement.
  * @param s2k_cipher symmetric cipher, used for key encryption. Pass NULL for default value.
  * See rnp_op_encrypt_set_cipher for possible values.
  * @return RNP_SUCCESS or error code if failed
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -1807,8 +1807,10 @@ rnp_result_t rnp_op_encrypt_set_expiration_time(rnp_op_encrypt_t op, uint32_t ex
  *                 via password provider.
  * @param s2k_hash hash algorithm, used in key-from-password derivation. Pass NULL for default
  *        value. See rnp_op_encrypt_set_hash for possible values.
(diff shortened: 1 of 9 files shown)


include/rnp/rnp.h
```diff
@@ -1807,8 +1807,10 @@ rnp_result_t rnp_op_encrypt_set_expiration_time(rnp_op_encrypt_t op, uint32_t ex
  *                 via password provider.
  * @param s2k_hash hash algorithm, used in key-from-password derivation. Pass NULL for default
  *        value. See rnp_op_encrypt_set_hash for possible values.
- * @param iterations number of iterations, used iin key derivation function. Pass 0 for default
- *        value.
+ * @param iterations number of iterations, used in key derivation function. Pass 0 for default
+ *        value. Only 256 distinct values within the range [1024..0x3e00000] can be encoded.
+ *        Thus, the number will be increased to the closest encodable value.
+ *        In case it exceeds the maximum encodable value, it will be decreased to the maximum encodable value.
```
ni4: I would add "See RFC 4880, chapter 3.7.1.3 for the encoding details".

src/tests/cipher.cpp
```diff
@@ -696,7 +696,27 @@ TEST_F(rnp_tests, s2k_iteration_tuning)
...
+    assert_int_equal(pgp_s2k_encode_iterations(1025), 1);
+    assert_int_equal(pgp_s2k_encode_iterations(1088), 1);
+    assert_int_equal(pgp_s2k_encode_iterations(1089), 2);
+    assert_int_equal(pgp_s2k_encode_iterations(2048), 16);
+    assert_int_equal(pgp_s2k_encode_iterations(MAX_ITER), 0xFF);
+    // decoding tests
+    assert_int_equal(pgp_s2k_decode_iterations(0), 1024);
+    assert_int_equal(pgp_s2k_decode_iterations(1), 1088);
+    assert_int_equal(pgp_s2k_decode_iterations(16), 2048);
+    assert_int_equal(pgp_s2k_decode_iterations(0xFF), MAX_ITER);
```
ni4: It would be good to add edge case(s) where value is larger then MAX_ITER in the encode call.

rrrooommmaaa:
I also refined tests to explicitly define exact or greater than comparison for more predictable behaviour

ronaldtse:
Well done! Merging 1000!! 🍾 🎆 😄

Issue #941, opened 2019-10-29
Ensure correct error handling for large MPIs


ni4:
## Description

While OpenPGP standard allows MPI lengths up to 64k bits, we allow only smaller ones.
Construct dummy key, signature and PKESK data files and make sure those are loaded with correct error reporting, without AV and range-check failures.
Separate test file should be added to rnp_tests suite.

ni4:
@killy631 would you like to take this issue? Cannot assign it to you until commented.

killy631:
Is there failed or demo?

killy631:
I meant - failed test in some suite..

ni4:
@killy631 Nope, there is no one (yet) - the idea is to build the test which will make sure that library works well in such cases (i.e. doesn't crash, has unexpected behavior and so on, and just reports an error).

ronaldtse:
@ni4 so @killy631 has just dropped off. This task is now open again.

ronaldtse:
@antonsviridenko is going to take this for trial. Welcome!

antonsviridenko:
ok, I've started working on it

antonsviridenko:
Which library functions are supposed to be tested? Only ones exposed in public API inside include/ directory or internal ones from src/librepgp? I used https://github.com/rnpgp/rnp/blob/master/src/tests/streams.cpp#L444 as a reference, but maybe I am moving into the wrong direction?

https://github.com/antonsviridenko/rnp/commit/459a4a31d39b0f4f401399415a420ff801753d42

ni4:
@antonsviridenko while all MPI load is done via `get_packet_body_mpi()`, single key load will not be good enough to check the whole - we should be sure that later on, if some code will be changed, nothing will fail. 
So I'd suggest loading of all key algorithms, with misc key lengths (PGP_MPINT_BITS + 1 should be handled as well as 64k bits).
Another thing which comes to my mind to check - large unknown curve oid.
Basically, pick all edge cases which you think would be worth checking to be protected against buffer overflow/memory overflow/other types of attacks.

antonsviridenko:
Keyrings in src/tests/data/keyrings directory, they should be considered read-only, am I right? For example, if I want to test importing keys of PGP_MPINT_BITS size, I should use some temporary keyring created in runtime, correct?

ni4:
@antonsviridenko yeah, those are used in multiple tests, and breaking them will most likely break those tests. Feel free to create separate folder for your data files named according to your test's name.

[...]

PR #1009, opened 2019-12-14
Fix signature parsing bugs

antonsviridenko:
Fixes 3 issues in signature processing discovered while working on https://github.com/rnpgp/rnp/issues/941


antonsviridenko:
I'm not sure if new separate test should be introduced or some existing one should be extended.
Basically, for this commit https://github.com/rnpgp/rnp/pull/1009/commits/7c1bc0f1ce52e9b758b6fd08357f1a1c8f44d8f9
it should be tested that signature packet is reported as invalid after parsing in case of any issues with signature MPI field.

Changes in other two commits are covered by "rnp_tests.test_ffi_op_verify_sig_count" as it segfaulted and then failed until these fixes were introduced. So maybe https://github.com/rnpgp/rnp/blob/master/src/tests/ffi.cpp can be updated with new function that tests loading signature with invalid MPI? Or new separate .cpp test file is better?

ni4:
@antonsviridenko with these changes some small additional test(s) in `ffi.cpp` will be okay. 
And it is always better to have a lot of smaller tests since those can be run simultaneously, and are easier to debug later on.

Issue #938, opened 2019-10-29
Check key packet length edge cases.


ni4:
## Description

Make sure key packet can be correctly read with 1-byte, 2-byte and 4-byte packet lengths.
Also test negative cases - when key packet length is very large (say, 3Gb), zero, 2 bytes, etc.
Add separate test for these cases to rnp_tests suite, using the modified existing key file.
Key loading should be done via the FFI interface (see `rnp_import_keys()` function).

ni4:
@msdev428 would you like to take this issue? Need your comment here to be able to assign it :) Thanks!

msdev428:
@ni4 sure, I will check and let you know. Thanks!

ni4:
@msdev428 great! Feel free to ask any questions here.

msdev428:
@ni4 do I need to create PR from `master` branch?

ni4:
@msdev428 no, please create a separate branch and then create PR from it. PR process in more details is described here: https://github.com/rnpgp/rnp/blob/master/docs/develop.adoc

msdev428:
@ni4 sorry for delay.
Can you advise me to solve this issue?
To simulate key packet length, then do I need to replace key length field in pgp binary key file, right?

ni4:
@msdev428 Yeah. Our implementation picks smallest possible packet header length. So you should cut existing header (see output of `rnp --list-packets keyfile`, prepend constructed header with different length field, and make sure key is loaded successfully.

ni4:
@msdev428 you should be able to create a PR from your own fork, please see https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork for the details.

msdev428:
@ni4 I created PR #950

ni4:
@msdev428 Thanks, I'll review it and add comments to the PR.

ni4:
Fixed in #1011

PR #1011, opened 2019-12-15
Add test for checking key pkt length while importing

ronaldtse:
Contination of #950

### DIFF
--- a/src/tests/ffi.cpp
+++ b/src/tests/ffi.cpp
@@ -5453,6 +5453,133 @@ TEST_F(rnp_tests, test_ffi_keys_import)
     rnp_ffi_destroy(ffi);
 }
 
+static std::vector<uint8_t>
+read_file_to_vector(const char *valid_key_path)
+{
+    std::ifstream        stream(valid_key_path, std::ios::in | std::ios::binary);
+    std::vector<uint8_t> contents((std::istreambuf_iterator<char>(stream)),
+                                  std::istreambuf_iterator<char>());
+    return contents;
+}
+
+/* shrink the length to 1 packet
+ * set packet length type as PGP_PTAG_OLD_LEN_1 and remove one octet from length header
+ */
+static std::vector<uint8_t>
+shrink_len_2_to_1(const std::vector<uint8_t> &src)
+{
+    std::vector<uint8_t> dst = std::vector<uint8_t>();
+    dst.reserve(src.size() - 1);
+    dst.insert(dst.end(),
+               PGP_PTAG_ALWAYS_SET |
+                 (PGP_PTAG_CT_PUBLIC_KEY << PGP_PTAG_OF_CONTENT_TAG_SHIFT) |
+                 PGP_PTAG_OLD_LEN_1);
+    // make sure the most significant octet of 2-octet length is actually zero
+    assert_int_equal(src[1], 0);
+    dst.insert(dst.end(), src[2]);
+    dst.insert(dst.end(), src.begin() + 3, src.end());
+    return dst;
+}
+
+/*
+ * fake a packet with len = 0xEEEE
+ */
+static std::vector<uint8_t>
+fake_len_EEEE(const std::vector<uint8_t> &src)
+{
+    std::vector<uint8_t> dst = std::vector<uint8_t>(src);
+    dst[1] = 0xEE;
+    dst[2] = 0xEE;
+    return dst;
+}
+
+/*
+ * fake a packet with len = 0x00
+ */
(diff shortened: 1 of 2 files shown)


rrrooommmaaa:
@ni4 I refactored the code to use vector for packet modification. Is that what you asked for before? I think that we should move this code from `ffi.cpp` I also think that `rnp_import_keys` is not an ideal method for testing packet parsing.

rrrooommmaaa:
@ni4 I also added a positive test for 4-byte length. To add a positive test for a single-byte test, I would need a public key that's less than 256 bytes, right? How can I achieve that?

ni4:
@rrrooommmaaa actually less then 192 bytes. For instance you may try EdDSA key, it is as small as 51 byte:
```
:off 0: packet header 0x9833 (tag 6, len 51)
Public key packet
    version: 4
    creation time: 1522760240 (Tue Apr  3 15:57:20 2018)
    public key algorithm: 22 (EdDSA)
    public key material:
        ecc p: 263 bits
        ecc curve: Ed25519
    keyid: 0xcc786278981b0728
```

src/tests/ffi.cpp
```diff
@@ -5451,6 +5451,121 @@ TEST_F(rnp_tests, test_ffi_keys_import)
...
 }
 
+static std::vector<uint8_t>
+read_file_to_vector(const char *valid_key_path, size_t allowed_size)
+{
+    FILE *  fp;
+    size_t  read_len;
+    uint8_t pkt[allowed_size + 1];
+    // open file for binary reading mode
+    fp = fopen(valid_key_path, "rb");
```
ni4: Let's use fstream or something similar here if we are aiming for C++.

src/tests/ffi.cpp
```diff
@@ -5451,6 +5451,121 @@ TEST_F(rnp_tests, test_ffi_keys_import)
...
+    assert_non_null(fp);
+    read_len = fread(pkt, 1, sizeof(pkt), fp);
+    assert_true(read_len <= allowed_size);
```
ni4: assert() in helper function would not allow to know which test actually caused it, so it would be better to return true/false and check it on upper level.

rrrooommmaaa:
@ni4 I added a positive 1-byte length test (using `keyrings/2/pubring.gpg`) and 
also updated len=0 test (erased extra octets from the stream so that the next packet would be read correctly).
What about len=EEEE and len=EEEEEEEE? Should I pad the packet with zeroes? I would need to implement callback.

src/tests/ffi.cpp
```diff
@@ -737,13 +737,13 @@ getpasscb_once(rnp_ffi_t        ffi,
 
 static bool
 getpasscb_inc(rnp_ffi_t        ffi,
-               void *           app_ctx,
-               rnp_key_handle_t key,
-               const char *     pgp_context,
-               char *           buf,
-               size_t           buf_len)
+              void *           app_ctx,
```
ni4: @rrrooommmaaa  Please rebase on master so these changes which are already merged will not appear in PR.

src/tests/rnp_tests.h
```diff
@@ -62,6 +62,8 @@ void rnpkeys_generatekey_verifykeyHomeDirOption(void **state);
 
 void rnpkeys_generatekey_verifykeyKBXHomeDirOption(void **state);
 
+void rnpkeys_generatekey_verifykeyNonexistingHomeDir(void **state);
```
ni4: This should be some rebase/merge artifact - this test was removed a long time ago.

ni4:
@rrrooommmaaa 
> What about len=EEEE and len=EEEEEEEE? Should I pad the packet with zeroes? I would need to implement callback.

No, I don't think this is needed at the moment.

rrrooommmaaa:
Closes #938

src/tests/ffi.cpp
rrrooommmaaa: I applied clang-format on this file. (It wasn't applied on `master`, looks like)

ni4: @rrrooommmaaa It was because this PR was started before the clang-format rule was added and run on the whole master. Anyway, now it looks good and almost ready to merge.

ni4:
@rrrooommmaaa Could you please squash 9e723123 with 631130a0 so we do not have extra commits?

ronaldtse:
Any comments @dewyatt ?

ronaldtse:
I approve this as well, but since this PR was created by me I can't "approve" it officially. Hence this message. Merging before 2020!

PR #1016, opened 2019-12-22
Remove some unused code.

ni4:
...to make codecov happier.


ronaldtse:
Ping @dewyatt for a review.

src/lib/pass-provider.h
```diff
@@ -26,7 +26,8 @@
 #ifndef RNP_PASS_PROVIDER_H
 #define RNP_PASS_PROVIDER_H
 
-#include <rnp/rnp_sdk.h>
+#include <stddef.h>
+#include <stdint.h>
```
dewyatt: I think these should technically be `cstddef` and `cstdint`, right?

ni4: @dewyatt Looks like it's not an easy question: https://stackoverflow.com/questions/5079325/should-i-include-stddef-h-or-cstddef-for-size-t

But, agree, let's stick to cstddef/cstdint notation as our code is aimed to be C++ sooner or later.

Issue #1001, opened 2019-12-11
Public keys, with signatures from unknown keys, are treated as invalid keys


kaie:
Alice's public key was signed by Bob.
Alice sends her key to Carol.

Carol uses OpenPGP software based on RNP.

Carol imports Alice's key, but doesn't have Bob's key.
Carol attempts to send an encrypted message to Alice.

Actual behavior:
Sending the encrypted message fails.
RNP reports that Alice's key is invalid.
[signature_check() /home/user/github/rnp/src/librepgp/stream-sig.cpp:1099] invalid or untrusted key
[validate_pgp_key_signature() /home/user/github/rnp/src/librepgp/stream-key.cpp:1571] bad signature
(line numbers are based on revision 369a687a04781dc177f30900cf3d9010695b42d7)

Expected behavior:
Carol should be able to send an encrypted message to Alice, even if Carol doesn't have Bob's key.

kaie:
To check the unknown signature is really causing the issue, I have used gnupg to edit Alice's key, removed Bob's signature, and imported this reduced key into RNP. This fixed the issue, sending the encrypted message to Alice was successful.

ni4:
@kaie thanks for noticing it. Within current logic signature from unknown key is considered as invalid, so the whole key is marked as invalid. And it seems to be too strict. Actually improving key trust settings is also part of the ongoing work.

kaie:
This patch allows me to use my usual keys.

```
diff --git a/src/librepgp/stream-key.cpp b/src/librepgp/stream-key.cpp
index 20c925d7..93fcd9fd 100644
--- a/src/librepgp/stream-key.cpp
+++ b/src/librepgp/stream-key.cpp
@@ -1474,7 +1474,8 @@ typedef struct validate_info_t {
  *         during the validation.
  */
 static rnp_result_t
-validate_pgp_key_signature(const pgp_signature_t *sig, validate_info_t *info)
+validate_pgp_key_signature(const pgp_signature_t *sig, validate_info_t *info,
+                           bool ignore_missing_signer_key)
 {
     rnp_result_t         res = RNP_ERROR_GENERIC;
     uint8_t              signer_id[PGP_KEY_ID_SIZE] = {0};
@@ -1493,6 +1494,9 @@ validate_pgp_key_signature(const pgp_signature_t *sig, validate_info_t *info)
     }
     sinfo.signer = rnp_key_store_get_key_by_id(info->keystore, signer_id, NULL);
[...]

PR #1017, opened 2019-12-23
Improve and relax key validation.

ni4:
Within current implementation when key is validated all signatures are checked, which is not correct.  For instance, if you'll import public key with valid self-signature but with some invalid certification from other key(s), you will not be able to use it ( see the issue #1001 )

This PR is aimed to improve those checks to have more logical behavior.


ni4:
@ronaldtse @dewyatt Could you please review this? Thanks!

ni4:
Ping @dewyatt @ronaldtse

PR #1028, opened 2020-01-13
Cancel implicit loading of pubring.

### LINKED ISSUE #11: Prune spurious 'keyring not found' output shown on every single command
Almost every command with `netpgp` shows up with this output preceding any action (if there is no `${HOME}/.gnupg`):

``` sh
${HOME}/.gnupg/pubring.gpg: No such file or directory
Can't read pubring /root/.gnupg/pubring.gpg
Can't read pub keyring
can't initialise
```

This should show useful output instead.

joke325:
Closes #11

### DIFF
--- a/src/rnp/rnp.cpp
+++ b/src/rnp/rnp.cpp
@@ -277,10 +277,16 @@ setcmd(rnp_cfg_t *cfg, int cmd, const char *arg)
     switch (cmd) {
     case CMD_ENCRYPT:
         rnp_cfg_setbool(cfg, CFG_ENCRYPT_PK, true);
+        if (rnp_cfg_getbool(cfg, CFG_ENCRYPT_SK)) {
+            rnp_cfg_setint(cfg, CFG_KEYSTORE_DISABLED, 0);
+        }
         newcmd = CMD_PROTECT;
         break;
     case CMD_SYM_ENCRYPT:
         rnp_cfg_setbool(cfg, CFG_ENCRYPT_SK, true);
+        if (!rnp_cfg_getbool(cfg, CFG_ENCRYPT_PK) && !rnp_cfg_getbool(cfg, CFG_SIGN_NEEDED)) {
+            rnp_cfg_setint(cfg, CFG_KEYSTORE_DISABLED, 1);
+        }
         newcmd = CMD_PROTECT;
         break;
     case CMD_CLEARSIGN:
@@ -289,6 +295,9 @@ setcmd(rnp_cfg_t *cfg, int cmd, const char *arg)
     case CMD_SIGN:
         rnp_cfg_setbool(cfg, CFG_NEEDSSECKEY, true);
         rnp_cfg_setbool(cfg, CFG_SIGN_NEEDED, true);
+        if (rnp_cfg_getbool(cfg, CFG_ENCRYPT_SK)) {
+            rnp_cfg_setint(cfg, CFG_KEYSTORE_DISABLED, 0);
+        }
         newcmd = CMD_PROTECT;
         break;
     case CMD_DECRYPT:
--- a/src/tests/cli_tests.py
+++ b/src/tests/cli_tests.py
@@ -1243,6 +1243,31 @@ class Misc(unittest.TestCase):
         run_proc(RNPK, ['--homedir', data_path('test_stream_key_load/g10'), '--list-keys', '--debug', '--all'])
         return
 
+    def test_pubring_loading(self):
+        test_dir = tempfile.mkdtemp(prefix='rnpctmp')
+        test_data = data_path('test_messages/message.txt')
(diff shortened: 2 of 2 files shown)


dewyatt:
It looks like this needs to be rebased on master now.

rrrooommmaaa:
@dewyatt It would be nice to assign `v0.13.2` tag when this is merged as I would like to include symmetric encrypt/decrypt test to Homebrew formula (because they said `rnp --version` is not a perfect test).

src/rnp/rnp.cpp
```diff
@@ -281,6 +281,7 @@ setcmd(rnp_cfg_t *cfg, int cmd, const char *arg)
         break;
     case CMD_SYM_ENCRYPT:
         rnp_cfg_setbool(cfg, CFG_ENCRYPT_SK, true);
+        rnp_cfg_setint(cfg, CFG_KEYSTORE_DISABLED, 1);
```
ni4: This will not be correct since user may use public-key and symmetric encryption simultaneously, or want to symmetrically encrypt and sign file.

ni4: Also, together with such changes, tests should be added which makes sure that issue is actually fixed (and do not have regression later on).

joke325: So what is the right "... one-line fix (via CFG_KEYSTORE_DISABLED) ..." as @dewyatt required above?

dewyatt: @joke325 Your fix is probably what I would have done, as it fixes the issue I referred to in my comment, but as @ni4 mentioned it probably introduces a separate issue, so maybe a few more lines will be required. :)

dewyatt:
> @dewyatt It would be nice to assign `v0.13.2` tag when this is merged as I would like to include symmetric encrypt/decrypt test to Homebrew formula (because they said `rnp --version` is not a perfect test).

@rrrooommmaaa In the mean time I believe you could do something like:
```
echo -ne "pass1\npass2\n" | rnpkeys --gen-key --numbits=1024 --userid=test --pass-fd=0
# now do a symmetric test (or pubkey)
```

ni4:
@rrrooommmaaa ...and do not forget to delete this key afterwards, and exactly this one (so existing keys in user keyring are not changed).

dewyatt:
> @rrrooommmaaa ...and do not forget to delete this key afterwards, and exactly this one (so existing keys in user keyring are not changed).

I think a temporary homedir would be the smartest route.

ronaldtse:
Ping @joke325 are we ready to merge this? Thanks!

joke325:
@ronaldtse   Unready to merge

src/tests/cli_tests.py
```diff
@@ -1243,6 +1243,27 @@ def test_debug_log(self):
...
 
+    def test_pubring_loading(self):
+        test_dir = tempfile.mkdtemp(prefix='rnpctmp')
+        test_data = data_path('test_messages/message.txt')
+        output = path.join(test_dir, 'output')
+        params = ['--symmetric', '--password', 'pass', '--homedir', test_dir, test_data, '--output', output]
+        
+        ret, _, err = run_proc(RNP, ['--encrypt'] + params)
+        if not (ret == 2 and 'wrong pubring path' in err):
+            raise_err('encrypt w/o pubring didn\'t fail', err)
```
dewyatt: nit: Use double quotes rather than escaping.

src/tests/cli_tests.py
```diff
@@ -1243,6 +1243,27 @@ def test_debug_log(self):
...
+
+        ret, _, err = run_proc(RNP, ['--sign'] + params)
+        if not (ret == 2 and 'wrong pubring path' in err):
+            raise_err('sign w/o pubring didn\'t fail', err)
```
dewyatt: nit: Use double quotes rather than escaping.

src/tests/cli_tests.py
```diff
@@ -1243,6 +1243,27 @@ def test_debug_log(self):
...
+        test_dir = tempfile.mkdtemp(prefix='rnpctmp')
+        test_data = data_path('test_messages/message.txt')
+        output = path.join(test_dir, 'output')
+        params = ['--symmetric', '--password', 'pass', '--homedir', test_dir, test_data, '--output', output]
+        
+        ret, _, err = run_proc(RNP, ['--encrypt'] + params)
+        if not (ret == 2 and 'wrong pubring path' in err):
+            raise_err('encrypt w/o pubring didn\'t fail', err)
+
+        ret, _, err = run_proc(RNP, ['--sign'] + params)
```
dewyatt: Maybe we should test `--clearsign` too

joke325:
@ronaldtse apparently I don't have write access to this repository so I cannot merge this PR

PR #1032, opened 2020-01-13
Add tests for handling large MPIs in RSA related operations. Partially addresses #941

antonsviridenko:
This PR adds tests for handling large MPIs in public and private RSA key packets, RSA signature packets and PKESK packets encrypted using RSA. Tests check if largest acceptable MPI of size 16384 bits is successfully processed and if attempts to load larger MPIs are properly rejected.

Remaining #941 tests for other algorithms  will follow in subsequent pull requests


src/librepgp/stream-parse.cpp
```diff
@@ -1880,7 +1881,6 @@ init_encrypted_src(pgp_processing_ctx_t *ctx, pgp_source_t *src, pgp_source_t *r
     }
 
     src->read = param->aead ? encrypted_src_read_aead : encrypted_src_read_cfb;
-    src->close = encrypted_src_close;
     src->finish = encrypted_src_finish;
     src->type = PGP_STREAM_ENCRYPTED;
```
ni4: These ones should be also moved, so stream is fully initialized before possible error exit.

src/tests/large-mpi.cpp
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright (c) 2017-2019 [Ribose Inc](https://www.ribose.com).
```
ni4: Let's add 2020 here as well.

src/librepgp/stream-parse.cpp
antonsviridenko: moving these lines makes 3 tests fail :)

```
--- a/src/librepgp/stream-parse.cpp
+++ b/src/librepgp/stream-parse.cpp
@@ -1872,7 +1872,11 @@ init_encrypted_src(pgp_processing_ctx_t *ctx, pgp_source_t *src, pgp_source_t *r
     }
     param = (pgp_source_encrypted_param_t *) src->param;
     param->pkt.readsrc = readsrc;
+
     src->close = encrypted_src_close;
+    src->read = param->aead ? encrypted_src_read_aead : encrypted_src_read_cfb;
+    src->finish = encrypted_src_finish;
+    src->type = PGP_STREAM_ENCRYPTED;
 
     /* Read the packet-related information */
     errcode = encrypted_read_packet_data(param);
@@ -1880,10 +1884,6 @@ init_encrypted_src(pgp_processing_ctx_t *ctx, pgp_source_t *src, pgp_source_t *r
         goto finish;
     }
 
-    src->read = param->aead ? encrypted_src_read_aead : encrypted_src_read_cfb;
-    src->finish = encrypted_src_finish;
-    src->type = PGP_STREAM_ENCRYPTED;
-
     /* Obtaining the symmetric key */
     have_key = false;
 

```

antonsviridenko: I guess `param->aead` is initialized by encrypted_read_packet_data() so this line
`src->read = param->aead ? encrypted_src_read_aead : encrypted_src_read_cfb;`
should be left on its place

ni4: @antonsviridenko Correct, this one should be set after the `encrypted_read_packet_data()`.

ni4:
@antonsviridenko We usually squash/rebase commits, which just fixed previous commits in PR, to keep master commit history as clean as possible.

rrrooommmaaa:
why are some files 'pgp' and some 'gpg'?

src/tests/large-mpi.cpp
```diff
@@ -0,0 +1,129 @@
...
+    assert_rnp_success(rnp_output_to_path(&output, "decrypted"));
+    assert_rnp_success(rnp_decrypt(ffi, input, output));
+    rnp_input_destroy(input);
+    rnp_output_destroy(output);
+    unlink("decrypted");
+
+    /* Load PKESK having 16385 bit MPI. Must fail. */
+    assert_rnp_success(
+      rnp_input_from_path(&input, "data/test_large_MPIs/message.enc.rsa16385.pgp"));
+    assert_rnp_success(rnp_output_to_path(&output, "decrypted"));
```
rrrooommmaaa: I would use `rnp_output_to_null` here, if the output is not needed

src/tests/large-mpi.cpp
```diff
@@ -0,0 +1,129 @@
...
+    /* Load RSA private key of PGP_MPINT_BITS size (16384). Must succeed. */
+    assert_rnp_success(
+      rnp_input_from_path(&input, "data/test_large_MPIs/rsa-priv-16384bits.pgp"));
+    assert_rnp_success(rnp_import_keys(ffi, input, RNP_LOAD_SAVE_SECRET_KEYS, NULL));
+    rnp_input_destroy(input);
+
+    /* Load PKESK encrypted by PGP_MPINT_BITS sized key. Must succeed. */
+    assert_rnp_success(
+      rnp_input_from_path(&input, "data/test_large_MPIs/message.enc.rsa16384.pgp"));
+    assert_rnp_success(rnp_output_to_path(&output, "decrypted"));
```
rrrooommmaaa: I would use `rnp_output_to_null` here, if the output is not needed

antonsviridenko:
>  Cannot retrieve metalink for repository: epel/x86_64. Please verify its path and try again
##[error]Process completed with exit code 1.

on centos7 / tests (sanitize, beta)

antonsviridenko:
@rrrooommmaaa 

> why are some files 'pgp' and some 'gpg'?

fixed

src/tests/large-mpi.cpp
```diff
@@ -0,0 +1,128 @@
...
+    assert_rnp_success(init_file_src(&keysrc, "data/test_large_MPIs/rsa-pub-65535bits.pgp"));
+    assert_rnp_failure(process_pgp_keys(&keysrc, &keyseq));
+    assert_int_equal(list_length(keyseq.keys), 0);
+    key_sequence_destroy(&keyseq);
+    src_close(&keysrc);
+
+    assert_rnp_success(rnp_ffi_create(&ffi, "GPG", "GPG"));
+    assert_rnp_success(
+      rnp_input_from_path(&input, "data/test_large_MPIs/rsa-pub-65535bits.pgp"));
+    assert_rnp_failure(rnp_import_keys(ffi, input, RNP_LOAD_SAVE_PUBLIC_KEYS, NULL));
```
rrrooommmaaa: I would test against a specific error code to make sure we're really getting the MPI parse error: `assert_int_equal(rnp_import_keys(ffi, input, RNP_LOAD_SAVE_PUBLIC_KEYS, NULL), RNP_ERROR_BAD_FORMAT);`

src/tests/large-mpi.cpp
```diff
@@ -0,0 +1,128 @@
...
+    rnp_input_destroy(input);
+
+    /* Load RSA pubkey of PGP_MPINT_BITS + 1 size (16385). Must fail. */
+    assert_rnp_success(
+      rnp_input_from_path(&input, "data/test_large_MPIs/rsa-pub-16385bits.pgp"));
+    assert_rnp_failure(rnp_import_keys(ffi, input, RNP_LOAD_SAVE_PUBLIC_KEYS, NULL));
```
rrrooommmaaa: I would test against a specific error code to make sure we're really getting the MPI parse error: `assert_int_equal(rnp_import_keys(ffi, input, RNP_LOAD_SAVE_PUBLIC_KEYS, NULL), RNP_ERROR_BAD_FORMAT);`

src/tests/large-mpi.cpp
```diff
@@ -0,0 +1,128 @@
...
+    pgp_source_t       keysrc = {0};
+    pgp_key_sequence_t keyseq;
+    rnp_ffi_t          ffi = NULL;
+    rnp_input_t        input = NULL;
+    rnp_input_t        signature = NULL;
+    rnp_op_verify_t    verify;
+
+    /* Load RSA pubkey packet with 65535 bit modulus MPI. Must fail. */
+    assert_rnp_success(init_file_src(&keysrc, "data/test_large_MPIs/rsa-pub-65535bits.pgp"));
+    assert_rnp_failure(process_pgp_keys(&keysrc, &keyseq));
```
rrrooommmaaa: I would test against a specific error code to make sure we're really getting the MPI parse error: `assert_int_equal(process_pgp_keys(&keysrc, &keyseq), RNP_ERROR_BAD_FORMAT);`

src/tests/large-mpi.cpp
```diff
@@ -0,0 +1,128 @@
...
+    assert_rnp_success(rnp_output_to_null(&output));
+    assert_rnp_success(rnp_decrypt(ffi, input, output));
+    rnp_input_destroy(input);
+    rnp_output_destroy(output);
+
+    /* Load PKESK having 16385 bit MPI. Must fail. */
+    assert_rnp_success(
+      rnp_input_from_path(&input, "data/test_large_MPIs/message.enc.rsa16385.pgp"));
+    assert_rnp_success(rnp_output_to_null(&output));
+    assert_rnp_failure(rnp_decrypt(ffi, input, output));
```
rrrooommmaaa: I would test against a specific error code to make sure we're really getting the MPI parse error:
`assert_int_equal(rnp_decrypt(ffi, input, output), RNP_ERROR_BAD_FORMAT);`

src/tests/large-mpi.cpp
```diff
@@ -0,0 +1,128 @@
...
+    rnp_ffi_t    ffi = NULL;
+    rnp_input_t  input = NULL;
+    rnp_output_t output = NULL;
+
+    assert_rnp_success(rnp_ffi_create(&ffi, "GPG", "GPG"));
+
+    /* Load RSA private key of PGP_MPINT_BITS + 1 size (16385). Must fail. */
+    assert_rnp_success(
+      rnp_input_from_path(&input, "data/test_large_MPIs/rsa-priv-16385bits.pgp"));
+    assert_rnp_failure(rnp_import_keys(ffi, input, RNP_LOAD_SAVE_SECRET_KEYS, NULL));
```
rrrooommmaaa: I would test against a specific error code to make sure we're really getting the MPI parse error:
`assert_int_equal(rnp_import_keys(ffi, input, RNP_LOAD_SAVE_SECRET_KEYS, NULL), RNP_ERROR_BAD_FORMAT);`

antonsviridenko:
@rrrooommmaaa I had this idea too, but it seems that it is not documented anywhere what error codes should be returned on particular errors.

See https://github.com/rnpgp/rnp/blob/master/include/rnp/rnp.h . Almost every API function's return value is described as `*  @return RNP_SUCCESS on success, or any other value on error`, error codes are defined only for rnp_op_verify_signature_get_status().

So checking against specific error codes may break the tests more than neccessary.

rrrooommmaaa:
@antonsviridenko Let's check with other developers where we're going with error codes, but:
tests shouldn't be resilient (as opposed to library functions).
The more often they break the better: the task of tests is to draw attention of developers even if it is a false alarm sometimes.
If we're talking about applications, they should be resilient, just check SUCCESS code or not.

dewyatt:
I would say test against a specific error code if one has been documented.
If one has not been documented and you feel confident that you can choose (or add) one that is specific enough and unlikely to change, document it, and test for it.

The reason we often test only for success or failure is that in many areas there's room for improvement on error codes. For example, we may currently return `RNP_ERROR_GENERIC` somewhere, but in the future we may return something more specific.

Eventually we will want to move away from `assert_rnp_success`/`assert_rnp_failure` and test for specific error codes.

antonsviridenko:
Adding new error codes can be a subject of another separate pull request, not this one :P to keep changes smaller and simplier.

For the intended purpose of #941 and this PR new specific error code RNP_ERROR_MPI_TOO_LARGE can be introduced.

ronaldtse:
One thing brought up is that we should document on the exact error codes that are possibly returned by a function to indicate the actual failure, to allow the user to gracefully handle them.

For most people generic success/error may be sufficient, but transparency is necessary for more accurate / advance usage of the API.

rrrooommmaaa:
I suggest to extend documentation for functions `rnp_import_keys`, `process_pgp_keys` and `rnp_decrypt` in this pull request, not to postpone it.

antonsviridenko:
@ronaldtse @ni4 So what is the final decision? I still think it is better to update documentation about return codes in a separate task, and all tests can be updated to check against specific return values after that, not only this one added by PR

ronaldtse:
> I still think it is better to update documentation about return codes in a separate task, and all tests can be updated to check against specific return values after that, not only this one added by PR

Agree, let's proceed with this PR, and file the relevant new issues. Thanks.

(Some builds for CentOS are failing)

antonsviridenko:
> (Some builds for CentOS are failing)

Yes, there was some issue with CentOS repositories, unrelated to code changes

https://github.com/rnpgp/rnp/pull/1032#issuecomment-574885792

Is there a way to re-run CI tests again?

ronaldtse:
You can use the "re-run checks" button at the top right corner in the failed action:

Screen Shot 2020-01-22 at 12 50 49 AM

antonsviridenko:
@ronaldtse I don't see this button, probably it is not available to me

ronaldtse:
Ah, can you check again? I just invited you to join the organization.

antonsviridenko:
@ronaldtse ok, it worked

ronaldtse:
Ping @dewyatt @rrrooommmaaa for one more review to merge. Thanks!

ni4:
@rrrooommmaaa @antonsviridenko Once there are 2 reviews feel free to rebase and merge the PR.
Regarding the FFI return codes - agree with @dewyatt , those should be tested against only after being documented. And there is another thing - since we should not change API, we cannot change error codes as well. So, if at one point function will be documented as 'function X return Y on case(s) Z', it should return that error code only in that case later on as well.

PR #1033, opened 2020-01-13
Fix rnpkeys secret key protection

dewyatt:
@ni4 Can you take a look and also add some `cli_tests`?

### DIFF
--- a/src/rnp/fficli.cpp
+++ b/src/rnp/fficli.cpp
@@ -855,19 +855,6 @@ cli_rnp_generate_key(rnp_cfg_t *cfg, cli_rnp_t *rnp, const char *username)
         (void) fprintf(stderr, "Failed to set hash algorithm.\n");
         goto done;
     }
-    if (rnp_op_generate_set_protection_cipher(genkey, rnp_cfg_getstr(cfg, CFG_KG_PROT_ALG))) {
-        (void) fprintf(stderr, "Failed to set protection algorithm.\n");
-        goto done;
-    }
-    if (rnp_op_generate_set_protection_hash(genkey, rnp_cfg_getstr(cfg, CFG_KG_PROT_HASH))) {
-        (void) fprintf(stderr, "Failed to set protection hash algorithm.\n");
-        goto done;
-    }
-    if (rnp_op_generate_set_protection_iterations(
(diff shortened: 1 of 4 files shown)


ni4:
@dewyatt I added commit with CLI test, however I'm not sure whether it will work with stdin password - it doesn't work locally, asking for password from tty. Will try to figure things out tomorrow.

dewyatt:
> @dewyatt I added commit with CLI test, however I'm not sure whether it will work with stdin password - it doesn't work locally, asking for password from tty. Will try to figure things out tomorrow.

Ah that's right, that won't work. Well the pass-fd test should be adequate for now, that at least confirms that we're generating protected keys. We can add an issue for the tty testing later.

ni4:
@dewyatt ok, so I'll force-push deleting the stdin test?

ni4:
...hm, looks like it works in virtual environment, without tty attached.

dewyatt:
> ...hm, looks like it works in virtual environment, without tty attached.

I would still drop the stdin test if it fails locally (which does make sense), we probably don't want tests that pass in CI and fail locally.

ni4:
@dewyatt Ok, agreed. Filed #1034 and deleting the test now.

src/rnp/fficli.cpp
```diff
@@ -902,24 +889,36 @@ cli_rnp_generate_key(rnp_cfg_t *cfg, cli_rnp_t *rnp, const char *username)
...
+    rnp_password_cb passcb;
+    void *          passctx;
+    char            password[MAX_PASSWORD_LENGTH];
+    if (rnp->passfp) {
+        passcb = &ffi_pass_callback_file;
+        passctx = rnp->passfp;
+    } else {
+        passcb = &ffi_pass_callback_stdin;
+        passctx = NULL;
+    }
```
ni4: What do you think about changing this to some FFI function call like rnp_request_password(ffi, char **password), which will use FFI's configuration instead of this (imagine, we'll add more ways to request password from CLI)?

dewyatt: I thought about that too since the context is already wrapped up in the ffi object. I'd rather file a separate issue for further consideration though like you suggested.

ni4: @dewyatt Ok, filed #1036

ronaldtse:
Thanks guys!

PR #1038, opened 2020-01-14
build: Fix version determination for master branch.

### LINKED ISSUE #1037: Fix version.cmake for master branch
## Description

Our initial release tag (0.9.0) is on a commit shared in the master branch, so rnp master ends up showing a version like:
`0.9.0+git20200114.618.8017915`
Which, although technically accurate, was intended to be more like:
`0.0.0+git20200114.8017915`

dewyatt:
Fixes #1037


dewyatt:
I may have to get back to this tomorrow, I'll re-add reviewers when ready.

dewyatt:
Ok this should be ok now. I added some tests to CI for the `version.cmake` module since it was already fairly complex, then I updated things based on feedback in https://github.com/rnpgp/rnp/issues/1037#issuecomment-574161255, then I had to fight with older versions of git and deal with the detached HEAD in CI.

ci/main.sh
```diff
@@ -5,6 +5,7 @@ set -eux
 
 : "${RNP_TESTS:=.*}"
 : "${LD_LIBRARY_PATH:=}"
+export RNP_OVERRIDE_BRANCH=master
```
ronaldtse: Should we set this here or in the job definition?

ronaldtse:
I wonder if other cmake projects would be interested in the versioning module? If so we could extract it into its own repo...

ci/main.sh
dewyatt: We could go either way, I'm thinking it's maybe easier here since it's one less env var for a developer have to set if they are running the CI scripts locally (which is sometimes useful if you're too impatient for the 2-core CI systems).

dewyatt:
> I wonder if other cmake projects would be interested in the versioning module? If so we could extract it into its own repo...

Possibly, I've seen similar scripts floating around that I didn't care for (various reasons) and I've seen blog posts of people writing these so it may be a lot of effort that we can save people. (Unless our solution sucks, but I think it's pretty decent)

ci/main.sh
ronaldtse: Got it. Thanks 👍

ronaldtse:
@dewyatt sounds good. Then do you want to extract the module out now, and then merge the updated code in?

dewyatt:
Ok I extracted it out etc.

ronaldtse:
Ping @ni4

ronaldtse:
Ping @dewyatt for conflict resolution.

ronaldtse:
Please feel free to merge once done @dewyatt. Thanks!

dewyatt:
> Please feel free to merge once done @dewyatt. Thanks!

Will do, just looking at something for a bit.

PR #1039, opened 2020-01-14
Fix issues, reported by coverity.

ni4:
I didn't fix all of the currently reported issues, but at least most of them.
Other looks like false positives.


dewyatt:
This reminds me I need to restore Coverity scans after the migration to GHA.

PR #1046, opened 2020-01-19
ci: Add GHA cache

dewyatt:
This saves a good chunk of time (~22% in my testing) for centos+ubuntu+macos when cache is available.
I also went ahead and reduced the ubuntu matrix because I don't believe there's any additional benefit in doing the sanitize/coverage/etc jobs there (in addition to centos).


ronaldtse:
Merging since this is just CI.

dewyatt:
I'm going to have to follow up and fix this, it breaks the build for macOS when run on a version of the github runner that differs from the one which created the cache. -_-

I'd say leave it for now, I'll try to get a fix done within a few hours.

PR #1047, opened 2020-01-19
Added the --keyring option to support custom keyring names.

### LINKED ISSUE #70: Rnpkeys : user is unable to create/append keys to custom keyring.
The user is unable to create new keys in a custom user provided keyring. 
`usage: ./rnpkeys --generate-key --keyring=max`
expected output: new keyring with name "maxpub & maxsec" is created in the HOMEDIR.
Test case; `rnpkeys_generatekey_verifykeyRingOptions`

joke325:
Closes #70

### DIFF
--- a/src/rnp/fficli.cpp
+++ b/src/rnp/fficli.cpp
@@ -1300,7 +1300,8 @@ rnp_path_compose_considering_keyring(const char *dir,
                 new_dir = ptr_dir;
                 new_subdir = NULL;
             } else {
-                if (!rnp_compose_path_ex(&ptr_subdir, NULL, subdir, keyring_dir.c_str(), NULL)) {
+                if (!rnp_compose_path_ex(
+                      &ptr_subdir, NULL, subdir, keyring_dir.c_str(), NULL)) {
                     return false;
                 }
                 new_subdir = ptr_subdir;
@@ -1315,7 +1316,8 @@ rnp_path_compose_considering_keyring(const char *dir,
         }
         new_filename = ptr_filename;
     }
-    result = rnp_path_compose(new_dir, new_subdir, filename ? new_filename : "", res, res_size);
+    result =
+      rnp_path_compose(new_dir, new_subdir, filename ? new_filename : "", res, res_size);
     free(ptr_dir);
     free(ptr_subdir);
     free(ptr_filename);
@@ -1412,7 +1414,8 @@ rnp_cfg_set_ks_info(rnp_cfg_t *cfg)
                 return false;
             }
         } else {
-            if (!rnp_path_compose_considering_keyring(homedir, subdir, NULL, keyring, pubpath, sizeof(pubpath))) {
+            if (!rnp_path_compose_considering_keyring(
+                  homedir, subdir, NULL, keyring, pubpath, sizeof(pubpath))) {
                 return false;
             }
         }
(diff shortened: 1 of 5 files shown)


dewyatt:
I haven't looked at this much, but personally I don't think I'd be satisfied with this approach.
1) `rnp_compose2` is not a very descriptive name.
2) `rnp_compose2` looks over-complicated and it's trivial to make it segfault.
3) We already have `rnp_compose_path` and `paths_concat`. We need to start using `std::filesystem::path` et al and drop this stuff, not add more of it. (but catch those pesky exceptions)

joke325:
@dewyatt I didn't observe any exceptions and it works with GPG21

dewyatt:
> @dewyatt I didn't observe any exceptions and it works with GPG21

The exceptions I was referring to were just basic STL ones, but it looks like we already have this issue in the code base, so we can track that separately.

I think my concern with GPG21 was that it will create keyrings that aren't readily usable with gpg, like `<prefix>private-keys-v1.d`. Our `--keyring` option is pretty different from gpg's overall. That may be fine though. @ni4 @ronaldtse opinions?

ni4:
@dewyatt GnuPG documentation tells `--secret-keyring file
This is an obsolete option and ignored. All secret keys are stored in the private-keys-v1.d directory below the GnuPG home directory.`, so we definitely should not use prefix for this one.

At first we should strictly define what actually `--keyring` option does. If it just adds prefix, then it should be called `--keyring-prefix`. However, then we would not be able to use keyring files with different extensions or endings, like pubring.pgp/pubkeys.asc/pubring.gpg, whatever else.
If it will have full keyring path, then we should also use `--secret-keyring` option.

Another thing - for GnuPG --keyring seems stackable, and has additionally --default-keyring option.

And I would definitely avoid adding new lines of tricky C code, and use std::string instead.

joke325:
MohitKumarAgniotri on May 6, 2017 wrote:
> The user is unable to create new keys in a custom user provided keyring.
> usage: ./rnpkeys --generate-key --keyring=max
> expected output: new keyring with name "maxpub & maxsec" is created in the HOMEDIR.

@ni4 @dewyatt nor @ronaldtse  didn't protest about the description of expected functionality.

The current PR just adds the prefix, in addition it can contain a relative or absolute path to the keyring.

@ni4 where and how I should use the std::string?

dewyatt:
> MohitKumarAgniotri on May 6, 2017 wrote:
> 
> > The user is unable to create new keys in a custom user provided keyring.
> > usage: ./rnpkeys --generate-key --keyring=max
> > expected output: new keyring with name "maxpub & maxsec" is created in the HOMEDIR.
> 
> @ni4 @dewyatt nor @ronaldtse didn't protest about the description of expected functionality.

When starting on a 3 year old issue, I'd say it's prudent to re-evaluate and check on whether things have changed. Sometimes we have higher priorities and things slip through the cracks. :(

> 
> The current PR just adds the prefix, in addition it can contain a relative or absolute path to the keyring.
> 
> @ni4 where and how I should use the std::string?

Generally if you find yourself using `strcpy`/`strcat`/`str*()`/`char*`, consider using a `std::string`. It does introduce exception concerns and most of our code is not exception-safe, but that's an issue we need to tackle regardless.

ni4:
Let's close this as not releveant anymore: we have `--homedir` and `--keyfile` options, which should be sufficient for all uses. If some edge case is not covered - we'd create a separate issue/PR.

PR #1055, opened 2020-01-22
Use single ERR_MSG() call to report errors and warning in CLI.

ni4:
Probably later on we should use some function instead of define, but at least now it will be more consistent, and later on easily replaceable.


ni4:
@rrrooommmaaa Thanks for noticing, please see the updated PR. Actually it would be better to remove that commented out code.

Issue #994, opened 2019-12-06
Possible rare error in test test_key_unlock_pgp.


ni4:
## Description

Under ubuntu with (normal, beta) configuration got this single errored test.
Under other oses and previously it worked well.

## Log

```
[==========] Running 1 test from 1 test suite.
2019-12-05T14:58:30.1893388Z [----------] Global test environment set-up.
2019-12-05T14:58:30.1893644Z [----------] 1 test from rnp_tests
2019-12-05T14:58:30.1893731Z [ RUN      ] rnp_tests.test_key_unlock_pgp
2019-12-05T14:58:30.1894073Z [signature_check_binding() /home/runner/work/rnp/rnp/src/librepgp/stream-sig.cpp:1210] subkey expired 64360857 seconds ago
2019-12-05T14:58:30.1894388Z [signature_check_binding() /home/runner/work/rnp/rnp/src/librepgp/stream-sig.cpp:1210] subkey expired 64360857 seconds ago
2019-12-05T14:58:30.1894722Z [signature_check_binding() /home/runner/work/rnp/rnp/src/librepgp/stream-sig.cpp:1210] subkey expired 64360858 seconds ago
2019-12-05T14:58:30.1895042Z [signature_check_binding() /home/runner/work/rnp/rnp/src/librepgp/stream-sig.cpp:1210] subkey expired 64360858 seconds ago
2019-12-05T14:58:30.1895369Z [signed_fill_signature() /home/runner/work/rnp/rnp/src/librepgp/stream-write.cpp:1088] wrong secret key password
2019-12-05T14:58:30.1895919Z [signed_dst_finish() /home/runner/work/rnp/rnp/src/librepgp/stream-write.cpp:1146] failed to calculate signature
2019-12-05T14:58:30.1896223Z [process_stream_sequence() /home/runner/work/rnp/rnp/src/librepgp/stream-write.cpp:1737] failed to finish stream
2019-12-05T14:58:30.1935158Z Good signature made Thu Dec  5 14:58:29 2019
2019-12-05T14:58:30.1935513Z using RSA (Encrypt or Sign) key 7bc6709b15c23a4a
2019-12-05T14:58:30.1935752Z 
2019-12-05T14:58:30.1936478Z pub   1024/RSA (Encrypt or Sign) 7bc6709b15c23a4a 2017-07-20 [SC]
2019-12-05T14:58:30.1936591Z       e95a3cbf583aa80a2ccc53aa7bc6709b15c23a4a
[...]

PR #1061, opened 2020-01-31
Rename pgp_content_enum to pgp_pkt_type_t

### LINKED ISSUE #785: Rename pgp_content_enum and it's elements
## Description

Historically it was pgp_content_* since it handled a lot of parser-related things. Now it's all about the packet tags.
So enum should be renamed to `pgp_pkt_tag_t` and constants to `PGP_PKT_SECRET_KEY` and so on.

ni4:
This PR fixes #785 by renaming ages old pgp_content_enum to more applicable pgp_pkt_type_t.
Also does some refactoring by using pgp_pkt_type_t instead of int.

### DIFF
--- a/src/lib/pgp-key.h
+++ b/src/lib/pgp-key.h
@@ -160,7 +160,7 @@ pgp_curve_t pgp_key_get_curve(const pgp_key_t *key);
 
 pgp_version_t pgp_key_get_version(const pgp_key_t *key);
 
-int pgp_key_get_type(const pgp_key_t *key);
+pgp_pkt_type_t pgp_key_get_type(const pgp_key_t *key);
 
 bool pgp_key_is_encrypted(const pgp_key_t *);
 
--- a/src/lib/pgp-key.h
+++ b/src/lib/pgp-key.h
@@ -273,7 +273,7 @@ pgp_subsig_t *pgp_key_get_subsig(const pgp_key_t *, size_t);
 
 void pgp_subsig_free(pgp_subsig_t *subsig);
 
-pgp_rawpacket_t *pgp_key_add_rawpacket(pgp_key_t *, void *, size_t, pgp_content_enum);
+pgp_rawpacket_t *pgp_key_add_rawpacket(pgp_key_t *, void *, size_t, pgp_pkt_type_t);
 
 pgp_rawpacket_t *pgp_key_add_key_rawpacket(pgp_key_t *key, pgp_key_pkt_t *pkt);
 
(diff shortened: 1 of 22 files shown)


src/lib/pgp-key.h
```diff
@@ -271,7 +271,7 @@ pgp_subsig_t *pgp_key_get_subsig(const pgp_key_t *, size_t);
 
 void pgp_subsig_free(pgp_subsig_t *subsig);
 
-pgp_rawpacket_t *pgp_key_add_rawpacket(pgp_key_t *, void *, size_t, pgp_content_enum);
+pgp_rawpacket_t *pgp_key_add_rawpacket(pgp_key_t *, void *, size_t, pgp_pkt_type_t);
```
rrrooommmaaa: why some function declarations have parameter names and some don't? Don't we want to make things uniform?

src/librekey/key_store_g10.cpp
```diff
@@ -1105,7 +1105,7 @@ rnp_key_store_g10_from_src(rnp_key_store_t *         key_store,
 {
     const pgp_key_t *pubkey = NULL;
     pgp_key_t        key = {0};
-    pgp_key_pkt_t    seckey = {0};
+    pgp_key_pkt_t    seckey = {};
```
rrrooommmaaa: Are we moving from `{0}` towards `{}` syntax?

src/lib/pgp-key.h
ni4: @rrrooommmaaa that's how historically it was done in NetPGP, and some new code just follows this rule. For sure code would be more readable with parameter names. However, there is some discussion regarding this:

https://wiki.sei.cmu.edu/confluence/display/c/API08-C.+Avoid+parameter+names+in+a+function+prototype

src/librekey/key_store_g10.cpp
ni4: @rrrooommmaaa this came out after moving from C99 to C++11. Within C code = {0} will fill struct with zeroes, while C++ would assign 0 to first parameter (and in this particular case - throw an error since first parameter is of pgp_pkt_type_t which cannot be assigned from int by default). For most struct's it still works since first parameter is of int or similar type.

rrrooommmaaa:
@ni4 As #1062 is merged, there is a need to update PKT constants in `pgp-key.cpp`

ronaldtse:
@ni4 once it's rebased please help merge it. Thanks!

ni4:
@ronaldtse Done rebase, let's wait for CI.

ni4:
Looks like we hit #994 here. Merging this, however it still needs attention at some point.

Issue #164, opened 2017-06-01
Add Python binding


ronaldtse:
Let's see if anyone is comfortable with doing this 😄

ni4:
I would be happy to do this :-)

PR #1067, opened 2020-02-10
Python binding - the initial version

joke325:
Issue #164
There is implemented the 'py_rnp' Python binding package and Python version of several tests from ffi.cpp


ronaldtse:
@joke325 thank you for this! Do you want to move this PR to https://github.com/rnpgp/py-rnp instead?

dewyatt:
What's the purpose of the "ffi bridge" and what advantages does this have over something simple like the below?

```python
#!/usr/bin/env python

from ctypes import CDLL, POINTER, c_void_p, c_size_t, c_uint32, c_uint64, c_char_p

rnp = CDLL('librnp-0.so')
rnp.rnp_version_string.argtypes = []
rnp.rnp_version_string.restype = c_char_p

print(rnp.rnp_version_string().decode('ascii'))

rnp.rnp_ffi_create.argtypes = [POINTER(POINTER(c_void_p)), c_char_p, c_char_p]
rnp.rnp_ffi_create.restype = c_uint32

rnp.rnp_ffi_destroy.argtypes = [POINTER(c_void_p)]
rnp.rnp_ffi_destroy.restype = c_uint32

rnp.rnp_input_from_path.argtypes = [POINTER(POINTER(c_void_p)), c_char_p]
rnp.rnp_input_from_path.restype = c_uint32

rnp.rnp_input_destroy.argtypes = [POINTER(c_void_p)]
rnp.rnp_input_destroy.restype = c_uint32

rnp.rnp_load_keys.argtypes = [POINTER(c_void_p), c_char_p, POINTER(c_void_p), c_uint32]
rnp.rnp_load_keys.restype = c_uint32

rnp.rnp_get_public_key_count.argtypes = [POINTER(c_void_p), POINTER(c_size_t)]
rnp.rnp_get_public_key_count.restype = c_uint32

ffi = POINTER(c_void_p)()
assert rnp.rnp_ffi_create(ffi, 'GPG'.encode('ascii'), 'GPG'.encode('ascii')) == 0

inp = POINTER(c_void_p)()
assert rnp.rnp_input_from_path(inp, 'key.asc'.encode('ascii')) == 0

count = c_size_t()
assert rnp.rnp_get_public_key_count(ffi, count) == 0
print(count.value)
rnp.rnp_load_keys(ffi, 'GPG'.encode('ascii'), inp, 1)
assert rnp.rnp_get_public_key_count(ffi, count) == 0
print(count.value)

assert rnp.rnp_input_destroy(inp) == 0

assert rnp.rnp_ffi_destroy(ffi) == 0
```

(I'm rusty on ctypes so this could probably be improved)

joke325:
@dewyatt (1) To load static libraries into a process is unfeasible. This shared library makes ffi independent from the build configuration.
(2) Alternatively, your code could look like this: `

    !/usr/bin/env python

    from py_rnp.librnp import Rnp, RNPD

    rnp = Rnp()
    outs = []
        
    print(rnp.rnp_version_string())
        
    assert rnp.rnp_ffi_create(outs, "GPG", "GPG") == 0
    assert rnp.rnp_input_from_path(outs, 'key.asc') == 0
    ffi, inp = outs[-2:]

    assert rnp.rnp_get_public_key_count(ffi, outs) == 0
    count = outs[-1]
    print(count)

    rnp.rnp_load_keys(ffi, "GPG", inp, RNPD.RNP_LOAD_SAVE_PUBLIC_KEYS)
    assert rnp.rnp_get_public_key_count(ffi, outs) == 0
    count = outs[-1]
    print(count)

    assert rnp.rnp_input_destroy(inp) == 0

    assert rnp.rnp_ffi_destroy(ffi) == 0`

joke325:
@ronaldtse The Pyton code can be moved into the dedicated repository.
The C++ changes have to be commited in here.

dewyatt:
> @dewyatt (1) To load static libraries into a process is unfeasible. This shared library makes ffi independent from the build configuration.

Right but it's generally accepted that if a user has not installed the shared library then they simply aren't using the FFI. Duplicating the entire interface (shim or not) into a separate shared object seems like significant overkill IMO. Also if the user has requested `BUILD_SHARED_LIBS=off` and we're building and installing a shared lib anyway, that sounds like a misfeature.

On a side note, this also includes some internal things that are not part of the public FFI.

joke325:
@dewyatt   It's pretty unusual that with static libraries I'm loosing an interface for many non-C languages. I guess a lot of users will by-pass this restriction by building (and installing of course) the both configurations.
The decision is on yours, the py_rnp may be easy refactored to librnp.so (by the way which exports much more symbols than the current implementation of the bridge)
If some symbols in py_rnp are not part of public FFI so let me know please which ones.

dewyatt:
> (by the way which exports much more symbols than the current implementation of the bridge)

I *think* we have an open issue for this but I could be wrong, either way restricting our exports is definitely a TODO item. If applications use any of our private interfaces (generally anything not in the installed headers), it's their own fault when it breaks IMO.

> If some symbols in py_rnp are not part of public FFI so let me know please which ones.

For the most part anything not in the installed headers (`include/rnp/rnp.h` and `include/rnp/rnp_err.h`). Like `SUBDIRECTORY_GNUPG`, etc.

Personally I couldn't approve this as it seems like too much unnecessary boilerplate to maintain for some small benefit and I'd rather see something more like [botan's approach](https://github.com/randombit/botan/tree/master/src/python), but let's see how @ni4 and @ronaldtse feel.

ni4:
@joke325 
I would agree with @dewyatt - 
- we should have single interface to the library, exposed via rnp.h (and some helper headers), like ruby-rnp does now. And from the comments above it should be pretty enough for Python.
- for sure we should not export constants/variables/functions which are out of those headers - probably there is some usage of them in CLI, but that will be removed.

P.S. Btw, since Python 2.x is deprecated now, we should stick to version 3.

joke325:
Reviewers have refused changes so I'm canceling this PR.

ronaldtse:
@joke325 unfortunately sometimes there is work that could be better aligned, but we can resolve this with better communication prior to commencing work. Let's communicate and discuss about approaches before taking on new issues. Thanks guys!

PR #1072, opened 2020-02-15
Fix secret key locking during signature import.

ni4:
Hit this while was working on subkey revocation: unlocked secret key is locked when some signatures/userids/subkeys are attached to it.
This PR fixes this issue.


ronaldtse:
Sorry @ni4 must have merged in the wrong order. Could you help fix the conflicts and merge? Thanks!

ni4:
@ronaldtse No problem, it would need rebase in either way - Git(Hub?) always fail when two PRs appends something to the same file.

PR #1077, opened 2020-02-24
Add FFI function and CLI command to revoke key and subkey.

ni4:
This PR adds the following functionality:
- rnp_key_revoke() FFI function, which allows to revoke key and subkey and all underlying code
- `rnpkeys --revoke-key` command, which uses this function.

Currently we have feature to export key revocation signature, this PR extends functionality to allow to revoke key and subkey directly (without the signature import).

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -3283,6 +3283,51 @@ rnp_key_export_revocation(rnp_key_handle_t key,
 }
 
 rnp_result_t
+rnp_key_revoke(
+  rnp_key_handle_t key, uint32_t flags, const char *hash, const char *code, const char *reason)
+{
+    if (!key || !key->ffi) {
+        return RNP_ERROR_NULL_POINTER;
+    }
+    if (flags) {
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+
+    pgp_key_t *exkey = get_key_prefer_public(key);
+    if (!exkey) {
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+    pgp_key_t *revoker = rnp_key_get_revoker(key);
+    if (!revoker) {
+        FFI_LOG(key->ffi, "Revoker secret key not found");
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+
+    pgp_signature_t *sig = NULL;
+    rnp_result_t     ret =
+      rnp_key_get_revocation(key->ffi, exkey, revoker, hash, code, reason, &sig);
+    if (ret) {
+        return ret;
+    }
+    pgp_sig_import_status_t pub_status = PGP_SIG_IMPORT_STATUS_UNKNOWN_KEY;
+    pgp_sig_import_status_t sec_status = PGP_SIG_IMPORT_STATUS_UNKNOWN_KEY;
+    if (key->pub) {
+        pub_status = rnp_key_store_import_key_signature(key->ffi->pubring, key->pub, sig);
+    }
+    if (key->sec) {
+        sec_status = rnp_key_store_import_key_signature(key->ffi->secring, key->sec, sig);
+    }
+    free_signature(sig);
+    free(sig);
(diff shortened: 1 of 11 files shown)


src/librekey/rnp_key_store.cpp
```diff
@@ -588,43 +588,97 @@ rnp_key_store_get_signer_key(rnp_key_store_t *store, const pgp_signature_t *sig)
...
+
+    pgp_key_t tmpkey = {};
+    if (!pgp_key_from_pkt(&tmpkey, &key->pkt) || !rnp_key_add_signature(&tmpkey, sig)) {
+        RNP_LOG("Failed to add signature to the key.");
+        pgp_key_free_data(&tmpkey);
+        return PGP_SIG_IMPORT_STATUS_UNKNOWN;
+    }
+
+    size_t expackets = pgp_key_get_rawpacket_count(key);
+    if (!(key = rnp_key_store_add_key(keyring, &tmpkey))) {
```
rrrooommmaaa: @ni4 I noticed that `rnp_key_store_add_key` actually has no effect (doesn't update) on `srckey->format == PGP_KEY_STORE_G10`. So I guess unit tests will fail with this keystore format. Is this behaviour desired?

ni4: @rrrooommmaaa pgp_key_from_pkt() sets format to PGP_KEY_STORE_GPG, so it will not fail in this case.

src/tests/cli_tests.py
```diff
@@ -1200,6 +1200,90 @@ def test_userid_escape(self):
...
+        ret, out, err = run_proc(RNPK, ['--homedir', RNPDIR, '--revoke-key', 'alice'])
+        if (ret == 0) or (len(out) > 0) or not re.match(r'(?s)^.*Revoker secret key not found.*Failed to revoke a key.*', err):
+            raise_err('Wrong revocation output')
+        # Import Alice's secret key and subkey
+        ret, _, _ = run_proc(RNPK, ['--homedir', RNPDIR, '--import', data_path('test_key_validity/alice-sub-sec.pgp')])
+        if ret != 0:
+            raise_err('Alice secret key import failed')
+        # Attempt to revoke without specifying a key
+        pipe = pswd_pipe(PASSWORD)
+        ret, out, err = run_proc(RNPK, ['--homedir', RNPDIR, '--revoke', '--pass-fd', str(pipe)])
```
rrrooommmaaa: @ni4 Is "revoke" a synonym for "revoke-key"? I can't find it in the map, nor in the usage text.

ni4: @rrrooommmaaa getopt_long() accepts abbreviation of the long command instead of the whole one, i.e. `--revok` will work as well. This doesn't seem to be documented in manual pages but can be seen from practice or glibc sources :)
Anyway, this line most likely will be subject to update once we have --revoke-cert/--revoke-userid.

ni4:
@dewyatt @ronaldtse Ping for review.

dewyatt:
Sorry, apparently I left this partially reviewed and forgot about it. Taking a look...

src/lib/rnp.cpp
```diff
@@ -3282,6 +3282,51 @@ rnp_key_export_revocation(rnp_key_handle_t key,
...
 
+rnp_result_t
+rnp_key_revoke(
+  rnp_key_handle_t key, uint32_t flags, const char *hash, const char *code, const char *reason)
+{
+    if (!key || !key->ffi) {
+        return RNP_ERROR_NULL_POINTER;
+    }
+    if (flags) {
+        return RNP_ERROR_BAD_PARAMETERS;
```
dewyatt, line 3293: I almost feel like we should be using `RNP_ERROR_NOT_IMPLEMENTED` for this case, but the cat is pretty much out of the bag on that since we do it this way elsewhere.

ronaldtse:
Thanks guys!

PR #1079, opened 2020-02-26
Add FFI function rnp_request_password().

### LINKED ISSUE #1035: Add memory cleaning function to FFI.
## Description

Currently CLI uses pgp_forget() function call to clean password data, which directly calls function from Botan.
However, CLI should be fully migrated to FFI usage (so CLI layer is not dependent on rnp library internals and underlying cryptography).

### LINKED ISSUE #1036: Add rnp_request_password() function to the FFI.
## Description

This function should request password via configured for FFI callback.
Please see the discussion: https://github.com/rnpgp/rnp/pull/1033#discussion_r366306832

ni4:
This PR implements function `rnp_request_password()` which requests password via configured FFI callback in case user needs it. Fixes #1036.
Also it adds function `rnp_buffer_clear()` to securely clear buffer (instead of pgp_forget() which belongs to lower layer).
Fixes #1035.

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -4697,6 +4697,14 @@ rnp_buffer_destroy(void *ptr)
     free(ptr);
 }
 
+void
+rnp_buffer_clear(void *ptr, size_t size)
+{
+    if (ptr) {
+        pgp_forget(ptr, size);
+    }
+}
+
 static pgp_key_t *
 get_key_require_public(rnp_key_handle_t handle)
 {
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -963,6 +963,29 @@ done:
     return ret;
 }
 
+rnp_result_t
+rnp_request_password(rnp_ffi_t ffi, rnp_key_handle_t key, const char *context, char **password)
+{
+    if (!ffi || !password || !ffi->getpasscb) {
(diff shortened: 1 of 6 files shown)


src/lib/rnp.cpp
```diff
@@ -963,6 +963,29 @@ rnp_supported_features(const char *type, char **result)
...
+{
+    if (!ffi || !password || !ffi->getpasscb) {
+        return RNP_ERROR_NULL_POINTER;
+    }
+
+    Botan::secure_vector<char> pass(MAX_PASSWORD_LENGTH, '\0');
+    bool                       req_res =
+      ffi->getpasscb(ffi, ffi->getpasscb_ctx, key, context, pass.data(), pass.size());
+    size_t pass_len = strlen(pass.data());
+    if (!req_res || !strlen(pass.data())) {
```
dewyatt: Can we just use `pass_len` here instead of calling `strlen` again?

ni4: @dewyatt Sure, somehow missed this. Please see the updated PR.

ni4:
Hm, looks like issue #994 is not that rare, let's restart the tests and see whether it happens again.

ni4:
...okay, PR #1084 should fix it.
@dewyatt Is it okay to merge now (given that test passes)?

ronaldtse:
Ping pending second approval to merge.

PR #1080, opened 2020-03-01
Added cli_rnp_escape_string() function to escape user strings

### LINKED ISSUE #1078: Escape non-printable characters in user output
## Description

GPG --list-keys escapes special characters in its output (note tab(\x09) and newline(\n):
`uid           [ unknown] }{][)^*.+(\x09\n|$@rnp`

Our current output is ugly:
```
uid           }{][)^*.+(
|$@rnp
```

We should escape non-printable chars (0x00-0x1F) in user output. It can be done as a separate function like cli_rnp_user_msg(), so we can later on use it in other places. (as from discussion #1014)

## Steps to Reproduce

Create a key with userid containing tab (0x09), newline (0x0A).

## Expected Behavior

The output should show \x09 and \n
## Actual Behavior

The output is actually instead of tabbed and line-fed

rrrooommmaaa:
Closes #1078


src/rnp/fficli.cpp
```diff
@@ -668,6 +668,29 @@ cli_key_usage_str(rnp_key_handle_t key, char *buf)
     return orig;
 }
 
+std::string
+cli_rnp_escape_string(const char *src)
```
ni4: As we are (rather slowly) moving to the C++, new code (especially on CLI level) should be as much as possible in C++, i.e. here please use std::string as parameter. Not sure about the escape_map.

PR #1083, opened 2020-03-09
Remove ssize_t usage.

### LINKED ISSUE #898: ssize_t isn't defined on Windows
The "signed size_t" type ssize_t is a POSIX type and is unavailable on Windows.

Although cross compilation worked, you're also using ssize_t in the public rnp.h header. You should change that to a portable size type.

ni4:
This PR removes ssize_t (which is not well available in Windows) from the whole library.
While internal usage was mostly around src_read() function and is pretty safe, it make a breaking change for FFI interface, so ruby-rnp side should be updated.

Fixes #898

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -171,9 +171,42 @@ typedef struct rnp_uid_handle_st *         rnp_uid_handle_t;
 typedef struct rnp_signature_handle_st *   rnp_signature_handle_t;
 
 /* Callbacks */
+/**
+ * @brief Callback, used to read data from the source.
+ *
+ * @param app_ctx custom parameter, passed back to the function.
+ * @param buf on successfull call data should be put here. Cannot be NULL,
+ *            and must be capable to store at least len bytes.
+ * @param len number of bytes to read.
+ * @param read on successfull call number of read bytes must be put here.
+ * @return true on success (including EOF condition), or false on read error.
+ *         EOF case is indicated by zero bytes read on non-zero read call.
+ */
 typedef bool rnp_input_reader_t(void *app_ctx, void *buf, size_t len, size_t *read);
+/**
+ * @brief Callback, used to close input stream.
+ *
+ * @param app_ctx custom parameter, passed back to the function.
+ * @return void
+ */
 typedef void rnp_input_closer_t(void *app_ctx);
+/**
+ * @brief Callback, used to write data to the output stream.
+ *
+ * @param app_ctx custom parameter, passed back to the function.
+ * @param buf buffer with data, cannot be NULL.
+ * @param len number of bytes to write.
+ * @return true if call was successfull and all data is written, or false otherwise.
+ */
 typedef bool rnp_output_writer_t(void *app_ctx, const void *buf, size_t len);
+
+/**
+ * @brief Callback, used to close output stream.
+ *
+ * @param app_ctx custom parameter, passed back to the function.
+ * @param discard true if the already written data should be deleted.
+ * @return void
+ */
 typedef void rnp_output_closer_t(void *app_ctx, bool discard);
 
 /**
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -171,10 +171,10 @@ typedef struct rnp_uid_handle_st *         rnp_uid_handle_t;
 typedef struct rnp_signature_handle_st *   rnp_signature_handle_t;
 
 /* Callbacks */
-typedef ssize_t rnp_input_reader_t(void *app_ctx, void *buf, size_t len);
-typedef void    rnp_input_closer_t(void *app_ctx);
-typedef bool    rnp_output_writer_t(void *app_ctx, const void *buf, size_t len);
-typedef void    rnp_output_closer_t(void *app_ctx, bool discard);
+typedef bool rnp_input_reader_t(void *app_ctx, void *buf, size_t len, size_t *read);
+typedef void rnp_input_closer_t(void *app_ctx);
+typedef bool rnp_output_writer_t(void *app_ctx, const void *buf, size_t len);
+typedef void rnp_output_closer_t(void *app_ctx, bool discard);
 
 /**
  * Callback used for getting a password.
--- a/src/examples/dump.c
+++ b/src/examples/dump.c
@@ -47,10 +47,15 @@ print_usage(char *program_name)
             basename(program_name));
 }
 
-static ssize_t
-stdin_reader(void *app_ctx, void *buf, size_t len)
+static bool
+stdin_reader(void *app_ctx, void *buf, size_t len, size_t *readres)
 {
-    return read(STDIN_FILENO, buf, len);
+    ssize_t res = read(STDIN_FILENO, buf, len);
+    if (res < 0) {
+        return false;
+    }
+    *readres = res;
+    return true;
 }
 
 static bool
--- a/src/librepgp/stream-common.cpp
+++ b/src/librepgp/stream-common.cpp
@@ -44,16 +44,17 @@
 #include "types.h"
 #include <algorithm>
 
-ssize_t
-src_read(pgp_source_t *src, void *buf, size_t len)
+bool
+src_read(pgp_source_t *src, void *buf, size_t len, size_t *readres)
 {
     size_t              left = len;
-    ssize_t             read;
+    size_t              read;
     pgp_source_cache_t *cache = src->cache;
     bool                readahead = cache ? cache->readahead : false;
 
     if (src->eof || (len == 0)) {
-        return 0;
+        *readres = 0;
+        return true;
     }
 
     // Do not read more then available if source size is known
@@ -66,7 +67,7 @@ src_read(pgp_source_t *src, void *buf, size_t len)
     // Check whether we have cache and there is data inside
     if (cache && (cache->len > cache->pos)) {
         read = cache->len - cache->pos;
-        if ((size_t) read >= len) {
+        if (read >= len) {
             memcpy(buf, &cache->buf[cache->pos], len);
             cache->pos += len;
             goto finish;
@@ -82,29 +83,28 @@ src_read(pgp_source_t *src, void *buf, size_t len)
     while (left > 0) {
         if (left > sizeof(cache->buf) || !readahead || !cache) {
             // If there is no cache or chunk is larger then read directly
-            read = src->read(src, buf, left);
-            if (read > 0) {
-                left -= read;
-                buf = (uint8_t *) buf + read;
-            } else if (read == 0) {
+            if (!src->read(src, buf, left, &read)) {
+                src->error = 1;
+                return false;
+            }
+            if (!read) {
                 src->eof = 1;
                 len = len - left;
                 goto finish;
-            } else {
-                src->error = 1;
-                return -1;
             }
+            left -= read;
+            buf = (uint8_t *) buf + read;
         } else {
             // Try to fill the cache to avoid small reads
-            read = src->read(src, &cache->buf[0], sizeof(cache->buf));
-            if (read == 0) {
+            if (!src->read(src, &cache->buf[0], sizeof(cache->buf), &read)) {
+                src->error = 1;
+                return false;
+            }
+            if (!read) {
                 src->eof = 1;
                 len = len - left;
                 goto finish;
-            } else if (read < 0) {
-                src->error = 1;
-                return -1;
-            } else if ((size_t) read < left) {
+            } else if (read < left) {
                 memcpy(buf, &cache->buf[0], read);
                 left -= read;
                 buf = (uint8_t *) buf + read;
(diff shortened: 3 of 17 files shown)


dewyatt:
Taking a second look at this I'm wondering if we should take a different approach. I don't like that we're potentially truncating `ssize_t` to `int` on POSIX systems.

ni4:
@dewyatt I also have some doubts about it (got hit with 32-bit int for packet lengths).
While packet length-reading code may be changed to `bool func(..., size_t *)`, it would be too much changes for src_read(). Also, since those things are about reading data and putting it to the memory, int should be enough at all.

dewyatt:
I think for MSVC (which is what thunderbird is probably building with), we could use:
```cpp
#if defined(_MSC_VER)
#include <BaseTsd.h>
typedef SSIZE_T ssize_t;
#endif
```

There's also the option of using `ptrdiff_t`/`intmax_t` but I'm not sure on ruby-ffi and python ctypes support there.

ni4:
@dewyatt Should we stick to ssize_t at all? It seems to be defined in a bit weird way to be able to return -1.
From another side - most modern systems define int as 32-bit at least, so it would be pretty enough for reads/writes (hope we'll not laugh on this message 20 years later).

dewyatt:
> @dewyatt Should we stick to ssize_t at all? It seems to be defined in a bit weird way to be able to return -1.
> From another side - most modern systems define int as 32-bit at least, so it would be pretty enough for reads/writes (hope we'll not laugh on this message 20 years later).

The problem I feel with using `int` is that `ssize_t read()` (for `file_src_read`) may return a positive value that does not fit in `int` on a given platform.

We could also change the interfaces to return a `bool` and take a `size_t *` param.

ni4:
> The problem I feel with using int is that ssize_t read() (for file_src_read) may return a positive value that does not fit in int on a given platform.

@dewyatt  src_read() (and `file_src_read()`) is mostly called internally and do not exceed 32K at the moment. It's not a 100% strong argument but on 99% cases this should work.
It would need some changes for src_read() to make it work as `bool/size_t *`, but it's one-time and should not be problematic.

We'd also need to change `rnp_input_reader_t`, introducing even more breaking change to the FFI, is it good to go?

dewyatt:
> @dewyatt src_read() (and `file_src_read()`) is mostly called internally and do not exceed 32K at the moment. It's not a 100% strong argument but on 99% cases this should work.

I understand that but if we're going to fix it we may as well do it right and it seems like maybe `bool/size_t*` is a decent approach IMO.

> It would need some changes for src_read() to make it work as `bool/size_t *`, but it's one-time and should not be problematic.
> 
> We'd also need to change `rnp_input_reader_t`, introducing even more breaking change to the FFI, is it good to go?

I think that may be the best route to take. The only real downstream use we have is ruby-rnp and we should be able to do a version check for this.

ni4:
@dewyatt Okay, let's then stick to `bool/*size_t` strategy. I will update the PR with necessary changes once ready, or drop a message if found something new against it.

src/examples/dump.c
```diff
@@ -47,10 +47,15 @@ print_usage(char *program_name)
             basename(program_name));
 }
 
-static ssize_t
-stdin_reader(void *app_ctx, void *buf, size_t len)
+static bool
+stdin_reader(void *app_ctx, void *buf, size_t len, size_t *readres)
 {
-    return read(STDIN_FILENO, buf, len);
+    int64_t res = read(STDIN_FILENO, buf, len);
```
dewyatt: I think we should use `ssize_t` here since that is what `read` returns according to POSIX.

src/rnp/fficli.cpp
```diff
@@ -1547,10 +1547,15 @@ cli_cfg_set_keystore_info(rnp_cfg_t *cfg)
     return true;
 }
 
-static ssize_t
-stdin_reader(void *app_ctx, void *buf, size_t len)
+static bool
+stdin_reader(void *app_ctx, void *buf, size_t len, size_t *readres)
 {
-    return read(STDIN_FILENO, buf, len);
+    int64_t res = read(STDIN_FILENO, buf, len);
```
dewyatt: Same here, `ssize_t`.

src/examples/dump.c
ni4: @dewyatt I was also thinking about it. But what will happen on system without ssize_t ?

src/rnp/fficli.cpp
ni4: @dewyatt Ok. I'm not finished with this PR yet (changing interface to bool/size_t*), will let you know once done. Probably will make some force-pushes before that.

dewyatt:
I'll need to get ruby-rnp updated to get CI passing.

I'm wondering, if the thunderbird build environment is using MSVC, is `read()` even available? Or, after fixing this missing `ssize_t` in the public interface, are we going to be encountering a multitude of other issues because they are building w/MSVC and not msys/etc?

ni4:
@dewyatt As I read just now in the mailing list, `Note we're using the javascript-ctypes interface to access the RNP C API. We don't link to it, instead we open the shared libraries at runtime.`.
So as far as I understand they build libraries with MinGW, and access FFI via JS. So most likely single showstopper is ssize_t in public API.

But, anyway, let's fix internals to size_t as well (I'm almost done with that) - it would be much less places with ssize_t variables and calls.

In Windows there is a native `_read()` call, which could be wrapped in simple define (as it is most likely done in MSYS/MinGW, including define for ssize_t as well).

ni4:
@dewyatt I finished with this PR, so feel free to check it out/update ruby-rnp once you have time.

include/rnp/rnp.h
```diff
@@ -171,10 +171,10 @@ typedef struct rnp_uid_handle_st *         rnp_uid_handle_t;
 typedef struct rnp_signature_handle_st *   rnp_signature_handle_t;
 
 /* Callbacks */
-typedef ssize_t rnp_input_reader_t(void *app_ctx, void *buf, size_t len);
-typedef void    rnp_input_closer_t(void *app_ctx);
-typedef bool    rnp_output_writer_t(void *app_ctx, const void *buf, size_t len);
-typedef void    rnp_output_closer_t(void *app_ctx, bool discard);
+typedef bool rnp_input_reader_t(void *app_ctx, void *buf, size_t len, size_t *read);
```
dewyatt: Can we add some docs for this too? I think my understanding is that the reader should always return true unless an error occurs (so EOF would be indicated by returning true but setting *read=0). Let me know if I'm mistaken though.

ni4: @dewyatt Agree, done. However now we seems to have problem with GTest...

dewyatt:
So I think the easiest way to do this is to ignore CI for now and merge this. After that I can push the update I have for ruby-rnp and re-run CI.

ni4:
@dewyatt Okay. Let's wait for the second approval? Ping @ronaldtse @rrrooommmaaa

src/librepgp/stream-common.cpp
```diff
@@ -173,62 +172,64 @@ src_peek(pgp_source_t *src, void *buf, size_t len)
...
     if (len < sizeof(sbuf)) {
-        return src_read(src, sbuf, len);
+        (void) src_read(src, sbuf, len, &res);
+        return;
     }
 
-    buf = calloc(1, std::min((size_t) PGP_INPUT_CACHE_SIZE, len));
+    void *buf = calloc(1, std::min((size_t) PGP_INPUT_CACHE_SIZE, len));
     if (!buf) {
-        return -1;
```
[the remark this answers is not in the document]
ni4: @rrrooommmaaa (here and below) src->error is managed in src_read() function, so independent on callback error will be set if callback returns it. Another point - this PR wasn't aimed to change the src->error handling, i.e. anyway it should not change it.

src/librepgp/stream-packet.cpp
```diff
@@ -91,139 +91,178 @@ get_packet_type(uint8_t ptag)
...
 
 bool
 stream_intedeterminate_pkt_len(pgp_source_t *src)
 {
     uint8_t ptag = 0;
-    if (src_peek(src, &ptag, 1) == 1) {
-        return !(ptag & PGP_PTAG_NEW_FORMAT) &&
-               ((ptag & PGP_PTAG_OF_LENGTH_TYPE_MASK) == PGP_PTAG_OLD_LEN_INDETERMINATE);
+    size_t  read = 0;
+    if (!src_peek(src, &ptag, 1, &read) || (read != 1)) {
```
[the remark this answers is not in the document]
ni4: @rrrooommmaaa Thanks, somehow missed this. Will push soon.

src/librepgp/stream-common.cpp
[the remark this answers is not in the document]
rrrooommmaaa: @ni4 This piece seems important to me. The previous version returned error code (-1) on memory allocation fault, this version doesn't return any error code, just exits prematurely (before moving the position of src, right?). Hence, we have indeterminate state after calling this function -- the position may or may not have moved. My solution is to put src in error state (at least), so the next read will be guaranteed to fail.

src/tests/streams.cpp
```diff
@@ -234,11 +233,15 @@ TEST_F(rnp_tests, test_stream_file)
...
-        assert_int_equal(src_read(&src, tmpbuf, filedatalen), filedatalen);
+        size_t read = 0;
+        assert_true(src_read(&src, tmpbuf, filedatalen, &read));
+        assert_int_equal(read, filedatalen);
         assert_int_equal(memcmp(tmpbuf, filedata, filedatalen), 0);
     }
     for (int i = 0; i < 5 * iterations; i++) {
-        assert_int_equal(src_read(&src, tmpbuf, 3), 3);
+        size_t read = 0;
+        assert_true(src_read(&src, tmpbuf, 3, &read));
```
rrrooommmaaa: Simplify by using `src_read_eq`?

src/tests/streams.cpp
```diff
@@ -234,11 +233,15 @@ TEST_F(rnp_tests, test_stream_file)
     /* read file back, checking the contents */
     assert_rnp_success(init_file_src(&src, filename));
     for (int i = 0; i < iterations; i++) {
-        assert_int_equal(src_read(&src, tmpbuf, filedatalen), filedatalen);
+        size_t read = 0;
+        assert_true(src_read(&src, tmpbuf, filedatalen, &read));
```
rrrooommmaaa: Simplify by using `src_read_eq`?

src/librepgp/stream-common.cpp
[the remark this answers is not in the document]
ni4: @rrrooommmaaa Yeah, it returned -1 which was never checked. So, basically, it would behave in the same way as before. From the 'strict logical' perspective memory allocation failure doesn't relate to the read failure, so it is not 100% correct to set error to 1 here.
And, anyway, on allocation failure all the further code will completely fail eventually.

src/tests/streams.cpp
ni4: @rrrooommmaaa I thought whether to replace it with src_read_eq(). While it doesn't make major difference, these two lines will test function src_read() and two cases (return true and read == 3), while src_read_eq() will test only this function and i'ts result.

src/librepgp/stream-common.cpp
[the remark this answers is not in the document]
rrrooommmaaa: @ni4 How about using a static buffer for skipping? If allocating 32K is too much for a static buffer, how about using a small fallback buffer in case the big buffer can't be alocated. Thus we can guarantee, that the src position will be skipped in any case.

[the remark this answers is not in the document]
rrrooommmaaa: ... and `src_read` and `src_peek` should exit with error immediately if `src->error` is set. Or would you recommend to do this as a separate PR?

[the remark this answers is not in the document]
ni4: @rrrooommmaaa I think it should be filed as a separate issue, since doesn't relate to this PR, and doesn't seem to be of high criticality. src->error is set to just be able to check it later on via src_error() function.

ronaldtse:
Are we waiting for @dewyatt to update ruby-rnp to pass the tests here?

dewyatt:
> Are we waiting for @dewyatt to update ruby-rnp to pass the tests here?

No, the plan is to merge this with CI failing, then I do the update to ruby-rnp when I have a reliable commit timestamp to reference (which I don't think I will until this is merged). There are other ways to do things but that's the way I was planning for.

ni4:
So we need just the second approval? @rrrooommmaaa do you have any other comments or this may be merged?

rrrooommmaaa:
@ni4 You haven't replied anything on the idea of having a small fallback buffer in case memory allocation fails.
If you insist, I can approve this and immediately file 2 new issues

ni4:
@rrrooommmaaa Actually I did:
> I think it should be filed as a separate issue, since doesn't relate to this PR, and doesn't seem to be of high criticality. src->error is set to just be able to check it later on via src_error() function.

Actually comments on this PR became a bit messy.
So please file separate issue(s), as we need to move forward on this.

Issue #1004, opened 2019-12-13
Add feature to update expiration date of the private key


ni4:
## Description

Currently it is not possible to update expiration date of the key/subkey. Checklist for the feature implementation:

 - [x] library code
 - [x] FFI
 - [x] FFI test

ni4:
Extracted CLI part to the separate issue #1132, so closing this.

PR #1094, opened 2020-04-07
Add FFI function to update key expiration time.

ni4:
This PR adds function `rnp_key_set_expiration()`, which allows to change key or subkey expiration time, as stated in issue #1004.
It doesn't close it since CLI part is not implemented yet (will be done later in separate PR).


src/lib/pgp-key.cpp
```diff
@@ -1023,6 +1043,94 @@ pgp_subsig_from_signature(pgp_subsig_t *subsig, const pgp_signature_t *sig)
...
+pgp_key_replace_signature(pgp_key_t *key, pgp_signature_t *oldsig, pgp_signature_t *newsig)
+{
+    pgp_subsig_t *subsig = NULL;
+    for (size_t i = 0; i < pgp_key_get_subsig_count(key); i++) {
+        subsig = pgp_key_get_subsig(key, i);
+        if (signature_pkt_equal(&subsig->sig, oldsig)) {
+            break;
+        }
+    }
+    if (!subsig) {
```
rrrooommmaaa, line 1095: Non-null `subsig` here doesn't mean that `signature_pkt_equal` matched. It may be a leftover from the last iteration.

ni4: @rrrooommmaaa Thanks for finding this out, will fix it.

ni4: @rrrooommmaaa Should be good now.

src/lib/pgp-key.cpp
```diff
@@ -1850,6 +1959,131 @@ pgp_key_add_userid_certified(pgp_key_t *              key,
...
+}
+
+bool
+pgp_subkey_set_expiration(pgp_key_t *sub,
+                          pgp_key_t *primsec,
+                          pgp_key_t *secsub,
+                          uint32_t   expiry)
+{
+    if (!pgp_key_is_subkey(sub)) {
+        RNP_LOG("Not a primary key");
```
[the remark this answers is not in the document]
ni4: @rrrooommmaaa Thanks, fixed.

ni4:
@rrrooommmaaa Thanks, but please next time do Rebase and Merge - this leaves cleaner history on master. The same should be done when working on own PRs - instead of merging in changes from master, rebase on master should be done.
cc @joke325

ronaldtse:
Next time please do rebase & merge -- cleaner history as mentioned by @ni4 . Thanks!

PR #1095, opened 2020-04-10
Correct key fields calculation and sigs validation.

### LINKED ISSUE #1020: Incorrect key flags check in transferable_subkey_bind()
## Description

As per RFC 4880 bis 8, if subkey is able to sign then primary key binding signature should be added as the subpacket to the subkey binding signature.
We doing this check as the following:
```
    realkf = (pgp_key_flags_t) binding->key_flags;
    if (!realkf) {
        realkf = pgp_pk_alg_capabilities(key->alg);
    }
```
Obviously, we should use `pgp_pk_alg_capabilities(subkey->alg)` here.

While it's easy to fix the code, the test which reproduces this case on level of function `pgp_generate_subkey` should be added.
Checks should be done for both set/unset `binding->key_flags` and with both sign-able/sign-unable subkey.

### LINKED ISSUE #1089: RNP and GPG report different key expiration times
## Description

For a key that was extended, RNP reports that the key has expired at an earlier time - not at the most recent expiration time.

## Steps to Reproduce

1. Import key 698e2885c1de74e32cd503ad295c746984af7f0c with rnpkeys
2. rnp --list 698e2885c1de74e32cd503ad295c746984af7f0c |grep EXPIRED

## Expected Behavior

GPG reports expiration date: 2020-03-17

## Actual Behavior

RNP reports expiration date 2018-04-12

### LINKED ISSUE #1092: test_ffi_signatures_dump test failure
## Description

I'm guessing a key expired

```
[ RUN      ] rnp_tests.test_ffi_signatures_dump
[signature_check_certification() /Users/runner/runners/2.165.2/work/rnp/rnp/src/librepgp/stream-sig.cpp:1203] key expired 532049 seconds ago
```

ni4:
This is a bit wide PR is part of the work on ability to update key expiration (as well as other fields).
Main feature here is moving from old incorrect way of key fields calculation (previously we just set key fields once new signature is added, without checking sigs validity and creation date).
Now it behave correctly by using data only from valid self-signatures, using the latest one as the source of key data (as key expiration, preferred algorithms and so on).

As a side effect of the changes different not so clearly related things were fixed here:
- issue #1092 (expired key)
- issue #1089 (wrong key expiration time)
- incorrect generation of subkey revocation signature
- issue #1020 (wrong subkey flags check)

Normally I would issue separate PRs for all these things, but now I'm a bit on rush with Thunderbird features implementation so allowed myself to put all these things into the single PR.

Fixes #1092, fixes #1089, fixes #1020.

### DIFF
--- a/src/lib/pgp-key.cpp
+++ b/src/lib/pgp-key.cpp
@@ -961,6 +961,125 @@ pgp_key_get_subsig(const pgp_key_t *key, size_t idx)
     return (pgp_subsig_t *) list_at(key->subsigs, idx);
 }
 
+static bool
+pgp_sig_is_certification(const pgp_subsig_t *sig)
+{
+    pgp_sig_type_t type = signature_get_type(&sig->sig);
+    return (type == PGP_CERT_CASUAL) || (type == PGP_CERT_GENERIC) ||
+           (type == PGP_CERT_PERSONA) || (type == PGP_CERT_POSITIVE);
+}
+
+static bool
+pgp_sig_self_signed(const pgp_key_t *key, const pgp_subsig_t *sig)
+{
+    /* if we have fingerprint let's check it */
+    if (signature_has_keyfp(&sig->sig)) {
+        pgp_fingerprint_t sigfp = {};
+        if (signature_get_keyfp(&sig->sig, &sigfp)) {
+            return fingerprint_equal(pgp_key_get_fp(key), &sigfp);
+        }
+    }
+    if (!signature_has_keyid(&sig->sig)) {
+        return false;
+    }
+    uint8_t sigid[PGP_KEY_ID_SIZE] = {0};
+    if (!signature_get_keyid(&sig->sig, sigid)) {
+        return false;
+    }
+    return !memcmp(pgp_key_get_keyid(key), sigid, PGP_KEY_ID_SIZE);
+}
+
+static bool
+pgp_sig_is_self_signature(const pgp_key_t *key, const pgp_subsig_t *sig)
+{
+    if (!pgp_key_is_primary_key(key) || !pgp_sig_is_certification(sig)) {
+        return false;
+    }
+
+    return pgp_sig_self_signed(key, sig);
+}
+
+static bool
+pgp_sig_is_direct_self_signature(const pgp_key_t *key, const pgp_subsig_t *sig)
+{
+    if (!pgp_key_is_primary_key(key) || (signature_get_type(&sig->sig) != PGP_SIG_DIRECT)) {
+        return false;
+    }
+
+    return pgp_sig_self_signed(key, sig);
+}
+
+static bool
+pgp_sig_is_key_revocation(const pgp_key_t *key, const pgp_subsig_t *sig)
+{
+    return pgp_key_is_primary_key(key) && (signature_get_type(&sig->sig) == PGP_SIG_REV_KEY);
+}
+
+static bool
+pgp_sig_is_subkey_binding(const pgp_key_t *key, const pgp_subsig_t *sig)
+{
+    return pgp_key_is_subkey(key) && (signature_get_type(&sig->sig) == PGP_SIG_SUBKEY);
+}
+
+static bool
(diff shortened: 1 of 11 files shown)


src/librepgp/stream-key.cpp
```diff
@@ -353,6 +353,54 @@ transferable_key_add_userid(pgp_transferable_key_t *key, const char *userid)
...
+                                  pgp_signature_t *       sig,
+                                  const pgp_key_pkt_t *   signer)
+{
+    if (!key || !uid || !sig || !signer) {
+        RNP_LOG("NULL parameter(s)");
+        return false;
+    }
+
+    rng_t rng = {};
+    if (!rng_init(&rng, RNG_SYSTEM)) {
```
dewyatt, line 368: The plan has been to drop `RNG_SYSTEM` at some point, so I'd rather see if we can pass the `rng_t` down through the various layers instead.

src/librepgp/stream-key.cpp
```diff
@@ -353,6 +353,54 @@ transferable_key_add_userid(pgp_transferable_key_t *key, const char *userid)
...
+                           pgp_signature_t *    sig,
+                           const pgp_key_pkt_t *signer)
+{
+    if (!key || !sig || !signer) {
+        RNP_LOG("NULL parameter(s)");
+        return false;
+    }
+
+    rng_t rng = {};
+    if (!rng_init(&rng, RNG_SYSTEM)) {
```
dewyatt, line 392: Same here

src/lib/pgp-key.cpp
```diff
@@ -961,6 +969,373 @@ pgp_key_get_subsig(const pgp_key_t *key, size_t idx)
...
+    /* validate self-signatures if not done yet */
+    if (key) {
+        pgp_subkey_validate_self_signatures(sub, key);
+    }
+    pgp_subsig_t *sig = pgp_key_latest_binding(sub, key);
+    /* subkey expiration */
+    sub->expiration = sig ? signature_get_key_expiration(&sig->sig) : 0;
+    /* subkey flags */
+    if (sig && signature_has_key_flags(&sig->sig)) {
+        sub->key_flags = sig->key_flags;
```
dewyatt, line 1252: It seems like this would leave an incorrect value here in the case where a `PGP_SIG_SUBPKT_KEY_FLAGS` subpacket was previously present and has now been removed, wouldn't it?

src/lib/pgp-key.cpp
```diff
@@ -961,6 +969,373 @@ pgp_key_get_subsig(const pgp_key_t *key, size_t idx)
...
+    }
+    if (signature_has_key_server_prefs(&subsig->sig)) {
+        uint8_t ks_pref = signature_get_key_server_prefs(&subsig->sig);
+        if (!pgp_user_prefs_set_ks_prefs(&subsig->prefs, &ks_pref, 1)) {
+            RNP_LOG("failed to alloc ks prefs");
+            goto error;
+        }
+    }
+    if (signature_has_key_server(&subsig->sig)) {
+        subsig->prefs.key_server = (uint8_t *) signature_get_key_server(&subsig->sig);
```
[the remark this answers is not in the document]
ni4: Thanks, fixed this.

src/lib/pgp-key.cpp
ni4: Agree, thanks for spotting this. Fixed.

src/librepgp/stream-key.cpp
ni4: @dewyatt What about moving this to the separate PR? Since it would require changing a lot of internal function declarations, as I think we don't want to make it global static variable.

dewyatt: @ni4 Sounds good to me

ni4:
@ronaldtse @rrrooommmaaa Could you review this please, so we can move forward?

ni4:
@ronaldtse Thanks! So merging this, considering that Windows fails because of GHA.

ronaldtse:
Thanks @ni4 !

PR #1096, opened 2020-04-11
Install automake for macos GHA.

ni4:
Looks like something changed in GHA macos runner, preventing correct installation of json-c. Attempting to fix it.

### DIFF
--- a/ci/before_install.sh
+++ b/ci/before_install.sh
@@ -4,7 +4,7 @@ set -ex
 . ci/utils.inc.sh
 
 macos_install() {
-  [ "${CI-}" = true ] || brew bundle
+  brew bundle
 }
 
 freebsd_install() {
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -12,6 +12,8 @@ task:
     LOCAL_BUILDS: /home/rnpuser/local-builds
     LOCAL_INSTALLS: /home/rnpuser/local-installs
   user_setup_script: |
+    # workaround 'pkg: repository meta has wrong version 2'
+    env ASSUME_ALWAYS_YES=yes pkg bootstrap -f
     pkg install -y sudo
     # add rnpuser
     pw useradd -n rnpuser -m


ni4:
@dewyatt Is this a correct way and place in script to install automake? And, still have issues with ruby-rnp.

ronaldtse:
@ni4 it seems that the automake part worked, but these two errors remain:

```
	 55 - rnp_tests.test_ffi_signatures_dump (Failed)
2554
	167 - ruby-rnp (Failed)
2555
```

I assume the latter is to be fixed by @dewyatt but any idea where the first comes from?

ni4:
@ronaldtse First one is because of test key expiration, I will make a separate PR fixing this. The second one is something on the ruby side, I guess. Did a PR in attempt to fix that here: https://github.com/rnpgp/ruby-rnp/pull/67

ci/before_install.sh
```diff
@@ -5,6 +5,7 @@ set -ex
 
 macos_install() {
   [ "${CI-}" = true ] || brew bundle
+  brew install automake
```
dewyatt, line 8: Instead of this just add it to the top-level Brewfile

dewyatt: Wait I just realized it's already in there -_-. Must be something else going on here...

dewyatt: I've been working on getting CI stable again with regards to the ruby-rnp failures, so I'll add this to the list. Hopefully I can get it finished today.

ni4: Thanks, @dewyatt !

dewyatt: Okay I see what's up here: https://github.blog/changelog/2020-04-15-github-actions-sets-the-ci-environment-variable-to-true/

Anyways, I'll get this stuff all fixed up.

dewyatt: I'm not really sure why I wrote it that way originally anyways, it would only install the bundle if CI was not true (which worked because it used to be blank until yesterday). That'll teach me to comment my code I guess. I'll probably just replace that with `brew bundle`. Hmm.

dewyatt:
I modified this to just a simple `brew bundle` and I'm going to go ahead and merge. I also added a workaround for an issue in Cirrus CI. Windows appears to be timing out but it shouldn't be related.

ni4:
Thanks, @dewyatt!

PR #1100, opened 2020-04-18
Pin to a specific commit of cmake-versioning

dewyatt:


dewyatt:
Merging since this is just CI and I think GHA is maybe having issues with Windows in general unfortunately.

PR #1105, opened 2020-04-23
Run tests verbose to check why windows CI hangs.

ni4:


ni4:
@dewyatt Yeah, nothing changed in codebase that dramatically. However, maybe it is related to some virtual environment changes, i.e. something which didn't hang before started to hang with some new software version (like asking something unexpected via console). Let's do another try and see if rnp_tests run separately from cli_tests.

ni4:
Closing this since source of problem was found and reproduced locally (but without luck to fix it yet :( )

PR #1109, opened 2020-04-27
Use std::vector to store signatures/uids/keys etc.

### LINKED ISSUE #1101: very slow listing of big (poisoned?) key
Download this key, which is a version of Werner Koch's key, but it might be a poisoned version. It has 3 user IDs, and each has around 1600 signatures:
https://kuix.de/misc/wk-full.asc

When listing the key with rnpkeys --list --with-sigs it is very very slow, while gpg lists it in less than 2 seconds.

```
mkdir /tmp/r2
rnpkeys --homedir=/tmp/r2 --import ~/Download/wk-full.asc
```

Speed of normal listing is acceptable, although it takes a couple of seconds:
```
rnpkeys --homedir=/tmp/r2 --list
```

Speed of listing with sigs isn't acceptable, it seems to get slower as it proceeds. Maybe you have some code that always needs to start iterating from the beginning, instead of proceeding from an offset?
```
rnpkeys --homedir=/tmp/r2 --list --with-sigs
```

ni4:
This would significantly speed things up for the large keys.
Also move us closer to the C++ approach: pgp_key_t now is treated as C++ object with constructors/destructors instead of direct memory copying/deletion/etc.

Fixes #1101

### DIFF
--- a/src/librekey/rnp_key_store.cpp
+++ b/src/librekey/rnp_key_store.cpp
@@ -490,11 +490,15 @@ rnp_key_store_add_subkey(rnp_key_store_t *keyring, pgp_key_t *srckey, pgp_key_t 
             RNP_LOG("failed to merge subkey");
             return NULL;
         }
-        pgp_key_free_data(srckey);
     } else {
-        oldkey = (pgp_key_t *) list_append(&keyring->keys, srckey, sizeof(*srckey));
+        oldkey = (pgp_key_t *) list_append(&keyring->keys, NULL, sizeof(*srckey));
         if (!oldkey) {
             RNP_LOG("allocation failed");
+            return NULL;
+        }
+        if (pgp_key_copy(oldkey, srckey, false)) {
+            RNP_LOG("key copying failed");
+            list_remove((list_item *) oldkey);
             return NULL;
         }
         if (primary && !pgp_key_link_subkey_grip(primary, oldkey)) {
@@ -521,7 +525,6 @@ rnp_key_store_add_key(rnp_key_store_t *keyring, pgp_key_t *srckey)
     pgp_key_t *added_key = rnp_key_store_get_key_by_grip(keyring, pgp_key_get_grip(srckey));
     /* we cannot merge G10 keys - so just return it */
     if (added_key && (srckey->format == PGP_KEY_STORE_G10)) {
-        pgp_key_free_data(srckey);
         return added_key;
     }
     /* different processing for subkeys */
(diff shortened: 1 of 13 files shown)


ronaldtse:
@ni4 is this ready to go?

ni4:
@ronaldtse Almost ready, it must wait for PR #1108 to be merged since based on it.

ronaldtse:
Windows failing again at timeout:

```
        Start 159: setupTestData
741
  1/169 Test #159: setupTestData .................................................................   Passed    0.21 sec
742
        Start 163: cli_tests-Encryption
743
        Start 162: cli_tests-Compression
744
```

src/tests/key-protect.cpp
```diff
@@ -75,13 +75,11 @@ TEST_F(rnp_tests, test_key_protect_load_pgp)
 
         pgp_key_t *tmp = NULL;
         assert_non_null(tmp = rnp_tests_get_key_by_id(ks, keyids[0], NULL));
-        assert_non_null(tmp);
 
         // steal this key from the store
-        key = (pgp_key_t *) calloc(1, sizeof(*key));
+        key = new pgp_key_t();
```
dewyatt, line 80: Style nit: I almost never see parens when there are no constructor args, but we don't have a rule on this

src/librekey/rnp_key_store.cpp
```diff
@@ -717,6 +719,7 @@ rnp_key_store_remove_key(rnp_key_store_t *keyring, const pgp_key_t *key)
     if (!list_is_member(keyring->keys, (list_item *) key)) {
         return false;
     }
+    key->~pgp_key_t();
```
dewyatt, line 722: I don't like the explicit destructor invocation. Is the plan to migrate the key store to use a vector instead and then get rid of this in the future?

ronaldtse:
Thanks @dewyatt! @ni4 can we address the minor concerns and merge? Thanks!

src/tests/key-protect.cpp
ni4: @dewyatt Thanks, fixing this.

src/librekey/rnp_key_store.cpp
ni4: @dewyatt Yeah, me too. The plan is to update this in a separate PR, it would be too heavy to migrate all lists to std::vector in a single PR.

ni4:
@ronaldtse Done, let's just wait for CI to finish.

ni4:
@dewyatt Hm, luckily it failed on Windows. From this: https://stackoverflow.com/a/5914697/452752 
It appears that `new pgp_key_t;` is not default-initialized compared to `new pgp_key_t();`.
So reverting change...

ni4:
@dewyatt Could we merge now, given that all tests are passing (but I still left braces in `new` call)?

dewyatt:
> @dewyatt Hm, luckily it failed on Windows.

Looks like the sanitize job caught it too. 

> It appears that `new pgp_key_t;` is not default-initialized compared to `new pgp_key_t();`.

Oops. I might need to brush up on my c++11.
Given how easy it is to make this mistake, I do wonder if we should address it in some way.
We could:
* Define our own constructor.
* Use member initializers like `pgp_fingerprint_t fingerprint {};` (more modern C++11).

That way either syntax would have the same behavior and be less surprising I think. Then we could also change all the `pgp_key_t pub = {}` to just `pgp_key_t pub`.

ni4:
@dewyatt I don't like the idea with constructor only because of one thing - it is easy to overlook it when adding new fields to the struct (at least till we use structs instead of class).
While seeing initializers on neighbours will make you type in default value.

Another question - should we update this PR, or make a separate issue? We'll need to type those curly braces for every field of every struct, actually.

dewyatt:
Ok let's just merge this and figure out the other stuff later

PR #1112, opened 2020-04-28
Used maximum password request tries.

### LINKED ISSUE #913: Limit password tries according to CLI/default parameters.
## Description

While we have option for maximum password request tries, it is not used anywhere inside and just stored in the config.
So code should be updated to use this parameter.

joke325:
Closes #913


ronaldtse:
The windows workflow here is broken, it can't checkout the ref... Can someone do a manual rebase?

dewyatt:
This is a good start but I think it's incomplete.
* Personally I'd prefer an argument name like `password-attempts` vs `num-tries`.
* I'm thinking `rnpkeys` should also accept and utilize this argument.
* I don't think `rnp` actually utilizes this, does it?

ronaldtse:
Closing — superseded by #2415 (Retry password request during decryption) which was merged.

PR #1113, opened 2020-04-28
Removed the version header from armored messages.

### LINKED ISSUE #871: Remove Version: header from armored messages.
## Description

As discussed in the OpenPGP mailing list, Version armor header should be removed to avoid some sort of attacks.

https://mailarchive.ietf.org/arch/msg/openpgp/KikdJaxvdulxIRX_yxU2_i3lQ7A

joke325:
Closes #871

### DIFF
--- a/.github/workflows/centos7.yml
+++ b/.github/workflows/centos7.yml
@@ -37,7 +37,12 @@ jobs:
             RNP_TESTS: rnp_tests
     env: ${{ matrix.env }}
     steps:
-      - uses: actions/checkout@v1
+      - run: |
+          yum -y install http://opensource.wandisco.com/centos/7/git/x86_64/wandisco-git-release-7-2.noarch.rpm
+          yum -y install git
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Add rnpuser
         run: |
           useradd rnpuser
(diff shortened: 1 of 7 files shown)


ronaldtse:
Failure in CentOS:

```
-- Configuring incomplete, errors occurred!
305
CMake Error at cmake/version.cmake:104 (file):
306
  file STRINGS file "/__w/rnp/rnp/version.txt" cannot be read.
307
Call Stack (most recent call first):
308
  CMakeLists.txt:41 (determine_version)
309

310

311
CMake Error at cmake/version.cmake:55 (message):
312
  Failed to extract version components.
313
Call Stack (most recent call first):
314
  cmake/version.cmake:107 (extract_version_info)
315
  CMakeLists.txt:41 (determine_version)
316

317

318
```

dewyatt:
> Failure in CentOS:
> 

I think it should be ok now. Initially we were encountering https://github.com/actions/checkout/issues/222 so I upgraded to `actions/checkout@v2`, but the version of git present in centos is too old for normal use with that action, so we end up missing the `.git` folder (it falls back to downloading via REST) which is where that error comes from. I added a step to install a newer git for centos, let's see how that goes.

ronaldtse:
Worked perfectly @dewyatt. Thanks!

Issue #1104, opened 2020-04-21
Failure in Windows build in master


ronaldtse:
## Description

In the master build the Windows build seems to fail by timeout; i.e.

```
[736]
+ cp /d/a/rnp/rnp/cmake/CTestCostData.txt /home/runneradmin/local-builds/rnp-build/Testing/Temporary
+ ctest -j2 -R '.*' --output-on-failure
Test project C:/hostedtoolcache/windows/Ruby/2.5.8/x64/msys64/home/runneradmin/local-builds/rnp-build
        Start 156: setupTestData
  1/166 Test #156: setupTestData .................................................................   Passed    0.24 sec
        Start 160: cli_tests-Encryption
        Start 159: cli_tests-Compression
```

## Steps to Reproduce

1. Build: https://github.com/rnpgp/rnp/runs/605970058?check_suite_focus=true

ni4:
@ronaldtse Looks like it's something with GHA (see comment https://github.com/rnpgp/ruby-rnp/pull/67#issuecomment-616207654).
At least recently merged PR #1097 didn't make any windows-related changes.

ronaldtse:
Got it, thanks! Will look out for the fix to come...

ronaldtse:
This is the latest failure: https://github.com/rnpgp/rnp/actions/runs/91859235

```
D:/a/rnp/rnp/src/lib/rnp.cpp: In function 'rnp_result_t key_to_json(json_object*, rnp_key_handle_t, uint32_t)':
437
D:/a/rnp/rnp/src/lib/rnp.cpp:6324:70: error: 'TRUE' was not declared in this scope
438
 6324 |     json_object *jsorevoked = json_object_new_boolean(key->revoked ? TRUE : FALSE);
439
      |                                                                      ^~~~
440
D:/a/rnp/rnp/src/lib/rnp.cpp:6324:77: error: 'FALSE' was not declared in this scope
441
 6324 |     json_object *jsorevoked = json_object_new_boolean(key->revoked ? TRUE : FALSE);
442
      |                                                                             ^~~~~
443
make[2]: *** [src/lib/CMakeFiles/librnp.dir/build.make:551: src/lib/CMakeFiles/librnp.dir/rnp.cpp.obj] Error 1
444
make[2]: Leaving directory '/home/runneradmin/local-builds/rnp-build'
445
make[1]: *** [CMakeFiles/Makefile2:1215: src/lib/CMakeFiles/librnp.dir/all] Error 2
[...]

PR #1114, opened 2020-04-29
Use true/false instead of TRUE/FALSE for json boolean.

ni4:
Latest MinGW uses json-c 0.14, where TRUE/FALSE constants were removed and replaced with 1/0.
So, changing those to normal true/false (so Windows CI may just hang instead of reporting compilation error).


ronaldtse:
This is related to the Windows failures of #1104, @dewyatt @rrrooommmaaa could we get one more review please? Thanks!

dewyatt:
> This is related to the Windows failures of #1104, @dewyatt @rrrooommmaaa could we get one more review please? Thanks!

I do not believe this is related, this was a failure in the build stage, we're still having timeouts during the test stage.

PR #1118, opened 2020-04-30
Correct key expiration check for multiple self-signatures.

### LINKED ISSUE #1115: Key validity extension signature is ignored/rejected, if done after expiration
We have an example key, that RNP treats as "expired".
GnuPG treats it as "never expires".

The key contains a self-signature, which defines an expiration date.
There's is an additional self-signature, made at a later time, which does not define an expiration.

I assume that GnuPG accepts the second signature, overriding the original expiration.

The second signature was made after the original expiration date.
When RNP processes the key, a log message is printed:

[signature_check() /home/user/moz/commcent/mozilla/comm/third_party/rnp/src/librepgp/stream-sig.cpp:1155] signature made after key expiration

I'm guessing that RNP decides to ignore the 2nd signature.

Searching the web, it appears that the common opinion is that extending a key after it expired is permitted.

Suggestion: Please allow self-signatures that are made after the key's expiration date.

I have been given an example key, which I can share in private, only, because the key owner prefers to have their user IDs be kept confidential.

ni4:
This PR fixes the case when key has multiple self-signatures, and the latest one doesn't have key expiration subpacket (marking key as non-expiring), while older one has already expired key.

Fixes #1115


ni4:
Merging this since we have two approvals.

PR #1127, opened 2020-05-08
tests: Drop Windows mkdtemp implementation.

dewyatt:
This used mktemp which will typically fail after 26 invocations in the same process.
This was mostly only an issue when invoking rnp_tests directly, not via CTest, due to the process handling therein.


ronaldtse:
Thanks @dewyatt!

PR #1128, opened 2020-05-09
Add Windows workflow badge.

ni4:
What about to have Windows badge as well (and utilizing the fact that it passes tests :) ) ?


dewyatt:
(I'm just going to merge this since it's just a doc change)

Issue #1134, opened 2020-05-12
crash at pgp_subkey_refresh_data


mkmelin:
## Description
Crash - see https://bugzilla.mozilla.org/show_bug.cgi?id=1637046
@ https://searchfox.org/comm-central/rev/32028983d85d767411a1487ed7b8394e00865b35/third_party/rnp/src/lib/pgp-key.cpp#1319

Looks like sub->revocation is null

0 librnp.so pgp_subkey_refresh_data(pgp_key_t*, pgp_key_t*) comm/third_party/rnp/src/lib/pgp-key.cpp:1319 context
1 librnp.so rnp_key_store_add_key(rnp_key_store_t*, pgp_key_t*) comm/third_party/rnp/src/librekey/rnp_key_store.cpp:530

ni4:
@mkmelin Thanks for the report! Do you use the latest master's commit?

ni4:
@mkmelin Sorry, no need to answer - we definitely should not call strlen() on possible NULL.

ni4:
Some more details:
- the same applies to the primary key
- we should add a test with both primary key/subkey revocation signatures which doesn't contain revocation reason subpacket.

pbrunschwig:
There's at least one more occurrence of that bug:
(see https://crash-stats.mozilla.org/report/index/bb8bc9ba-f139-469d-9bf6-49a600200523)

pgp_key.cpp, line 1388:

```
signature_get_revocation_reason(&sig->sig, &revocation->code, &revocation->reason);
        if (!strlen(revocation->reason)) {
            free(revocation->reason);
            revocation->reason = strdup(pgp_str_from_map(revocation->code, ss_rr_code_map));
        }
```

ni4:
@pbrunschwig @mkmelin This issue should be fixed since PR #1152 was merged. Could you please confirm this?

kaie:
@ni4 Building Thunderbird is a bigger task, and Patrick probably doesn't do it himself. I'm guessing that for testing, Patrick relies on the automatic nightly development builds that the Thunderbird project provides.

We don't pick up new RNP snapshots automatically, it's a manual process. @ni4 whenever you think you have made progress and now is a good time to pick up a new RNP snapshot, please let me know in some way, and we trigger a rebase to a newer RNP snapshot.

I've started that process yesterday https://bugzilla.mozilla.org/show_bug.cgi?id=1641612

kaie:
[...]

PR #1136, opened 2020-05-16
Add ability to strip unneeded signatures from keys.

### LINKED ISSUE #1006: Add feature to strip broken/unneeded key signatures
## Description

With the recent DoS attacks on keyservers, some applications have started to use stripped keys.
It would be good to have a way to give RNP a key with signatures, ask it to do a sanity check, and strip bad/broken or irrelevant signatures. Maybe I'd only want to keep the signatures made by one my own private keys.

ni4:
This PR adds FFI flags and functions which allows to strip unneeded signatures from the key.

Since that would be a bit too overcomplicated (and O(n^2)) if used with the previous rawpacket approach, it was changed to store rawpackets inside of the corresponding upper-level objects. 
Also those types were made more C++ for easier and cleaner memory management.

Still draft since requires to check for memory leaks/add FFI functions and tests.

Closes #1006

### DIFF
--- a/include/repgp/repgp_def.h
+++ b/include/repgp/repgp_def.h
@@ -490,4 +490,19 @@ typedef enum pgp_key_store_format_t {
     PGP_KEY_STORE_G10,
 } pgp_key_store_format_t;
 
+/**
+ * Key stripping flags. The meaning is as following:
+ * PGP_KS_INVALID_SIG : remove invalid signatures (including expired ones)
+ * PGP_KS_UNKNOWN_SIG : remove signatures issued by unknown keys
+ * PGP_KS_ONLYOWN_SIG : leave only self-signatures, made by the primary key
+ * PGP_KS_UNUSED_SIG  : remove valid but no longer meaningful self-signatures (i.e. old direct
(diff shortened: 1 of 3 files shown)


ni4:
@dewyatt Do you have an idea where Python3 was gone from Centos builds? Configuration seems to be the same as previous.

dewyatt:
> @dewyatt Do you have an idea where Python3 was gone from Centos builds? Configuration seems to be the same as previous.

I can only guess that python3 was previously pulled in as a dependency on something else, and that dependency no longer exists. So we probably just need to add `python3` explicitly to `linux_install_centos` in `before_install.sh`.

ronaldtse:
@ni4 is this soon good to go? Thanks!

ni4:
@ronaldtse Not yet, unfortunately. Due to new issues reported by TB team I'd first do PR #1152, then, based on it, fix issue #1145 & #1134, and then get back to this.

ronaldtse:
@ni4 when would we be able to merge this?

ni4:
@ronaldtse This one is to be replaced with #1382 and one other PR. So I'll close it once the second PR is merged.

PR #1138, opened 2020-05-16
Make hex2bin utilize rnp_hex_decode.

dewyatt:
The implementation of hex2bin had some bugs, mostly with handling whitespace.

### DIFF
--- a/src/lib/misc.cpp
+++ b/src/lib/misc.cpp
@@ -423,7 +423,14 @@ rnp_hex_encode(
 size_t
 rnp_hex_decode(const char *hex, uint8_t *buf, size_t buf_len)
 {
-    if (botan_hex_decode(hex, strlen(hex), buf, &buf_len) != 0) {
+    size_t hexlen = strlen(hex);
+
+    /* check for 0x prefix */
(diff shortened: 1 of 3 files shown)


dewyatt:
Added tests, but it seems there's an unrelated failure in CI with Windows:
```
loading package files...
error: could not open file /var/cache/pacman/pkg/msys2-runtime-3.1.4-1-x86_64.pkg.tar.zst: Child process exited with status 127
error: failed to commit transaction (cannot open package file)
Errors occurred, no packages were upgraded.
##[error]Process completed with exit code 1.
```

Not sure what's going on there but definitely unrelated.

ni4:
@dewyatt Thanks for the test. Yeah, given that msys2-runtime-3.1.4-1 was released just today (adjusted to timezones), it should be something wrong with it.

rrrooommmaaa:
@ni4 @dewyatt We have similar functions in `rnp/fficli.cpp`, like `hex_prefix_len` etc.
Can we move them to `lib/misc.cpp`?

ni4:
@rrrooommmaaa Yeah, there is some duplication since we separated CLI from the base library code, so CLI is allowed only to use the FFI interface to interact with the library.

ronaldtse:
Since the Windows failure is unrelated going to merge this first. Thanks!

PR #1140, opened 2020-05-16
tests: Fix up test_key_store_search_by_name.

dewyatt:
I'm not really sure on the last line here, could probably use another set of eyes on that


dewyatt:
Ah I forgot this depends on https://github.com/rnpgp/rnp/pull/1138 so if/when that is merged I'll rebase this

PR #1144, opened 2020-05-20
Allow wrong MPI bit count.

### LINKED ISSUE #1143: decrypt failure with error: wrong mpi bit count, failed to get rsa m
I regularly receive encrypted email from a bug tracking system.

Today I received a message that RNP cannot decrypt, it reports:
```
[get_packet_body_mpi() /home/user/github/rnp/src/librepgp/stream-packet.cpp:518] wrong mpi bit count
[stream_parse_pk_sesskey() /home/user/github/rnp/src/librepgp/stream-packet.cpp:1132] failed to get rsa m
```

Secret key is available. I compared with another encrypted email from the same source, same key was used.

GnuPG is able to decrypt the message correctly.

I'm attaching the message.

ni4:
While RFC tells that MPIs where bit count in header doesn't correspond to real contents as 'ill-formed', it doesn't tell whether we should allow those or discard.
From the discussion in #1143 it appears that GnuPG allows such packets, so we most likely (discussion is welcome!) should support them as well.
This should not be any sort of security issue since possible attacker may change bit count to valid value any time.

Fixes #1143.


ni4:
Okay, Windows build is still failing because of MSYS2 - see the issue here: https://github.com/msys2/MSYS2-packages/issues/1962

Tried described there workaround (upgrade pacman first) - didn't work, so reverted.

src/tests/ffi.cpp
```diff
@@ -6560,18 +6566,20 @@ TEST_F(rnp_tests, test_ffi_op_verify_sig_count)
     rnp_input_destroy(input);
 
     /* malformed detached signature, wrong bitlen in MPI  */
+    sigcount = 255;
     assert_rnp_success(rnp_input_from_path(&source, "data/test_messages/message.txt"));
     assert_rnp_success(
       rnp_input_from_path(&input, "data/test_messages/message.txt.sig.wrong-mpi-bitlen"));
     assert_rnp_success(rnp_op_verify_detached_create(&verify, ffi, source, input));
-    assert_int_equal(rnp_op_verify_execute(verify), RNP_ERROR_BAD_PARAMETERS);
+    assert_int_equal(rnp_op_verify_execute(verify), RNP_SUCCESS);
```
rrrooommmaaa, line 6574: we usually use `assert_rnp_success`

ni4: @rrrooommmaaa Agree, just copy-pasted piece of some old code, written before assert_rnp_success was introduced. Will fix and force-push.

ni4:
@rrrooommmaaa Just force-pushed. Any other comments?

ronaldtse:
Merging first, despite Windows failure.

PR #1155, opened 2020-05-30
ci: Fix Windows MSYS2.

dewyatt:


ronaldtse:
Since only GHA workflow on Windows was changed, the build failure on Cirrus (ruby-rnp) is irrelevant. Merging as this is a pure CI change. Thanks!

PR #1166, opened 2020-06-05
Log errored key ids during import/merge.

### LINKED ISSUE #1162: Log offending key ID on import failure
If import of a key block with rnp_import_keys with many keys is failing (because some key in the middle causes the import to return with failure), then it's currently impossible to find out which key was the offending key.

I suggest to dump the key ID of the offending key next to the error log describing the failure.

ni4:
This PR improves logging, making potential error discovery faster and easier.
Fixes #1162

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -1301,7 +1301,10 @@ rnp_import_keys(rnp_ffi_t ffi, rnp_input_t input, uint32_t flags, char **results
             continue;
         }
         if (validate_pgp_key_material(pgp_key_get_material(&key), &ffi->rng)) {
-            FFI_LOG(ffi, "warning! attempt to import key with invalid material.");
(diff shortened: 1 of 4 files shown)


ronaldtse:
@ni4 would it be better to do less preprocessor magic? Wouldn't the compiler optimize those anyway?

rrrooommmaaa:
> @ni4 would it be better to do less preprocessor magic? Wouldn't the compiler optimize those anyway?

inline function, probably?

ni4:
> @ni4 would it be better to do less preprocessor magic? Wouldn't the compiler optimize those anyway?

I don't like defines approach equally to code duplication, but in this case it gives the ability to correctly print source file/function/line number. When we'll migrate to C++ so all exceptions are caught this could be replaced with `const std::string &key.keyidhex()`.

ronaldtse:
@ni4 sounds good 👍

PR #1169, opened 2020-06-08
Update errored subkey dumping with primary's keyid.

ni4:
This PR updates #1166. It appears that during troubleshooting it's not easy to locate primary key's id using the GnuPG/other tools.


src/librekey/rnp_key_store.cpp
```diff
@@ -438,6 +438,7 @@ rnp_key_store_add_subkey(rnp_key_store_t *keyring, pgp_key_t *srckey, pgp_key_t
         /* in case we already have key let's merge it in */
         if (!rnp_key_store_merge_subkey(oldkey, srckey, primary)) {
             RNP_LOG_KEY("failed to merge subkey %s", srckey);
+            RNP_LOG_KEY("primary key %s", primary);
```
rrrooommmaaa, line 441: Let's do `"primary key is %s"` as in other places

ni4: @rrrooommmaaa Thanks, fixed and force-pushed.

PR #1172, opened 2020-06-16
Remove flag CLI_SEARCH_SUBKEYS_AFTER. Fixes #1153

### LINKED ISSUE #1153: Double key export via CLI
## Description

In some cases rnpkeys exports double key via `--export` command.
At first glance it doesn't happen without subkeys, and happens if there is at least one subkey.

## Steps to reproduce
```
mkdir .rnp
rnpkeys --homedir .rnp --import src/tests/data/test_key_validity/alice-sub-sec.pgp
rnpkeys --homedir .rnp --export Alice > alice.asc
rnp --list-packets alice.asc
```

antonsviridenko:
Fixes double export of keys having subkeys.
If the found key has subkeys, they are exported
anyway by rnp_key_export() with RNP_KEY_EXPORT_SUBKEYS flag set.

### DIFF
--- a/src/rnp/fficli.cpp
+++ b/src/rnp/fficli.cpp
@@ -1636,8 +1636,8 @@ stdout_writer(void *app_ctx, const void *buf, size_t len)
 bool
 cli_rnp_export_keys(cli_rnp_t *rnp, const char *filter)
 {
-    bool secret = rnp_cfg_getbool(cli_rnp_cfg(rnp), CFG_SECRET);
-    int  flags = CLI_SEARCH_SUBKEYS_AFTER | (secret ? CLI_SEARCH_SECRET : 0);
+    bool                          secret = rnp_cfg_getbool(cli_rnp_cfg(rnp), CFG_SECRET);
+    int                           flags = secret ? CLI_SEARCH_SECRET : 0;
     std::vector<rnp_key_handle_t> keys;
 
     if (!cli_rnp_keys_matching_string(rnp, keys, filter, flags)) {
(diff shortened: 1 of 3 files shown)


dewyatt:
Agreed w/Nick -- clang-format, add test, more descriptive commit msg (since we're not removing that flag entirely here, just the one specific use)

antonsviridenko:
Yeah, right, someone can think this flag is completely removed, did not think that way

antonsviridenko:
`and add test (most likely CLI), which ensures that we do not have regression.`

What exactly should this test check for, compare --list-packets output with expected output saved in the text file
or just check that key packet is exported once only?

ni4:
> What exactly should this test check for, compare --list-packets output with expected output saved in the text file
or just check that key packet is exported once only?

@antonsviridenko just compare to the text file, that would be enough. Please also note those `???` in already existing tests - on different systems with different time zones time values may differ.

antonsviridenko:
How can I properly run CLI tests? Looks like doing "make test" in a build directory does not run them, only C++ tests. And just running ./cli_tests.py directly prints lot of errors.

ni4:
> How can I properly run CLI tests? Looks like doing "make test" in a build directory does not run them, only C++ tests. And just running ./cli_tests.py directly prints lot of errors.

'ctest -V -R cli_tests' should work (once you are doing out-of-source build). Also you may limit test scope there, like 'cli_tests.Misc'

antonsviridenko:
ping @ni4 :)

PR #1174, opened 2020-06-17
FFI: Add functions to retrieve information about the recipients.

### LINKED ISSUE #1107: Encryption status from rnp_op_verify_*
When processing an encrypted message, prior to decrypting, it's unknown if a message is both encrypted and signed.

If the message uses the combined signature and encryption encoding, calling rnp_decrypt will not process the signature. Consequently I'm using the rnp_op_verify_* functions, which will optionally decrypt, and also allow retrieval of the signature information.

However, it seems there isn't a way to obtain status information about the decryption.

Did RNP have to decrypt the message?
Right now, I only can deduce that from the outer context, e.g. the message was received in a MIME message that said it's encrypted. But potentially that conclusion could be incorrect (message claims to be encrypted, but isn't).

Is there a reliable way to determine, if processing the data given to rnp_op_verify_* involved decryption?

Also it would be convenient to obtain status information about the decryption, e.g. which key ID was used to perform the decryption?

Unless there's other ways to obtain this information (without having to do an additional round of full processing with rnp_decrypt), would it be reasonable to add an API like the following?
```
rnp_result_t rnp_op_verify_get_decrypt_info(rnp_op_verify_t op, bool *was_decrypted, char **keyid);
```

Also, if it's initially unknown if there's a signature inside a message, the rnp_op_verify_* call might do decryption, only, without performing any signature verification. However, it seems possible to reliably detect this scenario, if rnp_op_verify_get_signature_count returns a zero count.

ni4:
This PR adds functions which allows to retrieve:
- list of keyids/algorithms of public keys message was encrypted to
- keyid/alg which was used for decryption
- information about passwords which can be used to decrypt message
- information about the password which was used for decryption

P.S. I don't 100% like name 'symencs' for password-related informations. Anyone have better suggestion?

Fixes #1107


ni4:
@kaie Would this interface be suitable for your needs?

src/lib/rnp.cpp
```diff
@@ -2665,6 +2707,85 @@ rnp_verify_dest_provider(pgp_parse_handler_t *handler,
...
+    handle->aalg = sesskey.aalg;
+}
+
+static void
+rnp_verify_on_recipients(const std::vector<pgp_pk_sesskey_t> &recipients,
+                         const std::vector<pgp_sk_sesskey_t> &passwords,
+                         void *                               param)
+{
+    rnp_op_verify_t op = (rnp_op_verify_t) param;
+    if (recipients.size()) {
```
rrrooommmaaa, line 2738: maybe `!empty()` is nicer?

src/lib/rnp.cpp
```diff
@@ -2665,6 +2707,85 @@ rnp_verify_dest_provider(pgp_parse_handler_t *handler,
...
+            FFI_LOG(op->ffi, "allocation failed");
+            return;
+        }
+        for (size_t i = 0; i < recipients.size(); i++) {
+            recipient_handle_from_pk_sesskey(&op->recipients[i], recipients[i]);
+        }
+    }
+    op->recipient_count = recipients.size();
+
+    if (passwords.size()) {
```
rrrooommmaaa, line 2751: maybe `!empty()` is nicer?

src/lib/rnp.cpp
ni4: @rrrooommmaaa Agree, thanks for pointing on this.

ni4:
> Looks like we don't have tests for symenc_count > 1 or recipient_count > 1.
Would be nice to add them.

Please see the updated PR.

PR #1175, opened 2020-06-17
ci: Add test for using cmake target

dewyatt:
I think I meant to do this also in #1173, it's similar to the pkgconfig test but uses the cmake target instead (for systems w/out pkgconfig, etc).


ronaldtse:
Merging since this is CI change.

PR #1177, opened 2020-06-19
Armor line length

### LINKED ISSUE #1099: Consider to use shorter armor lines
When exporting armored blocks, e.g. for public keys, RNP uses a fixed line length of 76 chars. I believe this is set in the following line:

```
third_party/rnp/src/librepgp/stream-armor.cpp:    param->llen = 76; /* must be multiple of 4 */
```

When attaching these in email, Thunderbird wraps it at position 73.

(Armored blocks produced by GnuPG wrap at position 64, and didn't cause Enigmail to wrap.)

It's not a big deal, it simply looks slightly ugly when looking at the message source (with alternating lines of 73 and 3 characters).

Not sure if I should ask you to change it to 64, too, or make it configurable. What do you think?

Low priority, because from the functional perspective its working fine.

rrrooommmaaa:
I have added `rnp_pipe` to pipe from input to output, as well as `rnp_output_armor_set_line_length` as siggested,
so it is now possible to do:
```
rnp_output_to_armor(...)
rnp_output_armor_set_line_length(...);
rnp_pipe(...);
```
Fixes #1099

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -2325,6 +2325,14 @@ RNP_API rnp_result_t rnp_identifier_iterator_destroy(rnp_identifier_iterator_t i
  */
 RNP_API rnp_result_t rnp_output_pipe(rnp_input_t input, rnp_output_t output);
 
+/** Set line length for armored output
+ *
+ *  @param output stream to configure
+ *  @param llen line length in characters [16..76]
+ *  @return RNP_SUCCESS on success, or any other value on error
+ */
+RNP_API rnp_result_t rnp_output_armor_set_line_length(rnp_output_t output, size_t llen);
+
 #if defined(__cplusplus)
 }
 
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -2317,6 +2317,14 @@ RNP_API rnp_result_t rnp_identifier_iterator_next(rnp_identifier_iterator_t it,
  */
 RNP_API rnp_result_t rnp_identifier_iterator_destroy(rnp_identifier_iterator_t it);
 
+/** Read from input and write to output
+ *
+ *  @param input stream to read data from
+ *  @param output stream to write data to
+ *  @return RNP_SUCCESS on success, or any other value on error
+ */
+RNP_API rnp_result_t rnp_output_pipe(rnp_input_t input, rnp_output_t output);
+
 #if defined(__cplusplus)
 }
 
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -7223,3 +7223,12 @@ rnp_output_pipe(rnp_input_t input, rnp_output_t output)
     output->keep = !ret;
     return ret;
 }
+
+rnp_result_t
+rnp_output_armor_set_line_length(rnp_output_t output, size_t llen)
+{
+    if (!output || !llen) {
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+    return armored_dst_set_line_length(&output->dst, llen);
+}
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -7212,3 +7212,14 @@ rnp_dearmor(rnp_input_t input, rnp_output_t output)
     output->keep = !ret;
     return ret;
 }
+
+rnp_result_t
+rnp_output_pipe(rnp_input_t input, rnp_output_t output)
+{
+    if (!input || !output) {
+        return RNP_ERROR_NULL_POINTER;
+    }
+    rnp_result_t ret = dst_write_src(&input->src, &output->dst);
+    output->keep = !ret;
+    return ret;
+}
--- a/src/librepgp/stream-armor.cpp
+++ b/src/librepgp/stream-armor.cpp
@@ -38,6 +38,8 @@
 #include "types.h"
 
 #define ARMORED_BLOCK_SIZE (4096)
+#define ARMORED_MIN_LINE_LENGTH (16)
+#define ARMORED_MAX_LINE_LENGTH (76)
 
 typedef struct pgp_source_armored_param_t {
     pgp_source_t *    readsrc;         /* source to read from */
@@ -789,10 +791,12 @@ armored_dst_write(pgp_dest_t *dst, const void *buf, size_t len)
         }
     }
 
+    /* this version prints whole chunks, so rounding down to the closest 4 */
+    auto adjusted_llen = param->llen & ~3;
     /* number of input bytes to form a whole line of output, param->llen / 4 * 3 */
-    inllen = (param->llen >> 2) + (param->llen >> 1);
+    inllen = (adjusted_llen >> 2) + (adjusted_llen >> 1);
     /* pointer to the last full line space in encbuf */
-    enclast = encbuf + sizeof(encbuf) - param->llen - 2;
+    enclast = encbuf + sizeof(encbuf) - adjusted_llen - 2;
 
     /* processing line chunks, this is the main performance-hitting cycle */
     while (bufptr + 3 <= bufend) {
@@ -802,8 +806,8 @@ armored_dst_write(pgp_dest_t *dst, const void *buf, size_t len)
             encptr = encbuf;
         }
         /* setup length of the input to process in this iteration */
-        inlend =
-          param->lout == 0 ? bufptr + inllen : bufptr + ((param->llen - param->lout) >> 2) * 3;
+        inlend = param->lout == 0 ? bufptr + inllen :
+                                    bufptr + ((adjusted_llen - param->lout) >> 2) * 3;
         if (inlend > bufend) {
             /* no enough input for the full line */
             inlend = bufptr + (bufend - bufptr) / 3 * 3;
@@ -946,6 +950,24 @@ finish:
     }
 
     return ret;
+}
+
+bool
+is_armored_dest(pgp_dest_t *dst)
+{
+    return dst->type == PGP_STREAM_ARMORED;
+}
+
+rnp_result_t
+armored_dst_set_line_length(pgp_dest_t *dst, size_t llen)
+{
+    if (!dst || llen < ARMORED_MIN_LINE_LENGTH || llen > ARMORED_MAX_LINE_LENGTH ||
+        !dst->param || !is_armored_dest(dst)) {
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+    auto param = (pgp_dest_armored_param_t *) dst->param;
+    param->llen = llen;
+    return RNP_SUCCESS;
 }
 
 bool
--- a/src/librepgp/stream-armor.cpp
+++ b/src/librepgp/stream-armor.cpp
@@ -961,7 +961,7 @@ is_armored_dest(pgp_dest_t *dst)
 rnp_result_t
 armored_dst_set_line_length(pgp_dest_t *dst, size_t llen)
 {
-    if (!dst || llen < ARMORED_MIN_LINE_LENGTH || llen > ARMORED_MAX_LINE_LENGTH ||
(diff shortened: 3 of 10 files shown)


include/rnp/rnp.h
```diff
@@ -2269,6 +2269,22 @@ rnp_result_t rnp_identifier_iterator_next(rnp_identifier_iterator_t it,
...
+ */
+rnp_result_t rnp_pipe(rnp_input_t input, rnp_output_t output);
+
+/** Set line length for armored output
+ *
+ *  @param output stream to configure
+ *  @param llen line length in characters
+ *  @return RNP_SUCCESS on success, or any other value on error
+ */
+rnp_result_t rnp_output_armor_set_line_length(rnp_output_t output, int llen);
```
ni4, line 2286: Why llen is int, not size_t? Is it supposed to be negative?
Also it would be good to document highest/lowest value (76 and something like 16 or so).

src/librepgp/stream-armor.cpp
```diff
@@ -979,50 +996,37 @@ rnp_dearmor_source(pgp_source_t *src, pgp_dest_t *dst)
 {
     rnp_result_t res = RNP_ERROR_BAD_FORMAT;
     pgp_source_t armorsrc = {0};
-    uint8_t      readbuf[PGP_INPUT_CACHE_SIZE];
-    size_t       read;
+    {
```
ni4, line 999: Why did you need to put this code into nested `{}` block? It doesn't seem to change anything.

src/librepgp/stream-common.h
```diff
@@ -353,4 +353,11 @@ void *mem_dest_own_memory(pgp_dest_t *dst);
  **/
 rnp_result_t init_null_dest(pgp_dest_t *dst);
 
+/** @brief reads from source and writes to destination
+ *  @param src initialized source
+ *  @param dst initialized destination
+ *  @return RNP_SUCCESS or error code
+ **/
+rnp_result_t rnp_pipe_source(pgp_source_t *src, pgp_dest_t *dst);
```
ni4, line 361: At this layer no need to use `rnp_` prefix. This could be better called `src_pipe()`, `dst_write_src()` or so on.

src/tests/ffi.cpp
```diff
@@ -4500,6 +4500,114 @@ TEST_F(rnp_tests, test_ffi_enarmor_dearmor)
...
+    std::string       data;
+
+    // enarmor with line_length = 64
+    data.clear();
+    {
+        uint8_t *    buf = NULL;
+        size_t       buf_size = 0;
+        rnp_input_t  input = NULL;
+        rnp_output_t output = NULL;
+        rnp_output_t armor_layer = NULL;
```
ni4, line 4517: Personally I'd prefer to have same-named variable to be declared in single place, without repeating in the each block. Also additional spacing makes code slightly less readable. However this is just my personal opinion.
@dewyatt What do you think?

include/rnp/rnp.h
```diff
@@ -2269,6 +2269,22 @@ rnp_result_t rnp_identifier_iterator_next(rnp_identifier_iterator_t it,
  */
 rnp_result_t rnp_identifier_iterator_destroy(rnp_identifier_iterator_t it);
 
+/** Read from input and write to output
+ *
+ *  @param input stream to read data from
+ *  @param output stream to write data to
+ *  @return RNP_SUCCESS on success, or any other value on error
+ */
+rnp_result_t rnp_pipe(rnp_input_t input, rnp_output_t output);
```
ni4, line 2278: It could be better understood by user if function name is more descriptive and sticked to `rnp_output_` prefix, like `rnp_output_pipe()`.

src/tests/ffi.cpp
```diff
@@ -4500,6 +4500,114 @@ TEST_F(rnp_tests, test_ffi_enarmor_dearmor)
     }
 }
 
+TEST_F(rnp_tests, test_ffi_customized_enarmor)
```
ni4, line 4503: I would suggest extending functionality of tests with following cases:
- make sure it works on some real data, like exporting large key
- check that export with every available line length in cycle (i.e. from 16 till 76), dearmoring it back.

src/tests/support.cpp
```diff
@@ -833,3 +833,28 @@ rnp_tests_key_search(rnp_key_store_t *keyring, const std::string &keyid)
     srch_userid.by.userid[sizeof(srch_userid.by.userid) - 1] = '\0';
     return rnp_key_store_search(keyring, &srch_userid, NULL);
 }
+
+size_t
+get_longest_line_length(const std::string &str)
+{
+    const char begin_tag[] = "-----BEGIN PGP MESSAGE-----\r\n";
+    // const char end_tag[] = "-----END PGP MESSAGE-----\r\n";
```
ni4, line 841: I guess this one was supposed to be removed?

src/tests/support.h
```diff
@@ -223,3 +223,4 @@ pgp_key_t *rnp_tests_get_key_by_id(rnp_key_store_t *  keyring,
                                    pgp_key_t *        after);
 pgp_key_t *rnp_tests_get_key_by_fpr(rnp_key_store_t *keyring, const std::string &keyid);
 pgp_key_t *rnp_tests_key_search(rnp_key_store_t *keyring, const std::string &keyid);
+size_t     get_longest_line_length(const std::string &str);
```
ni4, line 226: This function seems to be used only in single place, so can be moved there and made static.

src/tests/ffi.cpp
dewyatt: I get why one would want to limit the scope of variables like this, but I'd agree with Nick that it's probably not worth it here, and the blank lines are a bit much.

src/librepgp/stream-armor.cpp
rrrooommmaaa: @ni4 It releases 32K from the stack.
Without it, the stack should have capacity of 32K+32K = 64K

src/tests/ffi.cpp
rrrooommmaaa: @ni4 key dumps are not covered by `rnp_output_to_armor`. It's rather tricky to propagate this setting (and it wasn't required by the ticket)

src/librepgp/stream-armor.cpp
ni4: @rrrooommmaaa  But actually, since all bulk reading buffer was moved to rnp_pipe_source(), in this function it is enough to have buffer of `strlen(ST_CLEAR_BEGIN) + 1`, and so no need to add `{}` block.

src/tests/ffi.cpp
ni4: @rrrooommmaaa No need to propagate this to key export armoring - you can just export raw key to  `rnp_output_to_armor()` with modified line length. That would allow to test real usecase, as it was asked by TB team. It is hard to describe all the details in the issue description, but it is quite logical to add more thorough tests when we adding such new functionality. Say, who knows, whether code works well on 2kb armored message with line length of 50.

rrrooommmaaa: I have done this and other fixes too.

src/lib/rnp.cpp
```diff
@@ -3433,7 +3433,7 @@ rnp_key_export(rnp_key_handle_t handle, rnp_output_t output, uint32_t flags)
     // handle flags
     if (flags & RNP_KEY_EXPORT_ARMORED) {
         flags &= ~RNP_KEY_EXPORT_ARMORED;
-        armored = true;
+        armored = !is_armored_dest(&output->dst);
```
ni4, line 3436: I don't think we should decide for the user what to do in this case. Say, what if he really wants to put some binary data into armored stream, and then add armored public key.

src/librepgp/stream-armor.cpp
```diff
@@ -1006,23 +1028,8 @@ rnp_dearmor_source(pgp_source_t *src, pgp_dest_t *dst)
...
-            continue;
-        }
-        dst_write(dst, readbuf, read);
-        if (dst->werr) {
-            RNP_LOG("failed to output data");
-            res = RNP_ERROR_WRITE;
-            break;
-        }
-    }
+    res = dst_write_src(&armorsrc, dst);
```
ni4, line 1032: Previously we were logging an error if this operation fails, so we should log it here as well, so in case of error it would be easier to detect the problem.

src/librepgp/stream-armor.cpp
```diff
@@ -1032,30 +1039,13 @@ rnp_result_t
...
-            continue;
-        }
-        dst_write(&armordst, readbuf, read);
-        if (armordst.werr) {
-            RNP_LOG("failed to output data");
-            res = RNP_ERROR_WRITE;
-            break;
-        }
-    }
+    res = dst_write_src(src, &armordst);
```
ni4, line 1047: It would be good to log an error as well.

src/librepgp/stream-armor.cpp
ni4: And, some possible refactoring - as `armorsrc` is the only resource freed after goto to finish, we may safely return if `init_armored_src` failed, removing the goto label.

include/rnp/rnp.h
```diff
@@ -2306,6 +2306,22 @@ RNP_API rnp_result_t rnp_identifier_iterator_next(rnp_identifier_iterator_t it,
  */
 RNP_API rnp_result_t rnp_identifier_iterator_destroy(rnp_identifier_iterator_t it);
 
+/** Read from input and write to output
+ *
+ *  @param input stream to read data from
+ *  @param output stream to write data to
+ *  @return RNP_SUCCESS on success, or any other value on error
+ */
+rnp_result_t rnp_output_pipe(rnp_input_t input, rnp_output_t output);
```
ni4, line 2315: As of previously merged changes RNP_API should be prepended here and below.

rrrooommmaaa:
@ni4 I made all the requested fixes

ronaldtse:
Ping @ni4 for a quick check.

src/librepgp/stream-armor.cpp
```diff
@@ -979,17 +1001,17 @@ rnp_dearmor_source(pgp_source_t *src, pgp_dest_t *dst)
 {
     rnp_result_t res = RNP_ERROR_BAD_FORMAT;
     pgp_source_t armorsrc = {0};
-    uint8_t      readbuf[PGP_INPUT_CACHE_SIZE];
+    uint8_t      readbuf[strlen(ST_CLEAR_BEGIN) + 1];
     size_t       read;
 
-    if (!src_peek(src, readbuf, strlen(ST_CLEAR_BEGIN) + 1, &read) ||
+    if (!src_peek(src, readbuf, strlen(ST_CLEAR_BEGIN), &read) ||
```
ni4, line 1007: src_peek_eq() could be used here.

src/librepgp/stream-armor.cpp
```diff
@@ -948,6 +952,24 @@ init_armored_dst(pgp_dest_t *dst, pgp_dest_t *writedst, pgp_armored_msg_t msgtyp
...
+bool
+is_armored_dest(pgp_dest_t *dst)
+{
+    return dst->type == PGP_STREAM_ARMORED;
+}
+
+rnp_result_t
+armored_dst_set_line_length(pgp_dest_t *dst, size_t llen)
+{
+    if (!dst || llen < ARMORED_MIN_LINE_LENGTH || llen > ARMORED_MAX_LINE_LENGTH ||
```
ni4, line 964: It is preferred to put separate statements between `||` into brackets for better readability.

rrrooommmaaa:
@ni4 There are several ways to simplify the dearmoring function `rnp_dearmor_source`. Here's one (please see the latest commit).
Or we can even add a function `src_peek_string_eq`?

ni4:
@rrrooommmaaa Commit 3001f87 will change behavior and break things in following cases:
 - armored/cleartext message may be smaller then 128 bytes
 - there could be some addition whitespaces/eols before the actual header begin
So it should be better reverted, or moved to the separate issue and PR, covering all of these edge cases with tests.

PR #1178, opened 2020-06-25
Remove unused regex include from the rnp_key_store.cpp

ni4:
As title says.


ronaldtse:
Thanks @ni4 @dewyatt !

PR #1195, opened 2020-07-06
Add the fuzzing directory into CMake and add two simple fuzz targets

tomrittervg:
I'm compiling this in a bionic Docker container with `cmake -DENABLE_SANITIZERS=1 -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_SHARED_LIBS=on -DBUILD_TESTING=off ../rnp/`


ronaldtse:
@dewyatt could you have a look at this? This is currently a blocker for running OSS-Fuzz (thank you @tomrittervg !). In any case we will need this fixed prior to audit. Please let us know the priority of this in comparison to the other things we need to do. Thanks!

tomrittervg:
I updated the patch to put the fuzzers behind -DENABLE_FUZZERS=1 which makes all the builds pass :)

src/fuzzing/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
+# Copyright (c) 2018 Ribose Inc.
```
ni4, line 1: Probably, 2020 (and below instead of 2019)?

ronaldtse:
Moved the TODO to https://github.com/rnpgp/rnp/pull/1195

PR #1198, opened 2020-07-07
simplify dearmor

### LINKED ISSUE #1197: Simplify `rnp_dearmor_source`
## Description

As per discussion in PR #1177

rrrooommmaaa:
Closes #1197

### DIFF
--- a/src/librepgp/stream-armor.cpp
+++ b/src/librepgp/stream-armor.cpp
@@ -973,27 +973,23 @@ armored_dst_set_line_length(pgp_dest_t *dst, size_t llen)
 bool
 is_armored_source(pgp_source_t *src)
 {
-    uint8_t buf[128];
-    size_t  read = 0;
+    uint8_t buf[strlen(ST_ARMOR_BEGIN)];
 
-    if (!src_peek(src, buf, sizeof(buf), &read) || (read < strlen(ST_ARMOR_BEGIN) + 1)) {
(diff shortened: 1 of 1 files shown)


src/librepgp/stream-armor.cpp
```diff
@@ -973,48 +973,35 @@ armored_dst_set_line_length(pgp_dest_t *dst, size_t llen)
 bool
 is_armored_source(pgp_source_t *src)
 {
-    uint8_t buf[128];
-    size_t  read = 0;
+    uint8_t buf[strlen(ST_ARMOR_BEGIN)];
```
rrrooommmaaa, line 976: @ni4 you were saying previously that there might be EOL and other whitespace prepending `ST_ARMOR_BEGIN`, this is why `strstr` is used to determine the type of source? And this whitespace is limited by 100-something bytes? Is this a limitation by the protocol?

ni4: @rrrooommmaaa Yep. This is not a protocol limitation but heuristic, allowing to work in most of the cases. 128 bytes are picked to be enough for most of the cases.

ni4:
@rrrooommmaaa Would you have time to update this so we have less opened/abandoned PRs?

ni4:
Closing this as was resolved via the 4a2d42d9591ecf7428a738247181c988f6fc0ff2

PR #1203, opened 2020-07-13
Add FFI function to export autocrypt-compatible key.

ni4:
While this first planned as part of signature stripping work, I decided to do it as separate function due to requirements of autocrypt: be able to export single key with single uid and single subkey (see https://github.com/rnpgp/rnp/issues/1006#issuecomment-647536642).

Part if the issue #1006, and related to still draft PR #1136


ronaldtse:
Ping for an extra approval, @dewyatt @rrrooommmaaa

PR #1205, opened 2020-07-14
Correctly handle cases with non-detached signature input.

ni4:
This PR fixes possible crash, when non-detached input is fed to `rnp_op_verify_detached_create()`/`rnp_op_verify_execute()`.
Also it adds check whether detached input is expected in `process_pgp_source()`.


ronaldtse:
Ping for an extra review @dewyatt @rrrooommmaaa @antonsviridenko (all checks are green: the Hound build has hang and there's no way to restart it)

Issue #1181, opened 2020-06-25
Memory overhead for in-memory keyring is > factor 130


kaie:
I have a migrated keyring from GnuPG, which has 980 public keys in it.
Size of the file on disk is: 53 MB

In Thunderbird I'd like to load the keyring once using rnp_load_keys and keep it in memory, and work with is as necessary.

However, RNP requires a lot of additional memory to hold those keys.

I've compared loading the keyring, with the application's memory footprint when not loading it. With loading enabled, after loading, the virtual memory size is 7000 MB more, and the resident memory size is 5500 MB more.

I reproduced this using the rnpkeys application. Using rnpkeys --list-keys, I see similar numbers of memory consumption by the rnpkeys process.

Do you understand why RNP requires that much memory to hold the keys? Are there leaks? Or do you build extensive data structures? Even if you did, it's not clear why you would need that much memory.

@pbrunschwig @mkmelin

kaie:
May keys contain compressed data? Could I have poisoned keys, downloaded from keyservers, which are bloated?

kaie:
Using https://tech.michaelaltfield.net/2019/07/14/mitigating-poisoned-pgp-certificates/ for analysis, I don't seem to have poisoned keys. The largest key is 2047355 bytes, and it doesn't have a poison-like peak of signatures on a given date. Most is 280. Probably someone having attended key signing parties.

ni4:
@kaie Thanks for reporting, I see the comparable behavior when loading the huge keyring. We should not have such significant memory leaks since all tests are also run with allocation/memory sanitizers.

Most likely it is related to large memory overhead on C unions, used to store signature/signature subpacket/key info.
Since it was originally C code, instead of dynamic allocation we used statically sized mpi for those types, so current sizes of those are: `sig size: 4176, key size: 12720, subpkt size: 4200`. I.e. for 1000 keys, with 20 signatures each, and at least 3 subpackets each, that would be around 500Mb of memory. Multiplied by 2 since of public + secret keyring, and again by 2 because imported keys are stored in temporary keyring, that would give the same magnitude of memory as you've got.

Definitely requires fixing.

ni4:
@kaie PR #1207 with memory-related improvements is merged, so could you please check, once you have time, whether loading of huge keyring doesn't consume a lot of memory now? In my tests on macOS, keyring which used to occupy 2-3Gb, now takes 150-170Mb of memory.

kaie:
[...]

PR #1207, opened 2020-07-16
Reduce memory usage in library

ni4:
This PR is aimed to (at least partially) fix issue #1181 
It includes the following changes:
- use pointer to signature instead of static field for embedded signature subpacket (effectively reducing pgp_sig_subpkt_t size).
- parse signature material to corresponding structure only on validation: this would not add much time overhead, but save a bunch of kilobytes for the each signature
- use std::vector instead of list for signatures/signature subpackets, reducing memory allocations and fragmentation

### DIFF
--- a/src/lib/crypto/signatures.cpp
+++ b/src/lib/crypto/signatures.cpp
@@ -26,6 +26,7 @@
 
 #include <string.h>
 #include "crypto/signatures.h"
+#include "librepgp/stream-packet.h"
 #include "utils.h"
 
 /**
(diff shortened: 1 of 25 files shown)


ni4:
@dewyatt Do you have any idea about these errors:
` /__w/rnp/rnp/src/librekey/key_store_pgp.cpp:221: undefined reference to `std::vector<pgp_revoke_t, std::allocator<pgp_revoke_t> >::~vector()'`, which are fields of pgp_key_t struct ? 
I tried to think here and there and do not understand that at all. Sources seem to include pgp-key.h with all template declarations and so on, and it actually worked before and works on other builds.

ni4:
I extracted erroring commits to the separate branch, to take care of them later.

antonsviridenko:
How does using references instead of pointers in function arguments help in reducing memory usage?

ni4:
> How does using references instead of pointers in function arguments help in reducing memory usage?

@antonsviridenko  It doesn't affect memory usage, just refactors code where it comes close to other changes.

ni4:
@antonsviridenko @dewyatt @rrrooommmaaa Could we have merged this, so code audit will include these changes?

ronaldtse:
Ping @dewyatt @rrrooommmaaa @antonsviridenko for review, hoping to merge this ASAP for the audit. Thanks!

ni4:
@antonsviridenko Thanks! Feel free to 'rebase and merge'.

ni4:
Np, merging this since we have two approvals.

dewyatt:
> @dewyatt Do you have any idea about these errors:
> `` /__w/rnp/rnp/src/librekey/key_store_pgp.cpp:221: undefined reference to `std::vector<pgp_revoke_t, std::allocator<pgp_revoke_t> >::~vector()' ``, which are fields of pgp_key_t struct ?
> I tried to think here and there and do not understand that at all. Sources seem to include pgp-key.h with all template declarations and so on, and it actually worked before and works on other builds.

I just looked at this - maybe a bug in llvm 7, not sure, but a workaround might be to replace `pgp_key_t key = {};` with `pgp_key_t key;`, and `*key = {};` with `*key = pgp_key_t();`.

ni4:
> I just looked at this - maybe a bug in llvm 7, not sure, but a workaround might be to replace pgp_key_t key = {}; with pgp_key_t key;, and *key = {}; with *key = pgp_key_t();.

@dewyatt Thanks for the suggestion, it worked!

PR #1212, opened 2020-07-21
Add exception guard to FFI boundary.

### LINKED ISSUE #839: Do not allow exceptions beyond the FFI boundary
## Description

As we continue to transition to a more C++ based codebase, we are going to end up using things that may throw exceptions. We need to ensure that we catch exceptions within the FFI functions.

We can look at Botan's `ffi_guard_thunk` as one possible solution.

dewyatt:
This uses function try blocks to prevent C++ exceptions from propagating
across the FFI boundary. Some functions which return direct values
(rather than error codes) and do not throw are left as-is.

For now it just prints directly to `stderr` and uses a format similar to `RNP_LOG`, which I think is a good compromise considering that it should be a very rare occurrence.

Fixes #839

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -414,9 +414,35 @@ get_map_value(const pgp_map_t *map, size_t msize, int val, char **res)
     return RNP_SUCCESS;
 }
 
+static uint32_t
+ffi_exception(FILE *fp, const char *func, const char *msg, uint32_t ret = RNP_ERROR_GENERIC)
+{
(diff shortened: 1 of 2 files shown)


dewyatt:
> Maybe it would be nice to add some tests that throw various exceptions and check that everything works as expected

Added some tests. It only tests against a single FFI function really but it does test a few different exception types.

ni4:
@antonsviridenko Feel free to rebase and merge once there are two or more approvals.

PR #1222, opened 2020-07-28
Fix homebrew CI problem for macOS.

ni4:
As title said.
Issue reference: https://github.com/Homebrew/homebrew-bundle/issues/751


ni4:
Merging without reviews and despite the hanged Windows build since it's just only CI, and macos-related.

ronaldtse:
Thanks @ni4 !

PR #1223, opened 2020-07-28
Swallow the dump output to speed up fuzzing.

tomrittervg:
While this works, it might be possible to do this better but I'm not sure about how to use `rnp_dump_packets_to_output`.

### DIFF
--- a/src/fuzzing/dump.c
+++ b/src/fuzzing/dump.c
@@ -24,16 +24,7 @@
  * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <stdio.h>
-#include <unistd.h>
 #include <rnp/rnp.h>
-
-static bool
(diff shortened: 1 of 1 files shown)


tomrittervg:
I don't believe that test failure :)

src/fuzzing/dump.c
```diff
@@ -43,12 +40,7 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
     ret = rnp_input_from_memory(&input, data, size, false);
 
     rnp_output_t output = NULL;
-    ret = rnp_output_to_callback(&output, stdout_writer, NULL, NULL);
-    if (ret) {
-        fprintf(stderr, "failed to open stdout: error 0x%x\n", (int) ret);
-        rnp_input_destroy(input);
-        return 0;
-    }
+    ret = rnp_output_to_callback(&output, no_writer, NULL, NULL);
```
ni4, line 43: Actually, you may use here `rnp_output_to_null()`.

PR #1225, opened 2020-07-29
Use uint8_t when accessing B64DEC array during dearmor.

ni4:
Accidentally signed char/int was used, which may cause classic AV on invalid data.


ni4:
CC @tomrittervg

PR #1229, opened 2020-07-31
Fix memory leak during parsing of one-pass sig with invalid version.

ni4:
As title says.
CC @tomrittervg


ronaldtse:
2 approvals => merge.

PR #1230, opened 2020-07-31
Check string lengths when using strncmp for minimum bounds.

ni4:
Fix possible errors and out-of-bounds exceptions when processing malformed armored data.


src/librekey/key_store_g10.cpp
```diff
@@ -814,7 +814,8 @@ parse_protected_seckey(pgp_key_pkt_t *seckey, s_exp_t *s_exp, const char *passwo
         goto done;
     }
     sub_el = sub_element_at(alg, 0);
-    if (strncmp("sha1", (const char *) sub_el->block.bytes, sub_el->block.len) != 0) {
+    if ((sub_el->block.len != 4) ||
```
rrrooommmaaa, line 817: If the length is fixed,
it would be more evident to write
`(sub_el->block.len != 4) || memcmp("sha1", sub_el->block.bytes, 4)`
Agree?

src/librekey/key_store_g10.cpp
```diff
@@ -892,9 +893,10 @@ parse_protected_seckey(pgp_key_pkt_t *seckey, s_exp_t *s_exp, const char *passwo
             !sub_element_at(&sub_el->s_exp, 0)->is_block ||
             !sub_element_at(&sub_el->s_exp, 1)->is_block ||
             !sub_element_at(&sub_el->s_exp, 2)->is_block ||
+            (sub_element_at(&sub_el->s_exp, 0)->block.len != 4) ||
```
rrrooommmaaa, line 896: same here. Why not use `memcmp`?

src/librepgp/stream-armor.cpp
```diff
@@ -453,29 +453,35 @@ find_armor_header(const char *buf, size_t len, size_t *hdrlen)
...
+}
+
 static pgp_armored_msg_t
 armor_str_to_data_type(const char *str, size_t len)
 {
     if (!str) {
         return PGP_ARMORED_UNKNOWN;
     }
-    if (!strncmp(str, "BEGIN PGP MESSAGE", len)) {
+    if (str_equals(str, len, "BEGIN PGP MESSAGE")) {
```
rrrooommmaaa, line 468: What's wrong with `strncmp`? It should react on NULL character on both strings.

src/librekey/key_store_g10.cpp
```diff
@@ -965,9 +967,11 @@ g10_parse_seckey(pgp_key_pkt_t *seckey,
     }
 
     block = &sub_element_at(&s_exp, 0)->block;
-    if (!strncmp("private-key", (const char *) block->bytes, block->len)) {
+    if ((block->len == 11) &&
```
rrrooommmaaa, line 970: `memcmp` here?

src/librekey/key_store_g10.cpp
```diff
@@ -965,9 +967,11 @@ g10_parse_seckey(pgp_key_pkt_t *seckey,
...
+        !strncmp("private-key", (const char *) block->bytes, block->len)) {
         is_protected = false;
-    } else if (!strncmp("protected-private-key", (const char *) block->bytes, block->len)) {
+    } else if ((block->len == 21) &&
```
rrrooommmaaa, line 973: `memcmp` here?

src/librekey/key_store_g10.cpp
ni4: @rrrooommmaaa Agree, thanks for pointing to this.

src/librepgp/stream-armor.cpp
ni4: `strncmp()` will return 0 if str is shorter than "BEGIN PGP MESSAGE", or even empty.

ni4:

src/librepgp/stream-armor.cpp
rrrooommmaaa: @ni4 Not sure what you mean.
These tests pass:
```
    assert_int_not_equal(0, strncmp("aaa", "aaaa", 4));
    assert_int_not_equal(0, strncmp("aaaa", "aaa", 4));
    assert_int_equal(0, strncmp("aaaa", "aaaa", 4));
```

src/librepgp/stream-armor.cpp
```diff
@@ -453,29 +453,35 @@ find_armor_header(const char *buf, size_t len, size_t *hdrlen)
     return NULL;
 }
 
+static bool
+str_equals(const char *str, size_t len, const char *another)
+{
+    return (len >= strlen(another)) && !strncmp(str, another, len);
```
rrrooommmaaa, line 459: @ni4 and you should use `memcmp` here too

ni4: @rrrooommmaaa Thanks, fixed and re-pushed.

ronaldtse:
2 approvals => merged. Thanks guys!

PR #1232, opened 2020-08-02
Build with MSVC

### LINKED ISSUE #997: Add Windows workflow which uses Visual Studio instead of MSYS/MinGW
## Description

Currently rnp can be built only using MSYS/MinGW, however we should also provide a way to build and test it using the native Windows tools.

### LINKED ISSUE #1252: Cannot increment list iterator which is already past-the-end
## Description

It was detected that `key_iter_next_key()` is sometimes called again even if the previous call returned `false`.
Visual C++ triggers runtime exception.
https://stackoverflow.com/questions/53471376/cannot-increment-end-list-iterator-run-time-error-visual-c
I suggest to fix by comparing the iterator to `end` prior to incrementing.

### LINKED ISSUE #1259: `rnp_key_store_clear` throws runtime exception in MSVC
## Description

The line  `keyring->keys.clear();`  sometimes triggers runtime exception in MSVC.
Perhaps, some iterator is referencing a deleted object (incorrect order of destruction).
Does anyone have an idea where we could possibly have made a mess with iterators etc.?

```
rnp_tests.exe!std::_Iterator_base12::_Orphan_me() Line 1183
	at C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.26.28801\include\xmemory(1183)
rnp_tests.exe!std::_Iterator_base12::~_Iterator_base12() Line 1145
	at C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.26.28801\include\xmemory(1145)
rnp_tests.exe!std::_List_unchecked_const_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>,std::_Iterator_base12>::~_List_unchecked_const_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>,std::_Iterator_base12>()
rnp_tests.exe!std::_List_const_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>::~_List_const_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>()
rnp_tests.exe!std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>::~_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>()
rnp_tests.exe!std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>::~pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>()
rnp_tests.exe!std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>::`scalar deleting destructor'(unsigned int)
rnp_tests.exe!std::_Default_allocator_traits<std::allocator<std::_List_node<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>,void *>>>::destroy<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>>(std::allocator<std::_List_node<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>,void *>> & __formal, std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>> * const _Ptr) Line 677
	at C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.26.28801\include\xmemory(677)
rnp_tests.exe!std::_List_node<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>,void *>::_Freenode<std::allocator<std::_List_node<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>,void *>>>(std::allocator<std::_List_node<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>,void *>> & _Al, std::_List_node<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>,void *> * _Ptr) Line 318
	at C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.26.28801\include\list(318)
rnp_tests.exe!std::_List_node<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>,void *>::_Free_non_head<std::allocator<std::_List_node<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>,void *>>>(std::allocator<std::_List_node<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>,void *>> & _Al, std::_List_node<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>,void *> * _Head) Line 330
	at C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.26.28801\include\list(330)
rnp_tests.exe!std::list<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>,std::allocator<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>>>::clear() Line 1431
	at C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.26.28801\include\list(1431)
rnp_tests.exe!std::_Hash<std::_Umap_traits<pgp_fingerprint_t,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>,std::_Uhash_compare<pgp_fingerprint_t,std::hash<pgp_fingerprint_t>,std::equal_to<pgp_fingerprint_t>>,std::allocator<std::pair<pgp_fingerprint_t const ,std::_List_iterator<std::_List_val<std::_List_simple_types<pgp_key_t>>>>>,0>>::clear() Line 1282
	at C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.26.28801\include\xhash(1282)
rnp_tests.exe!rnp_key_store_clear(rnp_key_store_t * keyring) Line 227
	at C:\dev\rnp_forks\WINDOWS\rnp\src\librekey\rnp_key_store.cpp(227)
rnp_tests.exe!rnp_key_store_t::~rnp_key_store_t() Line 1027
	at C:\dev\rnp_forks\WINDOWS\rnp\src\librekey\rnp_key_store.cpp(1027)
[External Code]
rnp_tests.exe!rnp_ffi_destroy(rnp_ffi_st * ffi) Line 528
	at C:\dev\rnp_forks\WINDOWS\rnp\src\lib\rnp.cpp(528)
```

## Steps to Reproduce

run `test_ffi_key_iter` test

### LINKED ISSUE #1260: MSVC complains at attempts to close file descriptor twice
Some tests throw assertion failure when trying to close a file descriptor which is already closed:
- generatekeyECDSA_explicitlySetBiggerThanNeededDigest_ShouldSuceed
- generatekeyECDSA_explicitlySetSmallOutputDigest_DigestAlgAdjusted
- generatekeyECDSA_explicitlySetUnknownDigest_ShouldFail
- issue_1030_rnpkeys_secret_keys_unprotected
- rnpkeys_exportkey_verifyUserId
- rnpkeys_generatekey_testEncryption
- rnpkeys_generatekey_testExpertMode
.- rnpkeys_generatekey_testSignature
- rnpkeys_generatekey_verifySupportedHashAlg
- rnpkeys_generatekey_verifyUserIdOption
- rnpkeys_generatekey_verifykeyHomeDirOption
- rnpkeys_generatekey_verifykeyKBXHomeDirOption

I did some debugging on `generatekey_multipleUserIds_ShouldFail()` test.
Looks like `ask_expert_details()` closes the descriptor with
```
    if (pipefd[0]) {
        close(pipefd[0]);
    }
```
Stil, the same descriptor gets into `rnp->passfp` via `CFG_PASSFD`
and there there is an attempt to close `passfp`.

rrrooommmaaa:
Fixes #997 
Fixes #1252 
Fixes #1260 
Fixes #1259 

It builds with a lot of warnings but all tests pass.

FYI, WindowsSDK is required solely for `PathIsRelative()` function for `support.cpp` in `rnp_tests`

### DIFF
--- a/include/rnp/rnp_sdk.h
+++ b/include/rnp/rnp_sdk.h
@@ -41,13 +41,7 @@ extern "C" {
 
 #include <rnp/rnp_def.h>
 
-#ifndef PRINTFLIKE
-#define PRINTFLIKE(n, m) __attribute__((format(printf, n, m)))
-#endif
-
 typedef enum { RNP_HEX_LOWERCASE, RNP_HEX_UPPERCASE } rnp_hex_format_t;
-
-void rnp_log(const char *, ...) PRINTFLIKE(1, 2);
 
 int rnp_strcasecmp(const char *, const char *);
 
--- a/src/common/uniwin.h
+++ b/src/common/uniwin.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017-2020 [Ribose Inc](https://www.ribose.com).
+ * Copyright (c) 2020 [Ribose Inc](https://www.ribose.com).
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without modification,
@@ -23,36 +23,38 @@
  * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
  * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-/** File utilities
- *  @file
+#if !defined(_UNISTD_H) && defined(_MSC_VER)
+#define _UNISTD_H 1
+
+/* Taken partially from
+ * https://stackoverflow.com/a/826027/1202830
  */
 
-#include "file-utils.h"
+#include <io.h>
+#include "getoptwin.h"
+#include <direct.h> /* for _getcwd() and _chdir() */
 
-extern "C" {
-#ifdef _MSC_VER
-#include "uniwin.h"
+#ifdef _WIN64
+#define ssize_t __int64
 #else
-#include <sys/stat.h>
+#define ssize_t long
 #endif
-}
 
-bool
-rnp_file_exists(const char *path)
-{
-    struct stat st;
-    return stat(path, &st) == 0 && S_ISREG(st.st_mode);
-}
+#define STDIN_FILENO 0
+#define STDOUT_FILENO 1
+#define STDERR_FILENO 2
 
-/* return the file modification time */
-int64_t
-rnp_filemtime(const char *path)
-{
-    struct stat st;
+#define NOMINMAX 1 /* to retain std::min and std::max */
+#include <windows.h>
+#include <dirent.h> /* for S_ISREG and S_ISDIR */
 
-    if (stat(path, &st) != 0) {
-        return 0;
-    } else {
-        return st.st_mtime;
-    }
-}
+#define strncasecmp strnicmp
+#define strcasecmp stricmp
+
+#ifndef MAXPATHLEN
+#define MAXPATHLEN _MAX_PATH
+#endif
+
+typedef unsigned short mode_t;
+
+#endif
--- a/src/lib/crypto/mpi.cpp
+++ b/src/lib/crypto/mpi.cpp
@@ -40,12 +40,6 @@ to_buf(buf_t *b, const uint8_t *in, size_t len)
     return true;
 }
 
-const buf_t
-mpi2buf(pgp_mpi_t *val, bool uselen)
-{
-    return (buf_t){.pbuf = val->mpi, .len = uselen ? val->len : sizeof(val->mpi)};
-}
-
 bignum_t *
 mpi2bn(const pgp_mpi_t *val)
 {
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -45,7 +45,12 @@
 #include <rnp/rnp.h>
 #include <stdarg.h>
 #include <stdlib.h>
+#ifdef _MSC_VER
+#include "uniwin.h"
+#include <inttypes.h>
+#else
 #include <unistd.h>
+#endif
 #include <string.h>
 #include <sys/stat.h>
 #include <stdexcept>
@@ -487,9 +492,10 @@ try {
         goto done;
     }
 
-    ob->key_provider = (pgp_key_provider_t){.callback = ffi_key_provider, .userdata = ob};
-    ob->pass_provider =
-      (pgp_password_provider_t){.callback = rnp_password_cb_bounce, .userdata = ob};
+    ob->key_provider.callback = ffi_key_provider;
+    ob->key_provider.userdata = ob;
+    ob->pass_provider.callback = rnp_password_cb_bounce;
+    ob->pass_provider.userdata = ob;
     if (!rng_init(&ob->rng, RNG_DRBG)) {
         ret = RNP_ERROR_RNG;
         goto done;
@@ -7378,20 +7384,25 @@ static bool
 key_iter_next_key(rnp_identifier_iterator_t it)
 {
     // check if we not reached the end of the ring
-    it->keyp = std::next(it->keyp);
-    if (it->keyp != it->store->keys.end()) {
+    *it->keyp = std::next(*it->keyp);
+    if (*it->keyp != it->store->keys.end()) {
         it->uididx = 0;
         return true;
     }
     // if we are currently on pubring, switch to secring (if not empty)
-    if (it->store == it->ffi->pubring && rnp_key_store_get_key_count(it->ffi->secring)) {
+    if (it->store == it->ffi->pubring && !it->ffi->secring->keys.empty()) {
         it->store = it->ffi->secring;
-        it->keyp = it->store->keys.begin();
-    } else {
-        // we've gone through both rings
-        return false;
+        *it->keyp = it->store->keys.begin();
+        it->uididx = 0;
+        return true;
     }
-    return true;
+    // we've gone through both rings
+    it->store = NULL;
+    if (it->keyp) {
+        delete it->keyp;
+        it->keyp = NULL;
+    }
+    return false;
 }
 
 // move to next item (key or userid)
@@ -7405,7 +7416,7 @@ key_iter_next_item(rnp_identifier_iterator_t it)
         return key_iter_next_key(it);
     case PGP_KEY_SEARCH_USERID:
         it->uididx++;
-        while (it->uididx >= pgp_key_get_userid_count(&*it->keyp)) {
+        while (it->uididx >= pgp_key_get_userid_count(&**it->keyp)) {
             if (!key_iter_next_key(it)) {
                 return false;
             }
@@ -7430,7 +7441,7 @@ key_iter_first_key(rnp_identifier_iterator_t it)
         it->store = NULL;
         return false;
     }
-    it->keyp = it->store->keys.begin();
+    it->keyp = new std::list<pgp_key_t>::iterator(it->store->keys.begin());
     it->uididx = 0;
     return true;
 }
@@ -7447,12 +7458,11 @@ key_iter_first_item(rnp_identifier_iterator_t it)
         if (!key_iter_first_key(it)) {
             return false;
         }
-        while (it->uididx >= pgp_key_get_userid_count(&*it->keyp)) {
+        it->uididx = 0;
+        while (it->uididx >= pgp_key_get_userid_count(&**it->keyp)) {
             if (!key_iter_next_key(it)) {
-                it->store = NULL;
                 return false;
             }
-            it->uididx = 0;
         }
         break;
     default:
@@ -7465,7 +7475,7 @@ key_iter_first_item(rnp_identifier_iterator_t it)
 static bool
 key_iter_get_item(const rnp_identifier_iterator_t it, char *buf, size_t buf_len)
 {
-    const pgp_key_t *key = &*it->keyp;
+    const pgp_key_t *key = &**it->keyp;
     switch (it->type) {
     case PGP_KEY_SEARCH_KEYID: {
         const pgp_key_id_t &keyid = pgp_key_get_keyid(key);
@@ -7572,8 +7582,9 @@ try {
         return RNP_ERROR_GENERIC;
     }
     bool exists;
+    bool iterator_valid = true;
     while ((exists = json_object_object_get_ex(it->tbl, it->buf, NULL))) {
-        if (!key_iter_next_item(it)) {
+        if (!((iterator_valid = key_iter_next_item(it)))) {
             break;
         }
         if (!key_iter_get_item(it, it->buf, sizeof(it->buf))) {
@@ -7592,9 +7603,8 @@ try {
         *identifier = it->buf;
     }
     // prepare for the next one
-    if (!key_iter_next_item(it)) {
-        // this means we're done
-        it->store = NULL;
+    if (iterator_valid) {
+        key_iter_next_item(it);
     }
     ret = RNP_SUCCESS;
 
@@ -7611,6 +7621,9 @@ rnp_identifier_iterator_destroy(rnp_identifier_iterator_t it)
 try {
     if (it) {
         json_object_put(it->tbl);
+        if (it->keyp) {
+            delete it->keyp;
+        }
         free(it);
     }
     return RNP_SUCCESS;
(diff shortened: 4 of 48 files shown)


ronaldtse:
@rrrooommmaaa only one build failure left!! 😉

rrrooommmaaa:
@ronaldtse There are several tests failures yet, because MSVC seems to be more stringent in some cases.
I have opened two issues #1252 , #1259
Investigating into other issues.

ronaldtse:
@rrrooommmaaa thank you for the significant work! We probably should get this merged soon prior to conflicts... Is this ready for review?

.github/workflows/msvc.yml
```diff
@@ -0,0 +1,66 @@
+name: build
```
ronaldtse, line 1: @rrrooommmaaa you should change the `name:` value to `windows-msvc`? Better also to rename the file to `windows-mdvc.yml` for consistency (when we fetch the GHA badge).

rrrooommmaaa: Done

src/tests/generatekey.cpp
```diff
@@ -149,7 +151,10 @@ TEST_F(rnp_tests, rnpkeys_generatekey_testSignature)
 
                 /* Sign the file */
                 assert_true(cli_rnp_protect_file(&rnp));
-                close(pipefd[0]);
+                if (pipefd[0]) {
+                    close(pipefd[0]);
+                    pipefd[0] = 0;
```
antonsviridenko, line 156: 0 is a valid file descriptor in POSIX so it is not correct to use 0 here. While it is possible that standard input file descriptor is never closed in this context and never returned again as a pipe descriptor, but anyway -1 should be used instead. And all code pieces like `if (pipefd[0])` should be changed to `if (pipefd[0] != -1)`

include/rnp/rnp_sdk.h
```diff
@@ -41,13 +41,21 @@ extern "C" {
...
+#if _MSC_VER > 1400
+void rnp_log(_Printf_format_string_ const char *, ...);
+#else
+void rnp_log(__format_string const char *, ...);
+#endif /* FORMAT_STRING */
+#else  /* _MSC_VER */
+#ifndef PRINTFLIKE
+#define PRINTFLIKE(n, m) __attribute__((format(printf, n, m)))
+#endif /* PRINTFLIKE */
 void rnp_log(const char *, ...) PRINTFLIKE(1, 2);
```
ni4, line 57: We do not use this `rnp_log()` anywhere, it seems to be an artifact of the ancient times. So it, and PRINTFLIKE could be safely deleted.

src/common/uniwin.h
```diff
@@ -0,0 +1,62 @@
...
+#define _UNISTD_H 1
+
+/* This is intended as a drop-in replacement for unistd.h on Windows.
+ * Please add functionality as neeeded.
+ * https://stackoverflow.com/a/826027/1202830
+ */
+
+#include <stdlib.h>
+#include <io.h>
+#include "getoptwin.h"
```
ni4, line 11: Why do we need it as a separate file, while it is just `#include <getopt.h>`?

src/common/uniwin.h
```diff
@@ -0,0 +1,62 @@
...
+#include <process.h> /* for getpid() and the exec..() family */
+#include <direct.h>  /* for _getcwd() and _chdir() */
+
+#define srandom srand
```
ni4, line 15: srandom, random, R_OK, and many of other defines are not used anywhere in the codebase, so we probably should remove them.

src/lib/crypto/mpi.cpp
```diff
@@ -43,7 +43,11 @@ to_buf(buf_t *b, const uint8_t *in, size_t len)
 const buf_t
 mpi2buf(pgp_mpi_t *val, bool uselen)
 {
-    return (buf_t){.pbuf = val->mpi, .len = uselen ? val->len : sizeof(val->mpi)};
+    return
```
ni4, line 46: Hm, don't know how these survived code changes, but type buf_t and mpi2buf and all other functions are not used anywhere so could be safely removed.

src/lib/rnp.cpp
```diff
@@ -487,9 +492,16 @@ try {
         goto done;
     }
 
-    ob->key_provider = (pgp_key_provider_t){.callback = ffi_key_provider, .userdata = ob};
+    ob->key_provider =
+#ifndef _MSC_VER
+      (pgp_key_provider_t)
```
ni4, line 497: Just curious - which error does this throw? Seems to be legit code.

src/librepgp/stream-packet.cpp
```diff
@@ -1374,169 +1378,172 @@ signature_parse_subpacket(pgp_sig_subpkt_t &subpkt)
     bool oklen = true;
     bool checked = true;
 
-    switch (subpkt.type) {
```
ni4, line 1377: Do I correctly understand, MSVC doesn't allow range case labels? Then it would be cleaner to change `PGP_SIG_SUBPKT_PRIVATE_FIRST` and `PGP_SIG_SUBPKT_PRIVATE_LAST` to `PGP_SIG_SUBPKT_PRIVATE_100, 101... 110`, and use 10 case labels. This would also be useful together with undefined behavior sanitizer.

src/lib/rnp.cpp
rrrooommmaaa: This is legit C code and MSVC compiles it only if the extension is `.c`.
So we either have to move it out to `.c ` files or ... like this

src/common/uniwin.h
rrrooommmaaa: The distirbuted `getopt.h` file undefines `_BEGIN_EXTERN_C` and `_END_EXTERN_C` for some reason
and this may result in errors if those macros are required below the includsion point

src/librepgp/stream-packet.cpp
rrrooommmaaa: Exactly. Range switch turned out to be a non-standard feature.

src/lib/rnp.cpp
ni4: Another way is to change it to the following:
```
ob->key_provider.callback = ffi_key_provider;
ob->key_provider.userdata = ob;
```
which would do the same and take less code lines/look cleaner.

src/common/uniwin.h
rrrooommmaaa: @ni4 I have fixed other issues

src/tests/generatekey.cpp
rrrooommmaaa: Thanks for letting me know. I have fixed that.

rrrooommmaaa:
@ronaldtse I added a couple of files copied or based on from other people's projects: `FindWindowsSDK.cmake` and `uniwin.h`
Should we add Ribose copyright notice there?

antonsviridenko (approved):

src/common/uniwin.h
ni4: @rrrooommmaaa Great, thanks!

ni4 (commented):
LGTM, but let's wait for the answer to question about the copyright from @ronaldtse .

rrrooommmaaa:
ping @ronaldtse whether copyright notice is appropriate in `FindWindowsSDK.cmake` and `uniwin.h`

ronaldtse:
Sorry for the late reply. `uniwin.h` is fine to add the copyright statement. 

For `FindWindowsSDK.cmake`, because it is under the Boost license, we don't need to apply our own license, and must maintain its license in the file. As long as we don't change its header it is fine.

Thanks!

kaie:
s2k.cpp does not build in thunderbird's environment on Linux. 

error: use of undeclared identifier 'GetTickCount64'

kaie:
This builds:

```
diff --git a/third_party/rnp/src/lib/crypto/s2k.cpp b/third_party/rnp/src/lib/crypto/s2k.cpp
--- a/third_party/rnp/src/lib/crypto/s2k.cpp
+++ b/third_party/rnp/src/lib/crypto/s2k.cpp
@@ -31,7 +31,7 @@
 #include <botan/ffi.h>
 #include <stdio.h>
 #include "config.h"
-#ifdef HAVE_SYS_TIME_H
+#if !defined(_MSC_VER) || defined(HAVE_SYS_TIME_H)
 #include <sys/time.h>
 #else
 #include "uniwin.h"
@@ -153,7 +153,7 @@ pgp_s2k_encode_iterations(size_t iterati
 static uint64_t
 get_timestamp_usec()
 {
-#ifdef HAVE_SYS_TIME_H
+#if !defined(_MSC_VER) || defined(HAVE_SYS_TIME_H)
     // TODO: Consider clock_gettime
     struct timeval tv;
     ::gettimeofday(&tv, NULL);
```

rrrooommmaaa:
@kaie Would `
::GetTickCount64` be better?

kaie:
@rrrooommmaaa Why? GetTickCount64 appears to be a windows-only function, it's not available on Linux. The build rules need to ensure that the gettimeofday code is used on platforms other than Windows. This commit changed the code to fall back to GetTickCount64 whenever HAVE_SYS_TIME_H is undefined. It seems that the configure rules don't detect that it's available on Linux, and should be used.

rrrooommmaaa:
@kaie Right, I had no idea you're using MSVC on Linux

kaie:
We're not using MSVC on Linux !

rrrooommmaaa:
So, the problem is: `HAVE_SYS_TIME_H` is unset, whereas `sys/time.h` actually exists

kaie:
Roman, let me try to explain (what you did) it differently:

Previously, function get_timestamp_usec always called ::gettimeofday
This worked for us on all platforms that we're using, apparently.

Your patch changed that to say:
"Only use ::gettimeofday if HAVE_SYS_TIME_H is defined"

Linux has file sys/time.h
The symbol HAVE_SYS_TIME_H should be defined by the automatic configuration scripts you are using.
Apparently your config scripts have a bug, and don't find sys/time.h, and keep HAVE_SYS_TIME_H undefined.

Because HAVE_SYS_TIME_H is undefined, you are incorrectly falling back to the Windows code.

To fix this, you can either try to fix your configure scripts, and attempt to detect the presence of sys/time.h reliably.
(I don't know how to do that.)

The alternative solution (which I proposed above) is:
If you're not on Windows, always use the old code (include sys/time.h and use gettimeofday).

rrrooommmaaa:
@kaie I understand this completely. Obviously, fixing configuration scripts is the right thing to do.
As for use old code on non-Windows, it can be done simply `#ifndef _MSC_VER`

rrrooommmaaa:
To be more exact, use old code on non-MSVC

rrrooommmaaa:
I'll prepare a PR

ni4:
Yeah, we definitely should not use Windows-related functions as `#else` for other defines - those should be used only under Windows-related `#ifdef`'s.

jfx2006:
I've updated the Thunderbird build to check for sys/time.h so it will be set appropriately going forward. https://phabricator.services.mozilla.com/D90169

PR #1233, opened 2020-08-03
Update windows workflow to use msys2/setup-msys2@v2

ni4:
As title says.


ronaldtse:
Purely build related, merging with one approval.

Issue #1217, opened 2020-07-23
cannot import the result of gpg --export-secret-subkeys


kaie:
We received a request to support a configuration that uses an offline primary key.
We don't need to support generating such a scenario.
However, we want to support a scenario, in which the user uses gnupg to create their keys, and then imports all public keys, and then imports the secret keys of all subkeys, but not the secret key of the primary key.

Procedure used:
gpg --full-gen-key
select 4 (rsa sign only)

gpg --expert --edit-key
addkey
4
addkey
6
save

gpg --export --armor > pub.asc
gpg --export-secret-subkeys --armor > subkeys-sec.asc

Import pub.asc into Thunderbird -> works
Import subkeys-sec.asc into Thunderbird -> fails

Errors on console:
[get_packet_body_s2k() /home/user/moz/comm-esr78/mozilla/comm/third_party/rnp/src/librepgp/stream-packet.cpp:583] unknown s2k specifier
[stream_parse_key() /home/user/moz/comm-esr78/mozilla/comm/third_party/rnp/src/librepgp/stream-packet.cpp:2201] failed to read key protection
[process_pgp_key() /home/user/moz/comm-esr78/mozilla/comm/third_party/rnp/src/librepgp/stream-key.cpp:996] failed to parse key pkt at 0

Output of pgpdump of the secret key file starts with:

Old: Secret Key Packet(tag 5)(405 bytes)
        Ver 4 - new
        Public key creation time - Thu Jul 23 18:13:55 CEST 2020
        Pub alg - RSA Encrypt or Sign(pub 1)
        RSA n(3072 bits) - ...
        RSA e(17 bits) - ...
        Sym alg - Plaintext or unencrypted data(sym 0)
        GnuPG gnu-dummy (s2k 1001)

dewyatt:
Sounds like we should bump the priority of https://github.com/rnpgp/rnp/issues/1147 up

ni4:
@dewyatt Agree. I'll handle these, it should not be hard to parse this custom s2k. The only question is whether rnp will be okay with secret key without secret data.

kaie:
It's not yet fully working as expected.

I can import the subkeys. (I use a temp FFI handle.)

For the primary key, rnp_key_have_secret returns true. Shouldn't it return false?

Calling rnp_key_unprotect returns RNP_ERROR_BAD_PASSWORD, that might be ideal either, I'd rather expect something like RNP_ERROR_KEY_NOT_FOUND.

[...]

PR #1248, opened 2020-08-07
Add FFI functions to retrieve secret key protection info.

ni4:
This PR adds a bunch of FFI functions, which allows to get information about the key protection (and whether secret key is available at all, like cases where secret key is offloaded or stored on the smartcard).

Related issue #1217

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -1349,6 +1349,85 @@ RNP_API rnp_result_t rnp_key_is_retired(rnp_key_handle_t key, bool *result);
  **/
 RNP_API rnp_result_t rnp_key_is_locked(rnp_key_handle_t key, bool *result);
 
+/**
+ * @brief Get type of protection, used for secret key data.
+ *
+ * @param key key handle, cannot be NULL and should have secret part (see function
+ *            rnp_key_have_secret()).
+ * @param type on success protection type will be stored here. Cannot be NULL.
+ *             Must be freed by caller via rnp_buffer_destroy() call.
+ *             Currently defined values are:
+ *             - "None" : secret key data is stored in plaintext.
+ *             - "Encrypted" : secret key data is encrypted, using just CRC as integrity
+ *                 protection.
+ *             - "Encrypted-Hashed" : secret key data is encrypted, using the SHA1 hash as
+ *                 an integrity protection.
+ *             - "GPG-None" : secret key data is not available at all (this would happen if
+ *                 secret key is exported from GnuPG via --export-secret-subkeys)
+ *             - "GPG-Smartcard" : secret key data is stored on smartcard by GnuPG, so is not
+ *                 available
+ * @return RNP_SUCCESS on success, or any other value on error
+ */
+RNP_API rnp_result_t rnp_key_get_protection_type(rnp_key_handle_t key, char **type);
+
+/**
+ * @brief Get mode in which secret key data is encrypted.
+ *
+ * @param key key handle, cannot be NULL and should have secret part (see function
+ *            rnp_key_have_secret()).
+ * @param mode on success secret key protection mode name will be stored here. Cannot be NULL.
+ *             Must be freed by caller via rnp_buffer_destroy() call.
+ *             Currently defined values are:
+ *             - "None" : secret key data is not encrypted at all
+ *             - "Unknown" : it is not known how secret key data is encrypted, so there is no
+ *                 way to unlock/unprotect the key.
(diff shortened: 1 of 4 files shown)


src/lib/rnp.cpp
```diff
@@ -6164,6 +6178,130 @@ try {
...
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+    if (!pgp_key_has_encryption_info(key->sec)) {
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+
+    if (key->sec->pkt.sec_protection.s2k.specifier == PGP_S2KS_ITERATED_AND_SALTED) {
+        *iterations = pgp_s2k_decode_iterations(key->sec->pkt.sec_protection.s2k.iterations);
+    } else {
+        *iterations = 1;
```
dewyatt, line 6299: I think 0 or an error would make more sense.

ni4: @dewyatt here the only cases left for now are `PGP_S2KS_SIMPLE` and `PGP_S2KS_SALTED`, which use just 1 iteration of hashing. Does this make sense, or we should still return an error for such case?

include/rnp/rnp.h
```diff
@@ -1349,6 +1349,81 @@ RNP_API rnp_result_t rnp_key_is_retired(rnp_key_handle_t key, bool *result);
  **/
 RNP_API rnp_result_t rnp_key_is_locked(rnp_key_handle_t key, bool *result);
 
+/**
+ * @brief Get type of protection, used for secret key data.
+ *
+ * @param key key handle, cannot be NULL and should have secret part (see function
+ *            rnp_key_have_secret()).
+ * @param type on success protection type will be stored here. Cannot be NULL.
```
antonsviridenko, line 1357: When I read this, I thought the buffer should be allocated by the caller. But I see in the source that buffer is allocated using strdup() inside the function. So it should be explicitly mentioned that buffer is allocated by the function and that it should be released by rnp_buffer_destroy().

Also, is there a real reason for duplicating these strings - "None", "Encrypted", etc? They are static string literals, and therefore just pointer to them could be returned instead, and type parameter can be declared as `const char **type`. Is there any need that caller could modify contents of returned buffer later? This optimization could reduce memory usage.

include/rnp/rnp.h
```diff
@@ -1349,6 +1349,81 @@ RNP_API rnp_result_t rnp_key_is_retired(rnp_key_handle_t key, bool *result);
...
+ * @return RNP_SUCCESS on success, or any other value on error
+ */
+RNP_API rnp_result_t rnp_key_get_protection_type(rnp_key_handle_t key, char **type);
+
+/**
+ * @brief Get mode in which secret key data is encrypted.
+ *
+ * @param key key handle, cannot be NULL and should have secret part (see function
+ *            rnp_key_have_secret()).
+ * @param mode on success secret key protection mode name will be stored here. Cannot be NULL.
```
antonsviridenko, line 1377: same here

include/rnp/rnp.h
```diff
@@ -1349,6 +1349,81 @@ RNP_API rnp_result_t rnp_key_is_retired(rnp_key_handle_t key, bool *result);
...
+ */
+RNP_API rnp_result_t rnp_key_get_protection_mode(rnp_key_handle_t key, char **mode);
+
+/**
+ * @brief Get cipher, used to encrypt secret key data.
+ *        Note: this call will return an error if secret key data is not available or secret
+ *        key is not encrypted.
+ *
+ * @param key key handle, cannot be NULL and should have secret part.
+ * @param cipher on success cipher name will be stored here. See
```
antonsviridenko, line 1397: see previous comments

src/lib/rnp.cpp
dewyatt: @ni4 I thought I responded, but I think I missed the call to `pgp_key_has_encryption_info` and was thinking about the case of `PGP_S2KU_NONE`, so this is ok as-is.

include/rnp/rnp.h
ni4: @antonsviridenko Thanks for pointing at this, forgot to mention that string should be deallocated. Will fix it.

In all other function calls string constants (like algorithm names) are also duplicated. As I understand, the idea behind this is that caller shouldn't have any direct interpretable pointers to library internal things - all returned values are owned by caller.

ni4:
@antonsviridenko Please see the updated PR.

ronaldtse:
@dewyatt if approved please feel free to merge, thanks!

PR #1253, opened 2020-08-11
don't iterate when already past-the-end

### LINKED ISSUE #1252: Cannot increment list iterator which is already past-the-end

rrrooommmaaa:
Fixes #1252

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -7372,6 +7372,10 @@ FFI_GUARD
 static bool
 key_iter_next_key(rnp_identifier_iterator_t it)
 {
+    // emulate MSVC behaviour
+    if (it->keyp == it->store->keys.end()) {
+        throw new std::runtime_error("iterating past end of list");
+    }
     // check if we not reached the end of the ring
     it->keyp = std::next(it->keyp);
     if (it->keyp != it->store->keys.end()) {


rrrooommmaaa:
@ni4 I have prepared refactoring of the code that handles `rnp_identifier_iterator_st.keyp` in a proper way but haven't pushed it yet. Instead, I added the code to throw an exception when trying to advance the iterator past the end. It turns out, some portions of `test_ffi_key_iter` fail now, Do you want me to copy this pieces into a separate test? And should I replace exception with some other way to track the error?

ni4:
@rrrooommmaaa if I correctly understand you - FFI functions must not throw exceptions, and just check for errors/return corresponding error code. And unless there is a specific reason, behavior should not be changed, i.e. all old tests should be passing.

rrrooommmaaa:
@ni4 My question was about this:
> However this would require a test, which reproduces the case(s) in which error may occur.
We already have `test_ffi_key_iter` test that fails on Windows. What other tests would you want to have?

ni4:
@rrrooommmaaa Got it. Previously there was another code in this PR. Right now I think it would be more logical to close this PR and wait for MSVC changes finished and merged.

rrrooommmaaa:
This issue will be fixed as part of #1232

PR #1271, opened 2020-08-19
don't adjust line endings in test files

### LINKED ISSUE #1268: Add .gitattributes file to prevent CRLF/LF conversion for test data files.
## Description

Some tests have signatures over text files (example - `test_stream_signatures/source.txt`), which become invalid once git automatically changes EOL from LF to CRLF on Windows.
.gitattributes file should be added for these files, disabling EOL conversion (or marking it as fixed to LF).

rrrooommmaaa:
Closes #1268

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -18,8 +18,6 @@ jobs:
     if: "!contains(github.event.head_commit.message, 'skip ci')"
     timeout-minutes: 50
     steps:
-      - name: Disable git eol translation
-        run: git config --global core.autocrlf false
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
--- a/src/tests/data/.gitattributes
+++ b/src/tests/data/.gitattributes
@@ -1,1 +1,16 @@
-test_messages/message-32k-crlf.txt -text
+# for test_ffi_op_verify_sig_count
+test_messages/message-32k-crlf.txt			-text
+# for test_stream_signatures
+test_stream_signatures/source*.txt   			-text
+# for cleartext tests
+test_messages/message.txt            			-text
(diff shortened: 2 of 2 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+/src/tests/data/*/*.txt   -text
```
ni4, line 1: It would be preferred to put .gitattributes to /src/tests/data, so it will not add more mess to the root folder.
Also, I would suggest to not apply it to all *.txt files but only affected ones, in case, for instance, we'll want to test whether cleartext signing works fine with both line endings/setups.

ni4 (changes requested):

ronaldtse:
@rrrooommmaaa single failure on Windows:

```
ERROR: test_rnp_list_packets_edge_cases (__main__.Misc)
  File "D:/a/rnp/rnp/src/tests/cli_tests.py", line 1664, in test_rnp_list_packets_edge_cases
    compare_file_ex(data_path('test_key_edge_cases/key-empty-packets.txt'), out,
  File "D:/a/rnp/rnp/src/tests/cli_tests.py", line 170, in compare_file_ex
    raise_err(message)
  File "D:\a\rnp\rnp\src\tests\cli_common.py", line 34, in raise_err
    raise CLIError(msg, log)
cli_common.CLIError: <unprintable CLIError object>

Ran 17 tests in 55.387s

```

.gitattributes
rrrooommmaaa: @ni4 Done, please check it out

ronaldtse (approved):
LGTM! Thanks @rrrooommmaaa !

ni4 (approved):
LGTM! Thanks!

PR #1272, opened 2020-08-20
Correctly calculate packet nesting level during packet dumping.

### LINKED ISSUE #1262: 'too many OpenPGP packet layers' error on keyring dump.
## Description

PR #1241 incorrectly calculates packet nesting levels (incrementing it with each packet), so after 32 packets it reports an error and stops dumping.

Should be fixed, adding the corresponding test (for text and json output).

Example file which exposes this behavior: `keyrings/1/pubring.gpg`.

ni4:
Fixes #1262


antonsviridenko (approved):

ronaldtse (approved):
Thanks @ni4 !

PR #1282, opened 2020-08-29
Fix oss-fuzz invalid-enum-value issues

dewyatt:
Since we're using C++11 we can set an explicit type for these enums and that should fix these.
Testing locally with oss-fuzz shows all these to now be fixed.

EDIT: Proper wording since these are actual fixes

### DIFF
--- a/include/repgp/repgp_def.h
+++ b/include/repgp/repgp_def.h
@@ -313,7 +313,7 @@ typedef enum {
  *
  * \see RFC4880 5.2.1
  */
-typedef enum {
+typedef enum : uint8_t {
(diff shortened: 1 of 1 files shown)


antonsviridenko (approved):

antonsviridenko:
I assume this header file is used internally and not expected to be ever included in C source files, right?

ni4 (approved):
LGTM, thanks!
Didn't know about this trick.
It also fixes newly-added 25280, 25283, 25300.

ni4:
> I assume this header file is used internally and not expected to be ever included in C source files, right?

Exactly, all lib is built with C++11. C (and other sources) may and should import only `rnp/rnp.h`

PR #1286, opened 2020-08-31
Fix  oss-fuzz issues 25302 and 25292

ni4:
As title says.


ni4:
@ronaldtse @dewyatt @rrrooommmaaa @antonsviridenko ping for review - issue 25302 reported as fuzzer blocker, since happens too often.

ronaldtse (approved):
LGTM, thanks @ni4 !

src/tests/support.cpp
```diff
@@ -101,11 +101,12 @@ file_to_str(const std::string &path)
...
-    std::ifstream        file(path, std::ios::binary | std::ios::ate);
-    std::vector<uint8_t> vec(file.tellg());
+    std::ifstream file(path, std::ios::binary | std::ios::ate);
+    size_t        length = file.tellg();
     file.seekg(0, std::ios::beg);
-    vec.insert(
-      vec.begin(), std::istream_iterator<uint8_t>(file), std::istream_iterator<uint8_t>());
+
+    std::vector<uint8_t> vec(length);
+    file.read((char *) vec.data(), length);
```
rrrooommmaaa, line 109: judging by this example https://www.cplusplus.com/reference/istream/istream/read/
we should call assert_true(file) after read.
Also, we have `static read_file_to_vector()` function in `tests\ffi.cpp`
Can we update the source code to have only one function?

ni4: @rrrooommmaaa Thanks, somehow missed that we already have such function. Moved it to support.cpp and added assert.

rrrooommmaaa (approved):
LGTM!

PR #1288, opened 2020-08-31
Refactor usage of strcpy.

### LINKED ISSUE #1278: Get rid of `strcpy()`
`strcpy()` is a potential overflow target and we should really get rid of its usage.

ni4:
Fixes #1278

### DIFF
--- a/src/rnp/fficli.cpp
+++ b/src/rnp/fficli.cpp
@@ -1343,90 +1343,45 @@ done:
     return res;
 }
 
-/** @brief compose path from dir, subdir and filename, and store it in the res
- *  @param dir [in] null-terminated directory path, cannot be NULL
- *  @param subddir [in] null-terminated subdirectory to add to the path, can be NULL
- *  @param filename [in] null-terminated filename (or path/filename), cannot be NULL
- *  @param res [out] preallocated buffer
- *  @param res_size [in] size of output res buffer
+/** @brief compose path from dir, subdir and filename, and return it.
+ *  @param dir [in] directory path
+ *  @param subddir [in] subdirectory to add to the path, can be empty
+ *  @param filename [in] filename (or path/filename)
  *
- *  @return true if path constructed successfully, or false otherwise
+ *  @return constructed path
  **/
-static bool
-rnp_path_compose(
-  const char *dir, const char *subdir, const char *filename, char *res, size_t res_size)
+static std::string
+rnp_path_compose(const std::string &dir,
+                 const std::string &subdir,
+                 const std::string &filename)
 {
-    int pos;
-
-    /* checking input parameters for conrrectness */
-    if (!dir || !filename || !res) {
-        return false;
-    }
-
-    /* concatenating dir, subdir and filename */
-    if (strlen(dir) > res_size - 1) {
-        return false;
-    }
-
-    strcpy(res, dir);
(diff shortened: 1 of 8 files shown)


ronaldtse (approved):
LGTM thanks @ni4!

src/rnp/fficli.cpp
```diff
@@ -1347,109 +1347,59 @@ cli_rnp_keys_matching_strings(cli_rnp_t *                     rnp,
...
-
-    strcpy(res, dir);
-    pos = strlen(dir);
-
-    if (subdir) {
-        if ((pos > 0) && (res[pos - 1] != '/')) {
-            res[pos++] = '/';
+    std::string res = dir;
+    if (!subdir.empty()) {
+        if (!res.empty() && (res.back() != '/')) {
```
rrrooommmaaa, line 1357: should we support backslash on Windows?

src/rnp/fficli.cpp
```diff
@@ -1347,109 +1347,59 @@ cli_rnp_keys_matching_strings(cli_rnp_t *                     rnp,
  *
```
rrrooommmaaa, line 1347: looks like `NULL` in description is no longer appropriate

src/rnp/fficli.cpp
```diff
@@ -1347,109 +1347,59 @@ cli_rnp_keys_matching_strings(cli_rnp_t *                     rnp,
...
-    const char *subdir;
-
-    if (!defhomedir) {
-        subdir = NULL;
-    } else {
-        if ((subdir = rnp_cfg_getstr(cfg, CFG_SUBDIRGPG)) == NULL) {
-            subdir = SUBDIRECTORY_RNP;
-        }
+    if (!res.empty() && (res.back() != '/')) {
+        res.push_back('/');
```
rrrooommmaaa, line 1364: ... or even use some Windows function as `PathCombine` or `PathCchCombine` for Windows?

rrrooommmaaa (changes requested):
Overall good, but I have some concerns about backslash path separator on Windows

src/rnp/fficli.cpp
ni4: @rrrooommmaaa Most likely we should, but in another PR, since this one is aimed to remove `strcpy()` usage.

src/rnp/fficli.cpp
ni4: As documentation of other parameters :) Thanks, will update the PR.

src/rnp/fficli.cpp
ni4: @rrrooommmaaa I think that should be done in separate PR, which would check all of the path-related calls in the CLI.

antonsviridenko:
why `memcpy()` and not `strncpy()`?

ni4:
> why `memcpy()` and not `strncpy()`?

We do separate checks for the line length (to issue error messages, return false, and so on), so no need to use `strncpy()`.
Anyway, sooner or later this will be migrated to `std::string`

src/rnp/fficli.cpp
rrrooommmaaa: Ok, I added #1290

rrrooommmaaa (approved):
LGTM!

PR #1291, opened 2020-09-04
use mkstemp where possible

### LINKED ISSUE #1071: Annoying linker warning: stream-common.cpp:(.text+0x2666): warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
## Description

The function init_tmpfile_dest in src/librepgp/stream-common.cpp calls the deprecated mktemp function. That causes an annoying linker warning.
This is what the [man-page](https://linux.die.net/man/3/mktemp) has to say:

> Never use mktemp(). Some implementations follow 4.3BSD and replace XXXXXX by the current process ID and a single letter, so that at most 26 different names can be returned. Since on the one hand the names are easy to guess, and on the other hand there is a race between testing whether the name exists and opening the file, every use of mktemp() is a security risk. The race is avoided by mkstemp(3). 

## Steps to Reproduce

1. Link against a static librnp-0.a

## Expected Behavior

No warning

## Actual Behavior

A Warning:
```
rnpgp/lib/librnp-0.a(stream-common.cpp.o): In function `init_tmpfile_dest(pgp_dest_t*, char const*, bool)':
stream-common.cpp:(.text+0x2666): warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'

```

[This is the offending line](https://github.com/rnpgp/rnp/blob/5a39e2b1955c97c032a7b63df272397d935b40cd/src/librepgp/stream-common.cpp#L904).

rrrooommmaaa:
Fixes #1071


ronaldtse (approved):
LGTM thanks @rrrooommmaaa !

src/lib/CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ check_include_file_cxx(sys/time.h HAVE_SYS_TIME_H)
 check_include_file_cxx(unistd.h HAVE_UNISTD_H)
 check_include_file_cxx(sys/wait.h HAVE_SYS_WAIT_H)
 check_cxx_symbol_exists(mkdtemp "stdlib.h;unistd.h" HAVE_MKDTEMP)
+check_cxx_symbol_exists(mkdtemp "stdlib.h;unistd.h" HAVE_MKSTEMP)
```
ni4, line 54: I guess this supposed to be `check_cxx_symbol_exists(mkstemp...` ?

src/librepgp/stream-common.cpp
```diff
@@ -910,9 +925,22 @@ init_tmpfile_dest(pgp_dest_t *dst, const char *path, bool overwrite)
         RNP_LOG("failed to build file path");
         return RNP_ERROR_BAD_PARAMETERS;
     }
+#ifdef HAVE_MKSTEMP
+    int fd = mkstemp(tmp);
+    if (fd < 0) {
+        RNP_LOG("failed to create temporary file with tempate '%s'. Error %d.", tmp, errno);
+        return RNP_ERROR_WRITE;
+    }
+#else
```
ni4, line 934: Do I correctly understand that part after `#else` is supposed to run only under Windows? Then it would be cleaner to use windows-related define, with additional `#else #error Unsupported os` afterwards.

src/librepgp/stream-common.cpp
```diff
@@ -910,9 +925,22 @@ init_tmpfile_dest(pgp_dest_t *dst, const char *path, bool overwrite)
...
     mktemp(tmp);
```
ni4, line 935: Is `mktemp()` implementation secure enough in this case (i.e. under Windows) ? Did you consider to use something specific for Windows, like `GetTempFileNameA()`

ni4 (changes requested):

src/lib/CMakeLists.txt
rrrooommmaaa: @ni4 Exactly! Thanks for noticing. I we use Windows-specific define, we can leave this out, then?

src/librepgp/stream-common.cpp
rrrooommmaaa: @ni4 Yes, I have done some development this way. It entails massive complications, including splitting path and file name (using Windows SDK), temp file name becomes exactly 12 chars long, so the concept of renaming the temp file just by casting off the tail falls apart.  It's easier to implement `mkstemp` directly like this: https://stackoverflow.com/questions/6036227/mkstemp-implementation-for-win32

src/librepgp/stream-common.cpp
rrrooommmaaa: yes, I assumed `mkstemp` is available on all other platforms

src/lib/CMakeLists.txt
ni4: @rrrooommmaaa actually I'm not sure that `mkstemp()`, as part of  POSIX.1, is available on every non-Windows platform, so as for me it is better to use HAVE_MKSTEMP/windows-specific defines separately.

ni4: Another example - MinGW/MSYS2 will have both mkstemp and Windows defines enabled.

src/librepgp/stream-common.cpp
ni4: > It's easier to implement mkstemp directly like this:

Agree, it would be easier to have our own implementation then. Maybe as a separate issue.

src/common/mkstemp.cpp
```diff
@@ -0,0 +1,108 @@
+/* mkstemp extracted from libc/sysdeps/posix/tempname.c.  Copyright
+   (C) 1991-1999, 2000, 2001, 2006 Free Software Foundation, Inc.
+
+   The GNU C Library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Lesser General Public
```
antonsviridenko, line 5: Is it compatible with RNP license to include pieces of LGPL-licensed code? ping @ronaldtse

src/librepgp/stream-common.cpp
rrrooommmaaa: @ni4 I have implemented `rnp_mkstemp` (well, copied from GNU source code) for Windows MSVC only -- turns out that `mkstemp` is available on all other platforms -- no errors.

ni4: @rrrooommmaaa Thanks. Yeah, MinGW has it's own implementation. What about moving code to the file-utils.cpp instead of the separate file, for less clutter? Other then this (and license question) LGTM.

rrrooommmaaa: @ni4 Moved the function `rnp_mkstemp` to `file-utils`. It builds ok, the only failure is transient `homebrew` error, right?
Please review.
@ronaldtse we have GNU Lesser General Public License mentioned in `rnp_mkstemp` implementation in `file-utils.cpp`
Is it ok to commit like this?

ni4: @rrrooommmaaa LGTM, thanks! Restarted the macOS build, it happens with homebrew periodically. Feel free to merge once there is a reply from @ronaldtse

ni4 (approved):
LGTM!

src/librepgp/stream-common.cpp
ni4: ...and after the second approval of course, believed there is one already.

ronaldtse: @rrrooommmaaa the LGPL license is not compatible with our BSD license, which is well answered in this [SO post](https://opensource.stackexchange.com/questions/6698/can-i-use-code-from-an-lgpl-v2-licenced-project-in-a-bsd-3-clause-licenced-proje)

> You cannot use LGPL code in a BSD-licensed project. When you are using LGPL'ed code the LGPL behaves very similar to the GPL. You can then only publish your project under the LGPL or GPL (even though some individual files in your project that do not depend on the LGPLed parts may use compatible licenses such as BSD).

> A LGPL-licensed library can be used by a BSD-licensed project, as long as users (incl. all end users) are free to study and swap out the LGPLed parts of the software. This means either
> * dynamically linking to the LGPL library, or
> distributing your project in a way that allows it to be relinked with a different version.

> Note that the LGPL doesn't just bind you but also other people distributing your project. This would go beyond the usual requirements of the BSD license, so it is worth documenting these additional obligations in your project.

Could you re-write this portion or somehow seek alternative functions under the MSVC environment? Sorry about this but thanks...

ronaldtse: Or @rrrooommmaaa is it possible to utilize the original functions/macros without including it in our code base?

ronaldtse: Ping @rrrooommmaaa ?

rrrooommmaaa: I started to edit it, yes

rrrooommmaaa: As Windows filesystem is case-insensitive, it needed some changes anyway

rrrooommmaaa: This version has very littler of original code left, is it good enough?

ronaldtse: I think the new implementation can be considered an original implementation. Thank you @rrrooommmaaa !!

ronaldtse:
Thank you @rrrooommmaaa !!

PR #1295, opened 2020-09-07
Fix handling of 0 key expiration in key signatures.

### LINKED ISSUE #1293: encrypted_add_recipient, !userkey->valid, attempt to use invalid key as recipient
When attempting to encrypt to a certain recipient key, RNP reports "attempt to use invalid key as recipient".

The failing check is in encrypted_add_recipient, if (!userkey->valid)

I couldn't find out what's wrong with the key.
For privacy reasons I'll send the key to Nickolay by email.

The key has a user ID that doesn't any spaces to separate name from email address. It looks like this:
```user@example.com<user@example.com>```

But that doesn't seem to be the issue, I created a fresh key with the same property, which works fine.

ni4:
Fixes #1293


rrrooommmaaa (approved):
LGTM!

antonsviridenko (approved):

Issue #10, opened 2017-03-23
Heavily DRY out codebase


ronaldtse:
Some of the executables are compiled on duplicated code in the codebase. For example, `netpgpverify` has its own copy of `libverify.c` but there is another one in `libverify`.

All executables should utilize a single copy of common code.

ni4:
Do we really need libverify? As I understand it's goal is to have small/lightweight non-dependent of anything library which allows to check signatures.

ronaldtse:
We should get rid of libverify-- I don't think it's necessary to have a separate verification lib separate from the signing functions.

ni4:
@ronaldtse Since we do not have libverify for a while, and almost all old codebase is removed shouldn't we close this?

ronaldtse:
Agree. Thanks @ni4!

PR #1300, opened 2020-09-09
Add support for marker packet.

### LINKED ISSUE #1296: Implement support for marker packet.
## Description

It appears that some implementations (including Symantec Encryption Desktop) still issue marker packet for encrypted messages. We should support such messages as well.

ni4:
While this packet is obsolete and deprecated there are still implementations which continue to generate them.

Fixes #1296


nwalfield:
Where is the Marker packet marked as obsolete and deprecated?  https://tools.ietf.org/html/rfc4880#section-5.8 does not suggest this.  (Tag #10 was reused for the marker packet, which was previously a literal packet; tag #10 as a literal packet is obsolete.)

dewyatt:
> Where is the Marker packet marked as obsolete and deprecated? https://tools.ietf.org/html/rfc4880#section-5.8 does not suggest this. (Tag #10 was reused for the marker packet, which was previously a literal packet; tag #10 as a literal packet is obsolete.)

See RFC 4880 bis
https://www.ietf.org/id/draft-ietf-openpgp-rfc4880bis-10.html#name-marker-packet-obsolete-lite

dewyatt:
Actually I see what @nwalfield was saying, it's just confusing phrasing in the RFC but that packet isn't deprecated.

ni4:
Okay, it was just my wording, not the strict RFC language. I.e. in my opinion at the moment those packets doesn't seem to have any major/logical use, but feel free to correct me.

antonsviridenko (approved):

nwalfield:
@ni4: Fingerprinting OpenPGP messages and keys is difficult, because there is no magic number at a static offset.  It would be good if we established a convention of always including a marker packet at the beginning of TPKs/TSKs and messages.

src/librepgp/stream-dump.cpp
```diff
@@ -1152,6 +1152,16 @@ stream_dump_literal(pgp_source_t *src, pgp_dest_t *dst)
     return ret;
 }
 
+static rnp_result_t
+stream_dump_marker(pgp_source_t &src, pgp_dest_t *dst)
```
rrrooommmaaa, line 1156: A bit strange that `src` is by reference, whereas `dst` is a pointer, is it not?

src/librepgp/stream-packet.cpp
```diff
@@ -1077,6 +1077,23 @@ stream_write_signature(const pgp_signature_t *sig, pgp_dest_t *dst)
...
+rnp_result_t
+stream_parse_marker(pgp_source_t &src)
+{
+    pgp_packet_body_t pkt = {};
+    rnp_result_t      res = stream_read_packet_body(&src, &pkt);
+    if (res) {
+        return res;
+    }
+
+    if ((pkt.len != 3) || memcmp(pkt.data, "PGP", 3)) {
```
rrrooommmaaa, line 1089: I suggest to have a marker constant "PGP" and re-use it.

src/librepgp/stream-dump.cpp
```diff
@@ -1152,6 +1152,16 @@ stream_dump_literal(pgp_source_t *src, pgp_dest_t *dst)
...
+{
+    dst_printf(dst, "Marker packet\n");
+    indent_dest_increase(dst);
+    dst_printf(dst, "contents: %s\n", stream_parse_marker(src) ? "invalid" : "PGP");
```
rrrooommmaaa, line 1160: I suggest to have a marker constant "PGP" and re-use it.

src/librepgp/stream-dump.cpp
```diff
@@ -2070,6 +2082,18 @@ stream_dump_one_pass_json(pgp_source_t *src, json_object *pkt)
     return RNP_SUCCESS;
 }
 
+static rnp_result_t
+stream_dump_marker_json(pgp_source_t &src, json_object *pkt)
+{
+    rnp_result_t ret = stream_parse_marker(src);
+
+    if (!obj_add_field_json(
+          pkt, "contents", json_object_new_string(ret ? "invalid" : "PGP"))) {
```
rrrooommmaaa, line 2091: I suggest to have a marker constant "PGP" and re-use it.

src/tests/ffi.cpp
```diff
@@ -5449,6 +5449,46 @@ TEST_F(rnp_tests, test_ffi_pkt_dump)
...
     rnp_output_destroy(output);
 
+    // dump data with marker packet
+    assert_rnp_success(rnp_input_from_path(&input, "data/test_messages/message.txt.marker"));
+    assert_rnp_success(rnp_output_to_memory(&output, 0));
+    assert_rnp_success(
+      rnp_dump_packets_to_output(input, output, RNP_DUMP_MPI | RNP_DUMP_RAW | RNP_DUMP_GRIP));
+    assert_rnp_success(rnp_output_memory_get_buf(output, &buf, &len, false));
+    buf[len - 1] = '\0';
+    assert_non_null(strstr((char *) buf, "contents: PGP"));
```
rrrooommmaaa, line 5459: I suggest to have a marker constant "PGP" and re-use it.

rrrooommmaaa (changes requested):
Please see my thoughts.

src/librepgp/stream-dump.cpp
ni4: Yeah, it's because inner function calls use `pgp_dest_t *`parameter. And once we'll change them to `pgp_dest_t &`, we'll need to change 4 lines of code again, while within current declaration we'll need to change single line :)

src/librepgp/stream-packet.cpp
ni4: Thanks, will update.

src/librepgp/stream-dump.cpp
ni4: Here and below I think we should better stick to hard-typed text, to detect case when marker body constant is changed.

src/librepgp/stream-dump.cpp
rrrooommmaaa: I strongly believe newly-created functions should have nice signatures. You can call other functions with `&dst` for now.

ni4: @rrrooommmaaa Ok, re-pushed. Any other comments?

rrrooommmaaa (approved):
LGTM!

ni4:
There are 2 approvals and passing CI - merging.

PR #1302, opened 2020-09-10
Fix possible crash of CLI utilities when HOME is not set.

ni4:
As title says.

### DIFF
--- a/src/rnp/fficli.cpp
+++ b/src/rnp/fficli.cpp
@@ -1386,7 +1386,8 @@ rnp_cfg_set_ks_info(rnp_cfg_t *cfg)
     bool        defhomedir = false;
     std::string homedir = rnp_cfg_getstring(cfg, CFG_HOMEDIR);
     if (homedir.empty()) {
-        homedir = getenv("HOME");
+        const char *home = getenv("HOME");
+        homedir = home ? home : "";
         defhomedir = true;
     }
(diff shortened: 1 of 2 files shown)


antonsviridenko:
As homedir is important, maybe it's also makes sense to check if it exists and is accessible? Otherwise LGTM

antonsviridenko (approved):

rrrooommmaaa (commented):
there are a couple of other places with getenv("HOME"). Should we fix them too?
`tests/support.cpp:579` looks especially suspicious.

ni4:
> `tests/support.cpp:579` looks especially suspicious.

This one just for tests and will not cause problems since we set HOME explicitly in `rnp_tests.cpp`: 
`EXPECT_EQ(0, setenv("HOME", m_dir, 1));`.

One in `rnp_cfg_set_defkey` will not cause problems as well, just will not found gnupg's conf file.

rrrooommmaaa (approved):
LGTM.
I guess we're waiting for @ni4 response to Anton's question.

ni4:
> As homedir is important, maybe it's also makes sense to check if it exists and is accessible? Otherwise LGTM

I think this should be separated as another issue, will file it now.

ni4:
@antonsviridenko Is it good to merge now?

PR #1306, opened 2020-09-14
use GetTickCount64 on MSVC only

### LINKED ISSUE #1305: Incorrect fallback on Win API when HAVE_SYS_TIME_H is not detected properly
## Description

As described in discussion #1232 

## Steps to Reproduce

1.  Try to build in Thunderbird environment

## Expected Behavior

use functions from `sys/time.h`

## Actual Behavior

trying to fall back on Win API

rrrooommmaaa:
as `HAVE_SYS_TIME_H` isn't used anywhere else, instead of falling back on Win API, use Win API on  MSVC only
Fixes #1305


ronaldtse (approved):

ni4 (approved):
LGTM!

Issue #1303, opened 2020-09-11
Should/could we use C++ type instead of POD for FFI structs?


ni4:
## Description

We have some structs inside of FFI objects which sooner or later will become "more C++" and could include other C++ objects, including containers.
As an example: `rnp_op_encrypt_t`, which includes `rnp_ctx_t`, which now has `list recipients`. At some point we'll change recipients to `std::list`, and it must be initialized via constructor.
So we would not be able to use calloc() to allocate memory for `rnp_op_encrypt_t`, and use `new` and constructor instead.

Is there any problem on moving to `new` from the `calloc()` for `rnp_op_encrypt_t`?
Or should we keep POD-style for FFI structs and use pointers to `rnp_ctx_t` inside instead?

CC @dewyatt @rrrooommmaaa @antonsviridenko

ni4:
@antonsviridenko It doesn't as far as I know (and anyway we return pointer to opaque struct in API calls), but maybe I'm missing something deep/some possible side effect/whatever else.

dewyatt:
Shouldn't be an issue and I would say it's definitely the way forward

ni4:
Okay, thanks for replies!

ni4:
Closing this as resolved (we already use some C++ types for FFI structs).

PR #1307, opened 2020-09-17
Add support for unknown experimental s2k's

ni4:
Since unknown s2k may include not-yet-supported secret key fields, those should not be discarded during the load/save.
So, this PR:
- adds support of loading/saving of unknown s2k
- adds possible 'Unknown' result for `rnp_key_get_protection_type()`
- refactors code, which related to s2k storage, to C++ style, so we could rely on constructors/destructors.

Fixes partially #1303


antonsviridenko (approved):

src/lib/rnp.cpp
```diff
@@ -2477,32 +2441,25 @@ pgp_write_handler(pgp_password_provider_t *pass_provider,
...
-        if (!osig->create_set) {
-            sinfo.sigcreate = ctx->sigcreate;
-        }
-
-        if (!list_append(&ctx->signers, &sinfo, sizeof(sinfo))) {
-            return RNP_ERROR_OUT_OF_MEMORY;
+        if (!sig.create_set) {
+            sinfo.sigcreate = ctx.sigcreate;
         }
+        ctx.signers.emplace_back(sinfo);
```
rrrooommmaaa, line 2461: why not `push_back`?

src/librepgp/stream-ctx.cpp
```diff
@@ -74,19 +75,16 @@ rnp_ctx_add_encryption_password(rnp_ctx_t *    ctx,
     if (!pgp_s2k_derive_key(&info.s2k, password, info.key, sizeof(info.key))) {
         return RNP_ERROR_GENERIC;
     }
-    if (!list_append(&ctx->passwords, &info, sizeof(info))) {
-        pgp_forget(&info, sizeof(info));
+    try {
+        ctx.passwords.emplace_back(info);
```
rrrooommmaaa, line 79: why not `push_back`?

src/librepgp/stream-write.cpp
```diff
@@ -1644,8 +1633,8 @@ init_literal_dst(pgp_write_handler_t *handler, pgp_dest_t *dst, pgp_dest_t *writ
 {
     pgp_dest_packet_param_t *param;
     rnp_result_t             ret = RNP_ERROR_GENERIC;
-    int                      flen = 0;
     uint8_t                  buf[4];
+    size_t                   flen = 0;
```
rrrooommmaaa, line 1637: why is this moved?

src/librepgp/stream-ctx.cpp
ni4: @rrrooommmaaa In my understanding `emplace_back()` would directly construct object here, using the move constructor, while `push_back()` will first create temporary and then do copy/move. Is it wrong in this  case?

src/librepgp/stream-write.cpp
ni4: @rrrooommmaaa Just typo - first attempted to move it below, where it is used for the first time, but then realized that there is a goto before. Should I revert this to the initial location?

rrrooommmaaa: @ni4 would be much easier to track changes, evidently

src/librepgp/stream-ctx.cpp
rrrooommmaaa: @ni4 Just checking. We have lots of `push_back` calls throughout the app.
We even have `siginfos.push_back({})` which should probably be replaced with `emplace_back`
I'm not an expert in this field, but from what I read, `emplace_back` is useful when you pass a list of constructor arguments to it, so that it is constructed in place. But since `sinfo` is the object itself, I suspect "move construction" will happen in the same way as in `push_back` implementation. I'm just trying to figure out the general rule to use throughout the app.

ni4: @rrrooommmaaa My opinion on this is like 'call emplace_back() when we move object/construct from the parameter list or push_back() when we copy', but it could be subject to change if there are some arguments I'm missing. Maybe @dewyatt has also some thoughts on this topic?

src/librepgp/stream-write.cpp
ni4: Okk, will re-push.

src/librepgp/stream-ctx.cpp
dewyatt: I don't think there's an advantage to `emplace_back` in this case since we're inserting an existing object, not constructing and inserting. Since `emplace*` can have surprising behavior with implicit conversions, it may be best to stick with `push_back` except in cases where there is a clear advantage. (Effective Modern C++ has some guidance in Item 42).

ni4: @dewyatt Thanks, let's stick to `push_back()` in such cases then.

ni4: @rrrooommmaaa Changed to `push_back()` and re-pushed. Any other comments?

rrrooommmaaa (approved):
LGTM!

PR #1310, opened 2020-09-17
Add key and signature import fuzzer.

ni4:
These functions uses the same packet parsing code as already available fuzzers, but still used to process user-generated input with different code paths, so should be worth fuzzing as well.

Also it fixes some problems found during the fuzzing CI.


ronaldtse (approved):
LGTM, thanks @ni4 !

src/librepgp/stream-key.cpp
```diff
@@ -721,6 +726,10 @@ process_pgp_key_auto(pgp_source_t &          src,
                            PGP_PKT_SECRET_SUBKEY})) {
         ret = RNP_ERROR_READ;
     }
+    /* change error code if we didn't process anything at all */
```
rrrooommmaaa, line 729: Looks like a workaround. What is the purpose of this?

src/lib/crypto.cpp
```diff
@@ -198,6 +198,10 @@ key_material_equal(const pgp_key_material_t *key1, const pgp_key_material_t *key
 rnp_result_t
 validate_pgp_key_material(const pgp_key_material_t *material, rng_t *rng)
 {
+#ifdef FUZZERS_ENABLED
```
rrrooommmaaa, line 201: Increasing `fuzz-seconds` in `cifuzz.yml` didn't help?

src/librepgp/stream-key.cpp
ni4: @rrrooommmaaa This is actually to protect from the case where it is impossible to read anything from the stream, since this function could be used to 'read  till the end of stream, skipping bad format errors'.

src/lib/crypto.cpp
ni4: Yeah, it affects only run of all fuzzing attempts, not the single data block processing. Still decided to leave fuzz-seconds value higher for more confidence that there are no fuzzing blockers.

src/librepgp/stream-key.cpp
rrrooommmaaa: @ni4 don't we have `pgp_source_t::error` for this?

ni4: @rrrooommmaaa There is no error on the source in this case, since nothing is actually read - parser peeks few bytes to detect packet type on source, and then returns an error.

rrrooommmaaa: @ni4 so, is there actually an error? What is the line number where parser detects error? Maybe, it's better to set `pgp_source_t::error` there?

ni4: @rrrooommmaaa There is not error on source, but on source contents, i.e. everything is fine with reading but we cannot continue to process the data since contents are not something we expected.

antonsviridenko (approved):

ronaldtse:
Thanks guys!

PR #1053, opened 2020-01-21
#1014 weird userids

rrrooommmaaa:
Here are some key generation and encryption tests for weird userids (Issue #1014)
special characters (without \0) generation and encryption
unicode generation and encryption (without icase search)
`-r` (recipient) parameter accepts regex. In order to correctly quote all imaginable characters, I changed mode to `std::regex_constants::ECMAScript`, so that special characters are escaped this way: `-r [\x7D][\x7B][\x5D][\x5B][\x29][\x5E][\x2A][\x2E][\x2B][\x28][\x09][\x0A][\x7C][\x24]@rnp`
allowing begin `^` and end `$` anchors `^[\x7D][\x7B][\x5D][\x5B][\x29][\x5E][\x2A][\x2E][\x2B][\x28][\x09][\x0A][\x7C][\x24]@rnp$`

Issue #1031, opened 2020-01-13
Implement --password CLI parameter for rnpkeys.


ni4:
## Description

While rnp CLI has `--password` parameter, rnpkeys doesn't.
This may be useful in batch key generation when --pass-fd is not usable due to some reasons.

ni4:
Once this is implemented the following CLI test would be useful:
```
    def test_generate_protection_password(self):
        '''
        Generate key with RNP, using the --password parameter, and make sure key is encrypted
        '''
        clear_keyrings()
        params = ['--homedir', RNPDIR, '--password', 'password', '--userid', 'enc@rnp', '--generate-key']
        ret, out, err = run_proc(RNPK, params)
        if ret != 0:
            raise_err('key generation failed', err)
        # Check packets using the gpg
        params = ['--homedir', RNPDIR, '--list-packets', path.join(RNPDIR, 'secring.gpg')]
        ret, out, err = run_proc(RNP, params)
        match = re.match(RE_RNP_ENCRYPTED_KEY, err)
        if not match:
            raise_err('wrong encrypted secret key listing', err)
```

PR #1320, opened 2020-09-30
UTF-8 on Windows

### LINKED ISSUE #1312: Support utf-8/unicode in rnp_input_from_path() call.
## Description

Currently we do not interpret `char *path` parameter, and it works but only for ASCII paths (or, most likely, for paths which use system encoding).
If path contains any non-ASCII character, `rnp_input_from_path()` will fail, at least on Windows.

Possible solutions:
- assume that path* parameter is in utf-8 encoding in `rnp_input_from_path()`. However, this most likely will introduce incorrect behavior on non-utf8 *nix systems when non-utf8 path will be passed
- add function `input_from_path_ex()` with additional encoding parameter. This seems to overcoplicate things a bit, we should not support all of the possible encodings, at least yet.
- add some special processing if path in `rnp_input_from_path()` is prefixed with "utf8:" or "utf16:"
- another solution?

Related TB bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1665493

rrrooommmaaa:
Closes #1312 
Detailed description in the conversation.

### DIFF
--- a/src/common/file-utils.cpp
+++ b/src/common/file-utils.cpp
@@ -34,21 +34,177 @@
 #include <stdio.h>
 #include "uniwin.h"
 #include <errno.h>
-#include <locale>
-#include <codecvt>
-#include <random>
+#else
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#endif // !_MSC_VER
+#ifdef _WIN32
+#include <random> // for rnp_mkstemp
+#include "str-utils.h"
+#define CATCH_AND_RETURN(v) \
+    catch (...)             \
+    {                       \
+        errno = ENOMEM;     \
+        return v;           \
+    }
+#else
+#include <string.h>
+#endif
 #ifdef HAVE_FCNTL_H
 #include <fcntl.h>
 #endif
+#include <rnp/rnp_sdk.h>
+
+extern "C" {
+int
+rnp_unlink(const char *filename)
+{
+#ifdef _WIN32
+    try {
+        return _wunlink(wstr_from_utf8(filename).c_str());
+    }
+    CATCH_AND_RETURN(-1)
 #else
-#include <sys/stat.h>
-#endif // _MSC_VER
+    return unlink(filename);
+#endif
+}
 
 bool
 rnp_file_exists(const char *path)
 {
     struct stat st;
-    return stat(path, &st) == 0 && S_ISREG(st.st_mode);
+    return rnp_stat(path, &st) == 0 && S_ISREG(st.st_mode);
+}
+
+bool
+rnp_dir_exists(const char *path)
+{
+    struct stat st;
+    return rnp_stat(path, &st) == 0 && S_ISDIR(st.st_mode);
+}
+
+bool
+rnp_path_exists(const char *path)
+{
+    struct stat st;
+    return rnp_stat(path, &st) == 0;
+}
+}
+
+int
+rnp_open(const char *filename, int oflag, int pmode)
+{
+#ifdef _WIN32
+    try {
+        return _wopen(wstr_from_utf8(filename).c_str(), oflag, pmode);
+    }
+    CATCH_AND_RETURN(-1)
+#else
+    return open(filename, oflag, pmode);
+#endif
+}
+
+FILE *
+rnp_fopen(const char *filename, const char *mode)
+{
+#ifdef _WIN32
+    try {
+        return _wfopen(wstr_from_utf8(filename).c_str(), wstr_from_utf8(mode).c_str());
+    }
+    CATCH_AND_RETURN(NULL)
+#else
+    return fopen(filename, mode);
+#endif
+}
+
+int
+rnp_stat(const char *filename, struct stat *statbuf)
+{
+#ifdef _WIN32
+    static_assert(sizeof(struct stat) == sizeof(struct _stat64i32),
+                  "stat is expected to match _stat64i32");
+    try {
+        return _wstat64i32(wstr_from_utf8(filename).c_str(), (struct _stat64i32 *) statbuf);
+    }
+    CATCH_AND_RETURN(-1)
+#else
+    return stat(filename, statbuf);
+#endif
+}
+
+#ifdef _WIN32
+int
+rnp_mkdir(const char *path)
+{
+    try {
+        return _wmkdir(wstr_from_utf8(path).c_str());
+    }
+    CATCH_AND_RETURN(-1)
+}
+#endif
+
+int
+rnp_rename(const char *oldpath, const char *newpath)
+{
+#ifdef _WIN32
+    try {
+        return _wrename(wstr_from_utf8(oldpath).c_str(), wstr_from_utf8(newpath).c_str());
+    }
+    CATCH_AND_RETURN(-1)
+#else
+    return rename(oldpath, newpath);
+#endif
+}
+
+#ifdef _WIN32
+_WDIR *
+#else
+DIR *
+#endif
+rnp_opendir(const char *path)
+{
+#ifdef _WIN32
+    try {
+        return _wopendir(wstr_from_utf8(path).c_str());
+    }
+    CATCH_AND_RETURN(NULL)
+#else
+    return opendir(path);
+#endif
+}
+
+std::string
+#ifdef _WIN32
+rnp_readdir_name(_WDIR *dir)
+{
+    _wdirent *ent;
+    for (;;) {
+        if ((ent = _wreaddir(dir)) == NULL) {
+            return std::string();
+        }
+        if (wcscmp(ent->d_name, L".") && wcscmp(ent->d_name, L"..")) {
+            break;
+        }
+    }
+    try {
+        return wstr_to_utf8(ent->d_name);
+    }
+    CATCH_AND_RETURN(std::string())
+#else
+rnp_readdir_name(DIR *dir)
+{
+    dirent *ent;
+    for (;;) {
+        if ((ent = readdir(dir)) == NULL) {
+            return std::string();
+        }
+        if (strcmp(ent->d_name, ".") && strcmp(ent->d_name, "..")) {
+            break;
+        }
+    }
+    return std::string(ent->d_name);
+#endif
 }
 
 /* return the file modification time */
@@ -57,14 +213,14 @@ rnp_filemtime(const char *path)
 {
     struct stat st;
 
-    if (stat(path, &st) != 0) {
+    if (rnp_stat(path, &st) != 0) {
         return 0;
     } else {
         return st.st_mtime;
     }
 }
 
-#ifdef _MSC_VER
+#ifdef _WIN32
 static const char letters[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
 
 /** @private
@@ -76,49 +232,51 @@ static const char letters[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
 int
 rnp_mkstemp(char *tmpl)
 {
-    int       save_errno = errno;
-    const int mask_length = 6;
-    int       len = strlen(tmpl);
-    if (len < mask_length || strcmp(&tmpl[len - mask_length], "XXXXXX")) {
-        errno = EINVAL;
+    try {
+        int       save_errno = errno;
+        const int mask_length = 6;
+        int       len = strlen(tmpl);
+        if (len < mask_length || strcmp(&tmpl[len - mask_length], "XXXXXX")) {
+            errno = EINVAL;
+            return -1;
+        }
+        std::wstring tmpl_w = wstr_from_utf8(tmpl, tmpl + len - mask_length);
+
+        /* This is where the Xs start.  */
+        char *XXXXXX = &tmpl[len - mask_length];
+
+        std::random_device rd;
+        std::mt19937_64    rng(rd());
+
+        for (unsigned int countdown = TMP_MAX; --countdown;) {
+            unsigned long long v = rng();
+
+            XXXXXX[0] = letters[v % 36];
+            v /= 36;
+            XXXXXX[1] = letters[v % 36];
+            v /= 36;
+            XXXXXX[2] = letters[v % 36];
+            v /= 36;
+            XXXXXX[3] = letters[v % 36];
+            v /= 36;
+            XXXXXX[4] = letters[v % 36];
+            v /= 36;
+            XXXXXX[5] = letters[v % 36];
+
+            int flags = O_WRONLY | O_CREAT | O_EXCL | O_BINARY;
+            int fd =
+              _wopen((tmpl_w + wstr_from_utf8(XXXXXX)).c_str(), flags, _S_IREAD | _S_IWRITE);
+            if (fd != -1) {
+                errno = save_errno;
+                return fd;
+            } else if (errno != EEXIST)
+                return -1;
+        }
+
+        // We got out of the loop because we ran out of combinations to try.
+        errno = EEXIST;
         return -1;
     }
-    std::wstring_convert<std::codecvt_utf8<wchar_t>> utf8conv;
-    std::wstring tmpl_w = utf8conv.from_bytes(tmpl, tmpl + len - mask_length);
-
-    /* This is where the Xs start.  */
-    char *XXXXXX = &tmpl[len - mask_length];
-
-    std::random_device rd;
-    std::mt19937_64    rng(rd());
-
-    for (unsigned int countdown = TMP_MAX; --countdown;) {
-        unsigned long long v = rng();
-
-        XXXXXX[0] = letters[v % 36];
-        v /= 36;
-        XXXXXX[1] = letters[v % 36];
-        v /= 36;
-        XXXXXX[2] = letters[v % 36];
-        v /= 36;
-        XXXXXX[3] = letters[v % 36];
-        v /= 36;
-        XXXXXX[4] = letters[v % 36];
-        v /= 36;
-        XXXXXX[5] = letters[v % 36];
-
-        int flags = O_WRONLY | O_CREAT | O_EXCL | O_BINARY;
-        int fd =
-          _wopen((tmpl_w + utf8conv.from_bytes(XXXXXX)).c_str(), flags, _S_IREAD | _S_IWRITE);
-        if (fd != -1) {
-            errno = save_errno;
-            return fd;
-        } else if (errno != EEXIST)
-            return -1;
-    }
-
-    // We got out of the loop because we ran out of combinations to try.
-    errno = EEXIST;
-    return -1;
+    CATCH_AND_RETURN(-1)
 }
-#endif // _MSC_VER
+#endif // _WIN32
--- a/src/lib/misc.cpp
+++ b/src/lib/misc.cpp
@@ -86,7 +86,7 @@ __RCSID("$NetBSD: misc.c,v 1.41 2012/03/05 02:20:18 christos Exp $");
 #include "utils.h"
 #include "json_utils.h"
 
-#ifdef WIN32
+#ifdef _WIN32
 #define vsnprintf _vsnprintf
 #endif
 
@@ -400,20 +400,6 @@ rnp_compose_path_ex(char **buf, size_t *buf_len, const char *first, ...)
     char *path = vcompose_path(buf, buf_len, first, ap);
     va_end(ap);
     return path;
-}
-
-bool
-rnp_path_exists(const char *path)
-{
-    struct stat st;
-    return stat(path, &st) == 0;
-}
-
-bool
-rnp_dir_exists(const char *path)
-{
-    struct stat st;
-    return stat(path, &st) == 0 && S_ISDIR(st.st_mode);
 }
 
 bool
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -1782,7 +1782,7 @@ try {
     if (!ob) {
         return RNP_ERROR_OUT_OF_MEMORY;
     }
-    if (stat(path, &st) == 0 && S_ISDIR(st.st_mode)) {
+    if (rnp_stat(path, &st) == 0 && S_ISDIR(st.st_mode)) {
         // a bit hacky, just save the directory path
         ob->src_directory = strdup(path);
         if (!ob->src_directory) {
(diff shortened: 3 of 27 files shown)


rrrooommmaaa:
@ni4 @dewyatt @antonsviridenko Since there is no way to set utf-8 locale on Windows, here is the proposed solution:
`rnp_win_substitute_cmdline_args` reads wide char command line and converts it to multibyte utf-8,
and instead of calling file functions like `stat`, `mkstemp`, `unlink`, `open` etc. we use `rnp_stat`, `rnp_mkstemp`, `rnp_unlink`, `rnp_open`. Thus we'll have full support for Unicode parameters up to 0xFFFF (including file names)
The problem is still that we'll need to review all the places where we format strings as `%s` and where utf-8 data may occur to do conversion to wide char and use `%S` ? Also, due to `MAX_PATH` constants etc. we may incorrectly calculate string validity (it may be valid but appear longer).  What is your opinion on this? Is it a way to go?

ni4:
@rrrooommmaaa  Is this PR about an issue #1312? Could you please then add some explanation which solution did you pick up?
Also, is this supposed to work on non-utf8 Linux/macOS/other locales? Then some workflow tests should be added as well, to make sure it works correctly there.

rrrooommmaaa:
> @rrrooommmaaa Is this PR about an issue #1312? Could you please then add some explanation which solution did you pick up?
> Also, is this supposed to work on non-utf8 Linux/macOS/other locales? Then some workflow tests should be added as well, to make sure it works correctly there.

@ni4 Exactly. I picked the first scenario:
```
assume that path* parameter is in utf-8 encoding in rnp_input_from_path(). However, this most likely will introduce incorrect behavior on non-utf8 *nix systems when non-utf8 path will be passed
```
Note: this solution converts ALL the command line arguments of `rnp` and `rnpkeys` on Windows to UTF-8
So , UTF-8 is supported in all filenames, keyring paths, userids etc.
Non-Windows behaviour remains unchanged (the change is to translate command line for Windows and use wchar_t file functions on Windows). I added a cli test to create keyring in a UTF-8 directory. It succeeded on all platforms. I only had to add `LC_LANG` etc. in CentOS environment.
I'm just afraid `%s` won't work correctly with UTF-8 strings on Windows, so we have to review all the places where it can fail.
If you want me to add non-UTF8 locale tests on Ubuntu etc. we can do that in another PR (as we also need to translate uids to UTF-8 as required by RFC-4880 , so we need to read environment settings and do translations along the way?)

antonsviridenko:
> I'm just afraid %s won't work correctly with UTF-8 strings on Windows,

@rrrooommmaaa Why do you think so?

rrrooommmaaa:
@antonsviridenko Well, for this to work, we need to let Windows know it is working with UTF-8 data and it doesn't really like this codepage for whatever reason. It enjoys replacing suspicious characters with `?????`. If you know how to do auto-conversion and willing to provide a demo, please don't hesitate to.

antonsviridenko:
Ok, so issue is in printing UTF-8 encoded strings on Windows console, right? Not in %s itself?

rrrooommmaaa:
> Ok, so issue is in printing UTF-8 encoded strings on Windows console, right? Not in %s itself?

Exactly. We need to care what user sees as output

antonsviridenko:
Ok, so I think we need to print message strings to users in a uniform way. Now I see that some places in code use printf(), some others fprintf(stdout, ...) and probably most part of messages is printed using RNP_LOG() macro. So we can make it print all messages through RNP_LOG() or similar and in a Windows version of this macro UTF-8 string should be converted to wide character version before printing. Something like that.

ni4:
Looks like we needed to discuss this task in more details.
Actually, since we have both library and CLI, the requirements for their functionality are as following:
- FFI file/path related functions should support native encoding of the system it works on, working with zero-terminated C strings, like it is now. It could be UTF-8, CP-1251, whatever else. But this would not work with wide chars because of zeroes following ASCII chars.
- FFI file/path related function should support pre-encoded UTF8/UTF16/Unicode-16/whatever else to work with languages/systems where strings internally are unicode. We can easily convert between Unicode-16/Utf8/Utf16. But this would work only if underlying file system functions are also working with Unicode. So we should add functions like `rnp_input_from_path_ex`, where we can specify that input path has native/utf8/unicode16/whatever else encoding, and probably string length.
- we should not care about user ids and other strings in the FFI as those will be stored as they are in key packets, and it is a problem of the caller to correctly interpret data.
- CLI layer: we should not do any charset conversion of input/output parameters and use those as they are, including message to the user. It's his problem to correctly setup console. Except the case, if it is the case, when `argv` is actually widechar.

So, in short 
- we should not do any re-encoding of the data
- we should add FFI file/path functions which work with sorts of widechars because of problems with zeroes
- we should make sure that CLI work correctly on systems (if any) where argv is widechar.

rrrooommmaaa:
@ni4 Yes, the problem with CLI is that: Python3 calls `CreateProcessW` on Windows as opposed to `CreateProcessA` as Python2 did , and not only python does this, so Unicode-16 (wide chars) is de-facto standard for Windows. Now, as we use `getopt` which doesn't support wide char, and we have std::string throughout the whole application,the easiest way is to convert wide chars to utf8 before getopt parsing. Unicode may appear not only in file names but in UIDs too, so it is convenient to handle all the parameters (command line).

> But this would not work with wide chars because of zeroes following ASCII chars.

We don't cast `wchar_t *` to `char *`. We pack `wchar_t` into utf-8, then extract it back to `wchar_t *` when needed.

> we should not care about user ids and other strings in the FFI as those will be stored as they are in key packets, and it is a problem of the caller to correctly interpret data.

RFC4880 explicitly mentions utf-8 in many places, like this:
```
5.11. User ID Packet (Tag 13)
 A User ID packet consists of UTF-8 text that is intended to represent
 the name and email address of the key holder.
```
So whatever the current locale is set we should take care of CLI conversions.

src/common/cmdline-utils.cpp
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    std::vector<std::string> result;
+    result.reserve(arg_nb);
+
+    std::wstring_convert<std::codecvt_utf8<wchar_t>> utf8conv;
+    for (int i = 0; i < arg_nb; i++) {
+        auto utf8 = utf8conv.to_bytes(arg_w[i]);
+        result.push_back(utf8);
+    }
```
ni4, line 68: Shouldn't we wrap this in try ... catch so `LocalFree()` is called even when exception is raised?

src/common/cmdline-utils.cpp
```diff
@@ -0,0 +1,115 @@
...
+    char **argv_utf8_cstrs = new (std::nothrow) char *[argc_utf8];
+    if (!argv_utf8_cstrs) {
+        throw std::bad_alloc();
+    }
+    bool result = true;
+    for (int i = 0; i < argc_utf8; i++) {
+        argv_utf8_cstrs[i] = strdup(argv_utf8_strings[i].c_str());
+        if (!argv_utf8_cstrs[i]) {
+            result = false;
```
ni4, line 103: break afterwards?

src/common/file-utils.cpp
```diff
@@ -34,21 +34,136 @@
...
+#include <locale>
+#include <codecvt>
+#endif
+
+extern "C" {
+int
+rnp_unlink(const char *filename)
+{
+#ifdef _WIN32
+    std::wstring_convert<std::codecvt_utf8<wchar_t>> utf8conv;
```
ni4, line 61: Here and in the functions below:
Shouldn't we wrap C++ code in try..catch so no exception will go outside?

src/common/file-utils.cpp
```diff
@@ -34,21 +34,136 @@
...
+    return fopen(filename, mode);
+#endif
+}
+
+int
+rnp_stat(const char *filename, struct stat *statbuf)
+{
+#ifdef _WIN32
+    std::wstring_convert<std::codecvt_utf8<wchar_t>> utf8conv;
+    std::wstring filename_w = utf8conv.from_bytes(filename);
```
ni4, line 121: This and backward conversion snippet is used multiple times over the code, what about moving it to `std::wstring  wstr_from_utf8(...)`/`std::string wstr_to_utf8(...)`?

src/lib/misc.cpp
```diff
@@ -86,6 +86,12 @@ __RCSID("$NetBSD: misc.c,v 1.41 2012/03/05 02:20:18 christos Exp $");
 #include "utils.h"
 #include "json_utils.h"
 
+#ifdef _WIN32
+#include <windows.h>
+#include <locale>
+#include <codecvt>
+#endif
+
 #ifdef WIN32
```
ni4, line 95: Just curious - how WIN32 differs from _WIN32?

src/lib/rnp.cpp
```diff
@@ -60,6 +60,7 @@
 #include <botan/secmem.h>
 #include "ffi-priv-types.h"
 #include "file-utils.h"
+#include "cmdline-utils.h"
```
ni4, line 63: Why do we need cmdline-utils.h here? We do not process command line parameters on FFI/library layer.

src/librekey/rnp_key_store.cpp
```diff
@@ -58,45 +58,70 @@
...
+        auto dir = rnp_opendir(key_store->path.c_str());
         if (dir == NULL) {
             RNP_LOG(
               "Can't open G10 directory %s: %s", key_store->path.c_str(), strerror(errno));
             return false;
         }
 
-        while ((ent = readdir(dir)) != NULL) {
-            if (!strcmp(ent->d_name, ".") || !strcmp(ent->d_name, "..")) {
+        while ((ent = rnp_readdir(dir)) != NULL) {
```
ni4, line 87: As we make use only of ent->d_name here, what about moving most of this code to the function like `std::string rnp_readdir_name()`, and getting read of numerous ifdefs here, string conversions and so on?

src/librepgp/stream-common.cpp
```diff
@@ -400,14 +401,72 @@ file_src_close(pgp_source_t *src)
...
+    src->type = PGP_STREAM_FILE;
+    src->size = size ? *size : 0;
+    src->knownsize = !!size;
+
+    return RNP_SUCCESS;
+}
+
+#ifdef _WIN32
+rnp_result_t
+init_file_src_w(pgp_source_t *src, const wchar_t *path)
```
ni4, line 424: If we'll implent `std::string rnp_readdir_name()` as previous comment mentioned, we do not need this as we may safely call `rnp_open()` and have less system-dependent code/ifdefs.

src/common/cmdline-utils.h
```diff
@@ -0,0 +1,33 @@
...
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+ * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#if !defined(RNP_CMDLINE_UTILS_H_) && defined(_WIN32)
+#define RNP_CMDLINE_UTILS_H_
+
+bool rnp_win_substitute_cmdline_args(int *argc, char ***argv);
```
ni4, line 30: As this code is used only in CLI, I guess better place for it is in rnp.cpp, like fficli.h/cpp, rnpcfg.h/cpp.

ni4 (changes requested):
Mostly LGTM, except some code reuse/placing issue.

ni4:
@rrrooommmaaa Please see my review comments.

And to make sure we are on the same path, this PR, in short:
- allows CLI to work with wchar argv under Windows, storing all the stuff internally in UTF8
- expects UTF8 path-related parameters in FFI under the Windows
- do not change behavior on other systems, interpreting path-related parameters as they are represented on the system level
- adds some tests for UTF8 paths

Is my understanding correct?

src/lib/rnp.cpp
rrrooommmaaa: You're right, this should be removed.

src/lib/misc.cpp
rrrooommmaaa: `_WIN32` is defined by the compiler, `WIN32` is sort of overridable by user.
I use `_WIN32` everywhere, this `WIN32` is leftover from some other developer.
So, given that we're not planning to re-define it, it's all the same. I'll make it `_WIN32` for consistency.

src/librekey/rnp_key_store.cpp
rrrooommmaaa: this is done to avoid unnecessary conversion from wide char to utf-8 because `_wreaddir` returns wide char, and we need to open the file with `_wopen`, let it pass directly to `init_fle_src_w`

src/common/file-utils.cpp
rrrooommmaaa: Moved to `str-utils`

src/librepgp/stream-common.cpp
rrrooommmaaa: This is a shortcut, so we don't need to pack wide chart to utf-8 and unpack it back to potentially make the program less reliable

src/common/file-utils.cpp
rrrooommmaaa: Well, it's hard to define what `errno` these function should set in this case to mimick ordinary file functions...

src/common/cmdline-utils.cpp
rrrooommmaaa: Done

src/common/cmdline-utils.cpp
rrrooommmaaa: Done with `char *[]()` initialization

src/common/cmdline-utils.h
rrrooommmaaa: Put to `rnpcfg.h/cpp`

rrrooommmaaa:
@ni4 Correct,
this PR allows CLI to work with wchar argv under Windows, storing all the stuff internally in UTF8,
expects UTF8 path-related parameters in FFI under the Windows, using widechar file functions for Windows,
do not change behavior on other systems, interpreting path-related parameters as they are represented on the system level,
a unit test is provided with UTF8-encoded path,
a Python CLI encryption test is provided with UTF-8 file names.
More tests may be added as needed.

My points of concern are:
1) UTF8 file names may leak to RNP_(D)LOG under Windows as in this example:
`"Can't open G10 directory %s: %s", key_store->path.c_str(), strerror(errno));`
leading to garbage output. To fix this, I can convert `key_store->path.c_str()` to wide char on Windows, but I would also have to modify `%s` to `%S`.
2) Many functions check path lengths against MAXPATHLEN, MAX_PATH etc. using strlen, Now, as UTF-8 is multibyte, after conversion to wide char (under Windows) the path may become shorter, do we need to check this to avoid false alarms?
3) Do we also need to check CreateProcessA with codepages like cp-1251 on Windows? I haven't found a way to do this using Python3, should we use Python2 or some Powershell script to call CreateProcessA directly?

src/librekey/rnp_key_store.cpp
ni4: Wide char packing to utf-8 and unpacking is of linear complexity, and given that it would work on file names, i.e. short strings, performance benefits would be neglected with the more complex code/more code lines.

src/common/cmdline-utils.h
ni4: @rrrooommmaaa Probably better place for those would be fficli.h/cpp, as it already includes some helper functions.

src/librepgp/stream-common.cpp
ni4: @rrrooommmaaa I would still suggest to have less and more readable code but slightly more CPU cycles and allocation(s), as this is not a performance-critical code piece.

ni4:
> UTF8 file names may leak to RNP_(D)LOG (and also be displayed in some output listing files or uids) under Windows as in this example:
"Can't open G10 directory %s: %s", key_store->path.c_str(), strerror(errno));

I don't think that this has importance right now - user still may run rnp in non-Unicode console, or utf-8, whatever else, and we should not care about these cases as it is only diagnostic message(s). To fix it in 100% cases we'll end up with iconv().

> Many functions check path lengths against MAXPATHLEN, MAX_PATH etc. using strlen, Now, as UTF-8 is multibyte, after conversion to wide char (under Windows) the path may become shorter, do we need to check this to avoid false alarms?

This is also not something to fear about right now, but may be extracted to the separate issue to be checked later.

> Do we also need to check CreateProcessA with codepages like cp-1251 on Windows? I haven't found a way to do this using Python3, should we use Python2 or some Powershell script to call CreateProcessA directly?

Yeah, we'll need it to be compatible/do not crash on CP-1251 paths. This can be checked with some small WinAPI code snippet in rnp_tests:cli.cpp, which would call `CreateProcessA()` with rnp and some non-ASCII filenames.

src/common/file-utils.cpp
ni4: @rrrooommmaaa I don't think that we check `errno` in code, just success/failure. As this. code is used only by us we may freely skip setting errno or set it to some specific value, like ENOMEM

src/common/cmdline-utils.h
rrrooommmaaa: Done

src/common/file-utils.cpp
rrrooommmaaa: Done

src/librepgp/stream-common.cpp
rrrooommmaaa: @ni4 Done, should I delete `init_file_src_w` or let it stay?

ni4: @rrrooommmaaa As that code is not used I think it is better to delete it. Feel free to get this PR out of the Draft so it may be approved.

ni4 (approved):
LGTM!

ronaldtse:
@rrrooommmaaa is this PR ready to go? It's still in "draft" mode. Thanks!

rrrooommmaaa:
@ronaldtse I'm adding a test to call rnp/rnpkeys with CreateProcessA, as discussed.

rrrooommmaaa:
The `CreateProcessA` function test showed the following:
when we extract wide char command line via `GetCommandLineW` (we did this for compatibility with `CreateProcessW`), we get correct wide char representation of `CreateProcessA` input according to active ANSI code page (ACP). In the test, I have used 0x80 non-ASCII character in `--userid` parameter, that is € in CP-1252. So after all conversions we get everything in UTF-8, and € gets naturally into UID in UTF-8.
Now, if we want UIDs to appear "as they were in the encoding of Windows command line", this can be achieved via a hack (I tested that in PR #1053). When we check all command line arguments and see at least one non-ASCII character (>0x7F), we don't perform the substitution (because in the case `CreateProcessW`, Unicode characters get there as `????`, ASCII, and we are obliged to take from `GetCommandLineW`to get whatever is lost), making the variable `args_are_substituted`, for example global, and naming it appropriately. So we would use ordinary file functions inside `rnp_open` etc. for this mode. Files should be good in both modes, but user ids would be travelling "as they were". But do we really need user ids in encoding other than UTF-8, given that RFC4880 explicitly mentions UTF-8.
This "as is" mode was only useful for python2 tests I did for PR #1053, passing UTF-8 input to `run_proc` ignoring Windows ACP setting. But this isn't really Windows design, unless we need to keep compatibility with some existing setups.

rrrooommmaaa:
Looks like there is a problem passing passfd descriptor to child `rnpkeys` process in MSYS2. Can we by the way, introduce some other mechanism, like reading password from file by filename or add `--password` parameter to `rnpkeys`? This would greatly simplify testing.

rrrooommmaaa:
Or just disable `CreateProcessA` test for MSYS2?

ni4:
> Looks like there is a problem passing passfd descriptor to child `rnpkeys` process in MSYS2.
Yeah, I had the same issue during initial work on MSYS2 support. Please see the following question/answer, hope it helps:

https://stackoverflow.com/questions/58736093/use-pipe-and-fdopen-to-pass-data-from-python-script-to-c-application-in-wi

Switching to Windows python installation from the MSYS2 one helped.

> Can we by the way, introduce some other mechanism, like reading password from file by filename or add `--password` parameter to `rnpkeys`? This would greatly simplify testing.

You are welcome to take #1031 , it definitely needs to be implemented.

rrrooommmaaa:
> Now, if we want UIDs to appear "as they were in the encoding of Windows command line", this can be achieved via a hack (I tested that in PR #1053). When we check all command line arguments and see at least one non-ASCII character (>0x7F), we don't perform the substitution (because in the case `CreateProcessW`, Unicode characters get there as `????`, ASCII, and we are obliged to take from `GetCommandLineW`to get whatever is lost), making the variable `args_are_substituted`, for example global, and naming it appropriately. So we would use ordinary file functions inside `rnp_open` etc. for this mode. Files should be good in both modes, but user ids would be travelling "as they were". But do we really need user ids in encoding other than UTF-8, given that RFC4880 explicitly mentions UTF-8.
> This "as is" mode was only useful for python2 tests I did for PR #1053, passing UTF-8 input to `run_proc` ignoring Windows ACP setting. But this isn't really Windows design, unless we need to keep compatibility with some existing setups.

@ni4 The test now passes on Windows. What we decided about 8-bit "as is" UIDs?

ni4:
@rrrooommmaaa I'd say let's leave them as they are, and probably get back to them later via a separate issue/PR.

ronaldtse:
@rrrooommmaaa @ni4 are we good to go here?

rrrooommmaaa:
The tests succeeded for both `CreateProcessA` and `CreateProcessW` calls -- command line arguments were successfully converted to UTF8 (hence UIDs were saved correctly), file operations on Windows use wide char functions so no problems there too. But we need to take care of console output (as it is dependent on active code page and we can't just print UTF8 there).
image

ni4:
> We can merge the current update and place a separate issue on Windows console output?

Yeah, let's deal with console output in the separate issue. Main goal of this one was to support widechar in filenames, actually.

ronaldtse (approved):

ronaldtse:
Sounds good @ni4. In this case @antonsviridenko can you help merge when ready? Thanks!

Issue #1321, opened 2020-10-01
C++ coding style concerns


ni4:
## Description

As part of C++ refactoring I've got to the point where started moving signature_get_something/signature_set_somethhing/signature_do_something functions to the pgp_signature_t methods.
And since we do not have reference example in the code, I'd like to discuss do/not to do and other coding style concerns so we'd be stick to the single standard in future. Following is the `pgp_signature_t` declaration I have at the moment (with unrelated parts cut out), with questions and concerns below.

```C++
typedef struct pgp_signature_t {
    pgp_sig_type_t   type_;

    pgp_signature_t(const pgp_signature_t &src);
    pgp_signature_t(pgp_signature_t &&src);
    pgp_signature_t &operator=(pgp_signature_t &&src);
    pgp_signature_t &operator=(const pgp_signature_t &src);
    bool             operator==(const pgp_signature_t &src) const;
    bool             operator!=(const pgp_signature_t &src) const;
    ~pgp_signature_t();

    /* @brief Get signature's type */
    pgp_sig_type_t
    type() const
    {
        return type_;
    };
    void
    set_type(pgp_sig_type_t atype)
    {
        type_ = atype;
    };

    /* @brief Get v4 signature's subpacket of the specified type and hashedness */
    pgp_sig_subpkt_t *      get_subpkt(pgp_sig_subpacket_type_t stype, bool hashed = true);
    const pgp_sig_subpkt_t *get_subpkt(pgp_sig_subpacket_type_t stype,
                                       bool                     hashed = true) const;
    /* @brief Check whether v4 signature has subpacket of the specified type/hashedness */
    bool has_subpkt(pgp_sig_subpacket_type_t stype, bool hashed = true) const;
    /* @brief Check whether signature has signing key id (via v3 field, or v4 key id/key fp
     * subpacket) */
    bool has_keyid() const;
    /** 
     * @brief Get signature's signing key id
     * @param id reference to return key identifier
     * @return true on success (key id is available) or false otherwise
     */
    bool get_keyid(pgp_key_id_t &id) const;

    /**
     * @brief Add subpacket of the specified type to v4 signature
     * @param type type of the subpacket
     * @param datalen length of the subpacket body
     * @param reuse replace already existing subpacket of the specified type if any
     * @return reference to the subpacket structure or throws an exception
     */
    pgp_sig_subpkt_t &add_subpkt(pgp_sig_subpacket_type_t type, size_t datalen, bool reuse);

    /**
[...]

PR #1323, opened 2020-10-05
Refactor signature subpackets access code.

ni4:
To make it more C++, and add some test to check for correct subpackets fetching.


ni4:
msys2 tests failing because:
> The primary download server is currently (since Oct 2nd) down. We are working on bringing it back up. Pacman has preconfigured mirrors so package installations should work, albeit with timeouts.

ronaldtse (approved):
LGTM! Perhaps only merge this once msys2 is fixed?

ni4:
@ronaldtse Yeah, let's wait for few days, hopefully they'll bring the server up. Otherwise will need to find a workaround.

src/lib/types.h
```diff
@@ -347,6 +347,11 @@ typedef struct pgp_signature_t {
     bool             operator==(const pgp_signature_t &src) const;
     bool             operator!=(const pgp_signature_t &src) const;
     ~pgp_signature_t();
+
+    /* Get v4 signature's subpacket of the specified type and hashedness */
+    pgp_sig_subpkt_t *      get_subpkt(pgp_sig_subpacket_type_t stype, bool hashed = true);
```
rrrooommmaaa, line 352: It would be good to add this description here.
```
/* if hashed is false then accept any hashed/not hashed subpacket */
```
Better yet, why don't we create a table of packet type -> hashed requirements,
So that `PGP_SIG_SUBPKT_EMBEDDED_SIGNATURE` and `PGP_SIG_SUBPKT_ISSUER_KEY_ID` won't require it.
Otherwise, we'll have to copy this `hashed` parameter everywhere.

rrrooommmaaa (changes requested):
Please have a look at my suggestion.

antonsviridenko (approved):

src/lib/types.h
ni4: Thanks, updated the documentation.
Basically, except these two, all new subpackets are required to be hashed as per RFC. And as far as I remember version 5 of the key/signature will get rid of PGP_SIG_SUBPKT_ISSUER_KEY_ID, however new subpackets may be added at any time. So there is no need to have a table anywhere, default true parameter would be enough.

rrrooommmaaa: @ni4 So why don't we add `bool hashed = stype == ...` in the beginning of `get_subpkt`?
It's obvious you always call with `false` for certain types and always `true` with others.

ni4: @rrrooommmaaa It sort of logical layers separation, get_subpkt() method is on lower layer than caller and should not care about the OpenPGP logic.

rrrooommmaaa: @ni4 I double-checked and found out that `get_subpkt()` is both defined and called in the same file `librepgp/stream-sig.cpp`  -- doesn't appear as "layered" architecture. My concern is that when we add a new `get_subpkt(PGP_SIG_SUBPKT_ISSUER_KEY_ID)` call and forget to add `, false)`
So, main question: will it under any circumstances make sense to call `get_subpkt(PGP_SIG_SUBPKT_ISSUER_KEY_ID, true)`?
If not, why leave the code vulnerable to future edits?

ni4: @rrrooommmaaa 
> I double-checked and found out that get_subpkt() is both defined and called in the same file librepgp/stream-sig.cpp

Yeah, they are in the same file. Once we've got closer to C++-only code it would be logical to give each major class/struct their own place to live. The layers are in logic - all packet structs are dumb things, which are able to parse/write and return/set properties, but do not do any complex logic processing. That is done in signature generation/verification code, code used to fill up `pgp_key_t` data and so on.

> So, main question: will it under any circumstances make sense to call get_subpkt(PGP_SIG_SUBPKT_ISSUER_KEY_ID, true)?

Not sure if that happens anytime. But still we'll need it to be backwards compatible.

> If not, why leave the code vulnerable to future edits?

Not sure, If I understand how it would be vulnerable?  By default `get_subpkt()` still will use only hashed properties, and we'll make only an exception only for the explicit cases, like with key id or embedded primary key signature.

rrrooommmaaa: @ni4 So actually `signature_get_subpkt()` function should be left to call `get_subpkt` with correct flag based on packet type (as upper layer function).
Vulnerable means: currently we allow KEY_ID unhashed everywhere in the code, after a possible future update we may forget to add `, false` and it can lead to failures causing bad user experience

ni4: > So actually signature_get_subpkt() function should be left to call get_subpkt with correct flag based on packet type (as upper layer function).

get_subpkt() actually replaces signature_get_subpkt() and goes closer to C++, adding some more functionality on top.

> Vulnerable means: currently we allow KEY_ID unhashed everywhere in the code, after a possible future update we may forget to add , false and it can lead to failures causing bad user experience

Don't see any problem in this since we'll keep in mind this subpkt usage, cover it with tests and so on.

rrrooommmaaa: @ni4 As you suggested "layered" architecture, we can't add "hashed" requirement check to `get_subpkt` as it is in the wrong layer, but we can add it to `signature_get_subpkt` as it is the right layer. So why remove it? This layer has reasons to exists.
The more things we have to keep in mind, the more vulnerable the code is.

ni4: @rrrooommmaaa Hm, btw, `PGP_SIG_SUBPKT_ISSUER_KEY_ID` is currently used only in `signature_has_keyid()`/`signature_get_keyid()`, which perfectly isolate that calls (and after discussion in #1321 will become methods of `pgp_signature_t`). What is the reason of adding another function?

rrrooommmaaa (approved):
LGTM!

ni4:
Enough approvals and msys2 test is passing - so merging this.

Issue #1279, opened 2020-08-25
Upgrade Botan to 2.14.0


ronaldtse:
Botan is now at 2.14.0, probably time to upgrade @dewyatt @ni4 ? I remember we had an issue with it @dewyatt ?

ni4:
There is (or were?) an issue with Botan/MinGW, discussed here: https://github.com/rnpgp/rnp/issues/1104#issuecomment-623171383

Needs to recheck whether it was or wasn't gone with updated MinGW.
Also once building with MSVC will succeed we would not need MinGW for TB integration, and be compiled natively.

rrrooommmaaa:
The tests just succeeded in MSVC x64 (I'm going to add x32 build too), but vcpkg provides Botan version 2.12.1 there. Will it do?

ronaldtse:
There is the following issue in Botan 2.12.2:

>  2020-03-24: Side channel during CBC padding
> The CBC padding operations were not constant time and as a result would leak the length of the plaintext values which were being padded to an attacker running a side channel attack via shared resources such as cache or branch predictor. No information about the contents was leaked, but the length alone might be used to make inferences about the contents. This issue affects TLS CBC ciphersuites as well as CBC encryption using PKCS7 or other similar padding mechanisms. In all cases, the unpadding operations were already constant time and are not affected. Reported by Maximilian Blochberger of Universität Hamburg.
> Fixed in 2.14.0, all prior versions affected.

Still, it depends on whether the platform supports the later versions...?

ni4:
Update: re-checked https://github.com/rnpgp/rnp/issues/1279#issuecomment-680202719, CI still hangs with the latest msys2/mingw + latest Botan 2.14 package.

ronaldtse:
@ni4 @rrrooommmaaa can we have a re-check of this issue?

rrrooommmaaa:
On my MSVC system, cmake says `found suitable version "2.14.0", minimum required is "2.8.0"`, so we should change "required" version in CMake and it will work

ni4:
@rrrooommmaaa Could you please make a PR which updates required Botan version to 2.14.0 and makes sure that all CI works fine?

ronaldtse:
Sounds good guys — thanks!

PR #1324, opened 2020-10-07
Use secondary mirror if main MSYS2 packages server is down.

ni4:
As title says.

### DIFF
--- a/ci/before_install.sh
+++ b/ci/before_install.sh
@@ -86,8 +86,9 @@ msys_install() {
     mingw64/mingw-w64-x86_64-python3
   "
   pacman --noconfirm -S --needed ${packages}
-  pacman --noconfirm -U \
-    http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-libbotan-2.13.0-1-any.pkg.tar.xz
+  botan_pkg="mingw-w64-x86_64-libbotan-2.13.0-1-any.pkg.tar.xz"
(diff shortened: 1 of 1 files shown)


ci/before_install.sh
```diff
@@ -87,7 +87,10 @@ msys_install() {
   "
   pacman --noconfirm -S --needed ${packages}
   pacman --noconfirm -U \
-    http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-libbotan-2.13.0-1-any.pkg.tar.xz
+    http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-libbotan-2.13.0-1-any.pkg.tar.xz || \
```
rrrooommmaaa, line 90: I suggest to introduce an environment variable for botan version and re-use it in `cmake` dependency too,
As we have an outstanding task #1279

ni4: @rrrooommmaaa This PR is just to quick-fix msys2 CI, more complicated things should be done as part of that task. And we hardcode version 2.13 for botan here due to this problem: https://github.com/rnpgp/rnp/issues/1104#issuecomment-623171383

Anyway, it should be more clean to use variable for botan's package name instead of duplicating it, please see the updated PR.

antonsviridenko (approved):

rrrooommmaaa (approved):
LGTM!

ni4:
Merging this despite two CI failures - as it is only CI and corresponding msys2 build is passing.

PR #1328, opened 2020-10-15
macOS CI: fix homebrew openssl installation failure.

ni4:
As title says.

### DIFF
--- a/ci/before_install.sh
+++ b/ci/before_install.sh
@@ -5,6 +5,8 @@ set -ex
 
 macos_install() {
   brew update
+  # homebrew fails because `openssl` is a symlink while it tries to remove a directory.
+  rm /usr/local/Cellar/openssl || true
   brew bundle
 }
 


ronaldtse:
Maybe `brew unlink openssl`?

ni4:
> Maybe `brew unlink openssl`?

As I understand, `brew unlink` would unlink from `/usr/local/lib/`, while in this case somehow link from `/usr/local/Cellar/openssl` to `/usr/local/Cellar/openssl/1.0.2t` is created. Tried to found out where it comes from but didn't have a success, maybe overlooked something simple. Anyway, current run failed due to `Retrying download gem from https://rubygems.org/ due to error (4/4): Gem::RemoteFetcher::UnknownHostError no such name (https://rubygems.org/gems/jaro_winkler-1.5.4.gem)`, will restart it now.

ronaldtse:
Ah! It's all passing now. Merging.

PR #1337, opened 2020-10-31
Check userids for validity.

### LINKED ISSUE #1022: Implement userid validity checks.
## Description

Currently we have only key validity checking, however userid's should be also checked for validity.
Only non-revoked userid with valid self-signature should be considered as valid so corresponding key may be this userid.

### LINKED ISSUE #1126: Do not use non-self-certified userids.
## Description

Currently we doesn't check whether userid is valid itself (has at least one non-expired self-certification, and is not revoked). So functionality should be extended so:
- not signed userids are ignored and not accessible via userid enumeration;
- keys cannot be searched/listed by these userids.

ni4:
This PR adds userid validity checks.
Userid is now considered as valid when:
- it has valid, non-expired self-signature
- this self-signature doesn't expire the key itself
- there is no revocation for this userid.

This changes behavior to the following:
- key cannot be searched by non-valid userid
- primary key selection respects only valid userids
- however, non valid userids still are enumerated, and should be checked for validity via rnp_uid_is_valid().

Also it adds few more functions which allows to retrieve uid-related data via uid handle.

Fixes #1022 
Fixes #1126


ni4:
Single test is failing for ruby-rnp, it should be addressed via or together with https://github.com/rnpgp/ruby-rnp/pull/69

antonsviridenko (approved):

src/tests/key-store-search.cpp
```diff
@@ -82,23 +83,22 @@ TEST_F(rnp_tests, test_key_store_search)
     for (size_t i = 0; i < ARRAY_SIZE(testdata); i++) {
         pgp_key_id_t keyid = {};
         assert_true(rnp_hex_decode(testdata[i].keyid, keyid.data(), keyid.size()));
-        list seen_keys = NULL;
+        std::vector<pgp_key_t *> seen_keys;
```
rrrooommmaaa, line 86: I think it's more convenient to use `std::set` here

src/tests/key-store-search.cpp
```diff
@@ -82,23 +83,22 @@ TEST_F(rnp_tests, test_key_store_search)
...
         }
-        assert_int_equal(list_length(seen_keys), testdata[i].count);
-        list_destroy(&seen_keys);
+        assert_int_equal(seen_keys.size(), testdata[i].count);
     }
     // keyid search (by_name)
     for (size_t i = 0; i < ARRAY_SIZE(testdata); i++) {
-        list       seen_keys = NULL;
-        pgp_key_t *key = NULL;
+        std::vector<pgp_key_t *> seen_keys;
```
rrrooommmaaa, line 100: I think it's more convenient to use std::set here

src/tests/key-store-search.cpp
```diff
@@ -107,26 +107,27 @@ TEST_F(rnp_tests, test_key_store_search)
...
+            uid.valid = true;
+        }
+    }
     for (size_t i = 0; i < ARRAY_SIZE(testdata); i++) {
         for (size_t uidn = 0; testdata[i].userids[uidn]; uidn++) {
-            list        seen_keys = NULL;
-            pgp_key_t * key = NULL;
-            const char *userid = testdata[i].userids[uidn];
-            key = rnp_tests_key_search(store, userid);
+            std::vector<pgp_key_t *> seen_keys;
```
rrrooommmaaa, line 128: I think it's more convenient to use `std::set` here

rrrooommmaaa (changes requested):
LGTM, but I suggest to use `std::set` in places where we need to keep unique keys

src/tests/key-store-search.cpp
ni4: @rrrooommmaaa That would require to satisfy https://en.cppreference.com/w/cpp/named_req/Compare, and if I understand correctly, pgp_key_t will not. Given that it's just for tests I don't see the reason to spend time on implement that. And, actually, now it works the same way as it worked before. What do you think?

rrrooommmaaa: @ni4 Aren't we comparing pointers?

ni4: @rrrooommmaaa Oh, my fault, somehow mislooked (how?) pointerness, sorry. Fixed and force-pushed.

rrrooommmaaa (approved):
LGTM!

ronaldtse:
Seems that there are some failures with `ruby-rnp`. @dewyatt is there a way we can solve them once and for all?

dewyatt:
It looks like this changes the behavior of `rnp_locate_key` so that searching via a userid without a valid cert would no longer return results where it may have before. Am I understanding that correctly?
Should we be doing that kind of filtering or should we let the API client decide what to do?

ni4:
> It looks like this changes the behavior of `rnp_locate_key` so that searching via a userid without a valid cert would no longer return results where it may have before. Am I understanding that correctly?

Yes. Also userid will not be marked as primary if it is considered as invalid.

> Should we be doing that kind of filtering or should we let the API client decide what to do?

Usage of non-signed userid (or one with invalid signature) should be considered as a security issue, so I think we should not allow it to be used by default. However, API user still able to list all userids and check their validities.

dewyatt:
> > It looks like this changes the behavior of `rnp_locate_key` so that searching via a userid without a valid cert would no longer return results where it may have before. Am I understanding that correctly?
> 
> Yes. Also userid will not be marked as primary if it is considered as invalid.
> 
> > Should we be doing that kind of filtering or should we let the API client decide what to do?
> 
> Usage of non-signed userid (or one with invalid signature) should be considered as a security issue, so I think we should not allow it to be used by default. However, API user still able to list all userids and check their validities.

Shouldn't we at least throw some documentation updates in here?

ronaldtse:
> Shouldn't we at least throw some documentation updates in here?

Agree

ni4:
> > Shouldn't we at least throw some documentation updates in here?
> 
> Agree

Agree as well, just got lost with those CI issues. Update documentation and force-pushed.

ni4:
@dewyatt Should we merge this now? As I checked, only the single ruby-rnp tests is failed. And CIFuzz shots it's leg on the container building steps.

dewyatt:
> @dewyatt Should we merge this now? As I checked, only the single ruby-rnp test is failed. And CIFuzz shots it's leg on the container building steps.

May as well yes

PR #1340, opened 2020-11-05
Handle CR characters in the same way as GnuPG in text document and cleartext signatures.

### LINKED ISSUE #1265: Improve handling of CR characters within cleartext signatures.
## Description

We have some incompatibility with GnuPG when handling CR characters during cleartext signature generation/verification.

## Sample data

`0x31 0x30 0x64 0x0D`: when cleartext-signed with rnp, considered as BAD signature by GnuPG. However if signed with GnuPG, then signature is considered as valid by rnp.
`0x31 0x30 0x64 0x0D 0x0D`: BAD signature if signed with rnp and checked by GnuPG, and BAD signature if signed by GnuPG and checked by rnp.

More test cases could (and should) be added, if `\r` character is added to set of random characters in function `cli_common.py: random_text`.

## Solution

- when generating cleartext signature, we should transform CR characters to CRLF sequence, which should be more compatible with other implementations
- when checking cleartext signatures, we should treat CR without following LF just as another character, not any kind of EOL.

antonsviridenko:
Fixes #1265


ni4:
@antonsviridenko Please see the CIFuzz output (it also has artifact to reproduce the crash). This PR introduces heap buffer overflow, and most likely it causes the error described by you in the comment https://github.com/rnpgp/rnp/issues/1265#issuecomment-730590439

antonsviridenko:
macos tests fail to install python :(

ni4:
@antonsviridenko #1366 should fix macOS issues once approved/merged.

src/librepgp/stream-parse.cpp
```diff
@@ -797,40 +810,56 @@ signed_src_update(pgp_source_t *src, const void *buf, size_t len)
...
+    bool     max_linelen_warn = false;
+    /* we support LF and CRLF line endings */
     while (ch < end) {
-        /* continue if not reached CR or LF */
-        if ((*ch != CH_CR) && (*ch != CH_LF)) {
+        /* continue if not reached LF */
+        if (*ch != CH_LF) {
+            if (*ch != CH_CR && param->stripped_crs > 0) {
+                while (param->stripped_crs--)
+                    pgp_hash_list_update(param->txt_hashes, ST_CR, 1);
```
ni4, line 824: Please add `{}` around the `pgp_hash_list_update()`

ni4 (approved):
LGTM, except single coding style comment.

src/librepgp/stream-parse.cpp
```diff
@@ -797,40 +810,56 @@ signed_src_update(pgp_source_t *src, const void *buf, size_t len)
...
+
+                param->stripped_crs = 0;
+            }
+
+            if (!max_linelen_warn && (ch - linebeg) >= MAXIMUM_GNUPG_LINELEN) {
```
rrrooommmaaa, line 829: Do we assume here that the line starts since `buf` or later?
What would happen if a line started in a previous `signed_src_update` call?

antonsviridenko: Yes, right, data is processed in 32k blocks, not line by line, and line started in previous block could bypass this check. Going to fix that.

ni4: ...and add some test for this exact case (and probably 8k-16k-64k blocks) so we'll not regress in future )

antonsviridenko: @ni4 how exactly this test should look like? cli test that checks if warning message is printed on stderr?

ni4: @antonsviridenko It should just process test data file which exposes the mentioned behaviour (i.e. line started on a previous `signed_src_update()`), if this is possible.

src/librepgp/stream-parse.cpp
antonsviridenko: done

src/librepgp/stream-parse.cpp
antonsviridenko: Just to be clear, I'm talking about this piece (updated) only. It appears that github's web ui shows less lines mentioned than in email notifications

```
            if (!param->max_line_warn && param->text_line_len >= MAXIMUM_GNUPG_LINELEN) {
                RNP_LOG("Canonical text document signature: line is too long, may cause "
                        "incompatibility with other implementations. Consider using binary "
                        "signature instead.");
                param->max_line_warn = true;
            }
```
Otherwise the code already handles text lines that cross 32k block boundary, and it is covered by existing tests
https://github.com/rnpgp/rnp/blob/master/src/tests/ffi.cpp#L7028
and "cli_tests.SignDefault" where random text document up to 1M is generated each time and therefore with high probability many lines cross the block boundary.

antonsviridenko: And rnp CLI does not support detached text mode signatures yet, so it will be bit problematic to generate one :) If we sign that test file with text lines longer than 19995 characters using GnuPG, it will be verified as BAD signature by rnp. Because GnuPG forcefully inserts CRLF after 19995 chars.

ronaldtse (approved):
LGTM. Thanks @antonsviridenko ! I noticed you're still adding commits here, so please merge when ready.

ronaldtse:
Thank you @antonsviridenko !

PR #1342, opened 2020-11-13
Add FFI functions to retrieve key/uid revocation signature.

### LINKED ISSUE #1200: Provide new API to obtain the revocation date or revocation signature
I need the ability to obtain the date on which a key was revoked.

I can think of two alternatives, either a specific new API rnp_key_get_revocation_reason that returns a timestamp.

Alternatively, a more general API that uses rnp_key_get_signature_at, and then a new function that returns the signature type, like rnp_signature_get_type(rnp_signature_handle_t), then the application code loop the signatures, find the revocation signature, and obtain its timestamp using rnp_signature_get_creation.

ni4:
This PR fixes #1200 . 
Main goal is to add FFI functions to retrieve key/uid revocation signatures.
To make things simple it also refactors internal code so:
- a lot of C code is moved to C++, making things cleaner and easier.
- signatures now have calculated sigid field and stored in map, allowing faster access for keys with large amount of signatures (in case of merging, updating sings, whatever else)
- instead of keeping revocations list in key itself now revocations are stored in the corresponding uids.


antonsviridenko (approved):

src/lib/pgp-key.cpp
```diff
@@ -2117,11 +1966,151 @@ pgp_key_t::operator=(pgp_key_t &&src)
...
+    if (idx >= sigs_.size()) {
+        throw std::out_of_range("idx");
+    }
+    return get_sig(sigs_[idx]);
+}
+
+bool
+pgp_key_t::has_sig(const pgp_sig_id_t &id) const
+{
+    return sigs_map_.count(id);
```
rrrooommmaaa, line 2004: why not use `contains` method?

rrrooommmaaa: oh, it's too new (year 2020)!

src/lib/pgp-key.cpp
```diff
@@ -2117,11 +1966,151 @@ pgp_key_t::operator=(pgp_key_t &&src)
...
+        }
+    }
+    return res;
+}
+
+pgp_subsig_t &
+pgp_key_t::add_sig(const pgp_signature_t &sig, size_t uid)
+{
+    const pgp_sig_id_t sigid = sig.get_id();
+    if (sigs_map_.count(sigid)) {
```
rrrooommmaaa, line 2048: No need to check for key existence. map::erase will return 0 if key not found

src/lib/rnp.cpp
```diff
@@ -5675,33 +5670,42 @@ static rnp_result_t
...
             (*count)++;
         }
     }
     return RNP_SUCCESS;
 }
 
+static rnp_result_t
+rnp_key_return_signature(rnp_ffi_t               ffi,
+                         pgp_key_t *             key,
+                         pgp_subsig_t &          subsig,
```
rrrooommmaaa, line 5684: It's a bit confusing.
the new object captures the pointer `key` and puts it into the newly created object, 
it also captures the pointer `&subsig`
So let's have parameter of type `pgp_subsig_t *`

src/lib/types.h
```diff
@@ -106,7 +108,18 @@ template <> struct hash<pgp_fingerprint_t> {
...
 };
-} // namespace std
+
+template <> struct hash<pgp_sig_id_t> {
+    std::size_t
+    operator()(pgp_sig_id_t const &sigid) const noexcept
+    {
+        /* since signature id value is hash itself, we may use it's low bytes */
+        size_t res = 0;
+        std::memcpy(&res, sigid.data(), sizeof(res));
```
rrrooommmaaa, line 118: Can we add static_assert that pgp_sig_id_t is larger or equal to `size_t`'s size?

rrrooommmaaa (changes requested):
I put down a couple of suggestions to slightly improve the code.

src/lib/pgp-key.cpp
ni4: Thanks, will fix this.

src/lib/rnp.cpp
ni4: Thanks, fixed.

src/lib/types.h
ni4: Added (as well as to the pgp_fingerprint_t part)

ni4:

rrrooommmaaa (approved):
LGTM!

PR #1353, opened 2020-11-30
Increase timeout and run GHA on macOS 10.15 and 11.0

ni4:
Attempt to fix macOS CI timeout issue, and check whether GHA works under the Big Sur.


ni4:
@ronaldtse Could we merge this, given that it is CI, changes only macOS workflows, and those are passing?

ronaldtse:
Agree @ni4 , let's merge this first.

PR #1349, opened 2020-11-27
Refactor packet parsing to c++

ni4:
This PR refactors packet writing and parsing to the C++ methods instead of functions calls/pointers.
It would be split into the multiple PRs due to massive amount of changes.

PR #1354, opened 2020-11-30
Refactor base packets to C++

ni4:
This PR refactors the following packets to the C++ style parsing/writing:
- pgp_sk_sesskey_t
- pgp_pk_sesskey_t
- pgp_one_pass_sig_t
- pgp_userid_pkt_t

It borrows code from the #1349 to be easier to review/have less changes in a single PR.
Should be merged only after the #1350 as based on it.


ronaldtse (approved):
LGTM, thanks @ni4 !

src/lib/pgp-key.cpp
```diff
@@ -1472,8 +1472,8 @@ pgp_key_write_autocrypt(pgp_dest_t &dst, pgp_key_t &key, pgp_key_t &sub, size_t
             res = stream_write_key(&key.pkt, &memdst);
         }
 
-        res = res && stream_write_userid(&key.get_uid(uid).pkt, &memdst) &&
-              stream_write_signature(&cert->sig, &memdst);
+        key.get_uid(uid).pkt.write(memdst);
```
rrrooommmaaa, line 1475: Should it be `if (!res) { key.get_uid(uid).pkt.write(memdst); }`
to preserve original behaviour?

src/librepgp/stream-key.cpp
```diff
@@ -1453,6 +1463,53 @@ pgp_userid_pkt_t::~pgp_userid_pkt_t()
...
+
+    pgp_packet_body_t pkt(PGP_PKT_RESERVED);
+    rnp_result_t      res = pkt.read(src);
+    if (res) {
+        return res;
+    }
+
+    /* userid type, i.e. tag */
+    tag = (pgp_pkt_type_t) stag;
+    uid = (uint8_t *) malloc(pkt.size());
```
rrrooommmaaa, line 1503: What would happen if we call `parse` several times on the same object? Will we have memory leak?

rrrooommmaaa (changes requested):
I have some concerns about possible memory leaks

src/lib/pgp-key.cpp
ni4: Thanks, will update the PR. Actually it will be changed again in the next PRs, but let's keep things consistent.

src/librepgp/stream-key.cpp
ni4: While we should not do that twice, it is sort of incorrect logic, still let's free it before. Thanks.

ni4:

rrrooommmaaa (approved):
LGTM, though it's suspicious that `parse` methods may leave the object in partially-initialized state. Cant' we do this in a constructor with exception on failure?

ni4:
> LGTM, though it's suspicious that `parse` methods may leave the object in partially-initialized state. Cant' we do this in a constructor with exception on failure?

We need the return result from the parse() method, and via the result we may distinguish whether object is loaded successfully and may be used, or just dropped out.

ronaldtse:
@rrrooommmaaa @ni4 please merge when ready. Thanks!

antonsviridenko (approved):

PR #1355, opened 2020-11-30
Add test case for the oss-fuzz issues 25424

ni4:
This is not a real timeout but just a large compressed data packet processing, but still should be worth adding to not have possible regression in the future.


ronaldtse (approved):
LGTM!

rrrooommmaaa (approved):
LGTM!

PR #1357, opened 2020-12-03
Refactor signature packet to be more C++

ni4:
This PR refactors pgp_signature_t to use methods instead of C functions for packet parsing and writing.
Also it moves some large chunks of code to the more appropriate location.

It borrows code from the PR #1349


antonsviridenko (approved):

src/librepgp/stream-parse.cpp
```diff
@@ -859,51 +859,40 @@ signed_read_single_signature(pgp_source_signed_param_t *param,
...
-    ptype = get_packet_type(ptag);
-
+    int ptype = get_packet_type(ptag);
     if (ptype != PGP_PKT_SIGNATURE) {
         RNP_LOG("unexpected packet %d", ptype);
         return RNP_ERROR_BAD_FORMAT;
     }
 
     try {
         param->siginfos.push_back({});
```
rrrooommmaaa, line 875: don't we use `emplace_back()` in such cases?

src/librepgp/stream-parse.cpp
```diff
@@ -859,51 +859,40 @@ signed_read_single_signature(pgp_source_signed_param_t *param,
...
+        if (readsig.parse(*readsrc)) {
+            RNP_LOG("failed to parse signature");
+            siginfo.unknown = true;
+            if (sig) {
+                *sig = NULL;
+            }
+            return RNP_SUCCESS;
+        }
+        param->sigs.push_back(std::move(readsig));
+        siginfo.sig = &param->sigs.back();
```
rrrooommmaaa, line 887: Aren't you concerned with reallocations when vector grows?

src/librepgp/stream-sig.cpp
```diff
@@ -1281,3 +1466,340 @@ pgp_signature_t::matches_onepass(const pgp_one_pass_sig_t &onepass) const
...
+        RNP_LOG("cannot get enough bytes");
+        return RNP_ERROR_BAD_FORMAT;
+    }
+    /* length of hashed data, 5 */
+    if (buf[0] != 5) {
+        RNP_LOG("wrong length of hashed data");
+        return RNP_ERROR_BAD_FORMAT;
+    }
+    /* hashed data */
+    if (!(hashed_data = (uint8_t *) malloc(5))) {
```
rrrooommmaaa, line 1485: we can leak previous `hash_data` here.

src/librepgp/stream-sig.cpp
```diff
@@ -1281,3 +1466,340 @@ pgp_signature_t::matches_onepass(const pgp_one_pass_sig_t &onepass) const
...
+
+        subpkt.type = (pgp_sig_subpacket_type_t)(*buf & 0x7f);
+        subpkt.critical = !!(*buf & 0x80);
+        subpkt.hashed = hashed;
+        subpkt.parsed = 0;
+        memcpy(subpkt.data, buf + 1, splen - 1);
+        subpkt.len = splen - 1;
+
+        res = res && subpkt.parse();
+        subpkts.emplace_back(subpkt);
```
rrrooommmaaa, line 1562: `push_back(std::move )`?

rrrooommmaaa (changes requested):
I'm concerned about saving pointers to vector elements.

ronaldtse (approved):
LGTM, thank you @ni4 !

src/lib/pgp-key.cpp
```diff
@@ -1473,16 +1473,15 @@ pgp_key_write_autocrypt(pgp_dest_t &dst, pgp_key_t &key, pgp_key_t &sub, size_t
...
-        res = res && stream_write_signature(&cert->sig, &memdst);
-
         if (res && pgp_key_is_secret(&sub)) {
             pgp_key_pkt_t pkt(sub.pkt, true);
             res = stream_write_key(&pkt, &memdst);
         } else if (res) {
             res = stream_write_key(&sub.pkt, &memdst);
         }
-        res = res && stream_write_signature(&binding->sig, &memdst);
+        binding->sig.write(memdst);
```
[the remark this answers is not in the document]
ni4: Thanks, fixed.

src/librepgp/stream-parse.cpp
ni4: Actually it's leftover from some other PR, but let's fix it as well.

src/librepgp/stream-parse.cpp
ni4: That's why param->sigs is of `std::list` type - it doesn't invalidate pointers on addition/deletion. Got hit here with reallocations in one of the older PRs.

src/librepgp/stream-sig.cpp
ni4: Thanks, fixed.

src/librepgp/stream-sig.cpp
ni4: Agree, thanks for spotting this.

ni4:

src/librepgp/stream-parse.cpp
rrrooommmaaa: Good to hear that. Somehow missed that, sorry.

rrrooommmaaa (approved):
LGTM!

PR #1358, opened 2020-12-06
Improve performance of the signatures access.

ni4:
With old architecture retrieval of the single signature would cost O(N), making it quite slow for the large keys.
This PR adds fields which allows to enumerate direct-key/per-userid signatures much faster, giving ~15x performance improvement with huge test keyring.

### DIFF
--- a/src/lib/pgp-key.cpp
+++ b/src/lib/pgp-key.cpp
@@ -1403,7 +1403,7 @@ pgp_key_write_packets(const pgp_key_t *key, pgp_dest_t *dst)
     }
 
     /* write signatures on key */
-    size_t idx = pgp_key_write_signatures(dst, key, (uint32_t) -1, 0);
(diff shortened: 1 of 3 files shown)


antonsviridenko:
> giving ~15x performance improvement with huge test keyring.

What existing test could demonstrate this improvement?

ni4:
> > giving ~15x performance improvement with huge test keyring.
> 
> What existing test could demonstrate this improvement?

I tested only locally, with huge keyring loading and keys listing. I was thinking about adding some test which demonstrates that it's not of O(N^2) complexity, but that would require to generate huge keyring with large amount of keys, userids and certifications, which is not an fast and easy task right now. Probably we'll need to implement some key operations performance tests as a separate issue.

antonsviridenko (approved):

ronaldtse (approved):

Issue #515, opened 2017-10-19
Rename 'passphrase' things to 'password'


dewyatt:
## Description

https://github.com/riboseinc/rnp/issues/512 reminded me of this.

We seem to use the terms "password" and "passphrase" interchangeably in different parts of the code base (and user-facing portions).

There's arguably a difference in connotation but I think we're better off picking one and sticking to it.

As a native "U.S. English" speaker, I never really hear the term "passphrase" used outside of the context of cryptography, so I vote for using "password".

ronaldtse:
Me too, vote for "password"!

ni4:
Voting for 'password' too.

ronaldtse:
@dewyatt since we have enough people agreeing, it looks like a simple "search and replace" that perhaps can be done soon? Thanks!

ni4:
If nobody else want to take this I can do it.

ronaldtse:
Please go ahead @ni4 !

antonsviridenko:
To my understanding, 'passphrase' implies multiple words and therefore usually should be stronger than 'password' :)

PR #1362, opened 2020-12-15
Added --password option to rnpkeys

### LINKED ISSUE #1031: Implement --password CLI parameter for rnpkeys.

### LINKED ISSUE #1209: Please allow the creation of unprotected keys
## Description

We test RNP as part of the OpenPGP Interoperability Test Suite (https://tests.sequoia-pgp.org/) which includes a test to create keys.  `rnpkeys` used to be able to create keys with unprotected (sub)keys, but now it refuses to do so.  This means that we currently cannot create keys with RNP in our test suite.

(The technical reason for that is that the common interface that we use to drive the implementations that we test, the Stateless OpenPGP Command-Line Interface (https://gitlab.com/dkg/openpgp-stateless-cli) does not support protected keys, or any other operation that involves passwords.)

## Steps to Reproduce

1. `rnpkeys --generate-key --userid ...`

## Expected Behavior

Generate a key without encrypted (sub)keys.

## Actual Behavior

`rnpkeys` prompts for a password and the confirmation.

rrrooommmaaa:
Closes #1031
Closes #1209

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -1078,16 +1078,15 @@ rnp_request_password(rnp_ffi_t ffi, rnp_key_handle_t key, const char *context, c
     Botan::secure_vector<char> pass(MAX_PASSWORD_LENGTH, '\0');
     bool                       req_res =
       ffi->getpasscb(ffi, ffi->getpasscb_ctx, key, context, pass.data(), pass.size());
-    size_t pass_len = strlen(pass.data());
-    if (!req_res || !pass_len) {
+    if (!req_res) {
         return RNP_ERROR_GENERIC;
     }
-
-    *password = (char *) malloc(pass_len + 1);
+    size_t pass_len = strlen(pass.data()) + 1;
+    *password = (char *) malloc(pass_len);
     if (!*password) {
         return RNP_ERROR_OUT_OF_MEMORY;
     }
-    memcpy(*password, pass.data(), pass_len + 1);
+    memcpy(*password, pass.data(), pass_len);
     return RNP_SUCCESS;
 }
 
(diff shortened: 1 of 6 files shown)


ni4:
@rrrooommmaaa I'd suggest to add new parameter (together with missing pass-fd) to the usage message in `rnpkeys.cpp`.
We still need to write better one/update man page but this would be usable at least now.
Also I'd add test with empty password, to make sure that it works correctly - i.e. doesn't encrypt the generated key. Then we'd be able to close #1209 as well.

antonsviridenko (approved):

antonsviridenko:
> Also I'd add test with empty password, to make sure that it works correctly - i.e. doesn't encrypt the generated key.

What exactly do you mean here? No `--password` option passed or `--password ''` with empty string argument? In second case I'd expect that program encrypts generated keys using empty string as a password.

ni4:
> What exactly do you mean here? No `--password` option passed or `--password ''` with empty string argument? In second case I'd expect that program encrypts generated keys using empty string as a password.

No `--password` option would ask the password from the terminal, I mean empty password. Actually is there any reason to encrypt key with empty password? As another option we could use separate `--no-password` parameter.

rrrooommmaaa:
> No `--password` option would ask the password from the terminal, I mean empty password. Actually is there any reason to encrypt key with empty password? As another option we could use separate `--no-password` parameter.

I tested with `--password=` -- it correctly handles empty password and skips key protection step.

antonsviridenko:
Ok, the spec does not say anything about empty passphrases, so I assume they could be used. But GnuPG creates unprotected key when empty passphrase is supplied, so we are consistent in this behaviour.

Btw, why do we use the term "password" instead of "passphrase"?

antonsviridenko (approved):

ni4:
> Btw, why do we use the term "password" instead of "passphrase"?

Please see this for the discussion: https://github.com/rnpgp/rnp/issues/515

ni4 (approved):
LGTM, but please first wait for the #1366 to be merged and rebase on it, to make sure that macOS tests pass.

PR #1364, opened 2020-12-19
Refactor pgp_key_t fields access

ni4:
This PR refactors getters/setters for cleaner code.
Will be followed with other pgp_key_t functions refactoring, as it is too much changes for a single PR.

### DIFF
--- a/src/lib/pgp-key.h
+++ b/src/lib/pgp-key.h
@@ -167,11 +167,11 @@ struct pgp_key_t {
     const pgp_key_pkt_t &pkt() const;
     pgp_key_pkt_t &      pkt();
     void                 set_pkt(const pgp_key_pkt_t &pkt);
+
+    const pgp_key_material_t &material() const;
 };
 
 typedef struct rnp_key_store_t rnp_key_store_t;
-
-const pgp_key_material_t *pgp_key_get_material(const pgp_key_t *key);
 
 pgp_pubkey_alg_t pgp_key_get_alg(const pgp_key_t *key);
 
--- a/src/lib/pgp-key.h
+++ b/src/lib/pgp-key.h
@@ -130,10 +130,10 @@ struct pgp_key_t {
     pgp_fingerprint_t         primary_fp_{}; /* fingerprint of the primary key (for subkeys) */
     bool                      primary_fp_set_{};
     std::vector<pgp_fingerprint_t>
-      subkey_fps_{}; /* array of subkey fingerprints (for primary keys) */
+                    subkey_fps_{}; /* array of subkey fingerprints (for primary keys) */
+    pgp_rawpacket_t rawpkt_{};     /* key raw packet */
 
   public:
-    pgp_rawpacket_t        rawpkt{};     /* key raw packet */
     uint32_t               uid0{};       /* primary uid index in uids array */
     bool                   uid0_set{};   /* flag for the above */
     bool                   revoked{};    /* key has been revoked */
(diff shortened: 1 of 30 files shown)


src/librekey/rnp_key_store.cpp
```diff
@@ -442,25 +449,24 @@ rnp_key_store_add_subkey(rnp_key_store_t *keyring, pgp_key_t *srckey, pgp_key_t
         try {
             keyring->keys.emplace_back();
             oldkey = &keyring->keys.back();
-            keyring->keybyfp[pgp_key_get_fp(srckey)] = std::prev(keyring->keys.end());
+            keyring->keybyfp[srckey->fp()] = std::prev(keyring->keys.end());
```
rrrooommmaaa, line 452: Does this mean, if adding a mapping causes allocation exception, the `keyring->keys` will have an uninitialized element at the end?

src/librekey/rnp_key_store.cpp
```diff
@@ -499,7 +505,7 @@ rnp_key_store_add_key(rnp_key_store_t *keyring, pgp_key_t *srckey)
         try {
             keyring->keys.emplace_back();
             added_key = &keyring->keys.back();
-            keyring->keybyfp[pgp_key_get_fp(srckey)] = std::prev(keyring->keys.end());
+            keyring->keybyfp[srckey->fp()] = std::prev(keyring->keys.end());
```
rrrooommmaaa, line 508: Does this mean, if adding a mapping causes allocation exception, the `keyring->keys` will have an uninitialized element at the end?

rrrooommmaaa (changes requested):
LGTM, except for some minor comments.
Btw, can we create some overloads for sdk functions like `rnp_hex_encode`, `rnp_strhexdump_upper` to accept `std::vector` to avoid cluttering the code with `..data() ..size()`

antonsviridenko (approved):

src/lib/pgp-key.h
```diff
@@ -163,54 +164,73 @@ struct pgp_key_t {
...
+    bool             encrypted() const;
+    uint8_t          flags() const;
+    void             set_flags(uint8_t flags);
+    bool             can_sign() const;
+    bool             can_certify() const;
+    bool             can_encrypt() const;
+    /** @brief Get key's expiration time in seconds. If 0 then it doesn't expire. */
+    uint32_t expiration() const;
+    void     set_expiration(uint32_t expiry);
+    /** @brief Get key's creation time in seconds since Jan, 1 1980. */
```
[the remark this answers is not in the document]
ni4: Thanks, fixed.

src/librekey/rnp_key_store.cpp
ni4: @rrrooommmaaa Agree, thanks for spotting this. Fixed (and the one below as well).

src/tests/support.cpp
```diff
@@ -497,6 +497,13 @@ test_value_equal(const char *what, const char *expected_value, const uint8_t v[]
     return 0;
 }
 
+bool
+mpi_empty(const pgp_mpi_t &val)
+{
+    pgp_mpi_t zero{};
+    return (val.len == 0) && !memcmp(val.mpi, zero.mpi, PGP_MPINT_SIZE);
```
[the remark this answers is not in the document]
ni4: @antonsviridenko  I added this check for the case if, for instance, we'll clear mpi's length but forget to clean up the mpi contents, which would be a major security issue for the secret key's mpi.

And it will not affect performance at all as it is called from the few tests, and takes magnitudes less cycles then the single key generation operation.

ni4:
@rrrooommmaaa Does it look good now?

rrrooommmaaa (approved):
LGTM!

PR #1368, opened 2020-12-24
Refactor pgp_key_t validation/refreshing.

ni4:
This PR follows PR #1364 and PR #1367 with pgp_key_t refactoring.

PR #1373, opened 2020-12-27
Add FFI functions rnp_key_is_valid()/rnp_key_valid_till().

### LINKED ISSUE #1214: Add FFI function to retrieve key validity time.
## Description

Key/subkey validity time may depend on different signatures: certifications (and their revocations), direct-key signatures, key revocations, and so on.
This logic should be implemented into the single function, so user may understand whether he should believe, for instance, to the message, signed 2 years ago with the key which expired 1 year ago (or 3 years).

ni4:
First function would allow caller to check whether public key is considered to be valid for operations (encryption/verifications).
`rnp_key_valid_till()` would allow to fetch the latest timestamp when key is considered to be valid, so it may be used to validate old signatures.

Fixes #1214 

Should be rebased after #1368 is merged.

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -1420,6 +1420,29 @@ RNP_API rnp_result_t rnp_key_get_expiration(rnp_key_handle_t key, uint32_t *resu
 RNP_API rnp_result_t rnp_key_set_expiration(rnp_key_handle_t key, uint32_t expiry);
 
 /**
+ * @brief Check whether public key is valid. This includes checks of the self-signatures,
+ *        expiration times, revocations and so on.
+ *        Note: it doesn't take in account secret key, if it is available.
+ *
+ * @param key key's handle.
+ * @param result on success true or false will be stored here. Cannot be NULL.
+ * @return RNP_SUCCESS or error code on failure.
+ */
+RNP_API rnp_result_t rnp_key_is_valid(rnp_key_handle_t key, bool *result);
+
+/**
+ * @brief Get the timestamp till which key can be considered as valid.
+ *        Note: this will take into account not only key's expiration, but revocations as well.
+ *        For the subkey primary key's validity time will be also checked.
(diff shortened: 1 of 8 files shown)


include/rnp/rnp.h
```diff
@@ -1419,6 +1419,29 @@ RNP_API rnp_result_t rnp_key_get_expiration(rnp_key_handle_t key, uint32_t *resu
  */
 RNP_API rnp_result_t rnp_key_set_expiration(rnp_key_handle_t key, uint32_t expiry);
 
+/**
+ * @brief Check whether pbulic key is valid. This includes checks of the self-signatures,
```
antonsviridenko, line 1423: s/pbulic/public/

src/lib/pgp-key.cpp
```diff
@@ -1404,13 +1423,65 @@ pgp_key_t::is_protected() const
...
+            return 0;
+        }
+        const pgp_subsig_t &revsig = get_sig(revocation_.sigid);
+        if (revsig.sig.creation() > creation()) {
+            /* pick less time from revocation time and expiration time */
+            return std::min(revsig.sig.creation(), till);
+        }
+        return 0;
+    }
+    /* if key is not marked as expired then it wasn't valid at all */
```
antonsviridenko, line 1457: Why non-expired key cannot be valid here?

include/rnp/rnp.h
ni4: Thanks, fixed this.

src/lib/pgp-key.cpp
ni4: @antonsviridenko Because we checked for validity on the line 1442, and if key is not expired and is not revoked then it wasn't valid, i.e. has no valid self-signature/binding signature.

antonsviridenko (approved):

ronaldtse (approved):
LGTM @ni4 !

PR #1375, opened 2021-01-03
Add "remove-key" command to rnpkeys CLI.

### LINKED ISSUE #1322: Add command to delete key(s)/subkey(s) via CLI
## Description

We are missing key deletion functionality in CLI, so command `--remove-key` or `--delete-key` should be added.
Input to it could be userid, keyid/fingerprint (see recipient/signer list in `rnp` command).
If primary key is specified then it will be deleted together with subkeys, if subkey - then only subkey itself.
If there is corresponding secret key, it should additionally ask whether user really want to delete it unless `--force` parameter is added.

Output of the command should be like `3 secret(public) keys deleted successfully.`.

antonsviridenko:
Closes #1322 .


ni4 (approved):
LGTM!

src/rnp/fficli.cpp
```diff
@@ -1838,6 +1838,69 @@ cli_rnp_revoke_key(cli_rnp_t *rnp, const char *key)
...
+
+    if (secret && !rnp_cfg_getbool(cfg, CFG_FORCE)) {
+        char reply[10];
+
+        fprintf(rnp->userio_out,
+                "Key %s has coresponding secret key. Do you really want to delete it  (y/N)? ",
+                key);
+        fflush(rnp->userio_out);
+
+        if (fgets(reply, sizeof(reply), rnp->userio_in) == NULL) {
```
rrrooommmaaa, line 1884: I see no good reason to duplicate this piece of code from `rnp_get_output_filename`. Let's put it into a single static (for now) function in this file. I suggest to pass a string for output to this function without the `  (y/N)?` tail, so that it adds a tail and processes the input as it would be a `get yes or no` function.

rrrooommmaaa (changes requested):
Code looks correct, but I suggest not to duplicate input-processing code.

src/rnp/fficli.cpp
ni4: Agree, we should have some single point to get input from the user, like `std::string cli_rnp_get_user_input()`. And maybe something on top of it like `cli_rnp_get_confirmation(const std::string &msg)`, which would repeat asking for Y/N/Yes/No.

ronaldtse: Thanks guys! Once this is done let’s merge it.

antonsviridenko: @rrrooommmaaa please see updated code

src/rnp/fficli.cpp
```diff
@@ -198,6 +198,37 @@ ptimestr(char *dest, size_t size, time_t t)
...
+        va_start(ap, msg);
+        vfprintf(rnp->userio_out, msg, ap);
+        va_end(ap);
+        fprintf(rnp->userio_out, " (y/N) ");
+        fflush(rnp->userio_out);
+
+        if (fgets(reply, sizeof(reply), rnp->userio_in) == NULL) {
+            return false;
+        }
+        if (strlen(reply) > 0) {
```
rrrooommmaaa, line 217: y/N means N is the default choice.
However, when I press <Enter>, the choice is not recognized.
image

rrrooommmaaa (changes requested):
If we have the default option (N), it should react to empty input correctly.

Otherwise, the prompt should be `y/n`

src/rnp/fficli.cpp
antonsviridenko: @rrrooommmaaa fixed

PR #1378, opened 2021-01-07
Fix possible heavy memory usage on large amount of public key encrypted session keys.

ni4:
This PR:
- reduces memory usage by parsing PK-ESK entities only on demand
- reduces possible memory usage by not allowing endless amount of erroneous one-pass signatures in the signed stream.


rrrooommmaaa (approved):
LGTM!

antonsviridenko (approved):

ni4:
JFYI: Back-picked this to v0.14.0 since oss-fuzz is trained on this sample and doesn't allow to get through CI.

PR #1382, opened 2021-01-14
Add FFI functions to remove signatures/userids.

ni4:
This PR is part of the work on issue #1006.

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -1214,6 +1214,18 @@ RNP_API rnp_result_t rnp_signature_packet_to_json(rnp_signature_handle_t sig,
                                                   uint32_t               flags,
                                                   char **                json);
 
+/**
+ * @brief Remove a signature.
+ *
+ * @param key key handle, cannot be NULL.
+ * @param sig signature handle, cannot be NULL. Must be obtained via the key handle or one of
+ *            it's userids. You still need to call rnp_signature_handle_destroy afterwards to
+ *            destroy handle itself. All other handles of the same signature, if any, should
+ *            not be used after the call is made.
+ * @return RNP_SUCCESS if signature was successfully deleted, or any other value on error.
+ */
+RNP_API rnp_result_t rnp_signature_remove(rnp_key_handle_t key, rnp_signature_handle_t sig);
+
 /** Free signature handle.
  *
  * @param sig signature handle.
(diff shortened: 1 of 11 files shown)


include/rnp/rnp.h
```diff
@@ -1239,6 +1251,17 @@ RNP_API rnp_result_t rnp_uid_is_revoked(rnp_uid_handle_t uid, bool *result);
 RNP_API rnp_result_t rnp_uid_get_revocation_signature(rnp_uid_handle_t        uid,
                                                       rnp_signature_handle_t *sig);
 
+/**
+ * @brief Remove userid with all of it's signatures from the key
```
rrrooommmaaa, line 1255: Shouldn't it be "its" in English?

ni4: @rrrooommmaaa Yeah, thanks. Sometimes quick-typing produces weird results. Will fix it.

rrrooommmaaa: Actually, there are also occurrences in other places. Should we fix them or leave as they are?

ni4: @rrrooommmaaa It would be better to issue a separate PR, as this one is not related to language fixes.

src/tests/support.cpp
```diff
@@ -892,3 +892,79 @@ rnp_tests_key_search(rnp_key_store_t *keyring, const std::string &keyid)
...
+    assert_rnp_success(rnp_import_keys(*ffi, input, RNP_LOAD_SAVE_PUBLIC_KEYS, NULL));
+    assert_rnp_success(rnp_input_destroy(input));
+    assert_int_equal(unlink("pubring.gpg"), 0);
+}
+
+void
+reload_keyrings(rnp_ffi_t *ffi)
+{
+    rnp_output_t output = NULL;
+    assert_rnp_success(rnp_output_to_path(&output, "pubring.gpg"));
```
rrrooommmaaa, line 917: I usually prefer to use memory input/output, to not rely on the filesystem, avoid possible file collisions and speed up a bit. Are there any objections against it?

ni4: @rrrooommmaaa Agree, let's change to the memory input/output.

antonsviridenko (approved):

ronaldtse (approved):
LGTM!

PR #1387, opened 2021-01-20
Added instructions to build with MSVC

### LINKED ISSUE #1386: Update installation.adoc with Windows-MSVC instructions.
## Description

Currently it includes only instructions for the MinGW/MSYS2.

Related issue #1385

rrrooommmaaa:
These instructions embrace configuring with `cmake` and building with `MSBuild`, so `make` isn't required here.
Do we need to provide further steps to "deploy" or "install" the exe and dll files?
If so, what are the criteria to consider the rnp "installed" on Windows?
Close #1386

### DIFF
--- a/CMakeSettings.json
+++ b/CMakeSettings.json
@@ -11,7 +11,7 @@
       "buildCommandArgs": "",
       "ctestCommandArgs": "",
       "variables": [],
-      "cmakeToolchain": "${env.VCPKG_DIR}\\scripts\\buildsystems\\vcpkg.cmake",
+      "cmakeToolchain": "${env.VCPKG_ROOT}\\scripts\\buildsystems\\vcpkg.cmake",
       "intelliSenseMode": "windows-msvc-x64"
     },
     {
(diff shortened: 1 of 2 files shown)


ni4:
> Do we need to provide further steps to "deploy" or "install" the exe and dll files?
If so, what are the criteria to consider the rnp "installed" on Windows?

As for me, only developers who use rnp FFI, or 'power users' would build from source, but they do not need deploy/installation instructions.
For other users it would be preferable to distribute binaries in some way.

rrrooommmaaa:
Still, After these steps, the resulting files remain scattered across directories,
they need to be copied `bin` and `lib` directories which are already there.
I'm not sure whether we need to employ `make` for these purpose.
As I understand `make is not necessarily installed together with `cmake`, or is it?

ni4:
I thing we may leave binaries where they are for now, and update later, once needed.

ronaldtse:
I’d agree with @rrrooommmaaa that we needs some source installation documentation (and an install script (make?) to accompany) to help facilitate developers who use rnp as a library. 

They would likely want some “official documentation” on installing the whole package including binaries?

@rrrooommmaaa can you facilitate?

antonsviridenko (approved):

antonsviridenko:
> If so, what are the criteria to consider the rnp "installed" on Windows?

I think we can build installer binary, maybe using NSIS, and it will install files into C:\Program Files\rnp\ or some other place, like it is traditionally done on Windows.

ronaldtse (approved):

ronaldtse:
Looks great to me, thank you @rrrooommmaaa !

ronaldtse:
> I think we can build installer binary, maybe using NSIS, and it will install files into C:\Program Files\rnp\ or some other place, like it is traditionally done on Windows.

Agree! I think we should first fix up the Chocolatey package and add a downloadable installer package for Windows.

PR #1391, opened 2021-01-21
Fix parsing blank lines with whitespace characters.

### LINKED ISSUE #1381: Treat blank line according to RFC 4880 in cleartext signatures.
## Description

Currently we allow only really blank lines while parsing cleartext messages and computing signatures.
However, as per RFC 4880,  ` The blank line can either be zero-length or contain only whitespace, that is spaces (0x20), tabs (0x09) or carriage returns (0x0d).`.

We should change our behaviour accordingly.
Please see attached test message and a key, which expose this problem from the real world. Borrowed from the https://thunderbird.topicbox.com/groups/e2ee/Tdc01fcf0672d2234-M39a70a4d8bf54b2b919632fe

[Key file](https://github.com/rnpgp/rnp/files/5813950/autoaanetuk.txt)
[Cleartext signature](https://github.com/rnpgp/rnp/files/5813953/sig-embedded.txt)

antonsviridenko:
Fixes #1381. Now ASCII-armored messages that contain blank lines with white space characters (`spaces (0x20), tabs (0x09) or carriage returns (0x0d)`) can be properly parsed according to RFC.


ni4 (approved):
LGTM, however we'll need first to take care on CI hang under Windows.

ronaldtse (approved):
Thank you @antonsviridenko !

ronaldtse (approved):
LGTM, thanks @ni4 !

Issue #1394, opened 2021-01-22
cli_tests fail when running in parallel on Windows/MSVC


rrrooommmaaa:
## Description

Looks like one of `cli_tests` consistently hangs on 2 cores on MSVC build.

Let's try and run `cli_tests` with `-j1` and others with `-j${CTEST_PARALLEL}`, maybe?
Is there an easy filter to run all tests except those matching the pattern?

ni4:
@rrrooommmaaa Is it really a hang, not a timeout? cli_tests.Encryption test is the huge one, with a lot of key generations and so on.

rrrooommmaaa:
It hangs on various tests, not always on `cli_tests.Encryption`, probably some random cli test becomes a victim.
Sometime https://github.com/rnpgp/rnp/pull/1393/checks?check_run_id=1747921807 `Compression` passes in 369.18 sec, `Encryption` hangs
Sometime https://github.com/rnpgp/rnp/pull/1387/checks?check_run_id=1734829333 `Encryption` passes in 496.60 sec, `Compression` hangs.
Given that timeout is 3000 -- it does look like a hang (probably in botan library again).
@dewyatt Can you imagine what the reason can be?

ni4:
@rrrooommmaaa But it passed all CI when `CTEST_PARALLEL` was introduced?

rrrooommmaaa:
Maybe it passed with old botan library

ni4:
Yeah, looks like some locking occurs, see output at https://github.com/ni4/rnp/pull/12
If Compression test is run alone it takes ~270 seconds to finish.

PR #1397, opened 2021-01-26
Run cli_tests under MSVC in a single thread to avoid timeouts.

ni4:
This would work as a workaround until issue #1394 is fixed.

### DIFF
--- a/ci/test.msvc.sh
+++ b/ci/test.msvc.sh
@@ -10,7 +10,8 @@ mkdir -p "build/Testing/Temporary"
 cp "cmake/CTestCostData.txt" "build/Testing/Temporary"
 
 cd build
-ctest -j"${CTEST_PARALLEL}" -C Debug --output-on-failure
+ctest -j"${CTEST_PARALLEL}" -R rnp_tests -C Debug --output-on-failure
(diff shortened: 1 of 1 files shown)


ci/test.msvc.sh
```diff
@@ -10,7 +10,8 @@ mkdir -p "build/Testing/Temporary"
 cp "cmake/CTestCostData.txt" "build/Testing/Temporary"
 
 cd build
-ctest -j"${CTEST_PARALLEL}" -C Debug --output-on-failure
+ctest -j"${CTEST_PARALLEL}" rnp_tests -C Debug --output-on-failure
```
rrrooommmaaa, line 13: Looks like `ctest -j"${CTEST_PARALLEL}" -E "cli_tests" -C Debug --output-on-failure`
would do the trick. -E excludes those tests.

ni4: @rrrooommmaaa It seems to be much simpler - I forgot to add `-R` parameter. Now it should go fine.

antonsviridenko (approved):

rrrooommmaaa (approved):
LGTM!

PR #1413, opened 2021-02-11
Manpage cleanup

dkg:
This series improves the metadata in the manual page for librnp.

Note that with these changes, using a sensible manpage reader (like [man-db](https://nongnu.org/man-db/)), users can now do:

    man rnp_init

and it brings them to the correct page.


ni4:
Closing this as it is superseded with #1425

Issue #1414, opened 2021-02-12
C++ symbol leakage in shared object (librnp-0.so.0)


dkg:
#1135 pruned a bunch of symbol visibility from the shared object, which is great.  But as I'm packaging for debian, I've generated a [list of exposed symbols](https://salsa.debian.org/debian/rnp/-/blob/debian/main/debian/librnp-0-0.symbols), which include several C++-mangled symbols I'm pretty sure you do not want to expose:

```
_Z14rnp_log_switchv
_ZNSt10unique_ptrIN5Botan12HashFunctionESt14default_deleteIS1_EED1Ev
_ZNSt10unique_ptrIN5Botan12HashFunctionESt14default_deleteIS1_EED2Ev
_ZNSt12_Destroy_auxILb0EE9__destroyIP25pgp_transferable_subkey_tEEvT_S4_
_ZNSt12_Destroy_auxILb0EE9__destroyIP25pgp_transferable_userid_tEEvT_S4_
_ZNSt20__uninitialized_copyILb0EE13__uninit_copyIPK25pgp_transferable_userid_tPS2_EET0_T_S7_S6_
_ZNSt6vectorINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESaIS5_EE17_M_realloc_insertIJS5_EEEvN9__gnu_cxx17__normal_iteratorIPS5_S7_EEDpOT_
_ZNSt6vectorISt5arrayIhLm20EESaIS1_EE17_M_realloc_insertIJRKS1_EEEvN9__gnu_cxx17__normal_iteratorIPS1_S3_EEDpOT_
_ZNSt6vectorIhSaIhEE17_M_default_appendEm
_ZNSt6vectorIhSaIhEE17_M_realloc_insertIJRKhEEEvN9__gnu_cxx17__normal_iteratorIPhS1_EEDpOT_
_ZNSt8_Rb_treeI14pgp_pkt_type_tS0_St9_IdentityIS0_ESt4lessIS0_ESaIS0_EE24_M_get_insert_unique_posERKS0_
_ZNSt8_Rb_treeI14pgp_pkt_type_tS0_St9_IdentityIS0_ESt4lessIS0_ESaIS0_EE29_M_get_insert_hint_unique_posESt23_Rb_tree_const_iteratorIS0_ERKS0_
_ZZNSt8__detail18__to_chars_10_implImEEvPcjT_E8__digits
```

[...]

PR #1415, opened 2021-02-12
Make rnp_log_switch a C function, not C++

dkg:
This avoids exporting a C++-mangled symbol in the librnp shared object.

This addresses part of #1414.


ni4:
Closing as it is superseded with #1416 which was superseded with #1438

PR #1416, opened 2021-02-12
Avoid exposing rnp_log_switch in the dynamically-linked library

dkg:
rnp_log_switch is only used internally to librnp, and in the binaries
rnp and rnpkeys.

It deals with some internal state so that it can initialize lazily,
and so that a developer can tweak it manually with set_rnp_log_switch,
but this isn't intended for public functionality.

We solve this by exposing the functionality for handling
rnp_log_switch directly in its own compilation unit, and then
including that compilation unit in the objects that need it.

This means that set_rnp_log_switch will now be available to anyone who
includes that compilation unit, but it won't affect other copies of
that compilation unit in a dynamically-assembled executable.  However,
this is no loss compared to before: set_rnp_log_switch wasn't
accessible *at all* to a dynamically-linked rnp and rnpkeys before
this change, as it is not part of the public API.

Other situations may come up like this in the future: utility
functions that are used within librnp *and* directly by binaries like
rnp or rnpkeys, but are not (and should not) be exposed in the library
interface.  If this becomes more prevalent, the right thing to do is
to assemble a static object (maybe rnp-utils-private.a), which
comprises the compilation units that offer these utilities.  Link that
object into all the relevant places.

This will result in a slightly larger executable (one copy of the
private static lib on both sides of the shared object interface), but
it permits the compiler to inline small utility functions (rather than
calling into the dynamic library), and it exposes less in the library
interface.

If you want an example of another project that does something similar
with a private utility library included in both a shared object and a
binary that links against the shared object, take a look at
https://notmuchmail.org/

### DIFF
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -128,6 +128,7 @@ add_library(librnp-obj OBJECT
   fingerprint.cpp
   generate-key.cpp
   key-provider.cpp
+  log-switch.cpp
   misc.cpp
   pass-provider.cpp
   pgp-key.cpp
--- a/src/lib/log-switch.cpp
+++ b/src/lib/log-switch.cpp
@@ -1,0 +1,30 @@
+#include <stdlib.h>
+#include "utils.h"
+
+/* -1 -- not initialized
+    0 -- logging is off
+    1 -- logging is on
+*/
+int8_t _rnp_log_switch =
+#ifdef NDEBUG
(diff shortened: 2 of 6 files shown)


dkg:
Note that if this is done, then #1415 is probably not needed.  It also helps to resolve one of the exposed symbols mentioned in #1414.

ni4:
Closing as it is superseded with #1428 
Thanks again @dkg !

Issue #1408, opened 2021-02-11
intermittent test suite failures when run in parallel


dkg:
When i run the test suite for v0.14.0 in parallel on a machine with many cores, i get intermittent failures.

In particular, it looks like `cli_tests.py` is run multiple times concurrently, and sometimes two runs will try to produce the same output file at the same time.  During a failing test, i see this sort of output/warning fly by:

```
File '/tmp/x/rnp/build/src/tests/cleartext.rnp' already exists. Would you like to overwrite it (y/N)? Generating a new key...
```

I *want* to be able to run the test suite in parallel -- it's a large and comprehensive suite.  But i'm forced to run it serially to avoid these intermittent failures.

dewyatt:
This is a (probably undocumented) bug that occurs when the rnp build directory is in `/tmp`.

dkg:
i can confirm that i've only seen these failures when the build directory is in `/tmp`, but i'm mystified as to why that should be the case.  can you explain more?

dewyatt:
> i can confirm that i've only seen these failures when the build directory is in `/tmp`, but i'm mystified as to why that should be the case. can you explain more?

It's likely to be from:
https://github.com/rnpgp/rnp/blob/c04acfda7b5faf6bec2960cb8156fd93362411b0/src/tests/cli_tests.py#L751-L753

dkg:
thanks for the pointer.  #1420 appears to resolve the issue for me.

PR #1420, opened 2021-02-12
Avoid concurrent test failures in /tmp

dkg:
When built in /tmp with a parallel run of the test suite, the same
filename can be clobbered by different tests.

With this patch, it builds and tests clean.

Closes: #1408


ni4:
LGTM, but could you please create copy of this which merges into master instead of release?

ronaldtse:
@dkg I've changed to the master branch -- could you help rebase your patch on top of master? Thanks!

ni4:
@ronaldtse Probably it would be easier to close this PR and create a fresh one, WDYT? I can handle this, if needed.

dkg:
I'm fine with either approach.  I have not done a ton of testing with the master branch (esp. compared to the `release/0.x` branch), but #1426 should target it more cleanly, so i'm closing this one.  hopefully the changes won't be any problem to backport.

PR #1419, opened 2021-02-12
Produce more useful error reports if s2k_iteration_tuning test fails

dkg:
This doesn't necessarily solve the problem of a CPU core being
jittery, but it gives clearer detail if there is a failure.

Closes: #1418

An example failure is now:

```
15: Expected: ((static_cast<double>(iters_100) / iters_10)) > ((6)), actual: 5.93103 vs 6
15: [pgp_s2k_compute_iters() /home/dkg/src/rnp/rnp/src/lib/crypto/s2k.cpp:178] failed to create hash object
15: [  FAILED  ] rnp_tests.s2k_iteration_tuning (528 ms)
15: [----------] 1 test from rnp_tests (528 ms total)
15: 
15: [----------] Global test environment tear-down
15: [==========] 1 test from 1 test suite ran. (528 ms total)
15: [  PASSED  ] 0 tests.
15: [  FAILED  ] 1 test, listed below:
15: [  FAILED  ] rnp_tests.s2k_iteration_tuning
15: 
15:  1 FAILED TEST
````

PR #1424, opened 2021-02-15
Update master with commits from PR 1419.

ni4:
PR #1419 was merged into release, so we need it in master too.


ronaldtse (approved):

rrrooommmaaa (approved):
LGTM!

PR #1425, opened 2021-02-15
Add new man pages for rnp/rnpkeys/librnp.

ni4:
This PR adds new man pages for the RNP suite instead of outdated NetPGP ones.
Text of the man pages is now stored in markdown, and rendered to roff via pandoc.
Currently this is done manually since it seems to be an overload to add pandoc calls to the CMake files.
Also man pages are now copied during the install step.

Feel free to add any suggestions/corrections/share experience about the man pages authoring.


ni4:
@dkg would be thankful if you have time to take a look on this.

dkg:
@ni4, this looks great!  I haven't tried rendering the manpages directly with troff or anything, but the overall approach seems reasonable to me.

I'm not enough of a cmake guru to tell you how to do it, but it might be nice to include some sort of optional build step that would rebuild the roff files from the markdown.  I understand that you don't want to introduce a build dependency on pandoc, but for debian packaging, i'll probably re-generate the roff files from markdown during build anyway.  If you've got a particular way that you're preferring to invoke it, recording that invocation someplace in the build tooling will make it easier for downstream folks like me (and maybe for future upstream maintainers too).

thanks for doing this!

ronaldtse:
@ni4 I'd prefer not to include the rendered man pages in git but always re-generate them from source.

Re: packaging, I believe packagers should use published releases instead of git head, so it's not that much of an issue for them (the published release should have the rendered man pages).

dkg:
Just to clarify: I'm only one datapoint, but as a packager, i'm going to do my best to re-generate things using packaged tooling based on the "preferred form of modification" -- meaning, the underlying source.  so if you ship generated nroff and markdown, i'll probably ignore/discard the nroff and rebuild from the markdown.  I think (i hope) that's a good thing for both RNP and Debian and its downstream users.

I agree with @ronaldtse that it's generally not good form to commit generated files to an active revision control system -- better to only commit the source and the build instructions.

As for packagers using "published releases", where would we get them from that *aren't* based on the contents of the git repo?  The only place i've seen that i can get official "releases" from is github, which are just tagged releases.  Even the Download link on https://www.rnpgp.com/software/rnp/ redirects to the github tag page.

If there's some other place to retrieve published releases, i'd appreciate a pointer.

ronaldtse:
Thanks @dkg. For published releases, I just meant our GitHub releases page. It is possible to provide a slightly processed release archive that includes compiled man pages, but perhaps that's not even necessary...

ni4:
Thanks for the input. I'd also stick to 'have just source' strategy, it's quite logical. The only concern is the need to install `pandoc` for every rnp user to have the man pages, isn't this an overkill?

dkg:
Is it possible that you can use pandoc to generate html variants of the manpages, and then publish them automatically upon each release in a known URL?  that way, casual users-from-source can build and install rnp without worrying about pandoc (they can refer to the local .md files that they can read directly, or they can read them on the web), and packagers that are able to depend on a functional OS (most packagers) can just build-depend on pandoc.

note that this might mean needing some sort of build flag that controls whether the manpages are built or not (or autodetection based on the presence of pandoc), so that casual users-from-source don't end up with build errors due to a missing pandoc installation.

dkg:
(just wanted to note that users of packaged versions of rnp will *not* need pandoc installed to read the manpages, in case that wasn't clear, just like they won't need a C compiler to use `rnp` -- pandoc is only a build-dependency, not a runtime dependency)

ni4:
@dkg Thanks, addded cmake module/function to build man pages depending on pandoc availability, and issuing a warning if it is not available.

ronaldtse (approved):
Thanks @ni4!

ronaldtse:
Sorry for the lack of discussion but I've converted the man page flow to use AsciiDoc instead of Markdown.

Somehow the builds are still showing:
```
CMake Warning at cmake/Modules/AdocMan.cmake:47 (message):
  asciidoctor not found, man pages will not be generated.  Install
  asciidoctor or use the CMAKE_PROGRAM_PATH variable.
```

But we could first merge this and fix the CI in the #1437.

cmake/Modules/AdocMan.cmake
```diff
@@ -0,0 +1,117 @@
...
+    string(LENGTH ${CMAKE_SOURCE_DIR} CMAKE_SRC_LEN)
+    string(SUBSTRING ${SRC} 0 ${CMAKE_SRC_LEN} SRC_PREFIX)
+    string(COMPARE EQUAL ${CMAKE_SOURCE_DIR} ${SRC_PREFIX} _equal)
+    if (NOT _equal)
+      message(FATAL_ERROR "Cannot build DST path as SRC is outside of the CMake sources dir.")
+    endif()
+
+    # Strip '.adoc' from the output subpath
+    string(LENGTH ${SRC} SRC_LEN)
+    math(EXPR SUFFIX_LEN "${SRC_LEN} - ${CMAKE_SRC_LEN} - 3")
```
ni4, line 99: @ronaldtse Should not we have `- 5` here? Currently it generates file like `rnp.1.a`.

ronaldtse: Ha, I totally missed this function. Will fix (or if you can fix it...)

ni4: @ronaldtse No problem, I'll take care of it. Btw, what about to create separate repository (if there is no one yet) and put PandocMan.cmake, AdocMan.cmake there? I think people might find that helpful for other projects.

ronaldtse: I've just fixed it in ff551cd8

ronaldtse:
All builds are passing. @antonsviridenko @rrrooommmaaa can we have one more approval and merge? Thanks.

cmake/Modules/AdocMan.cmake
ronaldtse: > what about to create separate repository (if there is no one yet) and put PandocMan.cmake, AdocMan.cmake there? 

@ni4 this is a good idea. Let's do that.

antonsviridenko (approved):

cmake/Modules/AdocMan.cmake
ni4: > I've just fixed it in ff551cd

Thanks!

> @ni4 this is a good idea. Let's do that.

@ronaldtse  Do I have access to create a new repository (and what would be good name for it? `cmake-modules`?), or you'll do that? Also we may move `version.cmake` in that repository for less clutter.

ronaldtse: Of course you do, but I've just made it 😉  https://github.com/rnpgp/cmake-modules It's all yours!

ni4: @ronaldtse Thanks! :-)

src/lib/librnp.3.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+
+== OPTIONS
+
+*-o, --out-file*=_OUT_FILE_::
+  Write result to file _OUT_FILE_.
+
+*-c, --capture*::
+  Capture specimen if it's a picture of a life form.
+
```
ni4, line 37: @ronaldtse Does this came here by mistake or by intention :-) ?

ronaldtse: Complete mistake. Sorry. Can you help fix it?

ronaldtse: Thank you for spotting...

ni4: @ronaldtse Sure.

eli-schwartz:
> Just to clarify: I'm only one datapoint, but as a packager, i'm going to do my best to re-generate things using packaged tooling based on the "preferred form of modification" -- meaning, the underlying source. so if you ship generated nroff and markdown, i'll probably ignore/discard the nroff and rebuild from the markdown. I think (i hope) that's a good thing for both RNP and Debian and its downstream users.



> Is it possible that you can use pandoc to generate html variants of the manpages, and then publish them automatically upon each release in a known URL? that way, casual users-from-source can build and install rnp without worrying about pandoc (they can refer to the local .md files that they can read directly, or they can read them on the web), and packagers that are able to depend on a functional OS (most packagers) can just build-depend on pandoc.
> 
> note that this might mean needing some sort of build flag that controls whether the manpages are built or not (or autodetection based on the presence of pandoc), so that casual users-from-source don't end up with build errors due to a missing pandoc installation.


@dkg, just to note this is hostile to Gentoo, a from-source distro. Compiling pandoc is extremely painful and not even possible at all on many arches where we don't have GHC available. We don't want people to spend days just compiling manpages.

asciidoctor is not much better. ruby is unpopular, ruby *gems* have deep dependency trees, it adds considerable burden.

Gentoo also vehemently disputes "preferred form of modification", so you really should not go around telling upstream projects that a Debian policy is globally representative of all distros.

To put it bluntly, I only just realized today, you are personally the reason my system is missing the manpages for this package and I find it a bit... frustrating.

eli-schwartz:
> Sorry for the lack of discussion but I've converted the man page flow to use AsciiDoc instead of Markdown.

@ronaldtse, BTW in general I don't mind asciidoc, but the canonical tool for it is "asciidoc": https://github.com/asciidoc-py/asciidoc-py/

It is significantly less burdensome to install than the ruby based asciidoctor. This would be less of a problem if proper release tarballs were provided of course.

PR #1428, opened 2021-02-17
Add CMake option to disable Googletest download.

### LINKED ISSUE #1410: Support offline builds
Debian packages can't access the Internet when building.  We do want to run the test suites though.

the `googletest` debian package offers a reasonable, maintained version of `googletest` source, and it can be used instead of downloading `googletest` from the Internet.

It would be great if it was straightforward to tell the build process to look for `googletest` source in `/usr/src/googletest/googletest` instead of fetching it from the Internet. 

At the moment, I'm [patching `src/tests/CMakeLists.txt`](https://salsa.debian.org/debian/rnp/-/blob/debian/0.14.0-1/debian/patches/0001-avoid-fetching-files-or-using-git-during-build.patch#L140).  I can keep on patching it if you don't want to make it configurable, but i'd prefer either autodetection, or to just provide a simple configuration variable, instead of having to maintain a patch.

ni4:
This PR adds CMake option DOWNLOAD_GTEST, which is On by default.
Setting it to Off will search for preinstalled Googletest.

Fixes #1410

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -41,4 +41,36 @@ jobs:
         run: |
           set -x
           ./ci/run.sh
-
+  cmake-offline-googletest:
+    env:
+      DOWNLOAD_GTEST: Off
+      RNP_TESTS: rnp_tests
+    runs-on: ubuntu-latest
+    if: "!contains(github.event.head_commit.message, 'skip ci')"
+    timeout-minutes: 50
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Setup environment
+        run: |
+          . ci/gha/setup-env.inc.sh
+      - name: Build
+        run: |
+          set -x
+          ci/before_install.sh
+          sudo apt-get -y install googletest
+          mkdir -p "${LOCAL_BUILDS}/googletest-build"
+          pushd "${LOCAL_BUILDS}/googletest-build"
+          cmake /usr/src/googletest
+          sudo make -j2 install
+          popd
+          . ci/env.inc.sh
+          ./ci/install.sh
+      - name: tests
+        run: |
+          set -x
+          ./ci/run.sh
+          [ -d "${LOCAL_BUILDS}/rnp-build/src/tests" ]
+          [ ! -d "${LOCAL_BUILDS}/rnp-build/src/tests/googletest-build" ]
+          [ ! -d "${LOCAL_BUILDS}/rnp-build/src/tests/googletest-src" ]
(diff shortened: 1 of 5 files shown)


ni4:
@dkg could you please check whether this looks good as a solution for the #1410 ?

antonsviridenko (approved):

rrrooommmaaa (approved):
LGTM!

dkg:
This is great, but i don't think this actually addresses #1410.  #1410 is about `cmake/version.cmake`

dkg:
hm, i'm not sure that i can use this directly in debian without some additional patches.  In particular, it seems to assume that the GTest tooling is shipping as a pre-build library, one that is dynamically-linked to the executable.

However, see [debian documentation](https://sources.debian.org/src/googletest/1.10.0.20201025-1.1/debian/README.Debian/) for why this is not recommended.

Here's what i ended up using in debian after applying the patches in this series:

``` diff
diff --git a/src/tests/CMakeLists.txt b/src/tests/CMakeLists.txt
index 8386c7d..be75098 100644
--- a/src/tests/CMakeLists.txt
+++ b/src/tests/CMakeLists.txt
@@ -38,8 +38,10 @@ set_tests_properties(setupTestData PROPERTIES FIXTURES_SETUP testdata)
 include(GoogleTest)
 if (NOT DOWNLOAD_GTEST)
   # use preinstalled googletest
-  find_package(GTest REQUIRED)
-  set(GTestMain GTest::Main)
+  add_subdirectory(/usr/src/googletest/googletest
+                  ${CMAKE_CURRENT_BINARY_DIR}/googletest-build
+                  EXCLUDE_FROM_ALL)
+  set(GTestMain gtest_main)
 else()
   # download and build googletest
   configure_file(gtest-CMakeLists.txt.in googletest-download/CMakeLists.txt)
```

Without the above patch, i see the following errors during a build on a debian machine:

```
[…]
- Performing Test COMPILER_HAS_DEPRECATED_ATTR - Success
- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.2")
CMake Error at /usr/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:165 (message):
  Could NOT find GTest (missing: GTEST_LIBRARY GTEST_INCLUDE_DIR
  GTEST_MAIN_LIBRARY)
Call Stack (most recent call first):
  /usr/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:458 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.18/Modules/FindGTest.cmake:205 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  src/tests/CMakeLists.txt:41 (find_package)
```

ni4:
@dkg Thanks for the details. Does the solution in PR #1442  looks usable?

PR #1433, opened 2021-02-27
Add FFI function for batch signature removal.

### LINKED ISSUE #1006: Add feature to strip broken/unneeded key signatures

ni4:
This PR adds FFI function `rnp_key_remove_signatures()`, which would allow to quickly filter out unneeded signatures on key export or import.
Fixes #1006 

Also it refactors some internal and tests code as a side effect.

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -80,6 +80,17 @@ typedef uint32_t rnp_result_t;
 #define RNP_LOAD_SAVE_SINGLE (1U << 9)
 
 /**
+ * Flags for the rnp_key_remove_signatures
+ */
+
+#define RNP_KEY_SIGNATURE_INVALID (1U << 0)
+#define RNP_KEY_SIGNATURE_UNKNOWN_KEY (1U << 1)
+#define RNP_KEY_SIGNATURE_NON_SELF_SIG (1U << 2)
+
+#define RNP_KEY_SIGNATURE_KEEP (0U)
+#define RNP_KEY_SIGNATURE_REMOVE (1U)
+
+/**
  * Flags for output structure creation.
  */
 #define RNP_OUTPUT_FILE_OVERWRITE (1U << 0)
@@ -279,6 +290,24 @@ typedef void (*rnp_get_key_cb)(rnp_ffi_t   ffi,
                                const char *identifier_type,
                                const char *identifier,
                                bool        secret);
+
+/**
+ * @brief callback used to report back signatures from the function
+ *        rnp_key_remove_signatures(). This may be used to implement custom signature filtering
+ *        code or record information about the signatures which are removed.
+ * @param ffi
+ * @param app_ctx custom context, provided by application.
+ * @param sig signature handle to retrieve information about the signature. Callback must not
+ *            call rnp_signature_handle_destroy() on it.
+ * @param action action which will be performed on the signature. Currently defined are
+ *               RNP_KEY_SIGNATURE_KEEP an RNP_KEY_SIGNATURE_REMOVE.
+ *               Callback may overwrite this value.
+ *
+ */
+typedef void (*rnp_key_signatures_cb)(rnp_ffi_t              ffi,
+                                      void *                 app_ctx,
+                                      rnp_signature_handle_t sig,
+                                      uint32_t *             action);
 
 /** create the top-level object used for interacting with the library
  *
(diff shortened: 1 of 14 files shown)


include/rnp/rnp.h
```diff
@@ -79,6 +79,17 @@ typedef uint32_t rnp_result_t;
 #define RNP_LOAD_SAVE_PERMISSIVE (1U << 8)
 #define RNP_LOAD_SAVE_SINGLE (1U << 9)
 
+/**
+ * Flags for the rnp_key_remove_signatures
+ */
+
+#define RNP_KEY_SIGNATURE_INVALID (1U << 0)
+#define RNP_KEY_SIGNATURE_UNKNOWN_KEY (1U << 1)
+#define RNP_KEY_SIGNATURE_NON_SELF (1U << 2)
```
dewyatt, line 88: I know these are *somewhat* self-explanatory, but I still think there's slight ambiguity so can we document these more?

ni4: @dewyatt Defines are documented in the description of `rnp_key_remove_signatures()`. Anyway, let's rename it to the `RNP_KEY_SIGNATURE_NON_SELF_SIG`, looks better.

include/rnp/rnp.h
```diff
@@ -79,6 +79,17 @@ typedef uint32_t rnp_result_t;
...
+
+#define RNP_KEY_SIGNATURE_LEAVE (0U)
```
dewyatt, line 90: I would change this to `KEEP` rather than `LEAVE` if I understand what is doing correctly, it's a little clearer

ni4: Agree, sounds better.

dewyatt:
Personally I'd like to see a more generic `key_walk_signatures` interface here.
Something like:

```cpp
typedef void (*rnp_key_walk_signatures_cb)(rnp_ffi_t              ffi,
                                      void *                 app_ctx,
                                      rnp_signature_handle_t sig);
rnp_result_t rnp_key_walk_signatures(rnp_key_handle_t      key,
                                     uint32_t              flags, // filters
                                     rnp_key_signatures_cb sigcb,
                                     void *                app_ctx);
```

Then the callback could inspect a signature, determine what it wants to do (remove it, print some info out, store some stuff, do nothing, ...) and make API calls to do this (`rnp_key_signature_remove` or something).

This would be more easily expanded in the future and more flexible.
I haven't reviewed the code, so I'm just commenting on this interface (maybe there's a reason to not do it this way).

Or an iterator-like interface would be nice to see too. Is there a reason to not do this?

ni4:
> Personally I'd like to see a more generic `key_walk_signatures` interface here.
> Then the callback could inspect a signature, determine what it wants to do (remove it, print some info out, store some stuff, do nothing, ...) and make API calls to do this (`rnp_key_signature_remove` or something).
> 
> This would be more easily expanded in the future and more flexible.
> I haven't reviewed the code, so I'm just commenting on this interface (maybe there's a reason to not do it this way).
> 
> Or an iterator-like interface would be nice to see too. Is there a reason to not do this?

@dewyatt The main reason for the separate function is to be able to quickly cut-out all unneeded signatures. Removing those one-by-one would be of O(n^2) complexity, and say you have 1000 certifications on your main key and would like to export it with only the relevant self-sig. Or when you import large key and need to quickly clean it up.

However, we still may add separate `rnp_key_walk_signatures()` or iterator to be able to go through all the signatures and do whatever needed.

src/lib/pgp-key.h
```diff
@@ -343,6 +348,34 @@ struct pgp_key_t {
...
+    bool is_revocation(const pgp_subsig_t &sig) const;
+
+    /** @brief Check whether signature is userid revocation */
+    bool is_uid_revocation(const pgp_subsig_t &sig) const;
+
+    /** @brief Check whether signature is subkey binding */
+    bool is_binding(const pgp_subsig_t &sig) const;
+
+    /**
+     * @brief Validate key-related signature, where this is a signing key.
```
antonsviridenko, line 373: It is not clear what does this sentence mean.
I guess it is a replacement for `pgp_key_validate_signature()` which was documented this way:
> Validate key's signature.

ni4: Thanks, will update the description.

ni4:
@dewyatt @antonsviridenko Is it okay to go now?

antonsviridenko (approved):

dewyatt (approved):
The tests with stringstream seem a little weird to me but I assume there was good reason

ni4:
> The tests with stringstream seem a little weird to me but I assume there was good reason

It was just the shortest way to write things down (and in one line). Not the best solution, but may be easily fixed afterwards anyway.

PR #1434, opened 2021-02-28
Make CRC failures non-fatal on dearmoring

### LINKED ISSUE #1401: Consider to relax dearmoring code.
## Description

As reported in the https://bugzilla.mozilla.org/show_bug.cgi?id=1691307, GnuPG successfully imports armored key without valid CRC line.
As CRC doesn't add any security level, we should consider to do the same.

antonsviridenko:
Closes #1401

### DIFF
--- a/src/librepgp/stream-armor.cpp
+++ b/src/librepgp/stream-armor.cpp
@@ -62,6 +62,7 @@ typedef struct pgp_source_armored_param_t {
     unsigned brestlen;   /* number of bytes in brest */
     bool     eofb64;     /* end of base64 stream reached */
     uint8_t  readcrc[3]; /* crc-24 from the armored data */
+    bool     has_crc;    /* message contains CRC line */
     pgp_hash_t crc_ctx;  /* CTX used to calculate CRC */
 } pgp_source_armored_param_t;
 
@@ -154,6 +155,8 @@ armor_read_crc(pgp_source_t *src)
     param->readcrc[0] = (dec[0] << 2) | ((dec[1] >> 4) & 0x0F);
     param->readcrc[1] = (dec[1] << 4) | ((dec[2] >> 2) & 0x0F);
     param->readcrc[2] = (dec[2] << 6) | dec[3];
+
+    param->has_crc = true;
 
     src_skip(param->readsrc, 5);
     return src_skip_eol(param->readsrc);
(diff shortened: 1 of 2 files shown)


src/librepgp/stream-armor.cpp
```diff
@@ -386,7 +385,6 @@ armored_src_read(pgp_source_t *src, void *buf, size_t len, size_t *readres)
 
         if (memcmp(param->readcrc, crc_fin, 3)) {
```
rrrooommmaaa, line 386: I'm concerned about comparing `crc_fin` with potentially uninitialized `readcrc` (if CRC wasn't read successfully).
Or, to be more precise, the warning "CRC mismatch" will not appear in an indeterministic manner (every once in 2^24 cases, but still...)
What do you think, @ni4

rrrooommmaaa (commented):
Rare but indeterministic behaviour

src/librepgp/stream-armor.cpp
antonsviridenko: `readcrc` is initialized to zeroes in https://github.com/rnpgp/rnp/blob/master/src/librepgp/stream-common.cpp#L358 by `calloc()`

antonsviridenko: @rrrooommmaaa but at least "wrong crc line" will be printed anyway

rrrooommmaaa: My point is: "CRC mismatch" should never be printed if "Wrong CRC line" was printed
or "CRC mismatch" should be printed every time when "Wrong CRC line" is printed.
In this implementation it won't be printed when "Wrong CRC line" was signaled and crc_fin == 0, right?
So looks like a glitch.

ni4: Agree with @rrrooommmaaa - let's introduce some flag like `has_crc` and print different messages for different cases ("Warning: absent CRC", "Warning: malformed CRC", "Warning: CRC mismatch").
This may also help later in debugging.

ni4 (changes requested):
Please see my comment.

ni4 (approved):
LGTM now, thanks!

ronaldtse (approved):
Thanks @antonsviridenko !

PR #1438, opened 2021-03-02
Extract logging code to the separate file/header.

ni4:
This PR is based on PR #1416 by @dkg and additionally extracts all logging code to the separate code (this would be needed in later PR, which gets rid of lib includes for `rnpcfg.cpp` and `tui.cpp`).

See this comment for more details: https://github.com/rnpgp/rnp/pull/1416#issue-572227080

### DIFF
--- a/src/lib/logging.cpp
+++ b/src/lib/logging.cpp
@@ -1,0 +1,56 @@
+/*-
+ * Copyright (c) 2017-2021 Ribose Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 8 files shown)


ronaldtse (approved):
LGTM thanks @ni4 ! There's a remaining problem with fuzzing though.

src/lib/logging.cpp
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+#include "string.h"
+#include "logging.h"
+
+/* -1 -- not initialized
+    0 -- logging is off
+    1 -- logging is on
+*/
+int8_t _rnp_log_switch =
```
antonsviridenko, line 34: I guess this variable should not be referenced outside this file, so it makes sense to declare it as `static`

ni4: @antonsviridenko Thanks, fixed.

ni4:
@antonsviridenko Any further comments?

antonsviridenko (approved):

PR #1444, opened 2021-03-06
Remove CLI dependencies on internal library headers.

ni4:
This PR reorganises includes/headers so CLI now will not include any underlying library types, except define-only `defaults.h` and `logging.h` and common utility functions from `str-utils.h`/`file-utils.h`.


ronaldtse (approved):
Great cleanup, thanks @ni4 !

rrrooommmaaa (approved):
LGTM!

PR #1450, opened 2021-03-09
Add .editorconfig

ribose-jeffreylau:
This is for general development quality of life.

See: https://editorconfig.org/


ronaldtse (approved):

dewyatt (approved):
At a glance it looks like it won't conflict with clang-format so LGTM

PR #1455, opened 2021-03-10
Remove unused function bn_print_fp()

ni4:
Just some minor cleanup.

### DIFF
--- a/src/lib/crypto/bn.cpp
+++ b/src/lib/crypto/bn.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017-2018 Ribose Inc.
+ * Copyright (c) 2017-2021 Ribose Inc.
  * Copyright (c) 2012 Alistair Crooks <agc@NetBSD.org>
  * All rights reserved.
  *
@@ -27,9 +27,8 @@
 #include "bn.h"
 #include <botan/ffi.h>
 #include <stdlib.h>
+#include <assert.h>
 #include "utils.h"
-
-/**************************************************************************/
 
 /* essentiually, these are just wrappers around the botan functions */
 /* usually the order of args changes */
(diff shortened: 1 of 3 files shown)


src/lib/crypto/bn.cpp
```diff
@@ -39,38 +37,27 @@
 bignum_t *
 bn_bin2bn(const uint8_t *data, int len, bignum_t *ret)
 {
-    if (data == NULL) {
+    if (!data || !ret) {
```
dewyatt, line 40: Just a heads up, I don't think this is right, I think the original intention here was:
* if `!data` return NULL, as that's clearly a mistake. (should be an assertion here too)
* if `!ret` allocate a `bn` and use that for the function and return it
* if `ret` then use that `bn` and return it

ni4: @dewyatt Thanks, mislooked this. Re-pushing. Will RNP_LOG() be enough as we don't use assert's often?

ni4: Ok, let's add an assertion as well.

src/lib/crypto/bn.cpp
```diff
@@ -27,10 +27,9 @@
 #include "bn.h"
 #include <botan/ffi.h>
 #include <stdlib.h>
+#include <assert.h>
```
dewyatt, line 30: I think technically these C headers are deprecated and we should be using `cassert`, but I know we're doing this all over the codebase and it's maybe not a concern right now

dewyatt (approved):

src/lib/crypto/bn.cpp
ni4: You're right, just sometimes hard to think C++ when surrounded by the C code :)

ronaldtse (approved):

ronaldtse:
Thanks guys!

PR #1463, opened 2021-03-26
Move secure memory handling to the crypto folder.

ni4:
This PR adds secure_array type, which will be automatically cleared in the destructor, and replaces pgp_forget function with secure_clear. Also it moves all these functions to the crypto folder.

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -57,7 +57,6 @@
 #include "utils.h"
 #include "json_utils.h"
 #include "version.h"
-#include <botan/secmem.h>
 #include "ffi-priv-types.h"
 #include "file-utils.h"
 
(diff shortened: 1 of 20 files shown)


dewyatt:
I kinda vote against the use of the stack for secure storage, we should be using heap-allocated mlock'd memory ideally. Is the idea to move towards that after this?

ni4:
@dewyatt Yeah, I don't like that too, we should change to use allocated memory. Also we should change this for secure MPIs of the secret keys, which would require some work as `pgp_mpi_t` is now treated as opaque structure. So idea was to first quick-change current code, separating crypto, and later issue PR with more solid updates.

src/lib/rnp.cpp
```diff
@@ -5415,8 +5410,9 @@ try {
     }
 
     /* encrypt secret key if requested */
-    if (op->password) {
-        prov = {.callback = rnp_password_provider_string, .userdata = (void *) op->password};
+    if (op->password.size()) {
```
dewyatt, line 5413: I think `!op->password.empty()` is more common

src/lib/rnp.cpp
```diff
@@ -5433,11 +5429,7 @@ try {
     }
     ret = RNP_SUCCESS;
 done:
-    if (op->password) {
-        pgp_forget(op->password, strlen(op->password) + 1);
-        free(op->password);
-        op->password = NULL;
-    }
+    op->password.resize(0);
```
dewyatt, line 5432: Seems odd instead of `op->password.clear()` (should be same effect on modern STLs I believe)

ronaldtse (approved):
Thanks @ni4! Feel free to merge after addressing the comments!

antonsviridenko (approved):

src/lib/rnp.cpp
ni4: Thanks, @dewyatt. Will fix and force-push it now.

ni4:
Comments addressed, merging.

PR #1466, opened 2021-04-04
More tests for deamor, handling large (>=64 characters) lines. Closes #942

### LINKED ISSUE #942: Make sure dearmor correctly handles large lines
## Description

Construct corresponding data files and make sure rnp correctly handles following cases:
- too many spaces in front of armored message (>32k)
- too large header name/values
- too large single armored line.
Too large is about 32-64k, i.e. larger internal buffers size.

antonsviridenko:
Also now some amount of whitespace charactes is allowed in front of armored messages (Armor header should be completely located in the first 1024 bytes in order to be detected). Increased peek size from 128 to 1024 for better compatibility with GnuPG.


rrrooommmaaa (approved):
LGTM!

ni4 (approved):
LGTM, feel free to merge once Debian 9 build completes (or restart test if it fails after minute, that happens). Also feel free to ping me for review if it doesn't happen in a couple of days - may miss notifications sometimes.

PR #1475, opened 2021-04-09
Update man pages and packaging/bugreport email for the version 15.0

ni4:
We forgot to update rnp version in man pages, as well as packaging email in v0.15.0 release.
This PR is aimed to fix it (and probably we should then cherry pick it to the release branch/re-push tag?).

### DIFF
--- a/cmake/info.cmake
+++ b/cmake/info.cmake
@@ -26,7 +26,7 @@
 set(PACKAGE_VENDOR "Ribose Inc.")
 set(PACKAGE_URL "https://github.com/rnpgp/rnp")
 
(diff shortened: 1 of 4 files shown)


ronaldtse (approved):
Thanks @ni4!

ronaldtse:
Merging with one approval since these are  minor docs changes. 

@ni4 agree with cherry pick but maybe tag an increased version for consistency reasons?

ni4:
@ronaldtse Then we'll need to update man pages again and put 0.15.1 there :) Basically, would tag re-push cause any problems?

PR #1477, opened 2021-04-09
Fix man versions for the v0.15.0 release (release branch).

ni4:


ni4:
As from https://github.com/rnpgp/rnp/pull/1475#issuecomment-816563971 :

@ronaldtse Then we'll need to update man pages again and put 0.15.1 there :) Basically, would tag re-push cause any problems?

ronaldtse:
> Then we'll need to update man pages again and put 0.15.1 there :) Basically, would tag re-push cause any problems?

Don't think there would be problems...

I think the versions in man pages should probably be generated via CMake, not committed to the man page files.

ronaldtse:
@ni4 b88091dc implements the CMake versioning for man pages.

ni4:
@ronaldtse Thanks, this way it is much clear! LGTM, but cannot add me to the reviewers :)

antonsviridenko (approved):

ronaldtse:
@dewyatt can you have a quick look to see if this is good to go? Thanks!

ni4:
@dewyatt Is it good to go now, after merge to master?

ni4:
Ping @dewyatt

dewyatt (approved):

Xiretza:
> Basically, would tag re-push cause any problems?

Please don't do this again, it breaks everything downstream that expects tags to be immutable. Just create a bugfix release.

ni4:
@Xiretza Thanks for the input. Could you please provide an example of what can be broken? Is it some cmake build process or something similar?

Xiretza:
As an example, the AUR package (https://aur.archlinux.org/packages/rnp/) broke and had to be updated because the downloaded tarball suddenly had a different checksum. It also ruins reproducible builds because building the "old" 0.15.0 obviously produces different build products than the "new" 0.15.0, even though the version claims to be the same.

ni4:
@Xiretza Thanks for the clarification! Will not move tags anymore.

PR #1482, opened 2021-04-11
Make building man pages optional through option ENABLE_DOC.

### LINKED ISSUE #1481: Explicitly enable or disable building man pages in cmake configuration
## Description

As man pages are built using `asciidoctor`, current implementation just tries to autodetect presence of `asciidoctor` and depending on result either builds manpages or not. Now I am trying to create package for source-based Gentoo distro, and there building documentation is optional and configurable part, depending on user preferences. If user enables `doc` USE-flag either globally or specifically for this package, documentation will be built. And if it is enabled, additional build-time dependency (`asciidoctor` package in this case) should be pulled and installed first. So configuration of rnp build should be more deterministic. I propose to add new cmake variable like `ENABLE_DOC` or `ENABLE_MANPAGES`

## Expected Behavior

If documentation build is explicitly enabled (like -DENABLE_DOC=On), cmake should check for presence of `asciidoctor` and fail to continue if `asciidoctor` is not found. If it is explicitly disabled, then it should not check for `asciidoctor` at all and skip building manpages. Third option (like ENABLE_DOC=Auto) can be implemented and set as a default value in order to keep the current behavior.

## Actual Behavior

`asciidoctor` presence is automatically detected during build and there is no way to disable it or strictly require this dependency for succesfull build

antonsviridenko:
Closes #1481

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,12 @@ option(ENABLE_FUZZERS "Enable fuzz targets.")
 option(DOWNLOAD_GTEST "Download Googletest" On)
 option(DOWNLOAD_RUBYRNP "Download ruby-rnp and run related tests." On)
 
+# tri-state options
+set(TRISTATE_VALUES On Off Auto)
+
+set(ENABLE_DOC Auto CACHE STRING "Enable building documentation.")
+set_property(CACHE ENABLE_DOC PROPERTY STRINGS ${TRISTATE_VALUES})
+
 # so we can use our bundled finders
 set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/Modules")
 
@@ -137,7 +143,9 @@ if (ENABLE_SANITIZERS)
 endif()
 
 # adoc for man generation
-include(AdocMan)
+if (ENABLE_DOC)
+  include(AdocMan)
+endif()
 
 # everything else is in subdirs
 add_subdirectory(src/examples)
--- a/cmake/Modules/AdocMan.cmake
+++ b/cmake/Modules/AdocMan.cmake
@@ -44,7 +44,14 @@ find_program(ADOCCOMMAND_PATH
 )
 
 if(NOT EXISTS ${ADOCCOMMAND_PATH})
-  message(WARNING "AsciiDoc processor not found, man pages will not be generated. Install asciidoctor or use the CMAKE_PROGRAM_PATH variable.")
+  set(ADOC_MISSING_MSG "AsciiDoc processor not found, man pages will not be generated. Install asciidoctor or use the CMAKE_PROGRAM_PATH variable.")
+
+  string(TOLOWER "${ENABLE_DOC}" ENABLE_DOC)
+  if (ENABLE_DOC STREQUAL "auto")
+    message(WARNING ${ADOC_MISSING_MSG})
+  elseif(ENABLE_DOC)
+    message(FATAL_ERROR ${ADOC_MISSING_MSG})
+  endif()
 else()
   set(ADOCCOMMAND_FOUND 1)
 endif()
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -348,4 +348,6 @@ if (PKG_CONFIG_FOUND)
 endif()
 
 # Build and install man page
-add_adoc_man("${CMAKE_CURRENT_SOURCE_DIR}/librnp.3.adoc" ${RNP_VERSION})
+if (ENABLE_DOC)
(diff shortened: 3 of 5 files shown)


antonsviridenko:
Soo, as failed tests show, either asciidoctor should be installed in CI configurations, or building manpages should be disabled.

CMake's `option()` does not support three-state values, I've found possible workaround, not sure if it is suitable:

https://cmake.org/pipermail/cmake/2016-October/064342.html

antonsviridenko:
@ni4 any ideas how to fix failed tests? Where 'asciidoctor' can be added as a dependency exactly?

ni4:
@antonsviridenko Sorry for a late reply.
It could be better to use string variable instead of option, i.e.
```
if (NOT BUILD_DOC)
  set(BUILD_DOC "auto" CACHE STRING
    "Control building of documentation, which requires asciidoctor. Possible values are on, off and auto. Default is auto, which builds documentation only if asciidoctor is found."
    FORCE
  )
endif()
// not sure that one below will work - probably you'll need temporary variable
string(TOLOWER "${BUILD_DOC}" BUILD_DOC)

```

So fail only in case if BUILD_DOC is forcibly set to on.
I would suggest to not fail if asciidoctor is not found, and just issue a warning as it was before.
But, some tests could be added which checks whether man is built correctly under the windows, if asciidoctor is present.

ni4:
Accidentally hit return. For the MSYS test asciidoctor may be installed by adding it to packages in function `msys_install()`, `before_install.sh`.
For the MSVC build i'm not sure, documentation tells about Chocolatey or Rubyinstaller. Maybe @rrrooommmaaa could help here.

antonsviridenko:
My latest commit did not make any improvements :(

ping @rrrooommmaaa

ni4:
@antonsviridenko Yeah, since you still use Off/On for the BUILD_DOC, instead of adding relaxed 'auto' mode.

ni4:
@antonsviridenko Just came across possibly helpful cmake setup: https://stackoverflow.com/questions/39670879/cmake-variable-with-limited-valid-values

Basically, we may have On, Off or Auto value for the `BUILD_DOC` (default auto), and then:
- if it is Auto then build docs only if asciidoctor is available
- if it is On, then try to build docs and report a fatal error if asciidoctor is not found
- if it is Off then just do not build and search for asciidoctor.

antonsviridenko:
yeah, looks interesting

rrrooommmaaa (approved):
LGTM!

CMakeLists.txt
```diff
@@ -137,7 +143,9 @@ if (ENABLE_SANITIZERS)
 endif()
 
 # adoc for man generation
-include(AdocMan)
+if (ENABLE_DOC)
```
ni4, line 146: Will this work correctly if ENABLE_DOC is not upper-case, i.e. was set like -DENABLE_DOC=on or -DENABLE_DOC=off ?

cmake/Modules/AdocMan.cmake
```diff
@@ -44,7 +44,13 @@ find_program(ADOCCOMMAND_PATH
 )
 
 if(NOT EXISTS ${ADOCCOMMAND_PATH})
-  message(WARNING "AsciiDoc processor not found, man pages will not be generated. Install asciidoctor or use the CMAKE_PROGRAM_PATH variable.")
+  set(ADOC_MISSING_MSG "AsciiDoc processor not found, man pages will not be generated. Install asciidoctor or use the CMAKE_PROGRAM_PATH variable.")
+
+  if (ENABLE_DOC STREQUAL "Auto")
```
ni4, line 49: Will this work correctly if ENABLE_DOC was set in another case, like -DENABLE_DOC="auto" ?

ni4 (approved):
LGTM, just few concerns which I cannot check right now.
Feel free to merge once those concerns are excessive :)

cmake/Modules/AdocMan.cmake
antonsviridenko: It appears that no, STREQUAL is case-sensitive.
https://cmake.org/pipermail/cmake/2012-June/050651.html

Meh, things were easier in autotools, why did rnp switch to cmake :D

CMakeLists.txt
antonsviridenko: Yes, should work correctly


>    True if the constant is 1, ON, YES, TRUE, Y, or a non-zero number. False if the constant is 0, OFF, NO, FALSE, N, IGNORE, NOTFOUND, the empty string, or ends in the suffix -NOTFOUND. **Named boolean constants are case-insensitive**. If the argument is not one of these specific constants, it is treated as a variable or string and the following signature is used.

PR #1484, opened 2021-04-18
Check home directory, fixes #1304

### LINKED ISSUE #1304: Check whether HOME dir exists and is writable.
## Description

Separated from the PR https://github.com/rnpgp/rnp/pull/1302#issuecomment-691087258

antonsviridenko:


antonsviridenko:
ping @ni4

src/common/file-utils.cpp
```diff
@@ -109,6 +109,19 @@ rnp_fopen(const char *filename, const char *mode)
 #endif
 }
 
+int
+rnp_access(const char *path, int mode)
+{
+#ifdef _WIN32
+    try {
+        return _waccess(wstr_from_utf8(path).c_str(), mode);
```
rrrooommmaaa, line 117: Should we mask 0x1 bit as only values 0, 2, 4 and 6 are supported on Windows according to 
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/access-waccess?view=msvc-160
Also, the docs say, "it does not check the filesystem security settings",
meaning that if the directory needs administrative permissions `rnp_access` may return success.
Should we consider this situation?

src/rnp/fficli.cpp
```diff
@@ -1552,6 +1552,13 @@ rnp_cfg_set_ks_info(rnp_cfg &cfg)
         defhomedir = true;
     }
 
+    struct stat st;
+
+    if (rnp_stat(homedir.c_str(), &st) || rnp_access(homedir.c_str(), R_OK | W_OK)) {
+        ERR_MSG("Home directory '%s' does not exist or is not writeable!", homedir.c_str());
```
ni4, line 1558: Looks like 'writable' is more common spelling then 'writeable'.

ni4 (approved):
LGTM, just minor spelling comment.
Thanks!

src/rnp/fficli.cpp
ronaldtse: Ping, thanks @antonsviridenko !

src/common/file-utils.cpp
antonsviridenko: I think removing X_OK definition on Windows should be enough? So only supported values are defined as macros.

About second part,
looks like WinAPI implemenation of this task is bit complicated,
I've found some guide here
https://blog.aaronballman.com/2011/08/how-to-check-access-rights/
but in comments there are reports that it does not properly work on Windows 8 and newer...

antonsviridenko: And some Windows-specific test case must be added then,
where test home directory exists, but write access is not allowed for
Windows user running tests. This will require admin privileges...

antonsviridenko: @rrrooommmaaa @ni4 @ronaldtse should we implement checking Windows security permissions or keep as is?

ni4: @antonsviridenko Actually it's not just for tests - we should make sure that if user runs rnp (via FFI or CLI) on non-writeable home directory, it handles this case gracefully. I think this could be extracted to the separate task.

rrrooommmaaa: Actually, the only test of missing permissions for Unix directory gives quite a headache in CI tests (especially in docker containers). We need to create a special user there. Maybe it's enough to test it locally on WIndows?

ni4: @rrrooommmaaa We should make sure that this case is handled correctly, i.e. not crash/whatever else in this case. Adding user is just a few lines in GHA file, don't think this is a problem.

antonsviridenko: Is it ok to use WinAPI functions here?

antonsviridenko: ping @ni4 ^^
looks like we'll need to add linking to advapi32.dll
in order to use AccessCheck() and other related WinAPI functions

ni4: @antonsviridenko Sorry, missed your previous comment. I think it would be enough to keep it as it is now, without adding new dependencies and complications. Probably I used incorrect wording in previous comments - I meant, that we should make sure that RNP works ok in case if _waccess() returns 'writeable', but that requires administrative permissions or so on. Basically, even if we'd implement all complicated checks, we cannot 100% guarantee that write to the directory will be allowed.

antonsviridenko:
ping @ronaldtse

rrrooommmaaa:
> I meant, that we should make sure that RNP works ok in case if _waccess() returns 'writeable', but that requires administrative permissions or so on.

So has anyone actually tried what `_waccess` returns for a protected folder like `C:\Windows\`?

antonsviridenko:
> So has anyone actually tried what _waccess returns for a protected folder like C:\Windows\?

no, not yet :) going to try

antonsviridenko:
@rrrooommmaaa Unfortunately _waccess() returns success for write access. But it was  a good idea to check against known system files and directories. Pls see my latest commit.

rrrooommmaaa:
> Unfortunately _waccess() returns success for write access. But it was a good idea to check against known system files and directories. Pls see my latest commit.

Well, we can create an issue for later to solve this for Windows, as @ni4 suggested.
A smarter implementation to create a test file or whatever is relatively simple.

ronaldtse (approved):
Thanks @antonsviridenko !

ronaldtse:
Follow up issue has been created.

PR #1492, opened 2021-04-28
FFI: add function rnp_key_valid_till64()

ni4:
Currently function `rnp_key_valid_till()` returns uint32_t, which is good enough only till the year 2038.
While it is very unlikely that some implementation would produce keys which expire after the year 2038 (and actually till the year 2038 OpenPGP standard should be updated to correctly handle creation time), we still should have some interface for the future.

Fixes issue #1480.


src/lib/rnp.cpp
```diff
@@ -6421,6 +6421,28 @@ FFI_GUARD
...
+    }
+    uint64_t     res = 0;
+    rnp_result_t ret = rnp_key_valid_till64(handle, &res);
+    if (ret) {
+        return ret;
+    }
+    if (res == UINT64_MAX) {
+        *result = UINT32_MAX;
+    } else if (res > UINT32_MAX) {
+        *result = UINT32_MAX - 1;
```
nwalfield, line 6436: Why the `- 1`?

ni4: Because UINT32_MAX is reserved for 'never expires'.

nwalfield: Then you should probably do: `if (res >= UINT32_MAX)` to catch the case that `res` happens to equal `UINT32_MAX`.

ni4: Agree, missed that. Thanks.

antonsviridenko (approved):

ronaldtse (approved):

ronaldtse:
Thank you all!

PR #1496, opened 2021-05-04
Use RNP_FEATURE_* defines instead of raw string values.

ni4:
This way code should look better.


antonsviridenko (approved):

ronaldtse (approved):
LGTM thanks @ni4!

Issue #1316, opened 2020-09-22
RNP takes certificate expiration information from non-primary userid bindings


teythoon:
## Description

The following certificate is expired using a key expiration time
subpacket on the primary userid's binding signature.  There is a
conflicting subpacket on a non-primary userid's binding signature
that overrides the former subpacket, but shouldn't.  As
revocation is not well supported in the OpenPGP ecosystem,
liveness through e.g. certificate expiration is the most
important way to retire key material.

Relevant test: https://tests.sequoia-pgp.org/#Certificate_expiration

```
-----BEGIN PGP PUBLIC KEY BLOCK-----

xsDNBF2lnPIBDAC5cL9PQoQLTMuhjbYvb4Ncuuo0bfmgPRFywX53jPhoFf4Zg6mv
/seOXpgecTdOcVttfzC8ycIKrt3aQTiwOG/ctaR4Bk/t6ayNFfdUNxHWk4WCKzdz
/56fW2O0F23qIRd8UUJp5IIlN4RDdRCtdhVQIAuzvp2oVy/LaS2kxQoKvph/5pQ/
5whqsyroEWDJoSV0yOb25B/iwk/pLUFoyhDG9bj0kIzDxrEqW+7Ba8nocQlecMF3
X5KMN5kp2zraLv9dlBBpWW43XktjcCZgMy20SouraVma8Je/ECwUWYUiAZxLIlMv
9CurEOtxUw6N3RdOtLmYZS9uEnn5y1UkF88o8Nku890uk6BrewFzJyLAx5wRZ4F0
qV/yq36UWQ0JB/AUGhHVPdFf6pl6eaxBwT5GXvbBUibtf8YI2og5RsgTWtXfU7eb
SGXrl5ZMpbA6mbfhd0R8aPxWfmDWiIOhBufhMCvUHh1sApMKVZnvIff9/0Dca3wb
vLIwa3T4CyshfT0AEQEAAc0hQm9iIEJhYmJhZ2UgPGJvYkBvcGVucGdwLmV4YW1w
bGU+wsEXBBMBCgAqBYJfacc3BYkBQFRFBQsJCAcCBhUKCQgLAgQWAgMBAheAApkB
AhsDAh4BACEJEPv8yCoBXnMwFiEE0aZuGiOxgsmYD3iM+/zIKgFeczBmQQwAkOKL
[...]

Issue #1311, opened 2020-09-18
Expired OpenPGP Certificate not recognized as expired


nwalfield:
Consider the following key (attached):

```
$ sq packet dump primary-expired.pgp
Public-Key Packet, new CTB, 51 bytes
    Version: 4
    Creation time: 2020-09-18 11:57:00 UTC
    Pk algo: EdDSA Edwards-curve Digital Signature Algorithm
    Pk size: 256 bits
    Fingerprint: 8C7B 1E85 3474 17FC A304  D5EA E431 8B62 A47A A25C
    KeyID: E431 8B62 A47A A25C
  
Signature Packet, new CTB, 135 bytes
    Version: 4
    Type: DirectKey
    Pk algo: EdDSA Edwards-curve Digital Signature Algorithm
    Hash algo: SHA512
    Hashed area:
      Signature creation time: 2020-09-18 11:57:00 UTC (critical)
      Key expiration time: PT1S (critical)
      Symmetric algo preferences: AES256
      Hash preferences: SHA512
      Key flags: CS (critical)
      Features: MDC
    Unhashed area:
      Issuer: E431 8B62 A47A A25C
      Issuer Fingerprint: 8C7B 1E85 3474 17FC A304  D5EA E431 8B62 A47A A25C
    Digest prefix: 69F7
    Level: 0 (signature over data)
  
User ID Packet, new CTB, 45 bytes
    Value: Primary Expired <primary-expired@example.org>
  
Signature Packet, new CTB, 132 bytes
    Version: 4
    Type: PositiveCertification
    Pk algo: EdDSA Edwards-curve Digital Signature Algorithm
    Hash algo: SHA512
    Hashed area:
      Signature creation time: 2020-09-18 11:57:00 UTC (critical)
      Symmetric algo preferences: AES256
      Hash preferences: SHA512
      Primary User ID: true (critical)
      Key flags: CS (critical)
      Features: MDC
    Unhashed area:
      Issuer: E431 8B62 A47A A25C
      Issuer Fingerprint: 8C7B 1E85 3474 17FC A304  D5EA E431 8B62 A47A A25C
    Digest prefix: 1526
    Level: 0 (signature over data)
  
Public-Subkey Packet, new CTB, 56 bytes
    Version: 4
    Creation time: 2020-09-18 11:57:00 UTC
    Pk algo: ECDH public key algorithm
    Pk size: 256 bits
[...]

PR #1456, opened 2021-03-12
ci: Add CentOS8 support

ribose-jeffreylau:
This PR refactors the pipeline scripts <s>(with ideas taken from #1459 so as to minimize merge conflicts at a later stage for either branch)</s><b>(done)</b> in order to consolidate common build and install routines.

A pipeline for CentOS 8 has been added<s>, as well as for packaging RPMs in various CentOS and Fedora versions</s>.

<s>Signing</s><b>Support for building</b> of RPM packages will be added in a separate PR after this PR has been merged.

PR #1501, opened 2021-05-19
Fix updating of key expiration for keys with multiple uids.

### LINKED ISSUE #1497: Update self-certifications for all userids in pgp_key_set_expiration().
## Description

Currently we update expiration only for the latest self-signature, while we should update latest self-signature for all uids (and direct-key sigs as well). This may cause the problem like the one described here: https://bugzilla.mozilla.org/show_bug.cgi?id=1704182

ni4:
Main purpose of this PR was to fix issue #1497, however due to internal changes it also touches issues #1311, #1316 and #1315. Those would be closed with a separate PR which extends test coverage.

Now key expiration/flags behaviour would be changed:
- first we'll look for key expiration from the direct-key signature
- then for the latest self-certification for the primary uid (if any), and pick the more strict of both.
- in other cases we'll pick the latest self-signature from the any userid.

Same would apply to the key flags.

Also this PR fixes problems, found in `pgp_key_t::del_uid()` and `pgp_key_t::is_expired()`.

Fixes #1497


ni4:
@ribose-jeffreylau Do you have any quick idea why all centos7/8 buids are failing? Nothing seem to be changed since #1456 was merged, but it worked well.

ronaldtse:
@ni4 I believe @ribose-jeffreylau will revert tomorrow since it's holiday today, thanks!

ribose-jeffreylau:
@ni4 @ronaldtse Will be fixed in #1503.

ronaldtse:
Rebased on latest master.

ronaldtse:
The tests are still failing here; I wonder if this has something to do with the build cache?

ribose-jeffreylau:
It does look like a build cache issue.  I'm creating a fix here in #1504 .

ni4:
Thanks, @ribose-jeffreylau !

ni4:
@ribose-jeffreylau @ronaldtse Could we just create Docker containers with all installed pre-requisites, and use them (updating periodically) for CI instead of running complicated scripts each time/spending time on debugging?

ronaldtse:
@ni4 agree with the approach. 

@CAMOBAP would you happen to have time on this task?

ronaldtse (approved):

antonsviridenko:
src/tests/data/test_key_edge_cases/alice-3-uids-primary-boris.pgp

This file is added to the repo but not used anywhere in the tests code

ni4:
> src/tests/data/test_key_edge_cases/alice-3-uids-primary-boris.pgp
> This file is added to the repo but not used anywhere in the tests code

@antonsviridenko  Yeah, I had plans to use it in another PR to extend tests coverage. Do you think that having it in the repository may cause any issues (except additional data storage :) ) ?

antonsviridenko (approved):
ok then

PR #1505, opened 2021-05-24
Try to fix CI build cache issues

ribose-jeffreylau:
closes #1504

PR #1506, opened 2021-05-24
Refactor key protection to be more C++ and less redundant.

ni4:


ronaldtse:
Can we rerun the tests? #1505  is merged.

ni4:
@ronaldtse Sure, I'll rebase.

ronaldtse (approved):
Thanks @ni4 !

ronaldtse:
Ping @antonsviridenko for a quick check. Thanks!

src/lib/pgp-key.cpp
```diff
@@ -1618,56 +1609,47 @@ pgp_key_t::add_protection(pgp_key_store_format_t             format,
...
                    const std::string &                new_password)
 {
     if (!is_secret()) {
         RNP_LOG("Warning: this is not a secret key");
         return false;
     }
-    if (!decrypted_seckey.material.secret) {
+    bool ownpkt = &decrypted == &pkt_;
+    if (!decrypted.material.secret) {
         RNP_LOG("Decrypted seckey must be provided");
```
antonsviridenko, line 1626: maybe "secret key" instead of "seckey"?

antonsviridenko:
Do we have some tests that check if secret key is actually protected? Like: get a pointer to the buffer with secret key material, copy contents, try to protect the key, then try to check if key contents are still available at previously stored address? It maybe bit complicated with everything moved to C++-way of things

src/lib/pgp-key.cpp
ni4: @antonsviridenko Thanks, fixed.

ni4:
> Do we have some tests that check if secret key is actually protected? Like: get a pointer to the buffer with secret key material, copy contents, try to protect the key, then try to check if key contents are still available at previously stored address? It maybe bit complicated with everything moved to C++-way of things

We don't seem to have such tests, good idea, thanks. I'll update the PR.

ni4:
@antonsviridenko Is it good to go once CI finished?
cc @ronaldtse

src/lib/pgp-key.cpp
```diff
@@ -234,67 +234,59 @@ pgp_pk_alg_capabilities(pgp_pubkey_alg_t alg)
...
-        break;
-    case PGP_KEY_STORE_G10:
-        if (!g10_write_seckey(&memdst, seckey, password)) {
-            RNP_LOG("failed to write g10 seckey");
+    try {
+        switch (format) {
+        case PGP_KEY_STORE_GPG:
+        case PGP_KEY_STORE_KBX:
+            if (!write_sec_pgp(memdst, seckey, password)) {
+                RNP_LOG("failed to write seckey");
```
antonsviridenko, line 273: same issue, "secret key"
probably I missed it last time :)

src/lib/pgp-key.cpp
```diff
@@ -234,67 +234,59 @@ pgp_pk_alg_capabilities(pgp_pubkey_alg_t alg)
...
+                goto done;
+            }
+            break;
+        case PGP_KEY_STORE_G10:
+            if (!g10_write_seckey(&memdst, &seckey, password.c_str())) {
+                RNP_LOG("failed to write g10 seckey");
```
antonsviridenko, line 279: "seckey"

antonsviridenko (changes requested):

src/lib/pgp-key.cpp
ni4: Actually this message and previous one were not changed by this PR, but let's make things more understandable, will force-push shortly.

antonsviridenko (approved):

ronaldtse (approved):
Thanks @ni4 @antonsviridenko !

Issue #1457, opened 2021-03-19
FFI: add function to retrieve default subkey for the sign/encrypt operation.


ni4:
## Description

Most likely users will need some way to understand which subkey (or primary key itself) is picked for signing or encryption operation before executing it, so we should add function like `rnp_key_get_default_key(primary_key, operation_flags)`.

antonsviridenko:
So basically `find_suitable_key()` can be renamed to `rnp_key_get_default_key()` with some modifications?

ni4:
@antonsviridenko Not exactly, as new function should be a FFI one, i.e. set parameter to instance of rnp_key_handle_t (or NULL if there is no default key), and return RNP_SUCCESS/RNP_ERROR_*

antonsviridenko:
I've read related Thunderbird discussions, looks like more likely we need a way to configure subkey selection preferences. So users (API users and CLI users) can explicitly set how subkeys should be selected. Like most recent/most secure (based on key bits, not sure about sorting between different algorithms). Sorting order on algorithms can be also explicitly set by users. See https://www.openssl.org/docs/man1.1.1/man3/SSL_set_cipher_list.html as an example, OpenSSL users can set their preference order for algorithms using string parameter.

ni4:
> I've read related Thunderbird discussions, looks like more likely we need a way to configure subkey selection preferences. So users (API users and CLI users) can explicitly set how subkeys should be selected. Like most recent/most secure (based on key bits, not sure about sorting between different algorithms). 

Let's keep it simple right now, and use `find_suitable_key()` as it is at the moment. Just document it in the way "by default it selects latest valid subkey", and add some flags field (0 at the moment) so we may update the function later on.
If API user needs more complicated checks/behaviour, he may implement in the way he want by using other API calls (actually, that's how Thunderbird works at the moment).

antonsviridenko:
> so we should add function like `rnp_key_get_default_key(primary_key, operation_flags)`

I see integer flags parameter in one API function
https://github.com/rnpgp/rnp/blob/master/include/rnp/rnp.h#L1394
and in another case `usage flag` is a string parameter
https://github.com/rnpgp/rnp/blob/master/include/rnp/rnp.h#L761

which one should I prefer for this new function?

ni4:
@antonsviridenko hm, in this case it would be better to use string as user would need key for a single operation. Also please add uint32_t flags field, which is now zero, so we may update the behaviour in future.

[...]

Issue #1446, opened 2021-03-08
--ignore-time-conflict in rnp command line tool


thomas-roos:
--ignore-time-conflict known from gpg - which "GnuPG normally checks that the timestamps associated with keys and signatures have plausible values. However, sometimes a signature seems to be older than the key due to clock problems. This option makes these checks just a warning."
-> is there some option in rnp available?

ni4:
We do not have such option yet, sorry. But how often it happens?

ThomasRausU:
We want to replace gpg by rnp. We use it to check signature and encrypt update files for embedded devices. On those devices a correct RTC is not enforced. So our updates will fail. This is a mandatory option for us.

ronaldtse:
@ThomasRausU thanks for the suggestion, we should be able to make that work in the near-term roadmap. Thoughts @ni4 @dewyatt ?

ni4:
@ThomasRausU Thanks for the details, we should definitely support such usecase in our FFI and CLI. So, as I understand, with this option enabled rnp should ignore all the signature/key creation timestamps, and current timestamp inconsistency? 

Or, as key and its sigs are generated on the 'normal' device with correct RTC, we should ignore only the current timestamp when checking signatures?

ThomasRausU:
thx - the device creating the sigs and encrypting a file the RTC is correct. But on the embedded device which is decrypting and checking the  RTC is might not correct. So then there should be an option to ignore such failures.

nwalfield:
A better strategy would probably be to use the signature creation time as the reference time.  Then when you examine the certificate, you can still do the usual sanity checks and use the correct self signature (which it sounds like you might not be doing).

The app probably should probably also save the signature creation time in a DB and reject updates with an earlier signature creation time to avoid downgrade attacks.

ThomasRausU:
I dont understand correctly I think - what I wanna do is replace:
gpg --homedir /etc/gnupg/ --no-tty --pinentry-mode loopback --passphrase $SECRET --batch --yes --lock-never --no-auto-check-trustdb  --ignore-time-conflict --ignore-valid-from --decrypt /tmp/updatefile.idu | tar xz -C $TMPDIR
by
rnp --homedir=/etc/gnupg/ --password=$SECRET --decrypt /tmp/updatefile.idu | tar xz -C $TMPDIR

this is working correctly if the device RTC time is correct. If it's years ago it's not working.
[...]

PR #1520, opened 2021-06-04
Fix adding recipients. Fixes #1500.

### LINKED ISSUE #1500: RNP CLI: Show some error message if encryption failed due to expired recipient's key.
## Description

RNP CLI doesn't show any error message if encryption failed because of expired recipient's key.

## Steps to Reproduce

`rnp -r "expired_key" somefile.dat`

## Expected Behavior

Some message is shown, like `Encryption failed: recipient's key "expired_key" is expired.`

## Actual Behavior

No message at all (however, return code is correct and non-zero).

antonsviridenko:
`rnp_op_encrypt_add_recipient()` did not check return value from `find_suitable_key()` so in some cases NULL pointer could have been added to recipients lists, which caused failures on later stages.

Let's discuss if we need to print rejection reason for each key, primary and subkeys.

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -2317,6 +2317,9 @@ try {
                                        get_key_prefer_public(handle),
                                        &handle->ffi->key_provider,
                                        PGP_KF_ENCRYPT);
+    if (!key) {
+        key = get_key_prefer_public(handle);
+    }
     op->rnpctx.recipients.push_back(key);
     return RNP_SUCCESS;
 }
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -2461,6 +2461,14 @@ RNP_API rnp_result_t rnp_op_encrypt_create(rnp_op_encrypt_t *op,
                                            rnp_input_t       input,
                                            rnp_output_t      output);
 
+/**
+ * @brief Add recipient's public key to encrypting context.
+ *
+ * @param op opaque encrypting context. Must be allocated and initialized.
+ * @param key public key, used for encryption. Key is not checked for
+ *        validity or expiration.
+ * @return RNP_SUCCESS if operation succeeds or error code otherwise.
+ */
 RNP_API rnp_result_t rnp_op_encrypt_add_recipient(rnp_op_encrypt_t op, rnp_key_handle_t key);
 
 /**
--- a/src/rnp/fficli.cpp
+++ b/src/rnp/fficli.cpp
@@ -2287,6 +2287,7 @@ cli_rnp_encrypt_and_sign(const rnp_cfg &cfg,
     std::vector<rnp_key_handle_t> enckeys;
     std::vector<rnp_key_handle_t> signkeys;
     bool                          res = false;
+    rnp_result_t                  ret;
 
     rnp_op_encrypt_set_armor(op, cfg.get_bool(CFG_ARMOR));
 
@@ -2371,7 +2372,11 @@ cli_rnp_encrypt_and_sign(const rnp_cfg &cfg,
     }
 
     /* execute encrypt or encrypt-and-sign operation */
-    res = !rnp_op_encrypt_execute(op);
+    ret = rnp_op_encrypt_execute(op);
+    res = (ret == RNP_SUCCESS);
+    if (ret != RNP_SUCCESS) {
+        ERR_MSG("Operation failed: %s", rnp_result_to_string(ret));
+    }
 done:
     clear_key_handles(signkeys);
     clear_key_handles(enckeys);


antonsviridenko:
All centos checks fail with `Error: Unable to find a match: cmake3`

ronaldtse:
Ping @ribose-jeffreylau re: CentOS 8 failures on `cmake3`

ni4:
@antonsviridenko
I think we should not change behaviour in this way, as implementations which worked before by checking `rnp_op_encrypt_execute()` return code will fail now to get to that step. Also, there could be other reasons in the future why recipient's key is unusable for encryption, and we should not check all of these in `rnp_op_encrypt_add_recipient()` (functionality duplication). Another thing is issue #1446 - there are some cases when we should ignore times.

Instead I suggest the following changes:

- update documentation of `rnp_op_encrypt_add_recipient()`, telling that it doesn't do any checks on key expiration and that should be done manually.
- print error message in CLI, when encryption fails due to any reasons.
- once #1457 is implemented - add check in the CLI for default key/subkey expiration, telling that to the user and failing operation.

P.S. Btw, it looks like Github doesn't link issue to the PR if issue's number is on the title.

antonsviridenko:
@ni4 what about this part?

> rnp_op_encrypt_add_recipient() did not check return value from find_suitable_key() so in some cases NULL pointer could have been added to recipients lists, which caused failures on later stages.

https://github.com/rnpgp/rnp/blob/e43d5b1cfeba441e82e3c0153b9432cbb8131003/src/lib/rnp.cpp#L2323

Saving NULL pointer to recipients is clearly a bug to me, 
if we don't want to change behavior of `rnp_op_encrypt_add_recipient()`, then I think if `find_suitable_key()` returns NULL, `get_key_prefer_public(handle)` value should be saved into `recipients` std::list.

ni4:
> @ni4 what about this part?
> 
> > rnp_op_encrypt_add_recipient() did not check return value from find_suitable_key() so in some cases NULL pointer could have been added to recipients lists, which caused failures on later stages.
> 
> https://github.com/rnpgp/rnp/blob/e43d5b1cfeba441e82e3c0153b9432cbb8131003/src/lib/rnp.cpp#L2323
> 
> Saving NULL pointer to recipients is clearly a bug to me,
> if we don't want to change behavior of `rnp_op_encrypt_add_recipient()`, then I think if `find_suitable_key()` returns NULL, `get_key_prefer_public(handle)` value should be saved into `recipients` std::list.

Sorry, mislooked this. Yeah, let's store public key and let it be handled afterwards. Also this case (sign-only primary key + sign-only subkeys/invalid subkeys) should be covered with tests if it didn't yet.

src/lib/pgp-key.cpp
```diff
@@ -566,7 +566,7 @@ find_suitable_key(pgp_op_t            op,
     if (!key) {
         return NULL;
     }
-    if (key->flags() & desired_usage) {
+    if (key->valid() && key->flags() & desired_usage) {
```
ni4, line 569: Please add brackets here for better readability.

src/lib/rnp.cpp
```diff
@@ -2317,6 +2317,9 @@ try {
                                        get_key_prefer_public(handle),
                                        &handle->ffi->key_provider,
                                        PGP_KF_ENCRYPT);
+    if (!key) {
```
ni4, line 2320: Please add or update existing test which checks this case (i.e., for instance, attempt to add sign-only primary key with invalid or expired subkey for encryption), and check for failure of `rnp_op_encrypt_execute()` afterwards. That way we'll be sure that behaviour doesn't change in future/there are no memory leaks or other failures if incorrect key used for encryption attempt.

ni4 (changes requested):
Overall LGTM, just minor change and please add/update test. Thanks!

src/lib/rnp.cpp
antonsviridenko: @ni4 this case is already covered in https://github.com/rnpgp/rnp/blob/master/src/tests/ffi-enc.cpp#L852 - sign-only primary key and expired subkey

ni4: @antonsviridenko Thanks, then no new test is needed.
P.S. Don't be shocked by all red CI - it's because of key expiration used in tests, marking key as expired. I'll issue PR fixing that.

ronaldtse (approved):
Most tests should pass now on rebase?

antonsviridenko:
ruby-rnp tests fail...

ronaldtse:
@antonsviridenko the ruby-rnp failure, e.g.
https://github.com/rnpgp/rnp/pull/1520/checks?check_run_id=2922229337

Is identical to #1539 .

ni4:
@antonsviridenko Please rebase on the latest master. Issue is fixed within PR #1531

ni4 (approved):
LGTM, thanks! Feel free to rebase and merge once CI passes.

PR #1521, opened 2021-06-04
Fix installation of cmake for CentOS 8

ribose-jeffreylau:
It seems cmake3 has been renamed to cmake in the dnf repo.

GHA has been failing for CentOS 8 workflows.

See: https://github.com/rnpgp/rnp/runs/2743297015?check_suite_focus=true#step:6:1208

### DIFF
--- a/.github/workflows/centos8.yml
+++ b/.github/workflows/centos8.yml
@@ -39,8 +39,15 @@ jobs:
             GPG_VERSION: beta
           - BUILD_MODE: sanitize
             GPG_VERSION: stable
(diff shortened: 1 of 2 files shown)


ni4:
@ronaldtse @ribose-jeffreylau We have only single failure, where GnuPG beta compilation fails, so probably we may merge it.

ribose-jeffreylau:
@ni4 I've now disabled the failing test.  I've created #1522 for the issue.

ni4:
Thanks, @ribose-jeffreylau !

PR #1526, opened 2021-06-10
Fix incorrect key expiration calculation.

ni4:
This fixes previously uncovered case with non-zero key expiration time from direct-key signature with zero from primary uid certification.


ronaldtse (approved):
Thanks @ni4!

antonsviridenko (commented):
I see that changes in https://github.com/rnpgp/rnp/pull/1529 are identical to this one, probably one of PR-s was pushed by mistake twice.

antonsviridenko:
Or almost identical, except "case 15"...

ni4:
@antonsviridenko answered in #1529 - that one will be rebased on this one. Is this look good to you? Thanks!

antonsviridenko (approved):

PR #1534, opened 2021-06-18
Workaround MSYS2 CI issue, failing to checkout googletest via CMake

ni4:
Not sure whether this is correct solution, as we leave tests unbuildable on latest msys2/cmake combination without manual googletest sources download. However, this error seems to be on the CMake side and we'll just need to wait till they fix it.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -25,6 +25,8 @@ jobs:
     runs-on: windows-latest
     if: "!contains(github.event.head_commit.message, 'skip ci')"
     timeout-minutes: 70
+    env:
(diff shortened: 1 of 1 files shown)


ni4:
@ronaldtse What do you think about such solution? Or probably we should dig deeper into cmake internals to find exact source of the issue?

The root of the MSYS2 problem is that CMake generates external project code, where finally git is fed up with "HEAD^commit" parameter instead of "HEAD^{commit}".

antonsviridenko (approved):

ronaldtse (approved):

ronaldtse:
Thank you @ni4 !

PR #1536, opened 2021-06-23
Do not mark signature as invalid if key is expired

### LINKED ISSUE #1495: Do not mark signatures as invalid if key is expired.
## Description

If key is valid when data signature is created, then signature should be marked as valid even after key is expired or revoked.
This functionality should go to FFI and CLI, most likely via the helper function like `pgp_key_t->valid_at(timestamp)`.

ni4:
This PR fixes incorrect behaviour: if signature is made when key was valid (and now it is expired or revoked), signature was reported as invalid.

Fixes #1495


antonsviridenko (approved):

ronaldtse (approved):
Thanks @ni4 !

Issue #1542, opened 2021-06-27
Refactor CMake usage to best practices


ronaldtse:
The setup instructions of `googletest` suggests usage of the CMake `FetchContent` commands, but we're not doing that right now:
https://google.github.io/googletest/quickstart-cmake.html

We would want to take this chance to refactor the CMake configuration and build flow using current best practices.

ronaldtse:
Ping @thierryseegers

thierryseegers:
@ronaldtse Am I right in thinking that you'd ultimately like to get rid of as much workarounds/complexity as possible? I'm referring to these (there may be more):
- `GTEST_SOURCES` and whatever CMake code related to it.
- `DOWNLOAD_GTEST` and whatever CMake code related to it.
- The workaround for MSYS I see in `./github/workflows/windows.yml`.

ronaldtse:
@thierryseegers indeed, the less complexity we have, the easier to maintain. Thanks!

ni4:
JFYI: GTEST_SOURCES/DOWNLOAD_GTEST were added in support of offline builds, please see https://github.com/rnpgp/rnp/issues/1410 for discussion.

ronaldtse:
Thanks @ni4 for the reminder... let's see if we can simplify given this constraint of #1410.

thierryseegers:
@ronaldtse After a simple conversion of `ExternalProject` to `FetchContent`, a chain of annoyances ensued: 
1. When building the target of `FetchContent`, the current compilation options and flags will apply, in the case of an MSVC build, it picks up `/std:c++latest` from the root `CMakeLists.txt`. This fails compilation of `@c43f710` specified in the original `gtest-CMakeLists.txt.in`. It's a single compilation error in a header.
2. Curiously, on other platforms, `gmock` gets built when it appears it wasn't with the old way of doing things. I'm not positive I understand why. That also fails compilation with a single compilation error.

Issue no 2 is circumvented by forcing `googletest`'s `BUILD_GMOCK` option to `OFF`.

Issue no 1 is proving more complicated. Upgrading `googletest` to `release-1.11.0` (the latest) fixes the compilation problem for MSVC but brings in a new one on CentOS7. It appears that `std::is_trivially_copy_constructible` is not part of the libstdc++ found on CentOS7. From a cursory investigation, it looks like `/std:c++latest` may be needed only for one feature: designated initializers of structs (some examples found in `ecdh.cpp,50:57`).

[...]

PR #1547, opened 2021-06-29
Use FetchContent for googletest

thierryseegers:
This addresses issue #1542.

This replaces the use of an external CMake script that was invoked at configuration time to download googletest using `ExternalProject`. It is replaced with using `FetchContent`. The same URL and version is used. AFAICT, offline builds are still relevant so that hasn't been touched.

Some things of note:
1. The fetched content will be built with the "inherited" compilation options. That is, whatever options compilations are set globally (with `add_compile_options`, for example). 
1. I confirmed in passing that the flag `gtest_force_shared_crt` is still required. 
1. The target `gmock` target of that project would get built but fail to compile with MSVC. The version of googletest specified (@c43f710) has a line of code that is incompatible with `/std:c++latest`). Because of that, I explicitly disabled the building of the `gmock` target. It is not used anyway. On a tangential note, the fact that MSVC build has the option `/std:c++latest` set creates a discrepancy with the rest of the platforms which are built with `-std=c++11`.


ni4:
@thierryseegers rebasing on master would fix cmake test failure.

thierryseegers:
@ni4 Actually, if I'm not mistaken, the failure is just due to #1546 not being merged in yet because the failure is a check of similitude between the two `version.cmake` files and one has already been modified [here](https://github.com/rnpgp/cmake-versioning) but not yet on `master` which is what that PR will do.

ni4:
@thierryseegers Ah, sorry, you are right. Somehow got feeling that it is already merged.

ronaldtse:
@thierryseegers #1546 is now merged, could you help rebase? Thanks!

ronaldtse (approved):
Thanks for the clean up!

ronaldtse:
Will merge once passing since this is a pure CI change.

ronaldtse:
Merging since the #1546 merge to master is successful.

PR #1551, opened 2021-07-06
Respect timezone in key expiration time calculations, used in CLI.

ni4:
This PR fixes possible issues with key expiration calculations when timezone is different from UTC (and fixes some incorrect types usage as well).
I got those issues locally, and this was not covered by tests as it looks like all CI runners are in UTC.
Will file a separate issue to run some tests in non-UTC timezone.

P.S. Also it updates man page for rnpkeys, and fixes minor issue in rnp.1.adoc.


antonsviridenko (approved):

ronaldtse (approved):
LGTM!

PR #1555, opened 2021-07-09
Package using Cpack

thierryseegers:
This work (in progress) has a few things now that may be ready for discussion/adjustments/feedback.

The current additions are as follows:

#### CMakeLists.txt
The `ARCHIVE COMPONENT` of `librnp` is set to `development` otherwise the static library ends up in the "Unspecified" rpm file generated by CPack.

#### Ubuntu workflow
CPack's `DEB` generator is used on the Ubuntu platform to generate a Debian package. Of note:
1. The tool [`dpkg-shlibdeps`](https://man7.org/linux/man-pages/man1/dpkg-shlibdeps.1.html) needs to find the dependencies `libbotan-2.so.17` and `libjson-c.so.2`. Since these are installed in a non-standard directory (`$LOCAL_INSTALLS/[...]`), the paths must be explicitly passed to that tool. This is done via CPack's variable  [`CPACK_DEBIAN_PACKAGE_SHLIBDEPS_PRIVATE_DIRS`](https://cmake.org/cmake/help/latest/cpack_gen/deb.html#variable:CPACK_DEBIAN_PACKAGE_SHLIBDEPS_PRIVATE_DIRS) at the command line. Since CPack is invoked at the command line as a GitHub Action, these paths are exposed to `GITHUB_ENV` in `setup-env.inc.sh`.
That variable is available starting with CMake/CPack 3.20 which is the default on GitHub's `ubuntu-latest`.
1. A single package is created, at the end of the `tests` job. It will inherit whatever build properties are set in the context of that job.

#### CentOS7 workflow
CPack's `RPM` generator is used on the CentOS7 platform to generate two RPM packages. Of note:
1. I used the `pkgconfig-cmake-target` job to generate two sets of packages, one with a static lib and another with a dynamic lib. That job specifies a strategy matrix to do both.

#### CentOS8 workflow
The same modifications as for CentOS7's workflow were added but note that since its the job `pkgconfig-cmake-target` on which I originally grafted the packaging and that that job is currently disabled for CentOS8, then no package is produced.

#### Debian9 workflow
This resembles the packaging for Ubuntu since the same packager is used. Of note:
1. As mentioned above, in order to use CPack's `CPACK_DEBIAN_PACKAGE_SHLIBDEPS_PRIVATE_DIRS` variable, the version must be at least 3.20. CMake was already built from source for this workflow so I changed the versions specified in `install_functions.inc.sh`. I did not bump it within the root `CMakeLists.txt` since version 3.20 is only required for packaging, not for building but it maybe best to keep them in sync to avoid confusion.


ronaldtse:
Thanks for the great work @thierryseegers ! Could we expand this across all the other builds?

I wonder if we could incorporate botan/json-c in the CMake flow as well.

thierryseegers:
@ronaldtse I added the other Linux platforms. For macOS, I see that typically a `brew` package is provided but I don't think CPack has a generator for that format so I'm not sure what we should do for that platform. As for Windows (non-MSYS), there are some generators available. Would we expect the installer to include all necessary `dll`s? I'm not positive yet what are the options for MSYS.

ronaldtse:
@thierryseegers thanks for this.

macOS: we have a separate formula that compiles/installs RNP at https://github.com/rnpgp/homebrew-rnp so we don't need to do it in CPack.

The only remaining issue is that instead of "upload-artifact" for builds, we want to only upload on a new tag (using this action: https://github.com/softprops/action-gh-release). That said, the CPack jobs should always run.

thierryseegers:
@ronaldtse Open for review. The failed checks should be completely unrelated.

ronaldtse:
Thank you @thierryseegers !

ronaldtse (approved):

Issue #1556, opened 2021-07-10
rnp_key_get_subkey_at() implementation contains double conversion between hexadecimal and binary presentation of fingerprint


antonsviridenko:
## Description

https://github.com/rnpgp/rnp/blob/master/src/lib/rnp.cpp#L6169

```
    const pgp_fingerprint_t &fp = key->get_subkey_fp(idx);
    char                     fphex[PGP_FINGERPRINT_SIZE * 2 + 1] = {0};
    if (!rnp::hex_encode(fp.fingerprint, fp.length, fphex, sizeof(fphex))) {
        return RNP_ERROR_BAD_STATE;
    }
    return rnp_locate_key(handle->ffi, "fingerprint", fphex, subkey);

```
`rnp_key_get_subkey_at()` uses FFI function `rnp_locate_key()` to return subkey handle, which accepts textual hexadecimal representation of fingerprint and converts it back to binary presentation inside. `rnp_key_store_search()` can be used instead, which accepts fingerprint without excess conversions. While it is O(1) in complexity, it's still inefficient to do so and can slow down programs that call `rnp_key_get_subkey_at()` frequently.

## Expected Behavior

No unnecessary conversions should be performed.

PR #1557, opened 2021-07-11
Add new FFI function rnp_key_get_default_key().

### LINKED ISSUE #1457: FFI: add function to retrieve default subkey for the sign/encrypt operation.

antonsviridenko:
Closes #1457

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -97,6 +97,11 @@ typedef uint32_t rnp_result_t;
 #define RNP_OUTPUT_FILE_RANDOM (1U << 1)
 
 /**
+ * Flags for default key selection.
+ */
+#define RNP_KEY_SUBKEYS_ONLY (1U << 0)
+
+/**
  * User id type
  */
 #define RNP_USER_ID (1U)
@@ -1347,6 +1352,28 @@ RNP_API rnp_result_t rnp_key_get_subkey_count(rnp_key_handle_t key, size_t *coun
 RNP_API rnp_result_t rnp_key_get_subkey_at(rnp_key_handle_t  key,
                                            size_t            idx,
                                            rnp_key_handle_t *subkey);
+
+/** Get default key for specified usage. Accepts primary key
+ *  and returns one of its subkeys suitable for desired usage.
+ *  May return the same primary key if it is suitable for requested
+ *  usage and flag RNP_KEY_SUBKEYS_ONLY is not set.
+ *
+ *  @param primary_key handle of the primary key.
+ *  @param usage desired key usage i.e. "sign", "certify", etc,
+ *               see rnp_op_generate_add_usage() function description for all possible values.
+ *  @param flags possible values:  RNP_KEY_SUBKEYS_ONLY - select only subkeys,
+ *               otherwise if flags is 0, primary key can be returned if
+ *               it is suitable for specified usage.
+ *  @param default_key on success resulting key handle will be stored here, otherwise it
+ *                     will contain NULL value. You must free this handle after use with
+ *                     rnp_key_handle_destroy().
+ *  @return RNP_SUCCESS on success, RNP_ERROR_KEY_NOT_FOUND if no key with desired usage
+ *          was found or any other error code.
+ */
+RNP_API rnp_result_t rnp_key_get_default_key(rnp_key_handle_t  primary_key,
+                                             const char *      usage,
+                                             uint32_t          flags,
+                                             rnp_key_handle_t *default_key);
 
 /** Get the key's algorithm.
  *
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -6176,6 +6176,68 @@ try {
 FFI_GUARD
 
 rnp_result_t
+rnp_key_get_default_key(rnp_key_handle_t  primary_key,
+                        const char *      usage,
+                        uint32_t          flags,
+                        rnp_key_handle_t *default_key)
+try {
+    if (!primary_key || !usage || !default_key) {
+        return RNP_ERROR_NULL_POINTER;
+    }
+    uint8_t keyflag = 0;
+    bool    no_primary = false;
+    if (!str_to_key_flag(usage, &keyflag)) {
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+    if (flags & RNP_KEY_SUBKEYS_ONLY) {
+        no_primary = true;
+        flags &= ~RNP_KEY_SUBKEYS_ONLY;
+    }
+    if (flags) {
+        FFI_LOG(primary_key->ffi, "Invalid flags: %" PRIu32, flags);
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+    pgp_key_t *key = get_key_prefer_public(primary_key);
+    if (!key) {
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+    pgp_key_t *defkey = find_suitable_key(
+      PGP_OP_UNKNOWN, key, &primary_key->ffi->key_provider, keyflag, no_primary);
+    if (!defkey) {
+        *default_key = NULL;
+        return RNP_ERROR_NO_SUITABLE_KEY;
+    }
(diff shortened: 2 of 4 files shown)


include/rnp/rnp.h
```diff
@@ -1348,6 +1353,26 @@ RNP_API rnp_result_t rnp_key_get_subkey_at(rnp_key_handle_t  key,
                                            size_t            idx,
                                            rnp_key_handle_t *subkey);
 
+/** Get default key for specified usage. Accepts primary key
+ *  and returns one of its subkeys suitable for desired usage.
+ *  May return the same primary key if it is suitable for requested
+ *  usage and flag RNP_KEY_SUBKEYS_ONLY is not set.
+ *  available.
```
ni4, line 1360: I guess `available` came here by mistake?

include/rnp/rnp.h
```diff
@@ -1348,6 +1353,26 @@ RNP_API rnp_result_t rnp_key_get_subkey_at(rnp_key_handle_t  key,
...
+ *
+ *  @param primary_key handle of the primary key.
+ *  @param usage see rnp_op_generate_add_usage() function description.
```
ni4, line 1363: I would add here something like `Desired key usage, i.e. 'encrypt', 'sign', and so on. See rnp_op_generate_add_usage() for all possible values` so user may understand possible values from a single lookup.

include/rnp/rnp.h
```diff
@@ -1348,6 +1353,26 @@ RNP_API rnp_result_t rnp_key_get_subkey_at(rnp_key_handle_t  key,
...
+ *  @param flags possible values:  RNP_KEY_SUBKEYS_ONLY - select only subkeys,
+ *               otherwise if flags is 0, primary key can be returned if
+ *               it is suitabled for specified usage.
```
ni4, line 1366: I guess `suitabled`  == `suitable`?

include/rnp/rnp.h
```diff
@@ -1348,6 +1353,26 @@ RNP_API rnp_result_t rnp_key_get_subkey_at(rnp_key_handle_t  key,
...
+ *  @param default_key on success resulting key handle will be stored here, otherwise it
+ *         will contain NULL value. You must free this handle after use with
+ *         rnp_key_handle_destroy().
```
ni4, line 1369: Please add @return section with described RNP_SUCCESS and RNP_ERROR_KEY_NOT_FOUND.

src/lib/rnp.cpp
```diff
@@ -6173,6 +6173,64 @@ try {
...
+    uint8_t keyflag = 0;
+    bool    no_primary = false;
+    if (!str_to_key_flag(usage, &keyflag)) {
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+    if (flags & RNP_KEY_SUBKEYS_ONLY) {
+        no_primary = true;
+        flags &= ~RNP_KEY_SUBKEYS_ONLY;
+    }
+    if (flags) {
```
ni4, line 6194: I wondering whether should we add RNP_KEY_SECRET_ONLY, or (better?) search for secret key automatically with `sign`/`certify` usage?

src/lib/rnp.cpp
```diff
@@ -6173,6 +6173,64 @@ try {
...
+    if (!key) {
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+    pgp_key_t *defkey = find_suitable_key(
+      PGP_OP_UNKNOWN, key, &primary_key->ffi->key_provider, keyflag, no_primary);
+    if (!defkey) {
+        *default_key = NULL;
+        return RNP_ERROR_NO_SUITABLE_KEY;
+    }
+    pgp_key_search_t search = {(pgp_key_search_type_t) 0};
```
ni4, line 6208: Why not to reuse `rnp_locate_key()` here?

ni4 (changes requested):
Please see the review comments, thanks!

src/lib/rnp.cpp
antonsviridenko: Because it requires excess conversions binary->hex->binary, see https://github.com/rnpgp/rnp/issues/1556

src/lib/rnp.cpp
antonsviridenko: > or (better?) search for secret key automatically with sign/certify usage?

I think second option is better

antonsviridenko: @ni4 what condition should be used in this case, both public & secret must be available or only secret is enough?

src/lib/rnp.cpp
ni4: Posted a PR #1560 which should fix that, and solution may be reused in this PR.

src/lib/rnp.cpp
ni4: Only secret is enough, public key may be easily re-generated from secret (and actually must correspond to it, except secret fields availability).

antonsviridenko: Looks like find_suitable_key() should be modified then to include this option, there can be situation where primary key & some suitable subkeys are presented only by public part, but some other suitable subkey can have private part and not be selected as default according to current implementation.

ni4: @antonsviridenko Agree, some parameter like `is_secret` could be added to `find_suitable_key()`.

antonsviridenko: ok, I've implemented that parameter and now this new function returns expired subkey from test keyring # 1:
```
ssb   1024/DSA 1d7e8a5393c997a8 2017-07-20 [S] [EXPIRED 2017-11-20]
      c5b15209940a7816a7af3fb51d7e8a5393c997a8
```
Probably expiration is not checked somewhere for secret keys.

ni4: @antonsviridenko Yeah, initially secret keys were considered as valid as owner may always re-generate self-signatures. Right now I think that this should be changed, as may give some difficulties and problems. It's in a my pipeline.
However, adding FFI function should not change old behaviour as may bring some problems to already existing users. So probably for now let's leave the case with non-existing secret primary/subkeys (actually, right now we do not support public primary + secret subkey, we support only secret primary with special s2k, where secret key data is cut out).

antonsviridenko: @ni4 searching for secret keys breaks the test (which passed before), see my latest commit. So either we should revert this idea about secret keys or check for secret part presence on the key returned by original find_suitable_key() implemenation.

antonsviridenko: i.e. we still want to find non-expired keys, but their secret part should be present for specific usages (other than encrypt in this case, all others require secret key)

ni4: @antonsviridenko Within current implementation secret key is required twice - on the `find_suitable_key()` level and in the `rnp_key_get_default_key()`. You may rollback changes to the `find_suitable_key()`, so it will behave as previously, and require secret only on the `rnp_key_get_default_key()` level. Basically, we should first find key/subkey which is good enough by flags/validity, and then find it's secret part or return an error.

src/lib/rnp.cpp
antonsviridenko: Still not using rnp_locate_key_int(), as it allocates memory first, and presence of private key can be checked later on return, so in case private key is required by specified key usage but it is not present, memory must be free-d.

ni4:
@antonsviridenko Could you please rebase this on master? Now msvc and sanitize CI issues should be fixed.

ronaldtse (approved):
Thanks @antonsviridenko !

src/lib/rnp.cpp
ni4: @antonsviridenko This could be easily addressed by adding parameter like `require_secret` to the `rnp_locate_key_int()`. Anyway, let's merge this and address this later. Thanks!

ni4 (approved):
LGTM, thanks!

PR #1560, opened 2021-07-12
Refactor FFI internals to not do excessive hex encoding/decoding.

### LINKED ISSUE #1556: rnp_key_get_subkey_at() implementation contains double conversion between hexadecimal and binary presentation of fingerprint

ni4:
Fixes #1556

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -3545,24 +3545,9 @@ locator_to_str(const pgp_key_search_t *locator,
     return true;
 }
 
-rnp_result_t
-rnp_locate_key(rnp_ffi_t         ffi,
-               const char *      identifier_type,
-               const char *      identifier,
-               rnp_key_handle_t *handle)
-try {
-    // checks
-    if (!ffi || !identifier_type || !identifier || !handle) {
-        return RNP_ERROR_NULL_POINTER;
-    }
-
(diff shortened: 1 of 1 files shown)


antonsviridenko (approved):

ronaldtse (approved):
Thanks @ni4!

PR #1569, opened 2021-07-27
CLI: return non-zero exit code on unrecognized command line options.

### LINKED ISSUE #1548: CLI: rnp returns zero result for the unknown command.
## Description

Command like `rnp --something-wrong` returns 0, while should return non-zero result.
`rnpkeys` behaves correctly in this case.

antonsviridenko:
Fixes #1548


ronaldtse (approved):
Thank you @antonsviridenko !

ni4 (changes requested):
LGTM, but it would be good to have some simple test which makes sure that things work this way and there would be no regression in the future.

ni4 (approved):
LGTM, thanks @antonsviridenko !

PR #1580, opened 2021-08-08
docs: add installation instructions for Gentoo.

antonsviridenko:


ronaldtse (approved):

ronaldtse:
Merging with one approval since it is only docs.

PR #1588, opened 2021-08-13
ci: Make Cirrus CI skip CI also when 'skip ci' occurs elsewhere in the body of the commit message [skip ci]

ribose-jeffreylau:
The doc at https://cirrus-ci.org/guide/writing-tasks/#conditional-task-execution
says that it would skip CI if 'skip ci' is on the first line of the
commit message, but this behaviour would not match configs found in
GitHub workflows.  GitHub workflows would skip if 'skip ci' is found in any part of the commit message, not just the first line / last line / whatever.

This commit tries to match their behaviours.

It would also skip Cirrus CI if the only changes are in .github
workflows.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -4,6 +4,7 @@ freebsd_instance:
 task:
   name: build
   only_if: $CIRRUS_BRANCH == 'master' || $CIRRUS_BRANCH =~ 'release/.*' || $CIRRUS_PR != ''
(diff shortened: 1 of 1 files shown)


ni4:
@ribose-jeffreylau It would be also good to not trigger Cirrus if changes are made only to the `docs/**`

ribose-jeffreylau:
> @ribose-jeffreylau It would be also good to not trigger Cirrus if changes are made only to the `docs/**`

I agree.  Updated!

ni4 (approved):
LGTM, thanks, @ribose-jeffreylau !

antonsviridenko (approved):

PR #1591, opened 2021-08-16
Make tests pass after 2024

bmwiedemann:
Make tests pass after 2024

Tested on i586 and x86_64

Background:
As part of my work on reproducible builds for openSUSE, I check that software still gives identical build results in the future.
The usual offset is +15 years, because that is how long I expect some software will be used in some places.
This showed up 1 failing test in our `rnp` package build.
See https://reproducible-builds.org/ for why this matters.

This PR was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).

### DIFF
--- a/src/tests/utils-rnpcfg.cpp
+++ b/src/tests/utils-rnpcfg.cpp
@@ -128,7 +128,7 @@ TEST_F(rnp_tests, test_rnpcfg_get_expiration)
     assert_int_equal(get_expiration(exp.c_str(), &raw_expiry), 0);
     assert_int_equal(raw_expiry, rawtime - basetime);
     assert_int_equal(get_expiration("2100-01-01", &raw_expiry), 0);
-    assert_int_equal(get_expiration("2024-02-29", &raw_expiry), 0);
+    assert_int_equal(get_expiration("2124-02-29", &raw_expiry), 0);
     /* date in a past */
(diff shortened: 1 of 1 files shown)


antonsviridenko (approved):

antonsviridenko:
20 tests have failed...

bmwiedemann:
Interesting. These tests passed for me on openSUSE Tumbleweed.
```
[ RUN      ] rnp_tests.test_rnpcfg_get_expiration
/tmp/cirrus-ci-build/src/tests/utils-rnpcfg.cpp:130: Failure
Expected equality of these values:
  (get_expiration("3100-01-01", &raw_expiry))
    Which is: -3
  (0)
    Which is: 0
```

Maybe the test is good, but some of the implementation is bad?

ni4:
> Interesting. These tests passed for me on openSUSE Tumbleweed.
> ...
> Maybe the test is good, but some of the implementation is bad?

Is it possible that you checked with the version 0.15.1? This commit probably changed the behaviour: https://github.com/rnpgp/rnp/commit/2ae6479c9ed9e18b1e3b2573e58ca033de5635f7

Expiration time, which is calculated there, cannot exceed 32bit unsigned (as it is stored as 32 bit uint in OpenPGP packets).
So maximum value should be not more than ~125 years from now.

ni4 (approved):
LGTM, thanks @bmwiedemann !

PR #1550, opened 2021-07-05
Add OpenSSL backend.

ni4:
This PR is a preview version (not ready for production environment yet)  of the OpenSSL backend, which may be used instead of Botan. Requires at least OpenSSL v1.1.1.

Currently it runs just rnp_tests for centos 8. cli_tests support is planned for later as we cannot distinguish which backend is used by RNP right now (and which algorithms/curves are enabled).

PR #1593, opened 2021-08-16
Fix cmake warning with sm2, fixes 7923ede

andreasstieger:
Policy CMP0115 is not set: Source file extensions must be explicit.

@ni4 fixes #1550


ronaldtse (approved):
Thank you @andreasstieger !

ni4 (approved):
LGTM, thanks @andreasstieger !

PR #1393, opened 2021-01-22
Fix a bunch of GCC warnings.

ni4:
It fixes the following warnings:
- `warning: implicitly-declared 'pgp_key_t& pgp_key_t::operator=(const pgp_key_t&)' is deprecated [-Wdeprecated-copy]`
- `warning: this statement may fall through [-Wimplicit-fallthrough=]`
- `warning: 'void* memcpy(void*, const void*, size_t)' writing to an object of type 'pgp_transferable_key_t' {aka 'struct pgp_transferable_key_t'} with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Wclass-memaccess]`

PR #1602, opened 2021-08-17
Fix one implicit fallthrough warning, and consistent attribute.

andreasstieger:
Fix one implicit fallthrough warning, and consistent attribute. `[[fallthrough]]` is supported in clang, gcc since their C++11
implementations which we require. It is standard since C++17.

Fixes c177ea6c7568d5aa1a191cdee953b40b1d0178e5 from #1393


ni4 (approved):
LGTM, thanks @andreasstieger !

ronaldtse (approved):
Thank you @andreasstieger !

PR #1616, opened 2021-08-23
CLI: allow to use dash as stdout output specifier

### LINKED ISSUE #1314: rnp --encrypt ignores --output
## Description

rnp --encrypt ignores --output, even though rnp.1 suggests that it may be used with --encrypt.

## Steps to Reproduce

1. Generate or import a key
2. Create a file to encrypt: `echo huhu > huhu`
3. Encrypt the file: `rnp --homedir=/tmp/... --armor --encrypt --recipient ... --output=- huhu`

## Expected Behavior

Encrypted message is written to stdout.

## Actual Behavior

Encrypted message is written to `huhu.asc`:

```
% sq inspect huhu.asc
huhu.asc: Encrypted OpenPGP Message.

      Recipient: 7C2F AA4D F93C 37B2
```

ni4:
This PR updates CLI with the following changes:
- allows to use `-` as stdout output marker
- adds `--notty` option to read/write data from stdin/stdout only (useful in batch processing and tests)
- adds option `--overwrite` and changes behaviour of key/revocation export to use it instead of `--force` in case file already exists.
- partially refactors cli_tests to use `self.assertEqual()` and so on.

Fixes #1314

### DIFF
--- a/src/rnp/fficli.cpp
+++ b/src/rnp/fficli.cpp
@@ -1766,27 +1766,21 @@ cli_rnp_export_keys(cli_rnp_t *rnp, const char *filter)
     std::vector<rnp_key_handle_t> keys;
 
     if (!cli_rnp_keys_matching_string(rnp, keys, filter, flags)) {
-        fprintf(rnp->userio_out, "Key(s) matching '%s' not found.\n", filter);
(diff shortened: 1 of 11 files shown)


ronaldtse (approved):
Thanks @ni4 !

antonsviridenko (approved):

PR #1630, opened 2021-09-13
Fix minor coverity issues introduced with PR 1629

ni4:
Those doesn't seem to be critical, but still let's have a shorter list on coverity.
- related PR #1629


ronaldtse (approved):
Thank you @ni4 !

antonsviridenko (approved):

PR #1634, opened 2021-09-29
Limit number of message recipients and/or signatures.

ni4:
It doesn't seem that any implementation would use more then 16K recipients/signatures in a single stream.
Fixes some fuzzing out-of-memory and timeout issues.


antonsviridenko (approved):

ronaldtse (approved):
Thanks @ni4 !

PR #1635, opened 2021-09-29
Workaround sudo 1.9.8 issue on Cirrus FreeBSD 12.2 runner.

ni4:
sudo 1.9.8 has an issue, which isn't yet fixed in FreeBSD 12.2 packages collection, so this PR works around by using preinstalled 1.9.3p1.

With 1.9.8 CI gets the following:
```
sudo: (null): option "passprompt_override" does not take a value
sudo: error initializing audit plugin sudoers_audit
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -16,7 +16,10 @@ task:
   user_setup_script: |
     # workaround 'pkg: repository meta has wrong version 2'
     env ASSUME_ALWAYS_YES=yes pkg bootstrap -f
-    pkg install -y sudo
+    # FreeBSD 12.2 has sudo 1.9.3.p, while current 1.9.8 has bug which prevents CI to run:
+    # sudo: (null): option "passprompt_override" does not take a value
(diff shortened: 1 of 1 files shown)


ronaldtse:
Failing at a random place for centos8-openssl, re-running:
```
	 47 - rnp_tests.test_ffi_key_generate_eddsa (SEGFAULT)
1436
```

ni4:
> Failing at a random place for centos8-openssl, re-running:
> 
> ```
> 	 47 - rnp_tests.test_ffi_key_generate_eddsa (SEGFAULT)
> 1436
> ```

Thanks, now it works. Actually we were lucky to catch this - OpenSSL allows to generate 31-byte EdDSA secret key, which causes failure. Investigating which behaviour should be correct in this case.

ronaldtse (approved):
Happy to merge this first and then deal with #1636

ronaldtse:
CI change only requires one approval, merging.

PR #1637, opened 2021-09-30
Correctly handle short EdDSA secret key within OpenSSL backend.

### LINKED ISSUE #1636: OpenSSL allows to generate 31-byte EdDSA secret key causing test failure
> Failing at a random place for centos8-openssl, re-running:
> 
> ```
> 	 47 - rnp_tests.test_ffi_key_generate_eddsa (SEGFAULT)
> 1436
> ```

Actually we were lucky to catch this - OpenSSL allows to generate 31-byte EdDSA secret key, which causes failure. Investigating which behaviour should be correct in this case.

_Originally posted by @ni4 in https://github.com/rnpgp/rnp/issues/1635#issuecomment-931053565_

ni4:
Fixes #1636


antonsviridenko (approved):

ronaldtse (approved):
Thanks @ni4!

Issue #1595, opened 2021-08-17
Check for the OpenSSL available features when configuring


ni4:
## Description

Currently OpenSSL backend automatically disables all non-default features, however we should check for available ones for better flexibility.

ni4:
Required CMake module was added via PR #1692 , so now we should only check whether automatically disabled features are available.

ni4:
@andrey-utkin Also within fixing of this issue we could drop check from CMake which allows to build with OpenSSL in the debug mode.

ronaldtse:
Agree with @ni4. Thanks @andrey-utkin !

andrey-utkin:
A spec by @ni4 
> WIthin automatic feature detection we should by default use 'Auto' mode, when only supported by backend features are enabled. However, allow user to enable or disable desired features via the corresponding CMake options, throwing an error if user wants to enable unsupported by backend feature.

https://github.com/rnpgp/rnp/pull/1875#issuecomment-1181885579

Which I understand as follows:
No configuration flags:
Backend is Botan, check that core depended features are present, all flag-features available in installed Botan are enabled.

backend=botan: same as above

backend=openssl: check that core depended features are present, all flag-features available in installed OpenSSL are enabled.

Above + specific ENABLE_* features set: specific features override autodetected values.

ni4:
> A spec by @ni4
> 
> > WIthin automatic feature detection we should by default use 'Auto' mode, when only supported by backend features are enabled. However, allow user to enable or disable desired features via the corresponding CMake options, throwing an error if user wants to enable unsupported by backend feature.
> 
> [#1875 (comment)](https://github.com/rnpgp/rnp/pull/1875#issuecomment-1181885579)
> 
> Which I understand as follows: No configuration flags: Backend is Botan, check that core depended features are present, all flag-features available in installed Botan are enabled.
> 
> backend=botan: same as above
> 
> backend=openssl: check that core depended features are present, all flag-features available in installed OpenSSL are enabled.
> 
> Above + specific ENABLE_* features set: specific features override autodetected values.

Yeah, correct.

P.S. Do you really have a need to address issue in the commit title? It makes issue's comment log quite unreadable.

andrey-utkin:
I feel I do! It's a widespread practice and I'm glad for it, as software archeology is a regular part of the job :)
I don't know how to leave a link to the issue ticket in a commit message so that github doesn't pick it up but that it's not too obscure/tedious for people to pick up.
[...]

PR #1645, opened 2021-10-11
Update cli_tests to work with OpenSSL backend.

### LINKED ISSUE #1596: Enable cli_tests runner for OpenSSL backend.
## Description

Currently only rnp_tests are run since we cannot distinguish between openssl and botan backend in runtime.

ni4:
Currently OpenSSL backend just defaults to disabled AEAD, Twofish and Brainpool. This behaviour will be updated within #1595 .
As a side option this PR also speedups cli_tests for all other platforms by reducing S2K iterations for the test keys, used in encryption.

Fixes #1596


ronaldtse:
Ping @dewyatt @antonsviridenko for a review

PR #1646, opened 2021-10-12
Feature/allow skip tests and build static only

CAMOBAP:
- rnpgp/ruby-rnp#73

This PR introduces two features:
 - allow building only static variants of `botan` and `json-c` dependencies which will need to build embedded `rnp` library for `ruby-rnp` repository
 - allow skip tests execution 
 - update branch for `git clone ruby-rnp` (because default branch was changed from `master` to `main`)

Open questions:
 - why do we have `continue-on-error: true` for some workflows? should we remove it?
 - using the package manager for c++ it's a great idea, Are there any plans to use `vcpkg` for all platforms?

### DIFF
--- a/ci/lib/cacheable_install_functions.inc.sh
+++ b/ci/lib/cacheable_install_functions.inc.sh
@@ -84,8 +84,8 @@ install_jsonc() {
     local cpuparam=()
     [[ -z "$CPU" ]] || cpuparam=(--build="$CPU")
     local build_type_args=(
-        "--enable-$([ $USE_STATIC_DEPENDENCIES ] && echo 'static' || echo 'shared')"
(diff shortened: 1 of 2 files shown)


ni4:
> > why do we have continue-on-error: true for some workflows? should we remove it?
> 
> Probably used previously when those tests failed?

As far as I remember, if it was not set to true, it was harder to debug failing workflows, as first failure for matrix stopped all other tasks.

ni4:
@CAMOBAP JFYI: in RNP repository we stick to 'Rebase and merge' instead of merge, as this produces cleaner history/it is later easier to pick commits for the release branch.

CAMOBAP:
@ni4 thanks for note, I will follow  'Rebase and merge' approach for future PRs

ronaldtse:
Let's rebase this ASAP.

PR #1647, opened 2021-10-13
Feature/link shellcheck

CAMOBAP:


ronaldtse:
@CAMOBAP Spaces around = are quite dangerous in shell, I’d prefer keeping SC2191.

CAMOBAP:
@ribose-jeffreylau @ni4 should I "squash" several commits in PR to a single one before merge? Or just rebase will be enough?

ronaldtse:
> should I "squash" several commits in PR to a single one before merge? Or just rebase will be enough?

Would be better if we squash together commits that are meant to be together. For example if you have commits that correct/fix previous commits, they should be squashed so that the commit is "wholesome" itself.

ronaldtse:
@CAMOBAP if possible let's not have exceptions for shellcheck, it's just hard to manage them in the future. Thanks!

CAMOBAP:
Thanks, squashed+rebased, waiting for CI to pass

ronaldtse:
Thank you @CAMOBAP @ribose-jeffreylau @ni4 !

Issue #1651, opened 2021-10-14
LGTM alerts: Python test suite alerts


ronaldtse:
## Description

32 alerts in total. Too long to paste... See for yourself...

https://lgtm.com/projects/g/rnpgp/rnp/alerts/?mode=list&lang=python

Issue #1650, opened 2021-10-14
LGTM alerts: Declaration hides parameter C++


ronaldtse:
## Description
<img width="1231" alt="Screenshot 2021-10-14 at 10 29 36 PM" src="https://user-images.githubusercontent.com/11865/137338188-79bbc9e3-c62a-43f0-a2ee-52aa960acb1d.png">

<img width="1231" alt="Screenshot 2021-10-14 at 10 29 46 PM" src="https://user-images.githubusercontent.com/11865/137338216-7cec3ad0-0abe-4529-9419-a27481539115.png">

<img width="1228" alt="Screenshot 2021-10-14 at 10 29 53 PM" src="https://user-images.githubusercontent.com/11865/137338239-d62c7e22-b121-4706-bd48-55a4ab1a9214.png">

ni4:
Fixed via #1652, so closing.

PR #1652, opened 2021-10-14
Fix bunch of LGTM alerts for Python/C++ code.

ni4:
This would not fix all, but most of.
See
- #1650 
- #1651


ronaldtse:
Ping @antonsviridenko @ribose-jeffreylau @dewyatt

Issue #1654, opened 2021-10-18
DOWNLOAD_RUBYRNP [: -v: unary operator expected


CAMOBAP:
## Description

code `[ -v VARIABLE]` produce warning on macOS and it doesn't work

```
+ '[' -v GTEST_SOURCES ']'
ci/main.sh: line 89: [: -v: unary operator expected
+ '[' -v DOWNLOAD_GTEST ']'
ci/main.sh: line 90: [: -v: unary operator expected
+ '[' -v DOWNLOAD_RUBYRNP ']'
ci/main.sh: line 91: [: -v: unary operator expected
```

As result it's impossible to build `librnp` without `ruby-rnp` on macOS

## Steps to Reproduce

1. run build on macos

## Expected Behavior

With passed `env DOWNLOAD_RUBYRNP=OFF` it should not require download ruby-rnp` as it works on Linux

## Actual Behavior

It downloads `ruby-rnp` and it unnecessary during build rnp in scope of https://github.com/rnpgp/ruby-rnp repo

PR #1655, opened 2021-10-18
#1654 replace '[ -v' with '[ -n'

CAMOBAP:
- #1654

According to https://stackoverflow.com/a/39296723/902217 

> Note that the -v option was introduced in Bash 4.2

This may look minor but it's important for https://github.com/rnpgp/ruby-rnp to not repeat itself


ronaldtse:
Thanks @CAMOBAP !

Issue #1129, opened 2020-05-09
ci: Build with clang-cl


dewyatt:
## Description

I couldn't find any open issue for this and since the Thunderbird guys are building with clang-cl, we should probably add this to CI.

(I'm planning to tackle this myself, but if anyone is already working on this let me know!)

ni4:
@dewyatt #997 could be related (maybe not well formulated though). We definitely need an ability to make a native Windows build without MinGW/MSYS.

ni4:
@dewyatt What do you think about including this into v0.14 milestone? Since this is not urgent for TB since they have their own build scripts, but still would be nice to have in the release.

dewyatt:
> @dewyatt What do you think about including this into v0.14 milestone? Since this is not urgent for TB since they have their own build scripts, but still would be nice to have in the release.

This issue is mostly focused on CI so not related to a release per se. I don't see it as high priority but I don't see it as low priority either. If we're constantly breaking the build for the TB people because we don't have any similar build in CI, that's going to create some unnecessary friction.

ronaldtse:
I think it would be best to have the clang-cl build as soon as we can, because that would prevent us from being the source of errors to TB...

ronaldtse:
@dewyatt are you handling this? Or should we seek extra help? Thanks!

dewyatt:
> @dewyatt are you handling this? Or should we seek extra help? Thanks!

I looked at it briefly and observed that adding `-T ClangCL` leads to duplicate symbol errors. I came up with a quick workaround but it breaks all other CI.

I'd rather continue on py-rnp if someone else wants to do a proper fix here...

dewyatt:
(Unassigning but I can come back if we have trouble finding someone)

ronaldtse:
Thanks for updating this, @antonsviridenko @rrrooommmaaa would you have time to make this work?

ronaldtse:
@maxirmx would you have time to handle this? Thanks.

maxirmx:
@ronaldtse 
Yes, I can handle it.  No sure though what operating systems to support. Ubuntu ?

ronaldtse:
@maxirmx We will want to handle the platforms that Thunderbird builds RNP on. Do we have a list of those platforms somewhere?

maxirmx:
Clang issues are only indirectly related to rnp.
The first one:

**Target:**  CentOS 8
**Tests**   (normal, 2.3.1, clang, clang++)
[...]

Issue #1648, opened 2021-10-13
use vcpkg for all platforms


CAMOBAP:
## Description

It will be nice to use `vcpkg` for all our platforms to reduce the amount of shell script

## Expected Behavior

This will simplify and unify the build process

## Actual Behavior

For now, we have a huge amount of shell scripts related to the compilation

ronaldtse:
Agree with @CAMOBAP on vcpkg, will streamline our dependency management. Let’s get this done.

CAMOBAP:
@ronaldtse yep I already have some progress in this direction and will push first PR in next 1-2 days

ronaldtse:
Thank you @CAMOBAP , didn't know you were working on it! Next time please take the task so we won't put in double work. 👍

CAMOBAP:
@ronaldtse for the  next time I will add a comment that it's in progress, to not confuse others

In Progress

CAMOBAP:
I am still in the middle, here are some findings which I have so far:
 - local I tested building dependencies as static libraries and shared libraries

For static deps:
 - `vcpkg install --triplet x64-linux $(cat vcpkg.txt)`
 - `cmake -B ./build -DCMAKE_BUILD_TYPE=Release -DVCPKG_TARGET_TRIPLET=x64-linux -DCMAKE_TOOLCHAIN_FILE=${VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake .`

For shared deps:
Create proper triplet file `${VCPKG_ROOT}/triplets/x64-linux-dynamic.cmake` with content:

```
set(VCPKG_TARGET_ARCHITECTURE x64)
set(VCPKG_CRT_LINKAGE dynamic)
set(VCPKG_LIBRARY_LINKAGE dynamic)

set(VCPKG_CMAKE_SYSTEM_NAME Linux)
```

 - `vcpkg install --triplet x64-linux-dynamic $(cat vcpkg.txt)`
 - `cmake -B ./build -DCMAKE_BUILD_TYPE=Release -DVCPKG_TARGET_TRIPLET=x64-linux-dynamic -DCMAKE_TOOLCHAIN_FILE=${VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake .`


All this above is just for information, I will automate this in CI anyway

CAMOBAP:
Quick status update:
 - ubuntu dynamic linking with so works tested [GCC, clang], [Debug, Release], tests passing
[...]

Issue #1508, opened 2021-05-25
Investigate streamlining CI processes to reduce time on cross-platform debugging


ronaldtse:
From @ni4: 

> Could we just create Docker containers with all installed pre-requisites, and use them (updating periodically) for CI instead of running complicated scripts each time/spending time on debugging?

_Originally posted by @ni4 in https://github.com/rnpgp/rnp/issues/1501#issuecomment-845982728_

maxirmx:
This is done in #1968

PR #1666, opened 2021-11-01
Added Clang builds to Ubuntu, CentOS7, CentOS8 (+ ossl),  MacOS, Debian9 and Windows-msys2 workflows

maxirmx:
#1129


maxirmx:
@ronaldtse 
I added Clang builds to Ubuntu jobs and it works.
I would say that it is expected behavior since Clangs team maintains strong compatibility with gcc.
Is that what we are looking for or I misunderstand the task ?

ronaldtse:
@maxirmx yes it should pass as we know that Thunderbird is using this configuration. 

@kaie whom should we seek if we want to know what CI configurations we should test to ensure builds at TB work? Thanks!

.github/workflows/ubuntu.yml
```diff
@@ -88,12 +110,23 @@ jobs:
           [ -d "${LOCAL_BUILDS}/rnp-build/src/tests" ]
           [ -d "${LOCAL_BUILDS}/rnp-build/src/tests/googletest-build" ]
           [ ! -d "${LOCAL_BUILDS}/rnp-build/src/tests/googletest-src" ]
+          
   cmake-offline-googletest:
+    strategy:
+      matrix:
+        cc: ["gcc"]
```
ni4, line 117: I don't think that we should add clang test here - this test is just to check whether offline googletest cmake define works well.

.github/workflows/ubuntu.yml
```diff
@@ -131,14 +164,25 @@ jobs:
...
-    env:
-      DOWNLOAD_RUBYRNP: Off
-      RNP_TESTS: rnp_tests.hash_test_success
-    runs-on: ubuntu-latest
-    if: "!contains(github.event.head_commit.message, 'skip ci')"
-    timeout-minutes: 50
-    steps:
+   strategy:
+      matrix:
+        cc: ["gcc"]
```
ni4, line 171: Same here - no need to recheck with different compilers, as we are just making sure that CMake define works correctly.

.github/workflows/ubuntu.yml
```diff
@@ -164,6 +208,7 @@ jobs:
           ls -la "${RNP_INSTALL}/lib/"librnp*.so
           ! ctest -N | grep 'ruby' > /dev/null
           ! ls -la src/tests/ | grep 'ruby' > /dev/null
+
```
ni4, line 211: Do we need this additional empty line?

.github/workflows/ubuntu.yml
```diff
@@ -54,12 +65,23 @@ jobs:
...
           cd ${LOCAL_BUILDS}/rnp-build
           cpack -G DEB -D CPACK_DEBIAN_PACKAGE_SHLIBDEPS_PRIVATE_DIRS="${BOTAN_INSTALL}/lib;${JSONC_INSTALL}/lib;${GPG_INSTALL}/lib"
+          
   cmake-offline-googletest-src:
+    strategy:
+      matrix:
+        cc: ["gcc"]
+        cxx: ["g++"]
+        include:
+          - cc: clang
```
ni4, line 75: This test is just to check that we build correctly with Googletest from sources via CMake define, no need to check with different compilers.

ni4 (changes requested):
I'd suggest to remove redundant tests, where we (most likely) do not need to check different compilers.

.github/workflows/ubuntu.yml
maxirmx: Frankly speaking, I have no idea what is really required. In 99%+ cases everything that works with gcc will also work with clang. 
For other workflows I was debugging and fixing ci shell scripts which are .... well, a little bit difficult to maintain.
So if there is a list of tests that have to be extended to clang I would rather  delete extra staff before any reviews and save some time for everybody.

Thank you!

maxirmx: Considering this specific test, we check that offline googletest cmake define works well when CC/CXX is set to clang/clang++ which activates different toolchain.  As I mentioned above it really tests ci scripts - run.sh and others.

.github/workflows/ubuntu.yml
maxirmx: Same as above - it really tests ci scripts

.github/workflows/ubuntu.yml
maxirmx: Same as above - it really tests ci scripts

.github/workflows/ubuntu.yml
maxirmx: No, this additional empty line is optional. I will remove it. Thank you.

maxirmx:
Clang build and test for the following workflows:
- centos7
- centos8-ossl
- centos8
- debian9
- macos
- ubuntu

Fixed:
- #1670 
- #1669
- #1668 
- Packaging tests for centos8

maxirmx:
@ronaldtse @ni4 

I believe this PR is ready for merge.   All changes are listed in the previous comment:https://github.com/rnpgp/rnp/pull/1666#issuecomment-962404388 

Some general considerations:

1. This PR  is about changes to YAML and CI shell scripts so that they could generate correct clang configurations.  I reduced the number of tests as @ni4 asked in his review but I have some doubts that such reduction is consistent.

2. CI shell scripts are very difficult to understand, modify and maintain.  I believe they need refactoring and using package manager (or managers) as #1648  suggests would be a good approach.  #1508 will also work but it will force using specific (possibly outdated) versions of dependencies.

ronaldtse:
> 2. CI shell scripts are very difficult to understand, modify and maintain.  I believe they need refactoring and using package manager (or managers) as [use vcpkg for all platforms #1648](https://github.com/rnpgp/rnp/issues/1648)  suggests would be a good approach.  [Investigate streamlining CI processes to reduce time on cross-platform debugging #1508](https://github.com/rnpgp/rnp/issues/1508) will also work but it will force using specific (possibly outdated) versions of dependencies.

I agree. Let's merge this first and migrate to a package manager?

ronaldtse (approved):
Thank you @maxirmx !

ci/lib/install_functions.inc.sh
```diff
@@ -155,18 +155,19 @@ declare util_depedencies_yum=(
...
+   clang
+   gcc
+   gcc-c++
+   make
+   autoconf
+   automake
+   libtool
+   bzip2
+   gzip
+ )
```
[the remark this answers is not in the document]
maxirmx: I do not agree. I made identation consistent. Without extra whitespace these 4 lines have identation that differs from the code before and after.

.github/workflows/windows.yml
```diff
@@ -5,10 +5,13 @@ on:
     branches:
       - master
       - 'release/**'
+      - maxirmx-clang
```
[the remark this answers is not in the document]
maxirmx: It has been removed already.

.github/workflows/ubuntu.yml
```diff
@@ -87,13 +108,23 @@ jobs:
           ci/run.sh
           [ -d "${LOCAL_BUILDS}/rnp-build/src/tests" ]
           [ -d "${LOCAL_BUILDS}/rnp-build/src/tests/googletest-build" ]
-          [ ! -d "${LOCAL_BUILDS}/rnp-build/src/tests/googletest-src" ]
+          [ ! -d "${LOCAL_BUILDS}/rnp-build/src/tests/googletest-src" ]
```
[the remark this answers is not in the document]
maxirmx: Yes, thank you

.github/workflows/ubuntu.yml
maxirmx: These tests are commented

.github/workflows/ubuntu.yml
maxirmx: These tests are commented

.github/workflows/ubuntu.yml
maxirmx: These tests are commented

ci/lib/install_functions.inc.sh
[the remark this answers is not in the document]
maxirmx: Yes it had  double spaces before the changes and it has double spaces after changes.
I think there was a version in the middle that had wrong spacing.

ronaldtse:
Ping @ni4 @ribose-jeffreylau for one more review.

ci/env-msys.inc.sh
```diff
@@ -6,8 +6,8 @@
 export CFLAGS="${CFLAGS}"
 export CXXFLAGS="${CXXFLAGS}"
 export LDFLAGS="${LDFLAGS}"
-export CC=gcc
-export CXX=g++
+export CC=${CC-gcc} 
+export CXX=${CXX-g++}
```
ribose-jeffreylau, line 10: If `$CC` (and `$CXX`) are set to empty strings, after `CC=${CC-gcc}`, `$CC` would still remain as empty strings.  Is this intended?  (similar with `$CXX`)

ronaldtse:
@maxirmx it's okay to have this merged right now since it is only CI related.

ci/lib/install_functions.inc.sh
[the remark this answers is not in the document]
ribose-jeffreylau: I think the version here is the version that has wrong spacing :)

ci/lib/install_functions.inc.sh
```diff
@@ -522,6 +526,17 @@ linux_install_debian() {
     "${build_dependencies_deb[@]}" \
     "$@"
 
+  if [ "${CC-gcc}" = "clang" ]; then
```
ribose-jeffreylau, line 529: Same issue as above --- with `$CC` set to empty string, `"${CC-gcc}"` would still be empty string.  Is this intended?

ci/lib/install_functions.inc.sh
```diff
@@ -535,30 +550,43 @@ linux_install() {
...
     make
-    pkg-config
+    pkg-config 
     mingw64/mingw-w64-x86_64-cmake
-    mingw64/mingw-w64-x86_64-gcc
-    mingw64/mingw-w64-x86_64-json-c
     mingw64/mingw-w64-x86_64-python3
   )
 
+  if [ "${CC-gcc}" = "gcc" ]; then
```
ribose-jeffreylau, line 565: ditto

ci/lib/install_functions.inc.sh
```diff
@@ -535,30 +550,43 @@ linux_install() {
...
+              clang64/mingw-w64-clang-x86_64-libbotan
+              clang64/mingw-w64-clang-x86_64-libssp
+              clang64/mingw-w64-clang-x86_64-json-c
+              clang64/mingw-w64-clang-x86_64-libsystre
+   ) 
+  fi
+
   pacman --noconfirm -S --needed "${packages[@]}"
 
+  if [ "${CC-gcc}" = "gcc" ]; then
```
ribose-jeffreylau, line 582: ditto

ci/lib/install_functions.inc.sh
```diff
@@ -833,7 +866,7 @@ build_example_pkgconfig() {
   pushd "$(mktemp -d)" || return 1
 
   # shellcheck disable=SC2046
-  gcc "${rnpsrc}/src/examples/generate.c" -ogenerate $(pkg-config --cflags --libs $pkgflags librnp) $gccflags
+  ${CC-gcc} "${rnpsrc}/src/examples/generate.c" -ogenerate $(pkg-config --cflags --libs $pkgflags librnp) $gccflags
```
ribose-jeffreylau, line 869: ditto

ci/main.sh
```diff
@@ -32,11 +32,22 @@ prepare_build_prerequisites() {
...
   prepare_build_tool_env
 
-  export LD_LIBRARY_PATH="${GPG_INSTALL}/lib:${BOTAN_INSTALL}/lib:${JSONC_INSTALL}/lib:${RNP_INSTALL}/lib:$LD_LIBRARY_PATH"
+  export LD_LIBRARY_PATH="${GPG_INSTALL}/lib:${BOTAN_INSTALL}/lib:${JSONC_INSTALL}/lib:${RNP_INSTALL}/lib:${LD_LIBRARY_PATH-}"
 
-  # update dll search path for windows
   if [[ "${OS}" = "msys" ]]; then
+  # update dll search path for windows
     export PATH="${LOCAL_BUILDS}/rnp-build/lib:${LOCAL_BUILDS}/rnp-build/bin:${LOCAL_BUILDS}/rnp-build/src/lib:$PATH"
+    if [[ "${CC-gcc}" = "clang" ]]; then
```
ribose-jeffreylau, line 40: ditto

ci/main.sh
```diff
@@ -32,11 +32,22 @@ prepare_build_prerequisites() {
 prepare_test_env() {
   prepare_build_tool_env
 
-  export LD_LIBRARY_PATH="${GPG_INSTALL}/lib:${BOTAN_INSTALL}/lib:${JSONC_INSTALL}/lib:${RNP_INSTALL}/lib:$LD_LIBRARY_PATH"
+  export LD_LIBRARY_PATH="${GPG_INSTALL}/lib:${BOTAN_INSTALL}/lib:${JSONC_INSTALL}/lib:${RNP_INSTALL}/lib:${LD_LIBRARY_PATH-}"
```
ribose-jeffreylau, line 35: Would an improvement be:

```bash
export ...="...:${RNP_INSTALL}/lib${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}"
```

so that if `$LD_LIBRARY_PATH` is an empty string, there would not be a trailing colon?

ci/lib/install_functions.inc.sh
```diff
@@ -535,30 +550,43 @@ linux_install() {
...
-    libbz2-devel
     git
     automake
     autoconf
     libtool
     automake-wrapper
     gnupg2
     make
-    pkg-config
+    pkg-config
```
ribose-jeffreylau, line 560: There's a trailing white space here.

maxirmx:
@ronaldtse 
It now has a conflict with https://github.com/rnpgp/rnp/pull/1681
Sorry, I wanted to merge the opposite direction

.github/workflows/ubuntu.yml
```diff
@@ -132,38 +163,48 @@ jobs:
...
+         run: |
+           set -x
+           ci/install_cacheable_dependencies.sh
+       - name: tests
+         run: |
+           set -x
+           ci/run.sh
+           ls -la "${RNP_INSTALL}/lib/"librnp*.so
+           ! ctest -N | grep 'ruby' > /dev/null
+           ! ls -la src/tests/ | grep 'ruby' > /dev/null
```
ribose-jeffreylau, line 207: I think these lines have been indented by 11 white spaces?  Maybe an accident?

ribose-jeffreylau (changes requested):
@maxirmx I left some comments above.  Thanks!

ci/env-msys.inc.sh
ribose-jeffreylau: Also, there is a trailing white space on line 9.

maxirmx: This is protection against CC/CXX unset.    
CC=""  will force cmake default which would be gcc.   If CC is unset it will also force cmake default but it may fail at this line depending on shell settings.   

Actually it is an interesting issue related to CI logic consistency. 
CI logic is distributed accross _YAML-shell scripts-cmake_    If thess shell scripts are always called from GHA you can ensure that CC and other variablle are set in YAML.   Would you like to rely on this scenario only or hacve some extra logic that make scripts work reliably without GHA ?

ci/main.sh
maxirmx: If $LD_LIBRARY_PATH is unset the script won't fail

ci/lib/install_functions.inc.sh
maxirmx: CC="" will force cmake default which would be gcc. If CC is unset it will also force cmake default but it may fail at this line depending on shell settings.

ci/lib/install_functions.inc.sh
maxirmx: CC="" will force cmake default which would be gcc. If CC is unset it will also force cmake default but it may fail at this line depending on shell settings.

ci/lib/install_functions.inc.sh
maxirmx: CC="" will force cmake default which would be gcc. If CC is unset it will also force cmake default but it may fail at this line depending on shell settings.

ci/lib/install_functions.inc.sh
maxirmx: CC="" will force cmake default which would be gcc. If CC is unset it will also force cmake default but it may fail at this line depending on shell settings.

ci/main.sh
maxirmx: CC="" will force cmake default which would be gcc. If CC is unset it will also force cmake default but it may fail at this line depending on shell settings.

ci/lib/install_functions.inc.sh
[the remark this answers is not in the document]
maxirmx: I agree with everything you say about spacing and empty lines.

ci/env-msys.inc.sh
ribose-jeffreylau: I see, I didn't know about cmake's behaviour on empty string CC :)  Thanks!

Before your reply, I would have preferred `export CC=${CC:-gcc}` with the extra colon, so `CC` would actually explicitly take on the value of `"gcc"` instead of becoming an empty string when `CC` is originally unset.

Observe:

```console
$ unset TEST
$ echo "${TEST-this is a test}"
this is a test
$ TEST=
$ echo "${TEST-this is a test}"

$ echo  "${TEST:-this is a test}"
this is a test
```

> Actually it is an interesting issue related to CI logic consistency.
CI logic is distributed accross YAML-shell scripts-cmake If thess shell scripts are always called from GHA you can ensure that CC and other variablle are set in YAML. Would you like to rely on this scenario only or hacve some extra logic that make scripts work reliably without GHA ?

For now, I think we can safely assume that these scripts are always run within GHA, as long as they live inside `ci/`.

ci/lib/install_functions.inc.sh
ribose-jeffreylau: Here, when `CC` is an empty string, `${CC-gcc}` would actually remain an empty string.  [See my comment above.](https://github.com/rnpgp/rnp/pull/1666#discussion_r745410042)

ni4:
@maxirmx some other side notes:
- we got used to the branch naming like  `maxirmx-1129-build-with-clang-cl`
- we also prefer to rebase on master instead of merge, squashing together commits like 'fix of fix', 'fix pr comment'. Interactive rebase makes this easy and straightforward.

ci/env-msys.inc.sh
maxirmx: Actually I also put _export CC=${CC:-gcc}_ first but then I relized that if empty string comes from the upstream it does not hurt but may have some idea behind it.

ci/main.sh
ribose-jeffreylau: > If $LD_LIBRARY_PATH is unset the script won't fail

In which case `$LD_LIBRARY_PATH` would end up having a trailing `:` at the end.  I'm not familiar with the behaviour of `$LD_LIBRARY_PATH` with empty components, but if it's what's commonly accepted then I'm fine with it :)

maxirmx:
> @maxirmx some other side notes:

> * we got used to the branch naming like  `maxirmx-1129-build-with-clang-cl`

Does 'ni4-fix-rnp-hash-destructutor' follow this rule ?

ronaldtse:
@maxirmx we usually use `{username}-{topic}` as a development branch name under the rnp repo when it is shared. If the PR comes from a forked repository, the branch name isn't important.

ni4:
> > * we got used to the branch naming like  `maxirmx-1129-build-with-clang-cl`
> 
> Does 'ni4-fix-rnp-hash-destructutor' follow this rule ?

In this case there were no issue created for, so, yeah, no issue number.

PR #1667, opened 2021-11-02
Limit shared object/dylib symbols exports only to rnp_* FFI functions.

ni4:
Should fix issue #1414

### DIFF
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -244,6 +244,18 @@ set_target_properties(librnp
 
 if (BUILD_SHARED_LIBS)
   add_library(librnp-static STATIC $<TARGET_OBJECTS:librnp-obj> $<TARGET_OBJECTS:rnp-common>)
+  # Limit symbols export only to rnp_* functions.
+  if (APPLE)
+    # Check for Apple OSs and use -export_symbols_list
+    file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/librnp.symbols" "_rnp_*\n")
+    target_link_options(librnp PRIVATE -Wl,-exported_symbols_list "${CMAKE_CURRENT_BINARY_DIR}/librnp.symbols")
+    set_target_properties(librnp PROPERTIES LINK_DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/librnp.symbols")
+  elseif(NOT WIN32)
+    # Use --version-script otherwise
+    file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/librnp.vsc" "{\nglobal: rnp_*;\nlocal: *;\n};\n")
+    target_link_options(librnp PRIVATE -Wl,-O1 -Wl,--version-script=${CMAKE_CURRENT_BINARY_DIR}/librnp.vsc)
+    set_target_properties(librnp PROPERTIES LINK_DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/librnp.vsc")
+  endif()
 else()
   add_library(librnp-static ALIAS librnp)
 endif()
--- a/.github/workflows/centos8.yml
+++ b/.github/workflows/centos8.yml
@@ -96,6 +96,7 @@ jobs:
           set -euxo pipefail
           nm --defined-only -g $RNP_INSTALL/lib64/librnp*.so > exports
           grep -qv dst_close exports
+          grep -qv Botan exports
           grep -qw rnp_version_string exports
   pkgconfig-cmake-target:
     runs-on: ubuntu-latest


dkg:
This looks good to me (at least on the GNU/Linux side)

src/lib/CMakeLists.txt
```diff
@@ -244,6 +244,18 @@ set_target_properties(librnp
 
 if (BUILD_SHARED_LIBS)
   add_library(librnp-static STATIC $<TARGET_OBJECTS:librnp-obj> $<TARGET_OBJECTS:rnp-common>)
+  # Limit symbols export only to rnp_* functions.
+  if (APPLE)
+    # Check for Apple OSs and use -export_symbols_list
+    file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/librnp.symbols" "_rnp_*\n")
```
dkg, line 250: I don't claim to know much about macOS, but i'm a bit surprised to see the leading underscore here.  maybe the leading underscore is injected by the macOS toolchain?  [Apple documentation](https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryDesignGuidelines.html#//apple_ref/doc/uid/TP40002013-SW4) seems to suggest that it does, though.

src/lib/CMakeLists.txt
```diff
@@ -244,6 +244,18 @@ set_target_properties(librnp
...
+    target_link_options(librnp PRIVATE -Wl,-exported_symbols_list "${CMAKE_CURRENT_BINARY_DIR}/librnp.symbols")
+    set_target_properties(librnp PROPERTIES LINK_DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/librnp.symbols")
+  elseif(NOT WIN32)
+    # Use --version-script otherwise
+    file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/librnp.vsc" "{\nglobal: rnp_*;\nlocal: *;\n};\n")
```
dkg, line 255: Is there a reason to write this file via cmake `file(WRITE[…])` instead of just storing `librnp.vsc` directly in the source?

Tracking `librnp.vsc` in the source explicitly would make it easier to add things like version tags to new symbols going forward.

src/lib/CMakeLists.txt
```diff
@@ -244,6 +244,18 @@ set_target_properties(librnp
...
+    target_link_options(librnp PRIVATE -Wl,-O1 -Wl,--version-script=${CMAKE_CURRENT_BINARY_DIR}/librnp.vsc)
```
dkg, line 256: Why does this change add `-O1` to the linker?  is that necessary for this fix?  it seems problematic because this might override other explicit choices for optimization (e.g. default settings in other build systems)

dkg (changes requested):
I am happy to see this work happen -- but i think it could be a bit cleaner, as noted above.

- store the version-script and exported_symbols_list files explicitly in the source (and point to them in the source rather than in `${CMAKE_CURRENT_BINARY_DIR}`)
- don't introduce linker optimization in this change.

src/lib/CMakeLists.txt
ni4: @dkg thanks for the link, I was searching from the other side (for replacement of version script). As my development machine is Mac, I used try-and-see approach for this underscore and proven that it works. Probably I should update tests to check for the exported symbols for the macOS CI as well...

src/lib/CMakeLists.txt
ni4: The only reason is that librnp.vsc/librnp.symbols are almost one-liners, so it looked simpler to write it out from CMake instead of adding the file. However, probably it would be more clear to have those files in the source tree, agree.

dkg:
I've offered a variant PR over on #1671

src/lib/CMakeLists.txt
ni4: It's not necessary, just noticed the thing below, inspired by the Ulrich Drepper's article on shared libs, part regarding the relocation speedups. However if it may cause issues it should be removed...
```
[root@ee62582d45c9 rnp-build]# readelf -I src/lib/librnp.so

Histogram for bucket list length (total of 1017 buckets):
 Length  Number     % of total  Coverage
      0  577        ( 56.7%)
      1  332        ( 32.6%)     59.0%
      2  94         (  9.2%)     92.4%
      3  13         (  1.3%)     99.3%
      4  1          (  0.1%)    100.0%

Histogram for `.gnu.hash' bucket list length (total of 421 buckets):
 Length  Number     % of total  Coverage
      0  230        ( 54.6%)
      1  154        ( 36.6%)     66.4%
      2  33         (  7.8%)     94.8%
      3  4          (  1.0%)    100.0%
[root@ee62582d45c9 rnp-build]# readelf -I /usr/local/lib   
lib/     lib64/   libexec/ 
[root@ee62582d45c9 rnp-build]# readelf -I /usr/local/lib64/librnp.so

Histogram for bucket list length (total of 263 buckets):
 Length  Number     % of total  Coverage
      0  43         ( 16.3%)
      1  78         ( 29.7%)     16.5%
      2  75         ( 28.5%)     48.3%
      3  40         ( 15.2%)     73.7%
      4  15         (  5.7%)     86.4%
      5  9          (  3.4%)     96.0%
      6  2          (  0.8%)     98.5%
      7  1          (  0.4%)    100.0%

Histogram for `.gnu.hash' bucket list length (total of 197 buckets):
 Length  Number     % of total  Coverage
      0  64         ( 32.5%)
      1  68         ( 34.5%)     29.2%
      2  41         ( 20.8%)     64.4%
      3  19         (  9.6%)     88.8%
      4  1          (  0.5%)     90.6%
      5  3          (  1.5%)     97.0%
      6  0          (  0.0%)     97.0%
      7  1          (  0.5%)    100.0%
```

ni4:
Closing in favour of #1671 and #1674.

PR #1671, opened 2021-11-03
Limit shared object/dylib symbols exports only to rnp_* FFI functions.

dkg:
Note: i don't know whether the Apple variant works here, i've just
copy/pasted it from https://github.com/rnpgp/rnp/pull/1667.  In
particular, I'm a little bit surprised that there's whitespace between
exported_symbols_list and the path to the librnp.symbols, or at least
that there is no `-Wl,` prefix to the second chunk.

But i don't have a macOS dev environment set up so i can't confirm.

Closes: #1414


dkg:
This is mutually incompatible with #1667 -- only one should be merged!

ni4:
@dkg thanks, this actually looks more clear then #1667 .

> particular, I'm a little bit surprised that there's whitespace between
exported_symbols_list and the path to the librnp.symbols, or at least
that there is no -Wl, prefix to the second chunk.

I will re-check this, but as far as I remember passing `-exported_symbols_list="${CMAKE_CURRENT_SOURCE_DIR}/librnp.symbols"` caused some silent surprising linker error.

ni4 (approved):
LGTM, thanks, @dkg!

antonsviridenko (approved):

antonsviridenko:
I see that "neutral check" makes it look as unsuccessful?

Issue #1439, opened 2021-03-02
s390x and sparc64 "failures" in s2k_iteration_tuning : ratios < 6


dkg:
in the [s390x debian build](https://buildd.debian.org/status/fetch.php?pkg=rnp&arch=s390x&ver=0.14.0-6&stamp=1614653756&raw=0), with #1419 applied, the `s2k_iteration_tuning` test fails with a ratio of ~1.8:

```
 15/213 Test  #15: rnp_tests.s2k_iteration_tuning ................................................***Failed    0.51 sec
Note: Google Test filter = rnp_tests.s2k_iteration_tuning
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from rnp_tests
[ RUN      ] rnp_tests.s2k_iteration_tuning
./src/tests/cipher.cpp:665: Failure
Expected: ((static_cast<double>(iters_100) / iters_10)) > ((6)), actual: 1.82353 vs 6
[pgp_s2k_compute_iters() ./src/lib/crypto/s2k.cpp:178] failed to create hash object
[  FAILED  ] rnp_tests.s2k_iteration_tuning (506 ms)
[----------] 1 test from rnp_tests (506 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (506 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] rnp_tests.s2k_iteration_tuning

 1 FAILED TEST
```

on the [sparc64 buildd](https://buildd.debian.org/status/fetch.php?pkg=rnp&arch=sparc64&ver=0.14.0-6&stamp=1614655232&raw=0) we see a ratio of ~4.4:

```
 31/213 Test  #15: rnp_tests.s2k_iteration_tuning ................................................***Failed    0.64 sec
Note: Google Test filter = rnp_tests.s2k_iteration_tuning
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from rnp_tests
[ RUN      ] rnp_tests.s2k_iteration_tuning
./src/tests/cipher.cpp:665: Failure
Expected: ((static_cast<double>(iters_100) / iters_10)) > ((6)), actual: 4.42857 vs 6
[pgp_s2k_compute_iters() ./src/lib/crypto/s2k.cpp:178] failed to create hash object
[...]

PR #1672, opened 2021-11-04
Enable loosening the s2k_iteration_tuning test

dkg:
Some platforms or some specific machiens (e.g. the s390x and sparc64
debian build daemons) don't appear to consume a linearly-proportional
amount of wall clock time as the S2K iteration count increases.

This patch changes nothing substantive about the s2k_iteration_tuning
test, but enables specific machines or architectures to override the
expectation with a simple #define (e.g. by adding
-DS2K_MINIMUM_TUNING_RATIO=1.5 to CXXFLAGS).

I'm generally of the opinion that this kind of tuning is misleading or
even counterproductive when your threat model is modern hardware (the
iterated S2K count octet should probably just always be maxed out to
0xff, and even better would be to stop using this S2K mechanism and
adopt the upcoming argon feature).  So the right thing overall might
be to stop doing this tuning entirely.

But this targeted change would be sufficient for debian's
purposes. With this patch, we can easily tell specific build daemons
that they don't need to worry about a linear ratio between 10
iterations and 100 iterations of SHA512 and get on with the rest of
the test suite.

Closes: #1439

### DIFF
--- a/src/tests/cipher.cpp
+++ b/src/tests/cipher.cpp
@@ -634,6 +634,12 @@ TEST_F(rnp_tests, test_dsa_verify_negative)
     assert_int_equal(dsa_verify(&sig, message, h_size, key1), RNP_ERROR_SIGNATURE_INVALID);
 }
 
+// platforms known to not have a robust response can compile with
+// -DS2K_MINIMUM_TUNING_RATIO=2 (or whatever they need)
+#ifndef S2K_MINIMUM_TUNING_RATIO
+#define S2K_MINIMUM_TUNING_RATIO 6
+#endif
+
 TEST_F(rnp_tests, s2k_iteration_tuning)
 {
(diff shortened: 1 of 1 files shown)


ni4:
> I'm generally of the opinion that this kind of tuning is misleading or
even counterproductive when your threat model is modern hardware (the
iterated S2K count octet should probably just always be maxed out to
0xff, and even better would be to stop using this S2K mechanism and
adopt the upcoming argon feature). So the right thing overall might
be to stop doing this tuning entirely.

@dkg Agree, it seems wrong that security of message/key protected on slow device would be lower then on the modern CPU. From other side, it would not be convenient to wait 5-10 seconds for each key unlock on the slow device.
Filed issue #1675 for Argon.

ni4 (approved):
LGTM, thanks @dkg !

antonsviridenko (approved):

Issue #1664, opened 2021-10-29
test_sym_encryption__rnp_aead fails on armhf and armel when AEAD chunk size is too large


dkg:
on [armel](https://buildd.debian.org/status/fetch.php?pkg=rnp&arch=armel&ver=0.15.2-3&stamp=1635525726&raw=0) and [armhf](https://buildd.debian.org/status/fetch.php?pkg=rnp&arch=armhf&ver=0.15.2-3&stamp=1635525577&raw=0) (debian's two 32-bit arm architectures), the `test_sym_encryption__rnp_aead` test fails.  (warning: those build logs are ~60MiB in size and are a lot to render!)

they fail during that test like so:

```
/build/src/rnp/rnp --homedir /tmp/rnpctmpxr4myokp/.rnp -z 1 --aead=eax --aead-chunk-bits=30 --cipher AES256 --passwords 1 --pass-fd 4 -c /tmp/rnpctmpxr4myokp/cleartext.txt --output /tmp/rnpctmpxr4m
yokp/cleartext.enc
[encrypted_start_aead_chunk() ./src/librepgp/stream-write.cpp:363] failed to set ad
[process_stream_sequence() ./src/librepgp/stream-write.cpp:1731] failed to process data
Operation failed: Error writing file

ERROR
```

It looks like it might be the `--aead-chunk-bits=30` that causes the failure.  Since [the latest draft of crypto-refresh](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-04.html#name-aead-encrypted-data-packet-) limits the chunk size parameter to a value of 16, perhaps the tests with large chunks are unnecessary anyway?

dkg:
in debian, i've gotten the builds to stop failing on armel and armhf by dropping larger aead chunk sizes with [this patch](https://salsa.debian.org/debian/rnp/-/blob/debian/main/debian/patches/0002-Avoid-testing-chunk-sizes-greater-than-16.patch).

PR #1673, opened 2021-11-04
Avoid testing AEAD chunk sizes greater than 16

dkg:
In draft-ietf-openpgp-crypto-refresh-04, AEAD chunk size is capped at
16.  We don't need to test more than that, and it looks like tests of
chunk size 30 fail on 32-bit platforms like armel and armhf (though
they don't fail on i386 for some reason i don't understand).

Closes: #1664

### DIFF
--- a/src/tests/cli_tests.py
+++ b/src/tests/cli_tests.py
@@ -2705,8 +2705,7 @@ class Encryption(unittest.TestCase):
             CIPHERS.remove('TWOFISH')
         AEAD_C = list_upto(CIPHERS, Encryption.RUNS)
         AEAD_M = list_upto([None, 'eax', 'ocb'], Encryption.RUNS)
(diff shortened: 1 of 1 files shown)


ni4:
> though they don't fail on i386 for some reason i don't understand.

@dkg RNP uses `size_t` for chunklen, and sets it via `param->chunklen = 1L << (hdr[3] + 6);`.
So it definitely should fail if size_t is 32 bit. Could it be that size_t is 64-bit on i386 (as I understand it should be 32-bit, but Kernel: Linux 5.10.0-9-amd64 gives some doubts).

ni4 (approved):
LGTM, thanks @dkg!

Filed #1676 to limit chunk size in the codebase and fix possible 32-bit platform issues as well.

antonsviridenko (approved):

PR #1677, opened 2021-11-04
Refactor pgp_hash_t to rnp::Hash/rnp::CRC24 classes.

ni4:
This PR moves forward in the C++ transition, replacing old C code with C++ one, moving from pointers to references where applicable as well. Still a lot of things to change.


src/lib/crypto/hash.cpp
```diff
@@ -52,26 +23,6 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
-/*
- * Copyright (c) 2005-2008 Nominet UK (www.nic.uk)
```
antonsviridenko, line 56: I'm bit concerned about removing these copyright notices, I think it is still can be considered as a derived work. Better to keep them.

ronaldtse: @antonsviridenko we are slowly working through legacy inherited code and will remove inappropriate notices when those sections are fully revamped, guided by automatic code analysis.

antonsviridenko (approved):

ronaldtse (approved):

ronaldtse:
Thank you @ni4 !

Issue #1670, opened 2021-11-03
clang/rnp build issue (CentOS 7 workflow)


maxirmx:
Target: CentOS 7
Tests pkgconfig-cmake-target (no, clang, clang++), pkgconfig-cmake-target (yes, clang, clang++
Module: librepgp

```
In file included from /__w/rnp/rnp/src/librepgp/stream-common.cpp:50:
In file included from /__w/rnp/rnp/src/lib/crypto/mem.h:34:
In file included from /usr/include/botan-2/botan/secmem.h:12:
/usr/include/botan-2/botan/mem_ops.h:147:18: error: no member named 'is_trivially_copyable' in namespace 'std'
   static_assert(BOTAN_IS_TRIVIALLY_COPYABLE(T), "");
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/botan-2/botan/mem_ops.h:124:45: note: expanded from macro 'BOTAN_IS_TRIVIALLY_COPYABLE'
#define BOTAN_IS_TRIVIALLY_COPYABLE(T) std::is_trivially_copyable<T>::value
                                       ~~~~~^
/usr/include/botan-2/botan/mem_ops.h:147:46: error: 'T' does not refer to a value
   static_assert(BOTAN_IS_TRIVIALLY_COPYABLE(T), "");
                                             ^
/usr/include/botan-2/botan/mem_ops.h:124:67: note: expanded from macro 'BOTAN_IS_TRIVIALLY_COPYABLE'
#define BOTAN_IS_TRIVIALLY_COPYABLE(T) std::is_trivially_copyable<T>::value
                                                                  ^
/usr/include/botan-2/botan/mem_ops.h:145:19: note: declared here
template<typename T> inline void typecast_copy(uint8_t out[], T in[], size_t N)
                  ^
/usr/include/botan-2/botan/mem_ops.h:147:18: error: no member named 'value' in the global namespace
   static_assert(BOTAN_IS_TRIVIALLY_COPYABLE(T), "");
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/botan-2/botan/mem_ops.h:124:71: note: expanded from macro 'BOTAN_IS_TRIVIALLY_COPYABLE'
#define BOTAN_IS_TRIVIALLY_COPYABLE(T) std::is_trivially_copyable<T>::value
                                                                    ~~^
[...]

Issue #1669, opened 2021-11-03
clang/GnuPG v2.3.0-beta983 issue (CentOS 8 workflow)


maxirmx:
**Target:** CentOS 8
**Tests** (normal, beta, clang, clang++),  (sanitize, beta, clang, clang++)
**Module:** GnuPG v2.3.0-beta983
**Error:**
```
libcommon.a(libcommon_a-iobuf.o):(.bss+0x8): multiple definition of `iobuf_debug_mode'
t-iobuf.o:(.bss+0x0): first defined here
``` 
Other tests use GnuPG v2.2.24 whicj compiles without issues.

_Originally posted by @maxirmx in https://github.com/rnpgp/rnp/issues/1129#issuecomment-959545493_

ni4:
@maxirmx Probably we should use later beta (983 was picked just because it was the latest at the time when workflow was created).
Maybe this commits is related: https://dev.gnupg.org/rG6aff8a132815a84bab69401c1e7de96ec549fbf2

maxirmx:
If they do not support and/or test clang in their beta build it will be a real adventure to rely on it here

ronaldtse:
@maxirmx only RNP needs to be built with clang. Test modules like gnupg are installed using https://github.com/rnpgp/gpg-build-scripts and we don't care how they are built in this current repository. Thanks.

maxirmx:
I will enforce gnu toolchain in _ci/lib/cacheable_install_functions.inc.sh_ 
This will fix CentOS8/clang workflow

ronaldtse:
Thanks @maxirmx !

Issue #1668, opened 2021-11-03
clang/libgcrypt v1.9.3 issue  (CentOS 8 workflow)


maxirmx:
**Target:**  CentOS 8
**Tests**   (normal, 2.3.1, clang, clang++)
**Module:**  libgcrypt / 1.9.3
**Error:**
```
basic.c:315:16: error: inline assembly requires more registers than available
  asm volatile("movdqu %[data0], %%xmm0\n"
               ^
basic.c:315:16: error: inline assembly requires more registers than available
```

Other tests use  libgcrypt /v1.8.7  and  libgcrypt / v1.9.0-beta277  which are compiled without issues.

_Originally posted by @maxirmx in https://github.com/rnpgp/rnp/issues/1129#issuecomment-959510564_

ronaldtse:
@maxirmx we don't need GnuPG to work with clang, only RNP libraries. Thanks.

maxirmx:
same fix as #1669 (force gcc toolchain for GnuPg)

PR #1683, opened 2021-11-09
Added Clang builds to Ubuntu, CentOS7, CentOS8 (+ ossl), MacOS, Debian9 and Windows-msys2 workflows

maxirmx:
Clang build and test for the following workflows:

- centos7
- centos8-ossl
- centos8
- debian9
- macos
- ubuntu
- windows (msys)

Fixed:

- #1129 
- clang/rnp build issue (CentOS 7 workflow) #1670
- clang/GnuPG v2.3.0-beta983 issue (CentOS 8 workflow) #1669
- clang/libgcrypt v1.9.3 issue (CentOS 8 workflow) #1668

Not included in this PR comparing to the first attempt.
- Fix to 'centos8 / pkgconfig-cmake-target' job which was disabled in the upstream version. I have a fix for it, if you need it pls. create separate issue. It may go to another PR if so.
- All convinience changes that helped to run/debug CI scripts outside GHA environment

### DIFF
--- a/.github/workflows/debian9.yml
+++ b/.github/workflows/debian9.yml
@@ -62,7 +62,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v1
         with:
-          fetch-depth: 1
+          fetch-depth: 2
 
       - name: Setup noncacheable dependencies
         run: |
--- a/.github/workflows/debian9.yml
+++ b/.github/workflows/debian9.yml
@@ -9,6 +9,8 @@ on:
       - 'docs/**'
       - '**.adoc'
       - '**.md'
+      - '.github/workflows/*.yml'
+      - '!.github/workflows/debian9.yml'
   pull_request:
     paths-ignore:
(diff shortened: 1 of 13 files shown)


.github/workflows/debian9.yml
```diff
@@ -37,23 +39,30 @@ jobs:
...
           apt -y install git sudo
       
+# I386: actions/checkout@V2,  actions/cache@V2 actions/cache@V2
+#       give error about OCI container [https://github.com/actions/checkout/issues/334]
+#                                      [https://github.com/actions/runner/issues/1011]
       - name: Check out repository
         uses: actions/checkout@v1
         with:
-          fetch-depth: 1
+          fetch-depth: 10
```
maxirmx, line 65: There is some issue with actions/checkout@1 causing problems with fetch-depth: 1,2,3 ...
I was not able to find any quick explanation or support but fetch-depth: 10 works

PR #1695, opened 2021-11-12
#1648 building dependencies with vcpkg (skip ci)

CAMOBAP:
- #1648


CAMOBAP:
@ni4 @ribose-jeffreylau @maxirmx ubuntu is done, I plan to move forward to other platforms, I would like to get feedback on this moment, to make sure that I didn't break something important

Please don't hesitate to share your questions/feedback

ronaldtse:
Thanks @CAMOBAP , I think this looks reasonable.

maxirmx:
@CAMOBAP 
Looks good for me though I have a note.

You now have both 

```
 USE_STATIC_DEPENDENCIES: yes
```
and
```
triplet: [x64-linux, x64-linux-dynamic]
```

and it looks like some logical inconsistency.  I guess USE_STATIC_DEPENDENCIES does not have any sence now and can be removed.

CAMOBAP:
@maxirmx thanks for the feedback. I also was surprised that `USE_STATIC_DEPENDENCIES` was used for non-cacheable dependencies.

https://github.com/rnpgp/rnp/blob/master/ci/lib/install_functions.inc.sh#L276-L280

In my opinion `USE_STATIC_DEPENDENCIES` should not affect non-cacheable dependencies, how do you think?

maxirmx:
> @maxirmx thanks for the feedback. I also was surprised that `USE_STATIC_DEPENDENCIES` affects non-cacheable dependencies.
> 
> https://github.com/rnpgp/rnp/blob/master/ci/lib/install_functions.inc.sh#L276-L280
> 
> In my opinion `USE_STATIC_DEPENDENCIES` should not affect non-cacheable dependencies, how do you think?

I think it should not be a restriction as it appears to be. IMHO it shall be possible to use both sfatic and dynamic deoending on the application  context.

ribose-jeffreylau:
> @maxirmx thanks for the feedback. I also was surprised that `USE_STATIC_DEPENDENCIES` affects non-cacheable dependencies.
> 
> https://github.com/rnpgp/rnp/blob/master/ci/lib/install_functions.inc.sh#L276-L280
> 
> In my opinion `USE_STATIC_DEPENDENCIES` should not affect non-cacheable dependencies, how do you think?

To provide some context, I think I grouped `asciidoc` and `bundler` under "non-cacheable" as I wasn't sure how to properly cache their installations.  I believe that to be the only reason.

ni4:
@maxirmx @ribose-jeffreylau Does this PR have any value at this moment or we could safely close it now?

CAMOBAP:
@maxirmx unfortunatelly I didn't finish this PR(

BTW how do you think, is it makes sense to switch to some package manager (like vcpgk or conan) to simplify build scripts?

maxirmx:
> TW how do you think, is it makes sense to switch to some package manager (like vcpgk or conan) to simplify build scripts?

vcpkg actually rebuilds everything on each run. It works if it is possible to set up a cache and reuse the artifacts.  
However, this project uses several 32-bit workflows and GHA cache action does not exist for 32-bit workflows because cache action depends on Node.js and there is no 32-bit binary ditribution for it.
Also I do not think there is vcpkg version that reliably works with MSys
So, vcpkg can be applied only partially and won't cover the most demanding configurations

I cannot argue about conan, since I do not have any experience with it,

CAMOBAP:
@maxirmx thanks for the feedback

I propose to close this PR for now, and maybe return back to it later

PR #1698, opened 2021-11-15
Fix OpenSSL version detection in rnp::backend_version()

ni4:
This PR fixes (and slightly refactors) OpenSSL version detection code, which didn't work with dev/beta versions.
Now it should handle all possible cases as from `opensslv.h`.
Also it adds automatic backend detection in `cli_tests.py`.


ribose-jeffreylau (approved):

ronaldtse (approved):
Thanks @ni4 @ribose-jeffreylau !

PR #1699, opened 2021-11-15
Support parsing of 'None' algorithm in the compressed packet.

ni4:
While it's a bit awkward to have additional layer of uncompressed compressed stream, this configuration seems to be generated by other implementations.

I believe we had issue for this but cannot find it right now.


ronaldtse (approved):
Thank you @ni4 !

antonsviridenko (approved):

PR #1700, opened 2021-11-16
Fix coverity workflow which didn't run without RNP_TESTS variable.

ni4:
Currently workflow runs, but doesn't do anything due to:
```
+ ctest -j2 -R '' --output-on-failure
Test project /home/runner/work/rnp/rnp/builds/rnp-build
No tests were found!!!
```


ni4:
@ronaldtse  @ribose-jeffreylau Do you have any idea how to debug this change without merging to the master first?
We really need to finally get codecov back :)

ribose-jeffreylau:
@ni4 Maybe something like #1701?  Feel free to merge that into this branch.

ni4:
@ribose-jeffreylau Thanks! Will check that out.

ni4:
Argh, somehow messed up codecov and coverity (and seems it happened previously). Coverity doesn't need to run tests as it intercepts only compiler invocation, so empty RNP_TESTS is pretty correct.

PR #1703, opened 2021-11-17
Disable ruby-rnp download where applicable.

ni4:
We do not need to download it where we run only rnp_tests/cli_tests.


ribose-jeffreylau (approved):

antonsviridenko (approved):

ronaldtse:
Thanks guys. Given we have py-rnp now, we should probably treat ruby-rnp and py-rnp the same way. Will file new ticket.

PR #1706, opened 2021-11-22
Do not prematurely mark secret keys as valid.

ni4:
As it can go wrong in cases of dummy secret keys and so on.
This changes behaviour in tests.
Tests cover the case when new certification is generated by secret key which is not marked as valid.


ronaldtse (approved):
Thanks @ni4 !

ni4:
I think we may skip Codecov's report this time - on their site PR looks good, probably it doesn't refresh properly during force-push or something like that. Will keep an eye on it.

antonsviridenko (approved):

PR #1708, opened 2021-11-22
Upload only single report for all tests to the codecov to avoid mess.

ni4:
As I understand now, codecov uses only latest uploaded report from the PR, i.e. it can go messy for multiple test runs (and different for cli_tests and rnp_tests runs). So this PR changes to run single pass on centos8. Let's see how it goes.

Feel free to correct me if this works in a different way.

### DIFF
--- a/.github/workflows/centos7.yml
+++ b/.github/workflows/centos7.yml
@@ -71,18 +71,7 @@ jobs:
             CXX: clang++
             IMAGE: centos:7
 # Coverage can only been tested with the GNU compiler
-          - BUILD_MODE: coverage
-            GPG_VERSION: beta
-            RNP_TESTS: cli_tests
-            CC: gcc
-            CXX: g++
-            IMAGE: centos:7
-          - BUILD_MODE: coverage
-            GPG_VERSION: beta
-            RNP_TESTS: rnp_tests
-            CC: gcc
-            CXX: g++
-            IMAGE: centos:7
+# Upload only single report from centos8 to the codecov to avoid mess
     name: ${{ matrix.env.IMAGE }} Botan [mode ${{ matrix.env.BUILD_MODE }}; test type ${{ matrix.env.RNP_TESTS }}; CC ${{ matrix.env.CC }}; GnuPG ${{ matrix.env.GPG_VERSION }}]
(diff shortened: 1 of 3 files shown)


ni4:
@maxirmx Do you have any idea why centos8 workflow didn't run? Github documentation gives some doubts on whether `!` should be supported for `paths` and `paths-ignore`, or only for the first one. Thanks!

maxirmx:
I believe centos8 workflow did not run because it did not like line 74 
```
 RNP_TESTS: *
```


But it is also possible that I made a mistake with '!'

maxirmx:
@ni4 
I added quotes so that
```
RNP_TESTS: *
```
is now 
```
RNP_TESTS: "*"
```
and centos8 workflow has started.  

Whatever staff I have added with "!"  it is for "on push" only that are used by me only. You guys are doing builds "on pull request synchronize"

ni4:
@maxirmx Thanks! My fault with quotes. Second one is with a fact the GTest filter is regexp, not a glob :) Re-pushing now.

.github/workflows/centos7.yml
```diff
@@ -71,18 +71,19 @@ jobs:
...
-            CC: gcc
-            CXX: g++
-            IMAGE: centos:7
-          - BUILD_MODE: coverage
-            GPG_VERSION: beta
-            RNP_TESTS: rnp_tests
-            CC: gcc
-            CXX: g++
-            IMAGE: centos:7
+# Upload only single report from centos8 to the codecov to avoid mess
```
ronaldtse, line 74: @ni4 should we remove these commented out lines?

ronaldtse (approved):

ronaldtse:
@ni4 please feel free to merge once you address the question. Thanks!

.github/workflows/centos7.yml
ni4: @ronaldtse Sure, removing. And fixing one more thing :-)

ni4:
CI passes - merging. Most likely Codecov diff is because previously only rnp_tests run outcome was uploaded, with higher coverage.
Thanks all!

PR #1709, opened 2021-11-25
Add and use sha1collisiondetection.

ni4:
Adresses partially #1343


ronaldtse:
There's a warning in hash.cpp. Is this a concern?

Screenshot 2021-11-26 at 10 54 20 AM

ronaldtse (approved):

ni4:
> There's a warning in hash.cpp. Is this a concern?
> 

@ronaldtse Looks like LGTM do not understand correctly that in one case (SHA1) we use calloc, and in second case - new. Not sure whether we should use new/release for SHA1CD just to silence this warning.

Actually now I see that incorrect approach was used when refactoring from pgp_hash_t, reusing old one. We should use some factory-based thing, like Botan and cipher.hpp/cipher.cpp do.

ronaldtse:
@ni4 agree with the factory approach!

antonsviridenko (approved):

PR #1710, opened 2021-11-26
Refactor cli_tests to use unittest.TestCase assert methods.

### LINKED ISSUE #1610: cli_tests: use self.assertSomething() instead of current approach.
## Description

Currently we use old approach, before the `unittest`:
```
        ret, _, _ = run_proc(RNPK, ['--homedir', RNPDIR, '--import', data_path('test_key_validity/alice-sub-sec.pgp')])
        if ret != 0:
            raise_err('Alice secret key import failed')
```

However one below looks much cleaner and gives more information in case of error:

```
        ret, _, err = run_proc(RNPK, ['--homedir', RNPDIR, '--export-key', '--secret', 'alice'])
        self.assertNotEqual(ret, 0, 'Secret key export failed.')
        self.assertRegex(err, r'(?s)^.*Key\(s\) matching \'alice\' not found\.$', 'Wrong export output.')
```

ni4:
This started as attempt to quick-fix sonarlint comments, making code look scary in VS Code, and went too wild.
There are still some things to refactor, but that could be done in further PRs.
Fixes #1610

### DIFF
--- a/src/tests/cli_tests.py
+++ b/src/tests/cli_tests.py
@@ -78,7 +78,32 @@ WEIRD_USERID_TOO_LONG = 'x' * 125 + '@rnp' # totaling 129 (MAX_USER_ID + 1)
 KEY_ENCRYPT = 'encryption@rnp'
 KEY_SIGN_RNP = 'signing@rnp'
 KEY_SIGN_GPG = 'signing@gpg'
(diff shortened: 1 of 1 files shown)


ronaldtse (approved):
Thanks @ni4 !

antonsviridenko (approved):

Issue #1106, opened 2020-04-23
Do not create system rng is internal library code.


ni4:
## Description

Currently number of internal library functions creates rng:
`if (!rng_init(&rng, RNG_SYSTEM)) {...}`.
However, we should use the single one, which will be passed down through the function call list from the FFI.

PR #1712, opened 2021-11-30
Refactor certification generation functions.

### LINKED ISSUE #1012: rnp_key_add_uid calls signing twice.
## Description

Current code calls signing twice when adding uid and certificate to the secret key, while normally it should be just copied from the public key.

ni4:
This PR:
- improves flexibility of certification generation, which would help us later with key signing and so on.
- as a side effect, fixes #1012 
- adresses partially #1106


ronaldtse (approved):

antonsviridenko (approved):

antonsviridenko:
some issues with homebrew on macos

PR #1714, opened 2021-11-30
Correctly calculate key expiration on corner case with primary uid.

### LINKED ISSUE #1632: Correct key expiration calculation related to primary uid subpacket.
## Description

If some userid has two certifications, and older one marks key as primary, and the newer one doesn't, then RNP picks up key expiration from the older one. However, newer certification should be picked.
Also userid should not be marked as primary in this case, since primary flag is dropped by newer certification.

Details and test key are available here: https://bugzilla.mozilla.org/show_bug.cgi?id=1730979

ni4:
If uid had self-signature, marking key as primary, it was picked as most restricting for the key expiration calculation, even if uid had later self-signature without the primary flag.
- fixes #1632


ronaldtse (approved):
Thank you @ni4 !

antonsviridenko (approved):

antonsviridenko:
@ni4 this review request was missing in my github notifications for some reason, sorry

ni4:
@antonsviridenko Np, thanks!

PR #1717, opened 2021-12-13
Refactor signature checks to pgp_key_t methods.

ni4:
...consolidating code so we would be able to pass security context to the `pgp_key_t` and use it together with security profiles.


ronaldtse:
@ni4 there's an issue discovered in the fuzzer causing that build to fail.

ni4:
> @ni4 there's an issue discovered in the fuzzer causing that build to fail.

Yeah, it's due to exceptions thrown for unknown hash algorithm. Will update the PR once understand what's the better way to handle it.

ni4:
@ronaldtse Now it's green and ready for review :)

antonsviridenko (approved):

ronaldtse (approved):

ronaldtse:
Thank you very much @ni4!

PR #1720, opened 2021-12-16
Make pgp_signature_t::keyid() and pgp_signature_t::keyfp() noexcept.

ni4:
Fixes Coverity issues 1764240 and 1764241.


ronaldtse (approved):

ni4:
Looks like Codecov doesn't like that new lines are not covered with tests even while they were not covered before. So I think we may let it go as it is.

antonsviridenko (approved):

PR #1721, opened 2021-12-16
Use single RNG in library code and refactor RNG class.

### LINKED ISSUE #1106: Do not create system rng is internal library code.

ni4:
Within this PR:
- only single top-level RNG is used in underlying code, which fixes #1106 
- refactors POD rng_t to C++ class
- makes rnp_ffi_st C++ object instead POD, which will be usefull later

### DIFF
--- a/src/lib/crypto.cpp
+++ b/src/lib/crypto.cpp
@@ -152,7 +152,7 @@ pgp_generate_seckey(const rnp_keygen_crypto_params_t &crypto,
     seckey.sec_protection.s2k.usage = PGP_S2KU_NONE;
     seckey.material.secret = true;
     /* fill the sec_data/sec_len */
-    if (encrypt_secret_key(&seckey, NULL, NULL)) {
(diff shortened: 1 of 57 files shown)


ni4:
Again, codecov is not happy that some changed line are not covered buy previously those were not covered as well, and mostly are about unexpected exceptions.

ronaldtse (approved):

antonsviridenko (approved):

antonsviridenko:
Do we have some tests for RNG output? Looks like no. 
We can try to use "Dieharder test suite" https://webhome.phy.duke.edu/~rgb/General/dieharder.php
Or make some corner cases when /dev/urandom is not available or not functioning as expected on a specific platform.

ni4:
@antonsviridenko Thanks, answered in #1722.
Could we merge this or you have some more input?

PR #1734, opened 2022-01-11
Fix broken OpenSSL CI runs and corresponding undetected errors.

ni4:
OpenSSL crypto backend was accidentally deleted from the workflow, so for a while we didn't actually run OpenSSL tests.
This PR should fix this and accummulated errors, and make workflow file a bit less redundant.


ni4:
@ronaldtse @antonsviridenko codecov's failure should be handled via the issue #1736

antonsviridenko (approved):

ronaldtse (approved):
Thank you @ni4!

PR #1735, opened 2022-01-11
Force aligned access for sha1cd if it is compiled with sanitizers.

ni4:
As title says.


ni4:
Codecov seems to calculate coverage in a bit weird way in such cases.

antonsviridenko (approved):

ronaldtse (approved):
Thank you @ni4 !

PR #1737, opened 2022-01-13
Ease key material checks.

ni4:
As was discussed before via mailing/tickets, we decided to change behavior of the key material validation.
Previously it was done only during the key import, and via Botan's internal functions.
Within this PR:
- Old copyrights from ElGamal sources are replaced by actual one (most likely was copied from other source files during initial transition from the NetBSD codebase).
- ElGamal checks are changed to not use Botan's internal functions, as it was slow due to P primality test. Also it adds custom check for small subgroups.
- Public key is not checked during encryption/verification as it was considered that it is enough to do signature checks.
- Secret key is checked within the first secret key operation (i.e. when it is decrypted and available), afterwards validation result is cached.


antonsviridenko (approved):

ni4:
@ronaldtse Ping for review! The last PR before the v0.16.0 release.

ronaldtse (approved):
Thank you @ni4 !

PR #1740, opened 2022-01-19
Fix Cirrus CI run

ni4:
This PR:
- fixes CI run which was failing due to forced GnuPG gcc requirement. However, since cache key was based only on install.sh and ruby-rnp repository this did not appear for 3 or 4 months.
- updates cache key calculation
- update stable GnuPG and it's component versions to the LTS ones.


ribose-jeffreylau (approved):

ronaldtse (approved):

ronaldtse:
Thanks @ni4!

PR #1745, opened 2022-01-26
Accept armored messages without empty line after headers

### LINKED ISSUE #1289: Allow absence of extra empty line in armored messages.
## Description

Currently we require an extra empty line after the armor header and between base64 data, i.e.
```
-----BEGIN PGP PUBLIC KEY BLOCK-----

mQGNBFuJOtUBDACsrDfGPO9ZFB0TeFbAVJOkArNeBQtu/ShOSymuNrw+chhu8QtF
fHB935fc5AI1A0ls37DmSWnwhq8T06hpc8rBTZLTMkjutTUkBcUVbgPbyJSJzWEU
...
```

However, it doesn't seem to be strictly required by RFC, and GnuPG accepts data without an extra line:
```
-----BEGIN PGP PUBLIC KEY BLOCK-----
mQGNBFuJOtUBDACsrDfGPO9ZFB0TeFbAVJOkArNeBQtu/ShOSymuNrw+chhu8QtF
fHB935fc5AI1A0ls37DmSWnwhq8T06hpc8rBTZLTMkjutTUkBcUVbgPbyJSJzWEU
...
```

antonsviridenko:
Closes #1289

### DIFF
--- a/src/librepgp/stream-armor.cpp
+++ b/src/librepgp/stream-armor.cpp
@@ -659,6 +659,16 @@ armor_skip_line(pgp_source_t *src)
 }
 
 static bool
+is_base64_line(const char *line, size_t len)
+{
+    for (size_t i = 0; i < len && line[i]; i++) {
+        if (B64DEC[(uint8_t) line[i]] == 0xff)
+            return false;
+    }
+    return true;
+}
+
+static bool
 armor_parse_headers(pgp_source_t *src)
 {
     pgp_source_armored_param_t *param = (pgp_source_armored_param_t *) src->param;
@@ -676,6 +686,10 @@ armor_parse_headers(pgp_source_t *src)
             RNP_LOG("Too long armor header - truncated.");
             header[hdrlen] = '\0';
         } else if (hdrlen) {
+            if (is_base64_line(header, hdrlen)) {
+                RNP_LOG("Warning: no empty line after the base64 headers");
+                return true;
+            }
             src_skip(param->readsrc, hdrlen);
             if (rnp::is_blank_line(header, hdrlen)) {
                 return src_skip_eol(param->readsrc);
--- a/src/tests/data/test_messages/message.txt.cleartext-signed-nonewline
+++ b/src/tests/data/test_messages/message.txt.cleartext-signed-nonewline
@@ -1,0 +1,14 @@
+-----BEGIN PGP SIGNED MESSAGE-----
+Hash: SHA256
+This is test message to be signed, and/or encrypted, cleartext signed and detached signed.
+It will use keys from keyrings/1.
+End of message.
+-----BEGIN PGP SIGNATURE-----
+Version: rnp 0.9.0+git20191024.471.5201802
+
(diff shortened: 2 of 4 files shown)


ronaldtse (approved):
Thanks @antonsviridenko !

src/common/str-utils.cpp
```diff
@@ -77,6 +77,18 @@ is_blank_line(const char *line, size_t len)
     return true;
 }
 
+bool
+is_base64_line(const char *line, size_t len)
+{
+    for (size_t i = 0; i < len && line[i]; i++) {
+        if (!((line[i] >= 'A' && line[i] <= 'Z') || (line[i] >= 'a' && line[i] <= 'z') ||
```
ni4, line 84: For this purpose there is a table B64DEC in stream-armor.cpp, so check may be done via single lookup.

ni4 (changes requested):
`is_base64_line()` should reuse B64DEC table.

src/common/str-utils.cpp
antonsviridenko: ok

but `B64DEC' is declared as static, so either `is_base64_line()` should be moved to `stream-armor.cpp` and called only there, or `B64DEC` should be exported.

ni4: As `is_base64_line()` is used only in `stream-armor.cpp` it may be safely moved there.

ni4 (approved):
LGTM, thanks, @antonsviridenko !

src/librepgp/stream-armor.cpp
```diff
@@ -676,6 +686,14 @@ armor_parse_headers(pgp_source_t *src)
             RNP_LOG("Too long armor header - truncated.");
             header[hdrlen] = '\0';
         } else if (hdrlen) {
+            if (is_base64_line(header, hdrlen)) {
+                RNP_LOG("Warning: no empty line after the base64 headers");
+                if (param->type == PGP_ARMORED_CLEARTEXT) {
```
ni4, line 691: Hm, despite approval, we should probably cover these lines with tests? And why this should fail here? Because of absence of Hash: header?

antonsviridenko: https://github.com/rnpgp/rnp/issues/1289#issuecomment-763972963

ni4: Thanks, got it. Anyway, could you please add some small test for this case to make codecov happy? It can be done by modifying `message.txt.cleartext-signed` and adding test to `cli_tests.py` or updating `test_ffi_op_verify_sig_count`.

antonsviridenko: looks line this condition is not needed here at all, cleartext signatures are parsed in other place, `cleartext_parse_headers()` in `stream-parse.cpp`

antonsviridenko: more precisely, `is_base_64_line()` can return false, as cleartext message can contain any characters outside of Base64 alphabet, so execution does not reach this condition 
`if (param->type == PGP_ARMORED_CLEARTEXT) {`

ni4: Agree, feel free to remove that condition.

Issue #1538, opened 2021-06-23
Print just RSA instead of RSA (Encrypt or Sign) in key listing.


ni4:
## Description

As this may be confusing, giving a false assumption that key is capable of signing and encryption. Please see the discussion here: https://github.com/rnpgp/rnp/pull/1537#issuecomment-866900339

ni4:
Fixed via #1749

PR #1749, opened 2022-01-31
Print 'RSA' as algorithm name in CLI instead of 'RSA (Encrypt and Sign).

ni4:
This PR:
- fixes issue #1538
- refactors CLI key/sig dumping functions to not use packet dump in JSON (faster!).


ronaldtse (approved):
Excellent change -- always found that confusing!

antonsviridenko (approved):

PR #1754, opened 2022-02-03
CLI: add option `--source` to specify input data for the detached signature verification.

### LINKED ISSUE #919: Add CLI option to specify source for the detached signature verification.
## Description

Currently detached signature verification works only when signature is read from the file and named `source.dat.sig`/`source.dat.asc`.
Probably we should add option to specify source and/or signature path (including stdin for source).

ni4:
Fixes #919


antonsviridenko (approved):

ronaldtse (approved):

ronaldtse:
Thank you @ni4 !

PR #1755, opened 2022-02-03
FFI: Add function rnp_key_is_expired().

### LINKED ISSUE #1532: FFI: add function rnp_key_is_expired().
## Description

While currently user may manually check key's creation and expiration time, it would be handy to have a shortcut function instead of 3 other calls + calculations.

ni4:
Fixes #1532

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -5742,7 +5742,7 @@ rnp_buffer_clear(void *ptr, size_t size)
 static pgp_key_t *
 get_key_require_public(rnp_key_handle_t handle)
 {
-    if (!handle->pub) {
+    if (!handle->pub && handle->sec) {
         pgp_key_request_ctx_t request;
         request.secret = false;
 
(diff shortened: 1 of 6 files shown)


src/lib/rnp.cpp
```diff
@@ -6899,6 +6899,21 @@ try {
...
 
+rnp_result_t
+rnp_key_is_expired(rnp_key_handle_t handle, bool *result)
+try {
+    if (!handle || !result) {
+        return RNP_ERROR_NULL_POINTER;
+    }
+    pgp_key_t *key = get_key_prefer_public(handle);
+    if (!key) {
+        return RNP_ERROR_BAD_PARAMETERS;
```
ronaldtse, line 6910: Codecov says this line is not covered by tests.

ni4: Hm, wondering how to make it covered... Will think about.

antonsviridenko (approved):

ronaldtse (approved):
@ni4 will let you consider the right time to merge, thanks!

ni4:
Okay, now Codecov is happy so merging this. Thanks!

PR #1757, opened 2022-02-08
Display key/sig validity status in key listing.

### LINKED ISSUE #1528: CLI: print if key/subkey is invalid when listing keys.
## Description

Currently `rnpkeys` CLI shows only information whether key is expired or revoked. It should be updated to display whether key is invalid, i.e. unusable due to some reasons (now self-sigs, invalid self-sigs, and so on).

Same should be done for signatures, if `--with-sigs` is specified, and for the userids.

ni4:
This PR:
- fixes #1528 by displaying whether uid/key/sig is valid
- prints information about the all signatures (binding, direct-key, etc), not just uid certifications
- adds FFI function `rnp_signature_get_key_fprint()`.

Should be merged only after the #1755

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -1387,6 +1387,16 @@ RNP_API rnp_result_t rnp_signature_get_expiration(rnp_signature_handle_t sig,
  */
 RNP_API rnp_result_t rnp_signature_get_keyid(rnp_signature_handle_t sig, char **result);
 
+/** Get signer's key fingerprint from the signature.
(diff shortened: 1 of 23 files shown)


antonsviridenko (changes requested):
This PR contains changes already introduced in #1755 

i.e.  `rnp_key_is_expired()` implementation

ni4:
> This PR contains changes already introduced in #1755
> 
> i.e. `rnp_key_is_expired()` implementation

Yeah, it would not go without that function. That's why it should be merged after PR #1755  is merged )

antonsviridenko (approved):

PR #1759, opened 2022-02-08
Use windows-2019 instead of windows-latest for the Github workflows.

ni4:
Windows-latest now uses Windows 2022 with corresponding Visual Studiom which is not handled well by CI scripts.


ronaldtse (approved):

ni4:
Merging, thanks @ronaldtse !

PR #1767, opened 2022-02-17
Calculate and store default s2k iterations in the SecurityContext

### LINKED ISSUE #1639: Calculate default iterations count only once for FFI calls.
## Description

If default iteration count is requested, FFI calculates it for each key/subkey protection call.
This may lead to the situation where key and subkey have different iterations count.
This is not an error, just a some cosmetic defect.

ni4:
Fixes #1639 and slightly refactors internal code.

### DIFF
--- a/src/librepgp/stream-write.cpp
+++ b/src/librepgp/stream-write.cpp
@@ -2016,8 +2016,14 @@ rnp_raw_encrypt_src(pgp_source_t &        src,
     handler.ctx = &ctx;
     pgp_dest_t encrypted = {};
 
-    rnp_result_t ret = rnp_ctx_add_encryption_password(
-      ctx, password.c_str(), DEFAULT_PGP_HASH_ALG, DEFAULT_PGP_SYMM_ALG, 0);
+    rnp_result_t ret = RNP_ERROR_GENERIC;
+    try {
+        ret =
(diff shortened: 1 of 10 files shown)


src/librepgp/stream-write.cpp
```diff
@@ -2016,8 +2016,14 @@ rnp_raw_encrypt_src(pgp_source_t &        src,
...
 
-    rnp_result_t ret = rnp_ctx_add_encryption_password(
-      ctx, password.c_str(), DEFAULT_PGP_HASH_ALG, DEFAULT_PGP_SYMM_ALG, 0);
+    rnp_result_t ret = RNP_ERROR_GENERIC;
+    try {
+        ret =
+          ctx.add_encryption_password(password, DEFAULT_PGP_HASH_ALG, DEFAULT_PGP_SYMM_ALG);
+    } catch (const std::exception &e) {
+        RNP_LOG("%s", e.what());
+        goto done;
```
ronaldtse, line 2025: Codecov says "Added lines #L2023 - L2025 were not covered by tests". Not sure how to test this though.

ni4: This code is not used anywhere, as far as I remember I wrote it for some testing purposes. So probably we could skip this....

ronaldtse: Thanks!

ronaldtse (approved):

PR #1772, opened 2022-02-18
Limit AEAD chunk size bits to 16

### LINKED ISSUE #1676: Limit AEAD chunk sizes to 16.
## Description

As previously larger chunk sizes were documented, since of [draft-ietf-openpgp-crypto-refresh-04](https://datatracker.ietf.org/doc/html/draft-ietf-openpgp-crypto-refresh-04#section-5.16) 16 is the maximum chunk size:

An implementation **MUST** accept chunk size octets with values from 0 to 16.  An implementation **MUST NOT** create data with a chunk size octet value larger than 16 (4 MiB chunks).

So we should update code (internal, FFI, and CLI) accordingly: do not generate messages with chunks larger then 16, but possibly still accept older messages which were generated with larger chunk size, if any.

ni4:
Fixes #1676 
Also slightly refactors CLI code and extends test coverage.

### DIFF
--- a/src/common/str-utils.cpp
+++ b/src/common/str-utils.cpp
@@ -111,6 +111,18 @@ lowercase(char *s)
     return s;
 }
 
+bool
+str_to_int(const std::string &s, int &val)
+{
+    for (const char &ch : s) {
+        if ((ch < '0') || (ch > '9')) {
+            return false;
+        }
+    }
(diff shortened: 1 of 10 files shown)


ronaldtse (approved):
Thanks @ni4 !

antonsviridenko (approved):

PR #1773, opened 2022-02-24
Set default key expiration to 2 years and refactor some internals

### LINKED ISSUE #1553: CLI: set some default expiration time for the generated keys
## Description

Currently generated keys do not expire until `--expiration` option is used.
Some meaningful default should be applied (like 2 years), also it would be good to ask for it in expert key generation mode.

ni4:
This PR:
- fixes #1553 
- refactors function `rnp_generate_key_json()` for less code redundancy.


ronaldtse (approved):

andrey-utkin (commented):
LGTM

PR #1776, opened 2022-03-12
Fix edge case with missing crc24 and 3-byte last block.

### LINKED ISSUE #1775: In some cases RNP fails to dearmor data without a checksum
## Description

This would happend if input binary data size is multiply of 3, so there is no `=` at the end of it.

ni4:
Fixes #1775


antonsviridenko (approved):

ronaldtse (approved):

ronaldtse:
Thank you @ni4 and stay safe!

Issue #1730, opened 2022-01-05
Add support for base64-encoded autocrypt keys.


ni4:
## Description

Currently we support import/export of such keys only in binary format, however it could be handy to import/export base64 keys in one call.

Function `rnp_import_keys()` should be updated to allow just base64-encoded data (which is used to transfer autocrypt keys), without OpenPGP headers. Maybe with additional flag (RNP_ALLOW_BASE64?).

`rnp_key_export_autocrypt()` should be updated as well.

PR #1778, opened 2022-03-15
Allow the import of base64-encoded autocrypt keys.

ni4:
This RP adresses partially issue #1730. Second half, which allows to export base64-encoded autocrypt keys, will be added as a separate PR.
Also it refactors function `rnp_import_keys()` and separates FFI tests to different files for faster compilation.


antonsviridenko (approved):

ronaldtse (approved):
Thank you @ni4 !

PR #1791, opened 2022-03-29
PGP_FINGERPRINT_HEX_SIZE should be 1+2x binary size, not 1+3x

andrey-utkin:


ni4 (approved):
Thanks, @andrey-utkin! Could you please rebase it on the latest master as it fixes msys/clang CI issue?

ni4:
@andrey-utkin Thanks! We may skip codecov CI failure, as it happened due to the coverage change caused by the randomized input to some of the CLI tests.

antonsviridenko:
Isn't this file `include/repgp/repgp_def.h` a part of public API? And therefore, by changing this definition, we change public ABI between minor release. 

I guess `3x` was intended for placing spaces between individual octets.

andrey-utkin:
Good point about it being a part of public ABI.
Depending on the branching strategy (which is [somewhat unclear](https://github.com/rnpgp/rnp/issues/1796) ) it may or may not be OK to merge this to master any time.

I'm open to the possibility that 3x is actually what we need.

ni4:
> Isn't this file `include/repgp/repgp_def.h` a part of public API? And therefore, by changing this definition, we change public ABI between minor releases.
> 
> I guess `3x` was intended for placing spaces between individual octets.

The only public headers are `include/rnp.h` and `include/rnp_err.h`, which expose FFI interface. All other headers are to be considerd internal, and may be changed in any way. Present naming is leftover from the (unsuccessfull) try to separate parts of the library for key/data processing, and should be refactored out at some point.

antonsviridenko:
ok, approving and merging then

antonsviridenko (approved):

Issue #1275, opened 2020-08-25
Support decryption with hidden recipients.


duckdalbe:
RNP reports that it cannot decrypt if no Key ID of the Public-Key Encrypted Session Key Packets matches a known private key.

It would be great if it instead could check for "hidden" recipients. In case the Key ID `0x00000000` is present, RNP should try to decrypt with all known private keys. See <https://tools.ietf.org/html/rfc4880#section-5.1> (last paragraph of that section).

Background: Currently Thunderbird 78 an later fails to decrypt messages where the user is a hidden recipient. Such ciphertexts could e.g. be created by Enigmail, or in general with GnuPG's `hidden-encrypt-to` config option.  Downstream issue: <https://thunderbird.topicbox.com/groups/e2ee/T7e31aaa38399d4b3-Mefc2734288f0dfa17d01b823/fails-to-decrypt-with-recipient-key-id-0x00000000-hidden-recipient>.

ni4:
Thanks for reporting, we'll add corresponding functionality once get to this issue.

andrey-utkin:
I'd be interested to look into this.
Seems clear how to write a test for it, and then the test will guide my investigation.

ni4:
@andrey-utkin Feel free to take it!
The only thing is that I'd prefer to make this feature configurablte via the SecurityContext, so user may enable or disable this.
Feel free to ask additional questions if you have.

ni4:
> I will proceed by investigating it and in parallel fitting the test into some existing test suite.
> Pythonic one seems to be well-suited for such things, is that right?

I'd prefer to have FFI API test as well, since people who use library (like Thunderbird) could use approach which differs from a CLI one. Also it would be good to have test for different cases, like alread-preloaded secret keys, keys loaded by demand via key provider, trying multiple secret keys and so on.

andrey-utkin:
I got it working with the clumsiest hack ever. Cleaning up and refactoring now, will submit into a branch later.
The meaning of these hooks becomes less certain with this extension of functionality in mind:
```
typedef void pgp_on_recipients_func_t(const std::vector<pgp_pk_sesskey_t> &recipients,
                                      const std::vector<pgp_sk_sesskey_t> &passwords,
                                      void *                               param);
typedef void pgp_decryption_start_func_t(pgp_pk_sesskey_t *pubenc,
                                         pgp_sk_sesskey_t *symenc,
                                         void *            param);
```
[...]

Issue #1329, opened 2020-10-16
runtime error: call to function ffi_pass_callback_file


ni4:
## Description
Following error appears on centos sanitize build:

```
/__w/rnp/rnp/src/lib/rnp.cpp:1080:7: runtime error: call to function ffi_pass_callback_file(rnp_ffi_st*, void*, rnp_key_handle_st*, char const*, char*, unsigned long) through pointer to incorrect function type 'bool (*)(rnp_ffi_st *, void *, rnp_key_handle_st *, const char *, char *, unsigned long)'
2020-10-16T16:39:09.8868329Z 201: /__w/rnp/rnp/src/rnp/fficli.cpp:315: note: ffi_pass_callback_file(rnp_ffi_st*, void*, rnp_key_handle_st*, char const*, char*, unsigned long) defined here
```

ni4:
Strange since `ffi_pass_callback_file` is defined correctly:
```
static bool
ffi_pass_callback_file(rnp_ffi_t        ffi,
                       void *           app_ctx,
                       rnp_key_handle_t key,
                       const char *     pgp_context,
                       char             buf[],
                       size_t           buf_len)
```

ni4:
This could be false positive, as the issue with stdout_writer, see PR #1709. Possible solution:
```
#if defined(__clang__)
__attribute__((no_sanitize("undefined")))
#endif
```

andrey-utkin:
It seems like a trait of outdated Clang shipped with CentOS 7.

In PR 1792, CI checks on centos-based systems with "mode sanitize" and clang:

* centos7 / centos:7 Botan [mode sanitize; test type ; CC clang; GnuPG stabl... - FAIL
* centos7 / centos:7 Botan [mode sanitize; test type ; CC clang; GnuPG beta] - FAIL
* centos8-openssl / centos:stream8 OpenSSL [test type rnp_tests; mode sanitiz... CC clang... - PASS
* centos8 / centos:stream8 Botan [test type ; mode sanitize; CC clang; GnuPG stable]  - PASS
* centos8-openssl / centos:stream8 OpenSSL [test type cli_tests; mode sanitiz... CC clang... - PASS
* centos8 / centos:stream8 Botan [test type ; mode sanitize; CC clang; GnuPG - PASS

So I'd suggest to disregard the combination centos7 + clang + sanitize.

ni4:
[...]

Issue #1158, opened 2020-06-02
Update set_rnp_log_switch() to allow logging fd.


ni4:
## Description

Currently we are able to disable/enable logging. But it would be useful to be able to specify fd for logging output. That would allow to easily redirect log to file/pipe, and utilize this in tests and applications which use library.

rrrooommmaaa:
I can do this

ni4:
@rrrooommmaaa Thanks. But it is not of high priority right now.

dewyatt:
This is already available via `rnp_ffi_set_log_fd` for functions that use `FFI_LOG`, right? That was the original idea there, we just didn't update everything to use `FFI_LOG` yet.

ni4:
@dewyatt `FFI_LOG()` requires `rnp_ffi_t` instance, while `RNP_LOG` works on lower layers, where it will be unavailable. However, `rnp_ffi_set_log_fd()` could also set lower-layer global variable, what do you think about such approach?

dewyatt:
> @dewyatt `FFI_LOG()` requires `rnp_ffi_t` instance, while `RNP_LOG` works on lower layers, where it will be unavailable. However, `rnp_ffi_set_log_fd()` could also set lower-layer global variable, what do you think about such approach?

I lost track of this for a bit, but my thoughts are that it's probably not the best approach. Some appropriate logging context needs to be passed everywhere that logging is going to be done, even if seemingly burdensome. It's not OK to use a global or thread-local, it's only going to hinder debugging later on when trying to make sense of a log that has interleaved messages from multiple unrelated contexts within a given application (and we probably end up with a global mutex to synchronize that).

bqv:
Much better would be just a function pointer to call for logging

ni4:
@bqv Thanks for the note, definitely makes sense. To not break current workflow probably we should have both, adding some logging context as @dewyatt suggested.

PR #1792, opened 2022-03-29
[#1275] support decrypting messages with hidden recipients

andrey-utkin:


andrey-utkin:
* added FFI tests
* cleaned up the history and the code
* rebased on master

One thing I'm unsure about is how to go to further purify the runtime logs.
E.g. `elgamal_decrypt_pkcs1()` itself emits an error message straight to stderr. It is technically simple to just pass further the `bool speculative` function argument to switch that logging off, but this feels unnecessarily complicating the interface.
I'd appreciate any ideas of more subtle way to control emission of these error messages.

Oh and I still haven't implemented a SecurityContext switch to control this new feature.

andrey-utkin:
What I don't like is the name of a new subroutine I've created, `key_fetch_and_try()`. Name doesn't convey the meaning well, and there's already a similarly-named `encrypted_try_key()` which does a related bit of work.
Suggestions for better names or a different approach to factoring out that bit of code are welcome.

antonsviridenko:
LGTM so far, waiting for tests to be fixed

src/librepgp/stream-parse.cpp
```diff
@@ -2084,48 +2154,86 @@ init_encrypted_src(pgp_parse_handler_t *handler, pgp_source_t *src, pgp_source_t
...
+                 * or "speculative" Key ID.  In this case, the receiving implementation
+                 * would try all available private keys, checking for a valid decrypted
+                 * session key.  This format helps reduce traffic analysis of messages. */
+                rnp_identifier_iterator_t it = NULL;
+                rnp_key_handle_t          handle = NULL;
+                const char *              identifier = NULL;
+
+                /* Iterate through the keys */
+                rnp_ffi_t ffi = (rnp_ffi_t) handler->key_provider->userdata;
+                if (rnp_identifier_iterator_create(ffi, &it, "keyid")) {
```
ni4, line 2173: It would be faster to use fingerprint iterator here and in `rnp_locate_key()`, as in case of preloated keyring keys are stored in the map using the fingerprint as a key.

src/tests/ffi.cpp
```diff
@@ -5820,3 +5820,85 @@ TEST_F(rnp_tests, test_ffi_security_profile)
...
+{
+    // Secret keyrings can have many keys.
+    // When decrypting a message to a hidden recipient, each key has to be tried.
+    // There shouldn't be any error messages printed about decryption failures caused by the
+    // failed attempts.
+    rnp_ffi_t ffi = NULL;
+    assert_rnp_success(rnp_ffi_create(&ffi, "GPG", "GPG"));
+
+    // Log fd handing inspired by test_ffi_set_log_fd.
+    int log_fd = rnp_open("decrypt_hidden_to.log", O_RDWR | O_CREAT | O_TRUNC, 0777);
```
ni4, line 5869: It would be easier to use CLI/Python tests to check logging output, like this one:
```
        ret, _, err = run_proc(RNPK, ['--homedir', RNPDIR, '--import-keys', data_path('test_stream_key_load/ecc-25519-pub-4.b64')])
        self.assertEqual(ret, 1)
        self.assertRegex(err, r'(?s)^.*wrong base64 padding: ==zz.*Failed to init/check dearmor.*failed to import key\(s\) from .*, stopping.$')
```

ni4:
> What I don't like is the name of a new subroutine I've created, `key_fetch_and_try()`. Name doesn't convey the meaning well, and there's already a similarly-named `encrypted_try_key()` which does a related bit of work.
> Suggestions for better names or a different approach to factoring out that bit of code are welcome.

I added few PR comments, but will review it more carefully later, as it has a bunch of changes. Also need some time to think on the questions you asked )

src/librepgp/stream-parse.cpp
```diff
@@ -2084,48 +2154,86 @@ init_encrypted_src(pgp_parse_handler_t *handler, pgp_source_t *src, pgp_source_t
...
+              0,
+            };
+            if (pubenc.key_id == zeros) {
+                /* RFC 4880, 5.1:
+                 * An implementation MAY accept or use a Key ID of zero as a "wild card"
+                 * or "speculative" Key ID.  In this case, the receiving implementation
+                 * would try all available private keys, checking for a valid decrypted
+                 * session key.  This format helps reduce traffic analysis of messages. */
+                rnp_identifier_iterator_t it = NULL;
+                rnp_key_handle_t          handle = NULL;
```
ni4, line 2168: These sources must not have access to the FFI API, as are of lower layer. I.e. everythong from `include/rnp.h` and `lib/rnp.cpp` should not be accessed from other headers/cpp files (except tests and CLI code which is built on top of the FFI).
This check could be moved to the `lib/rnp.cpp`, and it should iterate via keys there.

Also code should work in two scenarios:
- secret keys are preloaded to the keyring
- keys are loaded by demand via the key provider (then it would be key provider's function to correctly handle request for zero-keyid secret key).

andrey-utkin:
Thanks for the comments Nickolay.

Meanwhile I was reviewing codecov CI check results.
The lines which are not covered now are error handling paths and the new disabled test.

An example of a line missing coverage:
```
@@ -1483,7 +1483,7 @@ encrypted_try_key(pgp_source_encrypted_param_t *param,
         pgp_fingerprint_t fingerprint;
         if (pgp_fingerprint(fingerprint, *seckey)) {
             if (!speculative) {
                 RNP_LOG("ECDH fingerprint calculation failed");  // <=====
             }
             return false;
         }
```
RNP_LOG() was there, what changed is that now it's conditional upon `!speculative`.
What this means is that we don't have any tests where this message is emitted and where that message actually is desirable!
But this is the kind of error log message for a condition which is not intended to happen normally, and it may be uneconomical to write a test just to target it. But if we want it, I guess the easiest way to do it is to build failure injection into the codebase itself. What are your thoughts on that?

ni4:
> But this is the kind of error log message for a condition which is not intended to happen normally, and it may be uneconomical to write a test just to target it. But if we want it, I guess the easiest way to do it is to build failure injection into the codebase itself. What are your thoughts on that?

@andrey-utkin you are right, most of these are left to have some diagnostic input in case when something unusal happens, so such codecov CI failures may be skipped. Things which must be covered should be triggered by external data or system configuration.

src/librepgp/stream-parse.cpp
```diff
@@ -1438,7 +1442,9 @@ encrypted_try_key(pgp_source_encrypted_param_t *param,
         err = rsa_decrypt_pkcs1(
           &ctx.rng, decbuf.data(), &declen, &encmaterial.rsa, &keymaterial->rsa);
         if (err) {
-            RNP_LOG("RSA decryption failure");
+            if (!speculative) {
```
ni4, line 1445: What do you think on replacing all of these `if (!speculative)` with some define, like `RNP_LOG_IF(speculative, blablabla)`?
Also, as security context is passed everywhere, we may add property `rnp::SecurityContext.no_log()` instead of passing `speculative` parameter to every function call.

src/librepgp/stream-parse.cpp
```diff
@@ -2084,48 +2154,86 @@ init_encrypted_src(pgp_parse_handler_t *handler, pgp_source_t *src, pgp_source_t
...
+                    if (rnp::hex_decode(identifier,
+                                        pubenc.key_id.data(),
+                                        pubenc.key_id.size()) != PGP_KEY_ID_SIZE) {
+                        assert(false);
+                        RNP_LOG("Internal error: failed to parse keyid %s", identifier);
+                        continue;
+                    }
+
+                    errcode = key_fetch_and_try(handler, param, pubenc, /*speculative=*/true);
+                    if (errcode == RNP_SUCCESS) {
```
ni4, line 2211: At some point we sticked to the shorter format - `if (!errcode) {...`

andrey-utkin: ok

src/librepgp/stream-parse.cpp
andrey-utkin: > What do you think on replacing all of these if (!speculative) with some define, like RNP_LOG_IF(speculative, blablabla)?

Seems too big a response for such a local need.

Would it work if all these "decryption failed" messages emitted *by the library* could be considered debug-level messages and thus be suppressed by default? I would expect that if the error condition bubbles up to the application level, the application could show a message itself. The downside is that some details of "why" may be lost (to mitigate which we'd need to extend the enum of status codes), but indiscriminate logging of low-level intermediate unsuccessful outcomes is not a good UX either.

This would achieve us the following:
* no need to pass `bool speculative` down the stack
* don't need to remove or change much in the way the library currently emits stuff to stderr
* could possibly pave the way to the adoption of a conventional policy of "no printing by libraries"

andrey-utkin:
Also the `rnp-ruby` test suite failure seems unrelated, maybe flakiness?

```
  1) #<Method: Rnp.s2k_iterations(hash:, msec:) /home/runner/work/rnp/rnp/installs/ruby-rnp/lib/rnp/misc.rb:196> returns a higher iterations count for MD5 vs SM3
     Failure/Error:
       expect(
         Rnp.s2k_iterations(hash: 'MD5', msec: MSEC)
       ).to be > Rnp.s2k_iterations(hash: 'SM3', msec: MSEC)

       expected: > 425984
            got:   163840
     # ./spec/misc_spec.rb:160:in `block (2 levels) in <top (required)>'

Finished in 17.82 seconds (files took 0.42674 seconds to load)
247 examples, 1 failure
```

andrey-utkin:
This "stderr clean" requirement doesn't come easy or fast.
It would be good to *just not print* from the libraries, but just wiping logging statements from the codebase feels barbaric.

src/librepgp/stream-parse.cpp
ni4: > Seems too big a response for such a local need. 

If rnp::SecurityContext.no_log() is an overkill, RNP_LOG_IF() would make code more readable, give less visual complexity. Also it would be easier to introduce changes in one place (for instance if we would like to add logging levels, whatever else).

> Would it work if all these "decryption failed" messages emitted by the library could be considered debug-level messages and thus be suppressed by default? 

Most of this logging is done to help to debug unexpected edge cases/configurations. I.e. we have problem which is already known then we'd report it to the caller, but in some cases caller need more information to understand what went wrong. 

However, we should not allow unlimited logging output as well - say, for instance, caller implemented something incorrectly and calls code which asks for the key in endless cycle. If log is redirected to the file then it will fill up the whole hard drive.

> could possibly pave the way to the adoption of a conventional policy of "no printing by libraries" 

User should be able to decide what to do with logs. We have issue #1158 related to this.

ni4:
> Here's some strange CI failure. It says type mismatch (`char const*` vs `const char *`?), but I don't see any in the sources. 

Yeah, most likely it's some compiler issue or something like that. I filed issue #1329 for this some time ago, but don't have a solution yet.

ni4:
> Also the `rnp-ruby` test suite failure seems unrelated, maybe flakiness?
> 
>   `1) #<Method: Rnp.s2k_iterations(hash:, msec:) /home/runner/work/rnp/rnp/installs/ruby-rnp/lib/rnp/misc.rb:196> returns a higher iterations count for MD5 vs SM3`

Yeah, this is due to timing issue, and happens sometimes. Just feel free to restart failed run in such case.

src/librepgp/stream-parse.cpp
andrey-utkin: Thanks for hinting at #1158. I totally agree with the argument from here that "Some appropriate logging context needs to be passed everywhere that logging is going to be done".

Regarding `SecurityContext.no_log()` and `RNP_LOG_IF()` i'd like to highlight that crypto backend functions such as `rsa_decrypt_pkcs1()` don't have neither `ffi` nor `SecurityContext` passed, yet OpenSSL-ish module logs into stderr via `RNP_LOG()`.

So we need to either pass that some context across the crypto backend boundary, or stop logs being emitted (by default; i think a special env var to enable is acceptable).

I am in favour of disabling, but I admit I have no insight into how bug reports get to us, how many, and how much back and forth with the reporters we can afford, and such.

ni4: > Regarding `SecurityContext.no_log()` and `RNP_LOG_IF()` i'd like to highlight that crypto backend functions such as `rsa_decrypt_pkcs1()` don't have neither `ffi` nor `SecurityContext` passed, yet OpenSSL-ish module logs into stderr via `RNP_LOG()`. (Botan's module doesn't though, making it not problematic!)

Yeah, those redundant logging calls are added to be ready for the unexpected cases, as OpenSSL backend is comparable new and could be covered not that good as Botan. `SecurityContext` sooner or later will be passed to those functions, once we add possibility to mark certain asym algorithms and key sizes as weak, like this is done for hashes.

> 
> So we need to either pass that some context across the crypto backend boundary, or stop logs being emitted (by default; i think a special env var to enable is acceptable).

We already have `RNP_LOG_CONSOLE` which would disable logging at all. However, in this particular case, we need to disable logging from the certain function set, not from the whole library.

> I am in favour of disabling, but I admit I have no insight into how bug reports get to us, how many, and how much back and forth with the reporters we can afford, and such.

All reports (except security-sensitive ones) are created/added as Github issues here.

andrey-utkin: > in this particular case, we need to disable logging from the certain function set, not from the whole library.

Could you please tell me which approaches to that would have your acceptance? Would make me happy if there are some easily fitting the scope of this task and PR.

After I wrote the above, I looked more closely into...

> We already have RNP_LOG_CONSOLE

And that was interesting. Changelog says

> Disabled logging by default in release builds and added support for environment variable `RNP_LOG_CONSOLE` to enable it back.

But that's not what I see. I've made a local release build, but the message from rsa_ossl.cpp is still being emitted. Is that a bug, or requirements changed since the changelog entry?

ni4: @andrey-utkin Sorry, somehow forgot to reply here timely.
1. What do you think about the following solution to disable logging: https://github.com/rnpgp/rnp/pull/1799 ?
2. Hm, as from `logging.cpp`, for the release build (if NDEBUG is not defined) it should be set only if thereis non-zero RNP_LOG_CONSOLE enivornment variable.

src/tests/cli_tests.py
```diff
@@ -636,6 +652,32 @@ def gpg_to_rnp_encryption(filesize, cipher=None, z=None):
...
+            expect_ok = recipient != 'unknown@rnp'
+            # Encrypt cleartext file with GPG
+            gpg_encrypt_file_hidden_to(src, dst, recipient, cipher, z, armor)
+            # Decrypt encrypted file with RNP
+            out, err = rnp_decrypt_file(dst, dec, PASSWORD, expect_ok)
+            if expect_ok:
+                compare_files(src, dec, RNP_DATA_DIFFERS)
+                self.assertEqual(out, '', 'Unexpected stdout contents')
+                err_lines = err.split('\n')
+                err_lines = list(filter(lambda x: x and "premature end of armored input" not in x, err_lines))
```
ni4, line 674: What about adding additional filter for CentOS 7 false-positive? That would allow CI to run.

andrey-utkin: Centos 7 UBSAN false-positive? I think it fails ruby-rnp suite at very least, so any filtering there doesn't change much.

ni4: Errm, don't see any ruby-rnp failures here in this PR, or do I misunderstand something? There are occassional ruby-rnp failures in s2k tuning, which involve time comparision.

andrey-utkin:
@ni4 @antonsviridenko I believe this is finally ready for merging.
May I ask you guys to have a thorough final review soonish?
Also, as this is my first work on this codebase, I would highly appreciate any feedback from you about how is it to work with me - here or in private email.

ni4:
@andrey-utkin Thanks, I'll review this later today. Usually we discuss issues here on Github, however if there are some questions which is handier to discuss via chat - feel free to drop a message in Skype.

antonsviridenko (approved):

src/lib/logging.cpp
```diff
@@ -39,6 +39,9 @@ static int8_t _rnp_log_switch =
 #endif
   ;
 
+/* Temporary disable logging */
+static size_t _rnp_log_disable = 0;
+
```
ni4, line 44: Could you please rebase on master as this code is already there?

src/librepgp/stream-parse.cpp
```diff
@@ -46,6 +48,7 @@
 #include "crypto/signatures.h"
 #include "fingerprint.h"
 #include "pgp-key.h"
+#include "ffi-priv-types.h"
```
ni4, line 51: We should not access this and `<rnp/rnp.h>` headers here, probably it was forgotten to be removed.

src/librepgp/stream-parse.cpp
ni4: @andrey-utkin Could you please address this comment?

ni4 (changes requested):
One major comment was not addressed - we cannot use FFI interface on the lower layer in `stream-parse.cpp`. To request a key we should handle zero keyid key request on the `lib/rnp.cpp` layer's crypto provider.

andrey-utkin:
> One major comment was not addressed - we cannot use FFI interface on the lower layer in `stream-parse.cpp`. To request a key we should handle zero keyid key request on the `lib/rnp.cpp` layer's crypto provider.

Thanks for the review.
Sorry, indeed I forgot about it.
I will revise the code in question.

andrey-utkin:
> One major comment was not addressed - we cannot use FFI interface on the lower layer in `stream-parse.cpp`. To request a key we should handle zero keyid key request on the `lib/rnp.cpp` layer's crypto provider.

So I've prototyped the way to have this try-all-keys work under search.
See branch autkin-1275-decrypt-hidden-to-under-search .
The case of any other search is knowingly broken right now, can be fixed.
The actual decryption to hidden recipient somehow doesn't work too, but probably due to UB or a problem in my code.
I am sharing this early now for discussion - whether I understood the suggestion correctly, does this approach still look right, etc
One problem which I stumbled upon is that during search the key objects we look at are `const`, and trying the key requires it to be non-const. When I just used copy-constructor of pgp_key_pkt_t to get a mutable object, my test emitted `src/librepgp/stream-parse.cpp:1354] checksum check failed` which I wasn't expecting, so I guess I am not supposed to just copy pgp_key_pkt_t.

ni4:
@andrey-utkin I'd suggest another approach, making key provider responsible to check for zero keyid:
- in `key_fetch_and_try()` just call a key provider, without checking for zero keyid.
- in `lib/rnp.cpp`, `ffi_key_provider` check whether zero keyid requested. On first call return a first secret key and store somewhere in userdata which key was returned. `rnp_identifier_iterator_t` could be useful here.
- back in `key_fetch_and_try()` attempt to start decryption. If it failed, and we have zero keyid - call key provider again.
- iterate it until we succeed to start decryption or key provider doesn't return any key.

Does this sound reasonable, or I missed some possible issues?

andrey-utkin:
So on the latest approach, there is a couple of complications I've noticed. Not that they are necessarily deal breakers.
1. Since we start putting the iteration position into `rnp_ffi_t`, we need a way to reset that position when the loop is completed, because I suspect `rnp_ffi_t` lifecycle is longer than that loop.
2. `find_key(try_key_provider=true)` tries `rnp_key_store_search()` and then resorts to `ffi->getkeycb()` (which can lead to more keys inserted into the key store) and then tries `rnp_key_store_search()` again. I suspect the newly inserted keys are not strictly appended so the old "last position" state is meaningless at that point. I see a few options there:
   (a) run `ffi->getkeycb()` first in such case and remember that we've done so;
   (b) remember not just the last position, but entire list of the keys already returned.

andrey-utkin:
WIP pushed into autkin-1275-decrypt-hidden-to-r3 branch. New tests pass, but there seems to be at least some memory corruption bug which leads to some tests hanging.

ni4:
> 1. Since we start putting the iteration position into `rnp_ffi_t`, we need a way to reset that position when the loop is completed, because I suspect `rnp_ffi_t` lifespan is longer than that loop.

Actually, in this case it would be preferred to add some wrapping struct, which includes link to `rnp_ffi_t` and iterator, and pass it via the `userdata`.

> 2. `find_key(try_key_provider=true)` tries `rnp_key_store_search()` and then resorts to `ffi->getkeycb()` (which can lead to more keys inserted into the key store) and then calls itself which results in `rnp_key_store_search()` again. I suspect the newly inserted keys are not strictly appended so the old "last position" state is meaningless at that point. I see a few options there:
>    (a) run `ffi->getkeycb()` first in such case and remember that we've done so;
>    (b) remember not just the last position, but entire list of the keys already returned.

If there is some custom key provider assigned then it should be task of the key provider to react on zero keyid. We'll just need to mention this in the `rnp_ffi_set_key_provider` documentation.

ni4:
@andrey-utkin Sorry for a long delay with a reply, needed to sort things out. The main problem here is that key_provider's `void *userdata` is stored in `rnp_ffi_st`. But we don't require to do it this way. Adding some: 
```
struct rnp_key_provider_param {
     rnp_ffi_t ffi;
     bool zero_keyid;
     rnp_identifier_iterator_t zero_it;
}
```
And handling it in `ffi_key_provider`, creating/destroying in `rnp_op_verify_execute/rnp_op_verify_destroy` (and saving to the `rnp_op_verify_t` field) should effectively fix all questions rised.
Also, as from the code, please note that zero-keyid is used only for the decryption, i.e. you should only look for the secret keys.

In case user sets he's own key provider via the `rnp_ffi_set_key_provider()` he should handle zero-keyid case in his own logic.

ronaldtse:
@andrey-utkin I know you're eager to get this merged -- does the above comment by @ni4 make sense?

ronaldtse:
Thanks @ni4 @andrey-utkin !

andrey-utkin:
Thanks @ronaldtse  @ni4 for your support and patience :)

ni4:
Closing this as it was superceded by #1940

PR #1803, opened 2022-04-13
Drop CI runs on centos7 with clang in sanitize mode

andrey-utkin:
Due to false positives.

See #1329.

### DIFF
--- a/.github/workflows/centos7.yml
+++ b/.github/workflows/centos7.yml
@@ -73,18 +73,7 @@ jobs:
             CC: clang
             CXX: clang++
             IMAGE: centos:7
-# Sanitize is only been tested with the clang compiler
-# see env-common.inc.sh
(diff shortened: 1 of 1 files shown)


ni4:
@andrey-utkin as centos7 has EOL 2024 (see https://github.com/rnpgp/rnp/issues/1329#issuecomment-1098309815 ), should we still remove those sanitize runs? cc @ronaldtse

ni4:
@ronaldtse If I understand correctly, RNP didn't get into CentOS 7 packages, and the only way to build it from sources? I.e. we'll just remove CI runs but will not drop CentOS 7 support, as we don't have anything in codebase/dependencies which doesn't build there.

ni4:
Let's close this as there is no major issues with Centos 7 CI Runs. We'll be able to do that later in a separate PR once needed.

PR #1815, opened 2022-04-29
Add test case for unencrypted EdDSA secret key with leading zeroes.

### LINKED ISSUE #1812: Check whether RNP correctly handles cleartext EdDSA secret keys with leading zeroes.
## Description

More details and sample key are available in this GnuPG ticket: https://dev.gnupg.org/T5120
Even if everything works we should add test case to not allow regression in the future.

ni4:
It worked well, but still it's good to cover it.

Fixes #1812


antonsviridenko (approved):

andrey-utkin (approved):
LGTM.

Codecov is crazy, that report doesn't make sense. It claims to have compared it to a baseline commit which makes sense, but apparently it has really compared it with something different.

ni4:
> Codecov is crazy, that report doesn't make sense. It claims to have compared it to a baseline commit which makes sense, but apparently it has really compared it with something different.

I thought so, but actually it isn't. See the PR #1817 - in CLI tests we use some random-generated input, which sometimes was hitting AEAD edge cases, and sometimes not. Will rebase as that PR is already merged.

ronaldtse (approved):

ronaldtse:
@ni4 we're waiting for a rebase here, right?

ni4:
@ronaldtse Actually not, but now I'm wondering why MSVC test fails, needs recheck so we don't miss something.

ni4:
Okay, let's try to restart and see whether that helps. Don't see where it may crash/seems crash happens after the test's run.

ni4:
Now it succeeded, so let's merge.

PR #1830, opened 2022-05-06
Fix typo --hom in test_rnpkeys_lists

jirutka:


antonsviridenko (approved):

ronaldtse (approved):
Thank you @jirutka -- what a great find 🤦 👍

ni4:
@jirutka Thanks! Actually, it worked with `--hom` as `getopt_long` allow partial command names.

jirutka:
> it worked with `--hom` as `getopt_long` allow partial command names.

No, it did not, that’s how I found it. That’s glibc-specific behaviour, `getopt_long` on musl libc doesn’t allow partial command names.

ni4:
Didn't know about the musl, thanks. Probably we should add some CI runner on top of musc libc.

PR #1834, opened 2022-05-10
[WIP] 1275 decrypt hidden to r3

andrey-utkin:
This is just to get CI runs for now.


ronaldtse:
@andrey-utkin the only test failing here is clang-format. Can you just run clang-format and clear the errors? Thanks.

PR #1844, opened 2022-05-26
Add FFI API and CLI options to override current time.

ni4:
This PR changes code to redirect all time request calls into the security context, allowing to override the current time. This would be useful for debugging and for test artifact creation.

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -551,6 +551,20 @@ RNP_API rnp_result_t rnp_request_password(rnp_ffi_t        ffi,
                                           const char *     context,
                                           char **          password);
 
+/**
+ * @brief Set timestamp, used in all operations instead of system's time. These operations
+ *        include key/signature generation (this timestamp will be used as signature/key
+ *        creation date), verification of the keys and signatures (this timestamp will be used
+ *        as 'current' time).
+ *        Please note, that exactly this timestamp will be used during the whole ffi lifetime.
+ *
+ * @param ffi initialized FFI structure
+ * @param time non-zero timestamp to be used. Zero value restores original behaviour and uses
+ *             system's time.
+ * @return RNP_SUCCESS or other value on error.
+ */
+RNP_API rnp_result_t rnp_set_timestamp(rnp_ffi_t ffi, uint64_t time);
+
 /** load keys
  *
  * Note that for G10, the input must be a directory (which must already exist).
--- a/src/lib/sec_profile.cpp
+++ b/src/lib/sec_profile.cpp
@@ -27,6 +27,7 @@
 #include "sec_profile.hpp"
 #include "types.h"
 #include "defaults.h"
+#include <ctime>
 #include <algorithm>
 
 namespace rnp {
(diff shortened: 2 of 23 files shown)


andrey-utkin (commented):
If deterministic testing is the whole point?
My first thought would be using https://github.com/wolfcw/libfaketime . Supposed to work (on Linux and macOS) without the need to pass the overriding timestamp throughout the codebase, which turned out to be a significant work.

As I see the timestamp passed around is geared towards replacing the calls to `time()` with a second-precision integer value, but `clock_gettime()` interface is generally widespread in software, and uses `struct timespec` so the internal mechanism might need to evolve in future.
Another, purely theoretical, risk is that some dependencies (I'm primarily concerned about botan and openssl) could have or obtain in future some compellingly smart functions which happen to query the timestamp directly from the system while not providing an interface to override it.

What you think?

ni4:
> If deterministic testing is the whole point?
My first thought would be using https://github.com/wolfcw/libfaketime . Supposed to work (on Linux and macOS) without the need to pass the overriding timestamp throughout the codebase, which turned out to be a significant work.

It's not just for testing, it's to be able to generate specific data files (like key created 2 years ago and expired) and so on. Also #1446 could be related. Adding additional dependency is quite an overhead for this.

> As I see the timestamp passed around is geared towards replacing the calls to time() with a second-precision integer value, but clock_gettime() interface is generally widespread in software, and uses struct timespec so the internal mechanism might need to evolve in future.

We don't need this (including cross-platform compatiblity overhead) as all timestamps in OpenPGP protocol have 1-second precision.

> Another, purely theoretical, risk is that some dependencies (I'm primarily concerned about botan and openssl) could have or obtain in future some compellingly smart functions which happen to query the timestamp directly from the system while not providing an interface to override it.

Crypto backends here are used just for the data processing, nothing related to the system date.

src/lib/sec_profile.cpp
```diff
@@ -174,4 +175,16 @@ SecurityContext::s2k_iterations(pgp_hash_alg_t halg)
...
+void
+SecurityContext::set_time(uint64_t time) noexcept
+{
+    time_ = time;
+}
+
+uint64_t
+SecurityContext::time() const noexcept
+{
+    return time_ ? time_ : ::time(NULL);
```
andrey-utkin, line 187: I am sure down the line somebody will want deterministic 1970-01-01 00:00:00 and will be frustrated :) Not that I'm worried too much about that, but maybe an extra "time-is-set" flag would keep some more people happy.

andrey-utkin (approved):

include/rnp/rnp.h
```diff
@@ -551,6 +551,17 @@ RNP_API rnp_result_t rnp_request_password(rnp_ffi_t        ffi,
                                           const char *     context,
                                           char **          password);
 
+/**
+ * @brief Set timestamp, used in all operations instead of system's time.
```
antonsviridenko, line 555: It would be nice to have more detailed description where this timestamp can be used, like new signatures, generated key timestamps, etc.

antonsviridenko (approved):

src/lib/sec_profile.cpp
ni4: @andrey-utkin We need to have a way to reset timestamp to the normal behaviour, so I used 0 value for this. Another way would be to add another function `rnp_unset_timestamp()`, but isn't that an overhead?

include/rnp/rnp.h
ni4: Thanks, updated.

src/lib/sec_profile.cpp
ronaldtse: @andrey-utkin any thoughts here? Thanks!

ronaldtse:
This PR has been approved and just pending resolution of some remaining comments.

PR #1846, opened 2022-05-31
Use factory-based approach for the rnp::Hash.

### LINKED ISSUE #1780: Use factory approach to create hash objects.
## Description

Currently we use borrowed from C behaviour, which causes following problems:
- more complicated code because of SHA1CD usage
- false alerts from the Coverity, caused by the previous one
- false alerts from memory sanitizers

This should be reworked to use something like `std::unique_ptr<rnp::Hash> Hash::Create(algorithm)`.

ni4:
This PR resolves problems of the old, C-struct based approach.
First commit is quite huge and should not happen, however I don't see any good way to do it via some smaller incremental commits.

Fixes #1780


ronaldtse (approved):
LGTM, thanks @ni4 !

antonsviridenko:
[src/lib/crypto/sha1cd/hash_sha1cd.h → src/lib/crypto/hash_crc24.hpp](https://github.com/rnpgp/rnp/pull/1846/files#diff-57ceef065deee750750a6bc2bab8957d2464313d539fa9083d3837318ec02a09)

How does this file rename make sense? SHA1 related header is replaced by something related to CRC24.

ni4:
@antonsviridenko It appears to be an issue of git's log command and it's algorithm for similarity detection, according to this answer: https://stackoverflow.com/a/15031787/452752

antonsviridenko (approved):

PR #1848, opened 2022-06-03
Refactor C path composition functions

ni4:
...and use C++ types/calls instead.


ni4:
We may ignore failing codecov this time again as it should be caused by less overall codelines/uncovered memory allocation failures.

ronaldtse (approved):

ronaldtse:
Thanks @ni4 !

antonsviridenko (approved):

PR #1851, opened 2022-06-09
Fix CID 1489624: uninitialized variable.

ni4:
As title says.


andrey-utkin (approved):
The issue identified and the fix are correct.

~Where CID 1489624 is from?~ It's from Coverity as Nickolay explained https://scan.coverity.com/projects/rnpgp-rnp

antonsviridenko (approved):

ronaldtse (approved):
Thank you @ni4 !

Issue #1612, opened 2021-08-22
FFI: add option to export armored revocation to rnp_key_export_revocation


ni4:
## Description

Currently it exports only binary revocation, however it would be handier to have armored flags.
Once implemented it should be used in `cli_rnp_export_revocation()`.

ni4:
Fixed via #1855

PR #1855, opened 2022-06-10
[#1612] FFI: add option to export armored revocation

andrey-utkin:
This addresses #1612.

No extra tests introduced so far, because this is a low-risk change and there are already existing tests for both
- FFI used to generate non-armored revocation signatures;
- CLI used to generate armored ones.
I guess we could introduce tests which compare the results with reference results, because currently the FFI tests discard the output without any checks due to `rnp_output_to_null()`.

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -4058,6 +4058,8 @@ try {
     if (!key || !key->ffi || !output) {
         return RNP_ERROR_NULL_POINTER;
     }
+    bool need_armor = flags & RNP_KEY_EXPORT_ARMORED;
+    flags &= ~RNP_KEY_EXPORT_ARMORED;
     if (flags) {
         return RNP_ERROR_BAD_PARAMETERS;
     }
@@ -4079,9 +4081,16 @@ try {
         return ret;
     }
 
-    sig.write(output->dst);
-    ret = output->dst.werr;
-    dst_flush(&output->dst);
+    if (need_armor) {
+        rnp::ArmoredDest armor(output->dst, PGP_ARMORED_PUBLIC_KEY);
+        sig.write(armor.dst());
+        ret = armor.werr();
+        dst_flush(&armor.dst());
+    } else {
+        sig.write(output->dst);
+        ret = output->dst.werr;
+        dst_flush(&output->dst);
+    }
     output->keep = !ret;
     return ret;
 }
--- a/src/tests/cli_tests.py
+++ b/src/tests/cli_tests.py
@@ -167,8 +167,9 @@ r'Secret subkey packet.*' \
 r'secret key material:.*' \
 r'encrypted secret key data:.*$'
 
-RE_RNP_REVOCATION_SIG = r'(?s)^.*' \
-r'packet header .* \(tag 2, len .*' \
+RE_RNP_REVOCATION_SIG = r'(?s)' \
+r':armored input\n' \
+r':off 0: packet header .* \(tag 2, len .*' \
 r'Signature packet.*' \
 r'version: 4.*' \
 r'type: 32 \(Key revocation signature\).*' \
@@ -1272,6 +1273,8 @@ class Keystore(unittest.TestCase):
         os.close(pipe)
         self.assertEqual(ret, 0)
         self.assertTrue(os.path.isfile(OUT_ALICE_REV))
+        with open(OUT_ALICE_REV, "rb") as armored:
+            self.assertRegex(armored.read().decode('utf-8'), r'-----END PGP PUBLIC KEY BLOCK-----\r\n$', 'Armor tail not found')
         # Check revocation contents
         ret, out, _ = run_proc(RNP, ['--homedir', RNPDIR, '--list-packets', OUT_ALICE_REV])
         self.assertEqual(ret, 0)
@@ -1333,6 +1336,8 @@ class Keystore(unittest.TestCase):
             self.assertEqual(ret, 0, 'Failed to export revocation with code ' + revcode)
             self.assertTrue(os.path.isfile(OUT_ALICE_REV), 'Failed to export revocation with code ' + revcode)
             # Check revocation contents
+            with open(OUT_ALICE_REV, "rb") as armored:
+                self.assertRegex(armored.read().decode('utf-8'), r'-----END PGP PUBLIC KEY BLOCK-----\r\n$', 'Armor tail not found')
             ret, out, _ = run_proc(RNP, ['--homedir', RNPDIR, '--list-packets', OUT_ALICE_REV])
             self.assertEqual(ret, 0, 'Failed to list exported revocation packets')
             self.assertNotEqual(len(out), 0, 'Failed to list exported revocation packets')
--- a/src/tests/ffi-key-sig.cpp
+++ b/src/tests/ffi-key-sig.cpp
@@ -488,14 +488,12 @@ TEST_F(rnp_tests, test_ffi_export_revocation)
     assert_rnp_success(rnp_key_export_revocation(
       key_handle, output, 0, "SHA256", "superseded", "test key revocation"));
     assert_rnp_success(rnp_output_destroy(output));
-    assert_rnp_success(rnp_output_to_path(&output, "alice-revocation.asc"));
-    assert_rnp_success(rnp_key_export_revocation(key_handle,
-                                                 output,
-                                                 RNP_KEY_EXPORT_ARMORED,
-                                                 "SHA256",
-                                                 "superseded",
-                                                 "test key revocation"));
-    assert_rnp_success(rnp_output_destroy(output));
+
+    /* check that the output is binary or armored as requested */
+    std::string data = file_to_str("alice-revocation.pgp");
+    assert_false(starts_with(data, "-----BEGIN PGP PUBLIC KEY BLOCK-----"));
+    assert_false(ends_with(strip_eol(data), "-----END PGP PUBLIC KEY BLOCK-----"));
+
     /* make sure FFI locks key back */
     bool locked = false;
     assert_rnp_success(rnp_key_is_locked(key_handle, &locked));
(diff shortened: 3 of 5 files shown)


antonsviridenko:
I think it over-complicates implementation of `rnp_key_export_revocation()`, makes it do multiple things at once, while armoring can be achieved by combining multiple FFI calls.

src/lib/rnp.cpp
```diff
@@ -4049,18 +4049,15 @@ rnp_key_get_revocation(rnp_ffi_t        ffi,
 
 rnp_result_t
 rnp_key_export_revocation(rnp_key_handle_t key,
-                          rnp_output_t     output,
+                          rnp_output_t     external_output,
```
ni4, line 4052: Why do we need parameter name change here?

src/lib/rnp.cpp
```diff
@@ -4049,18 +4049,15 @@ rnp_key_get_revocation(rnp_ffi_t        ffi,
...
                           uint32_t         flags,
                           const char *     hash,
                           const char *     code,
                           const char *     reason)
 try {
-    if (!key || !key->ffi || !output) {
+    if (!key || !key->ffi || !external_output) {
         return RNP_ERROR_NULL_POINTER;
     }
-    if (flags) {
```
ni4, line 4061: We should check for wrong flags here as well, i.e. that flags doesn't contain anything except `RNP_KEY_EXPORT_ARMORED`. For instance, please see how this is done in `rnp_add_security_rule()`. Btw, `test_ffi_export_revocation()` must check this at line 444, so I surprised why it passes.

src/lib/rnp.cpp
```diff
@@ -4079,10 +4076,23 @@ try {
         return ret;
     }
 
+    rnp_output_t output = external_output;
```
ni4, line 4079: As we are moving to C++ internals, we should use here ArmoredDest instead of C `rnp_output_t`, as this ensures deallocation handling due to exceptions and so on. Please see how this is implemented in the function `rnp_key_export_autocrypt()`.

ni4 (changes requested):
Please see the review comments. Also, we should extend tests for the following cases:
- make sure that CLI-exported revocation is armored (by adding this check in `cli_tests.test_export_revocation`, including check for `-----END PGP MESSAGE` as we must make sure that output is correctly flushed.
- make sure that without RNP_KEY_EXPORT_ARMORED FFI function's output is binary to not allow regressions in the future.

ni4:
> I think it over-complicates implementation of `rnp_key_export_revocation()`, makes it do multiple things at once, while armoring can be achieved by combining multiple FFI calls.

It's for the user convenience - while internally we may implement this once and in a few codelines ensuring all checks, each user would need to understand all the logic, do all of the calls, checking for possible error cases/failures and so on.

src/lib/rnp.cpp
andrey-utkin: To disambiguate with the other "output" entity which was either the original, or the armoring one.
I reverted this change while addressing your other comment.

src/lib/rnp.cpp
andrey-utkin: Addressed.

andrey-utkin:
> Please see the review comments. Also, we should extend tests for the following cases:

Addressed the comments around code changes.
Extra tests will come soon.

src/lib/rnp.cpp
andrey-utkin: Addressed.

src/tests/ffi-key-sig.cpp
```diff
@@ -468,18 +474,49 @@ TEST_F(rnp_tests, test_ffi_export_revocation)
...
+    assert_rnp_success(rnp_locate_key(ffi, "userid", "Alice <alice@rnp>", &key_handle));
+    rnp_output_t output = NULL;
+    assert_rnp_success(rnp_output_to_path(&output, "alice-revocation.pgp"));
+
     /* unlocked key - must succeed */
     assert_rnp_success(rnp_key_unlock(key_handle, "password"));
-    assert_rnp_success(rnp_key_export_revocation(key_handle, output, 0, "SHA256", NULL, NULL));
-    assert_rnp_success(rnp_output_destroy(output));
+    assert_rnp_success(
+      rnp_key_export_revocation(key_handle, output, flags, "SHA256", NULL, NULL));
```
andrey-utkin, line 507: > Added line #L507 was not covered by tests

I can't see how this warning and other by codecov could be right.
I can imagine a cause for it though: there are two calls, `rnp_key_export_revocation()` and `assert_rnp_success()`, and as I recall from debugging practice the compiler attributes both of them to the first line, 506. So Codecov is like "hey, i've never encountered running the line 507 according to my traces!".
We could check this guess by temporarily violating the coding style on purpose and refactoring the code to be in single long lines.
Does the above make sense to you?

ni4: In this case we may ignore Codecov's warning and stick to the linting rules.

src/tests/cli_tests.py
```diff
@@ -1272,6 +1273,8 @@ def test_export_revocation(self):
         os.close(pipe)
         self.assertEqual(ret, 0)
         self.assertTrue(os.path.isfile(OUT_ALICE_REV))
+        with open(OUT_ALICE_REV, "rb") as armored:
```
ni4, line 1276: You may use helper function `file_text()` here for less code.

src/tests/ffi-key-sig.cpp
```diff
@@ -438,48 +443,69 @@ TEST_F(rnp_tests, test_ffi_export_revocation)
...
+
+    rnp_output_destroy(output);
+    assert_rnp_success(rnp_key_handle_destroy(key_handle));
+    assert_rnp_success(rnp_ffi_destroy(ffi));
+}
+
+TEST_F(rnp_tests, test_ffi_export_revocation_failure_modes)
+{
+    for (bool armor : {false, true}) {
+        test_ffi_export_revocation_failure_modes_fn(armor);
```
ni4, line 478: This style would lead to the problems once we got to some failure: we just would not know which run (armor false or true) lead to the error reported by some line number. So it's preferred to duplicate code, as these are just tests, for better clarity.

src/tests/ffi-key-sig.cpp
```diff
@@ -508,20 +533,46 @@ TEST_F(rnp_tests, test_ffi_export_revocation)
 
     /* check signature contents */
     pgp_source_t src = {};
-    assert_rnp_success(init_file_src(&src, "alice-revocation.pgp"));
+    pgp_source_t file_src = {};
+    pgp_source_t armored_src = {};
+    assert_rnp_success(init_file_src(&file_src, "alice-revocation.pgp"));
```
ni4, line 538: Same here: better duplicate tests code then make it complicated to debug afterwards.
Also, as these are FFI tests, we should not use internal functions from the lower layer (pgp_source_t and so on).
Reason for this is that at some point it would be desirable to run FFI tests suite against the built library instead of compiling tests together with lower layer sources.

We have helper functions `file_to_str()`/`file_to_vec()` in the `support.h`.

src/tests/ffi-key-sig.cpp
```diff
@@ -508,20 +533,46 @@ TEST_F(rnp_tests, test_ffi_export_revocation)
...
+        src_close(&armored_src);
     assert_int_equal(sig.type(), PGP_SIG_REV_KEY);
     assert_true(sig.has_subpkt(PGP_SIG_SUBPKT_REVOCATION_REASON));
     assert_true(sig.has_keyfp());
     assert_int_equal(sig.revocation_code(), PGP_REVOCATION_SUPERSEDED);
     assert_string_equal(sig.revocation_reason().c_str(), "test key revocation");
-    assert_int_equal(unlink("alice-revocation.pgp"), 0);
 
+    // Check that the output is binary or armored as requested
+    std::ifstream inf("alice-revocation.pgp", std::ios::binary | std::ios::ate);
```
ni4, line 557: Feel free to use `file_to_str()` here for less coding.

ni4 (changes requested):
Please see my comments (some of theme I added as comments, not as the part of the review).

src/lib/rnp.cpp
```diff
@@ -4058,6 +4058,8 @@ try {
     if (!key || !key->ffi || !output) {
         return RNP_ERROR_NULL_POINTER;
     }
+    bool need_armor = flags & RNP_KEY_EXPORT_ARMORED;
+    flags &= !RNP_KEY_EXPORT_ARMORED;
```
ni4, line 4062: We should not use logical not here (!) and use bitwise not (~) instead.

andrey-utkin: Oops, my bad. Thanks for pointing out.

src/tests/ffi-key-sig.cpp
andrey-utkin: I see what you mean.
In this case, I'll do this
```c
TEST_F(rnp_tests, test_ffi_export_revocation_failure_modes)
{
    test_ffi_export_revocation_failure_modes_fn(false);
}
TEST_F(rnp_tests, test_ffi_export_revocation_failure_modes_armored)
{
    test_ffi_export_revocation_failure_modes_fn(true);
}
TEST_F(rnp_tests, test_ffi_export_revocation)
{
    test_ffi_export_revocation_fn(false);
}
TEST_F(rnp_tests, test_ffi_export_revocation_armored)
{
    test_ffi_export_revocation_fn(true);
}
```

ni4: We do not need to run most of thesel checks (like `assert_rnp_failure(rnp_key_export_revocation(key_handle, NULL, flags, "SHA256", NULL, NULL));`) for armored/binary revocations. Also rewriting existing tests so heavily could lead to unexpected regressions. So i'd suggest just to extend existing tests with single armored revocation export check (and making sure that old test exports binary data on success).

src/tests/cli_tests.py
andrey-utkin: Thanks, good to know.
But, I'm a bit concerned that file_test() definition does `replace('\r\r', '\r')`. Is that right?
Until this is clarified, I'll leave the code as it was.

src/tests/ffi-key-sig.cpp
ni4: > I've disambiguated the scenarios.

Please see my previous comment - we do not need to run all bunch of tests for armored/binary output.

> I see, it's testing that the revocation certificate meets the expectations, but does so using internal API.

Yeah, you are right (missed that we had lowerl-layer dependency already) - let's leave it as is and add a ticket for that. It should be easier to do this check via CLI tests, by exporting revocation and running `rnpkeys --list-packets` on it.

src/tests/cli_tests.py
ni4: Not sure what it was aimed for. It came via this commit: https://github.com/rnpgp/rnp/commit/56d9ae9d258a8603eeeac7bb41e637020e0a0eb9

Probably some mistyping. Anyway, it worked for now )

src/tests/ffi-key-sig.cpp
andrey-utkin: I agree to not duplicate entire "failure modes" test case, see 1aa06b8f8acb2dd49d53142b25f3ec14633321bb

But for the case when the revocation is successfully exported, I want to test that it's functional, and for that the test case imports it, changing the state of the keyring. I found that for all the existing assertions following the importing the revocation to stand, the revocation has to be imported exactly once. So I replicated the entire testcase, to get the same keyring lifecycle. Do you agree with this reasoning?

ni4: @andrey-utkin Basically, it has just three steps - load initial keyring, export revocation, import it back and make sure that key becomes revoked. Just unload keys after the previous steps and all is fine.

src/tests/ffi-key-sig.cpp
```diff
@@ -469,17 +472,47 @@ TEST_F(rnp_tests, test_ffi_export_revocation)
...
+    assert_rnp_failure(rnp_key_export_revocation(key_handle,
+                                                 output,
+                                                 RNP_KEY_EXPORT_ARMORED,
+                                                 "SHA256",
+                                                 "superseded",
+                                                 "test key revocation"));
+
+    rnp_output_destroy(output);
+    assert_rnp_success(rnp_key_handle_destroy(key_handle));
+    assert_rnp_success(rnp_ffi_destroy(ffi));
```
ni4, line 484: With these changes we completely missing the case(s) when user attempts to successfully export revocation after few initial failures. So, again, please do not overwrite test cases but just extend them.

ni4 (changes requested):
Please do not overwrite test cases but just extend those, as this could (potentially) lead to unexpected regressions.

src/tests/ffi-key-sig.cpp
andrey-utkin: I have just noticed the above message, will address now.

andrey-utkin:
CI checks pass, except for Codecov, which I think is a false positive of the kind I've described above.

ni4 (approved):
LGTM, thanks! Yeah, Codecov somehow got crazy about those lines. Merging.

PR #1856, opened 2022-06-13
CLI: do not confuse user with 'no keyring file' messages and default gnupg key.

### LINKED ISSUE #1558: When `~/.gnupg` exists without `~/.rnp`, rnpkeys reads the former for default key but does not load other keys
## Description

This is from v0.15.1.

```sh
# ~/.gnupg exists with keys, ~/.rnp doesn't exist
$: rnpkeys --list-keys
rnp: default key set to "333C4477426DC542610211FB5719063A2D8998FA"
wrong pubring path
Key(s) not found.
```

Once I move away `~/.gnupg`, it works properly.
```
$: mv ~/.gnupg{,.bak}
$: rnpkeys --list-keys
wrong pubring path
Key(s) not found.
```

Question: Should we support only a single key directory or multiple?

rnpkeys should do one or multiple of the following:

1. Display a message on which key directory it is loading from ("Keyring detected at path ~/.gnupg/secring.gpg, using it")

2. When `~/.gnupg` exists, there are a few modes of operation (we need to discuss)
    * only operate from `~/.gnupg/`, perhaps give a warning: "WARNING: RNP loading non-native keyring from `~/.gnupg`. To migrate to the native RNP key collection format, run `rnpkeys import --home=~/.gnupg`".
    * prompt the user to migrate to `~/.rnp` and leave `~/.gnupg` alone in case we break something there.
    * Tell user in order to use `~/.gnupg` as directory they will need to configure a setting or always use `rnpkeys --homedir=~/.gnupg [command]`

### LINKED ISSUE #1559: When `~/.rnp` doesn't exist, `rnpkeys` prints out "wrong pubring path" which is uninformative
## Description

Right after installing RNP, the first run of `rnpkeys` give this output which is incredibly uninformative...

```sh
$ rnpkeys --list-keys
wrong pubring path
...
```

## Expected Behavior

More informative message needed, e.g.
"Initializing ~/.rnp/ directory as keystore..."

## Actual Behavior

What behavior did you observe instead?

```sh
$ rnpkeys --list-keys
wrong pubring path
...
```

ni4:
This PR changes CLI behaviour on non-existent keyrings to the following:
- do not display `warning: keyring at path '/Users/user/.rnp/pubring.gpg' doesn't exist.` message on empty homedir
- do not parse GnuPG conf, extracting the default key as it may not exist in RNP's keyring
- if homedir is empty then display message about key generation/import from the GnuPG.

Fixes #1558 
Fixes #1559

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -380,18 +380,19 @@ RNP_API rnp_result_t rnp_ffi_set_pass_provider(rnp_ffi_t       ffi,
  */
 RNP_API rnp_result_t rnp_get_default_homedir(char **homedir);
 
-/** try to detect the formats and paths of the homedir keyrings
- *
+/** Try to detect the formats and paths of the homedir keyrings.
  * @param homedir the path to the home directory (example: /home/user/.rnp)
  * @param pub_format pointer that will be set to the format of the public keyring.
  *        The caller should free this with rnp_buffer_destroy.
+ *        Note: this and below may be set to NULL in case of no known format is found.
  * @param pub_path pointer that will be set to the path to the public keyring.
  *        The caller should free this with rnp_buffer_destroy.
  * @param sec_format pointer that will be set to the format of the secret keyring.
  *        The caller should free this with rnp_buffer_destroy.
  * @param sec_path pointer that will be set to the path to the secret keyring.
  *        The caller should free this with rnp_buffer_destroy.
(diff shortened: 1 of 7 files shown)


antonsviridenko (approved):

ronaldtse (approved):

andrey-utkin:
LGTM (late, but still).

PR #1858, opened 2022-06-14
Add and use function rnp_fdopen which would not cause AV with wrong fd on Windows.

ni4:
As title says.

### DIFF
--- a/src/common/file-utils.cpp
+++ b/src/common/file-utils.cpp
@@ -110,6 +110,16 @@ rnp_fopen(const char *filename, const char *mode)
 #endif
 }
 
+FILE *
+rnp_fdopen(int fildes, const char *mode)
+{
+#ifdef _WIN32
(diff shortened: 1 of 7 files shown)


ronaldtse (approved):
Thanks @ni4 !

andrey-utkin (approved):
I am not very knowledgeable about WIN32 API, but from a brief research, this patch seems to be a response to the following two problems and is legit.

1. Invalid fdno causes a runtime exception on Windows
https://gist.github.com/novocaine/09d5c00e67fd0aa13cfc#file-gistfile1-diff-L48
2. POSIX `fdopen()` is deprecated in Windows API (["because it doesn't follow the Standard C rules for implementation-specific names"???](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fdopen?view=msvc-170)) and developers are recommended to call `_fdopen()`.

PR #1861, opened 2022-06-21
Separate security rules for data and key signatures.

### LINKED ISSUE #1819: Add option to apply security rule only to certain signatures.
## Description

Currently hash security rules apply to all signatures. However, it may be desirable to distinguish between key signatures and data signatures, so we should add corresponding flag to the `rnp_add_security_rule()`.
Probably it could also worth to allow SHA1 for key's signatures by default.

ni4:
This PR updates security rules support to allow separate rules for data and key signatures (see #1819 for details).
Also it changes default SHA-1 rule to allow it for keys till Jan, 19 2024 (but still do not allow for data signatures produced after the Jan, 19 2019).

Fixes #1819


antonsviridenko (approved):

andrey-utkin (approved):

ronaldtse:
Two approvals for merging. Thanks @ni4 !

Issue #1498, opened 2021-05-06
Delete temporary folders for rnp_tests if test succeeds.


ni4:
## Description

Currently test folders (from rnp_tests) are not deleted automatically, which may lead to the /tmp bloated with rnp-gtest-* dirs.
Those should be deleted in case of successful test run, and left in case of test's failure, so we may check what went wrong/upload an artifact in the CI.

ni4:
Fixed via #1866

Issue #1430, opened 2021-02-22
Update GHA and cli_tests.py to upload artifacts on gpg vs rnp failure.


ni4:
## Description

This would help to debug rarely-appearing issues like #1379

PR #1866, opened 2022-06-27
[#1498] Fix tmpdir cleanup

andrey-utkin:
A lot of temporary directories are always left around after the tests even when all they all pass. There are a few sources of that, and this patchset addresses them all.

Bug: https://github.com/rnpgp/rnp/issues/1498

### DIFF
--- a/ci/env-common.inc.sh
+++ b/ci/env-common.inc.sh
@@ -26,3 +26,6 @@ fi
 BOTAN_MODULES=$(<ci/botan-modules tr '\n' ',')
 
 export BOTAN_MODULES
+
+# Don't clean up tempdirs when in CI runners to save time. Unset to disable.
+export RNP_KEEP_TEMP=1
--- a/src/tests/cli_tests.py
+++ b/src/tests/cli_tests.py
@@ -4344,6 +4344,8 @@ if __name__ == '__main__':
     if LEAVE_WORKING_DIRECTORY:
         # -w must be removed as unittest doesn't expect it
         sys.argv.remove('-w')
+    else:
+        LEAVE_WORKING_DIRECTORY = os.getenv('RNP_KEEP_TEMP') is not None
 
     LVL = logging.INFO
     if "-d" in sys.argv:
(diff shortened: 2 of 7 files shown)


ni4:
@andrey-utkin Could you please add check whether run is made under the GHA, so only local runs would be cleaned up automatically? (this should be helpful: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables )

Basically, current centos8 failure is something because of what we should not cleanup CI automatically, and upload artifacts once #1430 is implemented.

ni4 (changes requested):
Please see a separate comment.

andrey-utkin:
For now I've added a way to preserve the tempdirs by setting an env var `KEEPTEMP`.

I'd rather make GHA workflows declare explicitly that they want this, by setting that env var, because this simple concept will likely outlive the distinction between GHA and local runs.

We can add this setting in `.github/workflows/*.yml` right now, basically reverting to the previous behaviour in GHA. But I imagine if we start writing actual artifact retention logic, we'd handle it from the context of a failed assertion, which should have the knowledge of what the workdir is, and before the workdir is removed, so the setting is not necessary.

ni4:
> For now I've added a way to preserve the tempdirs by setting an env var `KEEPTEMP`.
> 
> I'd rather make GHA workflows declare explicitly that they want this, by setting that env var, because this simple concept will likely outlive the distinction between GHA and local runs.

I doubt that it would be the heaviest problem once migrating away from GHA :-) But, okay, let's use this approach, but it would be better to rename variable to something starting with `RNP_`, like. `RNP_KEEP_TEMP` or so on.

antonsviridenko (approved):

ni4:
@ronaldtse Thanks for bringing this up, somehow missed this. Should be related to test file leftovers...

antonsviridenko:
yeah, probably some issue introduced by this PR...

ni4:
Seems to hang on `SHFileOperationA` call, continuing with debugging...

ni4:
@ronaldtse Now it should be good to go. Issue should be in some hanging GUI, displayed during the `SHFileOperationA()`, maybe some readonly file or similar.

ronaldtse:
Thank you @ni4 ! This is perfect, merging.

Issue #1617, opened 2021-08-23
Test Fedora


ribose-jeffreylau:
Fedora 35 has been released, and should be tested.

andrey-utkin:
Fedora 36 has been released since, and I guess we want to test on it, too!
Fedora 35 might be still relevant to us.
Even older releases, 28 and 34, may be as well interesting to us as they are the bases for RHEL 8 and 9 respectively.
And we want to test them with both Botan and OpenSSL backends, if it's not too much to ask of Github Actions.

andrey-utkin:
Raised a blocker https://github.com/riboseinc/yum/issues/9 in relation to adding CentOS 9 to CI :)

ronaldtse:
@andrey-utkin I think RNP should probably provide its own yum repository instead of using the riboseinc one. What do you think?

andrey-utkin:
> @andrey-utkin I think RNP should probably provide its own yum repository instead of using the riboseinc one. What do you think?

RNP probably doesn't need to right now.
The importance of this internal package repository is very small. I'd be grateful to be able to lean on the riboseinc repo and not take on maintenance of another RPM repo.

In this case, these `ribose-automake116*` packages play purely instrumental role of build prerequisites for custom-building GnuPG for CI runs. There are ways to avoid relying on an RPM repo - we could write a Dockerfile which builds it from source in place and preserves the container image for reuse.

andrey-utkin:
PR https://github.com/rnpgp/rnp/pull/1870 merged.
Closing this ticket.
CentOS 9 is not enabled in CI though.

PR #1870, opened 2022-07-07
[#1617] Add Fedora 35, 36, CentOS 9 to CI

andrey-utkin:


ni4:
@andrey-utkin I don't think we need such extensive runs for all fedora35/36 combinations.
For our purposes for now it could be enough to have just OpenSSL backend, g++ compiler, sanitize/normal builds. Also we may run both rnp_tests/cli_tests in a single run, using the stable GnuPG release.

cc @ronaldtse

ronaldtse:
> @andrey-utkin I don't think we need such extensive runs for all fedora35/36 combinations. For our purposes for now it could be enough to have just OpenSSL backend, g++ compiler, sanitize/normal builds. Also we may run both rnp_tests/cli_tests in a single run, using the stable GnuPG release.

I agree because Fedora 36 uses OpenSSL 3.0, which RNP support is currently under development, so we won't be able to add the Fedora 36 workflow to the current main/master yet. It will have to go into the OpenSSL 3.0 PR (in the future).

andrey-utkin:
Please review, finally all CI tests are passing.

PR #1871, opened 2022-07-08
Bogus change to test CI status on release branch

andrey-utkin:


ni4:
@andrey-utkin these CI failures are related to changes in GHA/used OS images/whatever else. Those are merged into master already so you just would need to cherry-pick those as well.

Issue #1872, opened 2022-07-11
CI failures on release/0.x branch as of v0.16.0


andrey-utkin:
## Description

The configured CI jobs (Github Actions) fail in many ways on 0.16.0 release itself, which makes it difficult to build a next patch release on top of it.

## Steps to Reproduce

1. Create a branch based on v0.16.0 git tag (or release/0.x branch) and add a trivial cosmetic change to the codebase, e.g. a no-op statement to src/lib/rnp.cpp
2. push the branch
3. create a draft pull request with release/0.x branch as a base

## Expected Behavior

All CI checks pass

## Actual Behavior

Many CI checks fail. This is my PR demonstrating the issue: https://github.com/rnpgp/rnp/pull/1871
> 27 failing, 22 successful, 1 skipped, and 1 neutral checks

https://github.com/rnpgp/rnp/runs/7252596896?check_suite_focus=true
> Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist

Many runs have this ruby-rnp test suite error:

`Failure/Error: expect(@json["expiration"]).to be 63072000`

Windows:
> could not find any instance of Visual Studio.

> D:/a/rnp/rnp/src/common/str-utils.cpp:30:10: fatal error: 'cstddef' file not found

"build" (freebsd?)
> pkg: No packages available to install matching 'ruby26' have been found in the repositories

andrey-utkin:
Nickolay said there are fixes on master which should be cherry picked.
https://github.com/rnpgp/rnp/pull/1871#issuecomment-1180251635

ni4:
This is something that would always happen as time flies: expiring keys, updated CI images and runners, changed timestamps and so on. So I don't see a cure from such failures. If you don't have any further questions - feel free to close it as not planned.

andrey-utkin:
There's no cure, but scheduled runs help with discovering it earlier ;)
I am going to use this ticket to denote the prerequisite work to backporting of the #1859 patches onto the release branch.

andrey-utkin:
The fix comes in three pieces.
1. The minor-stable-release branch (`release/0.x`?) should take three backport commits. https://github.com/rnpgp/rnp/pull/1874
2. The master branch of RNP repo should take the increment of version.txt https://github.com/rnpgp/rnp/pull/1873
3. ruby-rnp repo should change to distinguish master from the stable branch correctly: https://github.com/rnpgp/ruby-rnp/pull/81

[...]

PR #1873, opened 2022-07-11
[#1872] version.txt: increment to 0.17.0

andrey-utkin:
This is to make it possible to differentiate between 0.16.0 and current
master. A specific case where this is needed is the CI problems with
ruby-rnp test suite in the PR backporting specific features onto 0.16.0
to make a 0.16.1: https://github.com/rnpgp/rnp/pull/1869

This should be merged at the same time with a change for ruby-rnp:
https://github.com/rnpgp/ruby-rnp/pull/81

Bug: https://github.com/rnpgp/rnp/issues/1872


ni4 (changes requested):
This MUST NOT be merged at it goes strongly against the logic - we have v0.16.0+ sources on master and we never ever had v0.17.0 release yet.

andrey-utkin:
Fine, but we still need to somehow tell the release from a master branch.
A couple other ways I see here. Does any of them look good to you?

1. Make the master branch have the semblance of "unstable release branch", with odd minor number, e.g. 0.17.x, then when there's time to branch from that, convert it to the even-numbered 0.18.x.

2. A special token in `version.txt` on live branches. For example, ffmpeg does it this way:
As a part of release branching procedure, right after the branching, the value of `RELEASE` file changes from `4.2.git` to `4.3.git` for master and `4.2` for the release branch.
https://git.ffmpeg.org/gitweb/ffmpeg.git/blobdiff/1ee3c984b91e0241068d1c093d222ecec2e6052c..f66f3ca2c18f00bf4fc17f7a5d346abb78be41de:/RELEASE
https://git.ffmpeg.org/gitweb/ffmpeg.git/blobdiff/0a8a96c25194b550f3b87902e1b7f061738ae3cd..c1ebaffba9fdc8948bce54b96c347ff960d1440c:/RELEASE

ni4:
In this particular case we got problems on ruby-rnp side since planned next version was v0.17.0. Having knowledge that key expiration time commits would be included into v0.16.1, we may just fix https://github.com/rnpgp/ruby-rnp/pull/81/files with the `Rnp.version >= Rnp.version("0.16.1")` check.

ni4:
> > In this particular case we got problems on ruby-rnp side since planned next version was v0.17.0. Having knowledge that key expiration time commits would be included into v0.16.1, we may just fix https://github.com/rnpgp/ruby-rnp/pull/81/files with the `Rnp.version >= Rnp.version("0.16.1")` check.
> 
> While this would fix it for v0.16.1 and its descendants, it won't fix it for release/0.x branch and any PRs targeting it, which ties our hands in preparing the release.

But how in practice we could need something inbetween v0.16.0 and v0.16.1?

ronaldtse:
> > But how in practice we could need something inbetween v0.16.0 and v0.16.1?
> 
> During release preparation, for example, for PRs backporting individual features.

Agree.

ni4:
Closing this as it was handled in a different way.

PR #1875, opened 2022-07-11
[#1595] OpenSSL backend features detection

andrey-utkin:


ni4:
@andrey-utkin At the moment I see two issues:
1) AEAD is not AES (and our OpenSSL backend implementation doesn't support AEAD algos at all, we have separate issue for this).
2) WIthin automatic feature detection we should by default use 'Auto' mode, when only supported by backend features are enabled. However, allow user to enable or disable desired features via the corresponding CMake options, throwing an error if user wants to enable unsupported by backend feature.

ni4:
> Sorry, I am not an expert on neither AEAD nor OpenSSL features, it was my best guess based on a brief online research. What would be the right set of OpenSSL features to require?

Np, it's ok. At this moment I'd suggest to automatically detect/enable/disable only features which are not available in some OpenSSL distributions (and disable AEAD which is not available from us at the moment).

> Oh, I see. Would it be OK if I submit it as a separate PR, to leave this one trivially simple?

We may start from currently known 'problematic' features (Blowfish, Twofish, SM2, IDEA, AEAD) and then add additional features via separate PRs (ENABLE_ELGAMAL, ENABLE_DSA, ENABLE_CAMELLIA, whatever else would come).

src/tests/cli_tests.py
```diff
@@ -841,9 +841,6 @@ def rnp_check_features():
     RNP_BRAINPOOL = re.match(r'(?s)^.*Curves:.*brainpoolP256r1.*brainpoolP384r1.*brainpoolP512r1.*', out)
 
     RNP_IDEA = re.match(r'(?s)^.*Encryption:.*IDEA.*', out)
-    # Check that everything is enabled for Botan:
```
ronaldtse, line 844: Just wanted to check why we had this before and whether the intention is still present today. Thanks!

ronaldtse (approved):
LGTM

antonsviridenko (approved):

PR #1879, opened 2022-07-17
Add missing codes to rnp_result_to_string().

### LINKED ISSUE #1878: Missing result codes in rnp_result_to_string()
## Description

FFI function `rnp_result_to_string()` does not return string results for all defined status/error codes.
At this moment 31 error/status/result codes are defined in https://github.com/rnpgp/rnp/blob/master/include/rnp/rnp_err.h#L32
but only 26 of them are handled by `rnp_result_to_string()`

## Expected Behavior

All defined codes should have corresponding string value in `rnp_result_to_string()`

## Actual Behavior

Codes
```
RNP_ERROR_RNG
RNP_ERROR_SIGNATURE_EXPIRED
RNP_ERROR_SIGNATURE_UNKNOWN
RNP_ERROR_VERIFICATION_FAILED
```
are missing.

antonsviridenko:
Fixes #1878

### DIFF
--- a/src/tests/ffi.cpp
+++ b/src/tests/ffi.cpp
@@ -5949,6 +5949,8 @@ TEST_F(rnp_tests, test_result_to_string)
         for (code = range.first; code <= range.second; code++) {
             result_string = rnp_result_to_string(code);
 
+            assert_int_not_equal(strcmp("Unsupported error code", result_string), 0);
+
             auto search = stringset.find(result_string);
 
             /* Make sure returned error string is not already returned for other codes */
--- a/src/tests/ffi.cpp
+++ b/src/tests/ffi.cpp
@@ -27,6 +27,8 @@
 #include <fstream>
 #include <vector>
 #include <string>
+#include <set>
+#include <utility>
 
 #include <rnp/rnp.h>
 #include "rnp_tests.h"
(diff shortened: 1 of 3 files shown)


ronaldtse (approved):
Thank you @antonsviridenko !

src/tests/ffi.cpp
```diff
@@ -5923,3 +5925,36 @@ TEST_F(rnp_tests, test_ffi_security_profile)
...
+
+    /* Cover all defined error code ranges,
+     * check that each defined
+     * code has corresponding unique string */
+
+    std::vector<std::pair<rnp_result_t, rnp_result_t>> error_codes = {
+      {RNP_ERROR_GENERIC, RNP_ERROR_NULL_POINTER},
+      {RNP_ERROR_ACCESS, RNP_ERROR_WRITE},
+      {RNP_ERROR_BAD_STATE, RNP_ERROR_SIGNATURE_UNKNOWN},
+      {RNP_ERROR_NOT_ENOUGH_DATA, RNP_ERROR_EOF}};
```
andrey-utkin, line 5946: I'm afraid this is not going to catch the return of this bug, as the new codes will be added after the range ends, as those at the end of the range are not special sentinel entries but actual error codes.

One way I see to nail it is parsing rnp_err.h and generating a test program *at testing time* to exercise each defined error code.

src/tests/ffi.cpp
```diff
@@ -5923,3 +5925,36 @@ TEST_F(rnp_tests, test_ffi_security_profile)
...
+
+    for (auto range = error_codes.begin(); range != error_codes.end(); ++range) {
+        for (code = range->first; code <= range->second; code++) {
+            result_string = rnp_result_to_string(code);
+
+            auto search = stringset.find(result_string);
+
+            /* Make sure returned error string is not already returned for other codes */
+            assert_true(search == stringset.end());
```
andrey-utkin, line 5955: We should also check that the string is not `Unknown error`.
And not NULL while we're at that.

andrey-utkin (approved):
Thanks for spotting and fixing this.

I don't believe the added test, as is, has any value for the future, please see my comment.
Also I think a comment at the enum definition could help to remind the developers to update `rnp_result_to_string()`.

src/tests/ffi.cpp
```diff
@@ -5923,3 +5925,36 @@ TEST_F(rnp_tests, test_ffi_security_profile)
...
+     * check that each defined
+     * code has corresponding unique string */
+
+    std::vector<std::pair<rnp_result_t, rnp_result_t>> error_codes = {
+      {RNP_ERROR_GENERIC, RNP_ERROR_NULL_POINTER},
+      {RNP_ERROR_ACCESS, RNP_ERROR_WRITE},
+      {RNP_ERROR_BAD_STATE, RNP_ERROR_SIGNATURE_UNKNOWN},
+      {RNP_ERROR_NOT_ENOUGH_DATA, RNP_ERROR_EOF}};
+
+    for (auto range = error_codes.begin(); range != error_codes.end(); ++range) {
```
ni4, line 5948: Why not `for (auto &range: error_codes)` ?

src/tests/ffi.cpp
ni4: +1 for `Unknown error.`. Check for NULL is done automatically by not crashing on conversion from `char *` to `std::string`.

src/tests/ffi.cpp
ni4: Parsing of the `rnp_err.h` seem to be an overkill, based on possible impact of this issue.

ni4 (approved):
LGTM, thanks! Feel free to merge as comments are addressed.

src/tests/ffi.cpp
antonsviridenko: "Unknown error" is returned in two cases, as a string result corresponding to RNP_ERROR_GENERIC parameter and as a default value for unknown codes.

antonsviridenko: So, "Unknown error" is one of the possible values. Probably we can return different strings for default unknown code and RNP_ERROR_GENERIC, but it can break compatibility with existing applications.

antonsviridenko: @ni4 what do you think about "Unknown error" comments above?

ni4: @antonsviridenko we may change message for the unknown error code to something like "Unsupported error code". Existing implementations should not rely on error codes which we didn't return at all.

PR #1884, opened 2022-07-26
Custom Docker images to speed up CI

andrey-utkin:
This changeset introduces the first custom Docker container intended for RNP's CI.
This Fedora 35 CI config intends to replace the existing config fedora35-ossl.yml (which is currently broken due to riboseinc repo signature issue).
There seems to be no way to trigger GHA to run this CI configuration before it is merged to master.
There is also a script ci/docker/fedora35.test-locally which shows how to use this container for local testing.

### DIFF
--- a/.github/workflows/fedora35-lean.yml
+++ b/.github/workflows/fedora35-lean.yml
@@ -1,0 +1,72 @@
+name: fedora35-lean
+
+on:
+  push:
+    branches:
+      - master
+      - 'release/**'
+    paths-ignore:
+      - '/*.sh'
+      - '/.*'
+      - '/_*'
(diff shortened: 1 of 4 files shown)


andrey-utkin:
Dear reviewers, could you please merge this?
Current CI status is irrelevant - neither the product codebase nor the actual CI configuration was changed here.
What matters is that after this is merged, the `fedora35-lean` GHA workflow will start working, and we want to see how it will work in practice.

ronaldtse (approved):

ronaldtse:
Since this is a CI change, one positive review is sufficient for merging. Merging now. Thank you @andrey-utkin !

PR #1885, opened 2022-07-30
Use latest commit of GoogleTest.

### LINKED ISSUE #1726: Upgrade bundled Google Test version
## Description
Google Test version downloaded by rnp build scripts fails to build with GCC 11
```
[ 11%] Building CXX object src/lib/CMakeFiles/librnp-obj.dir/__/librepgp/stream-write.cpp.o
In file included from /home/odsk/Ribose/rnp/build/_deps/googletest-src/googletest/src/gtest-all.cc:42:
/home/odsk/Ribose/rnp/build/_deps/googletest-src/googletest/src/gtest-death-test.cc: In function ‘bool testing::internal::StackGrowsDown()’:
/home/odsk/Ribose/rnp/build/_deps/googletest-src/googletest/src/gtest-death-test.cc:1301:24: error: ‘dummy’ may be used uninitialized [-Werror=maybe-uninitialized]
 1301 |   StackLowerThanAddress(&dummy, &result);
      |   ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
/home/odsk/Ribose/rnp/build/_deps/googletest-src/googletest/src/gtest-death-test.cc:1290:13: note: by argument 1 of type ‘const void*’ to ‘void testing::internal::StackLowerThanAddress(const void*, bool*)’ declared here
 1290 | static void StackLowerThanAddress(const void* ptr, bool* result) {
      |             ^~~~~~~~~~~~~~~~~~~~~
/home/odsk/Ribose/rnp/build/_deps/googletest-src/googletest/src/gtest-death-test.cc:1299:7: note: ‘dummy’ declared here
 1299 |   int dummy;
      |       ^~~~~

```
This is a known bug, already fixed in upstream
https://github.com/google/googletest/issues/3219
## Steps to Reproduce

1. Configure build with options `-DBUILD_TESTING=on -DCMAKE_BUILD_TYPE=Debug -DDOWNLOAD_GTEST=On`
2. Use GCC 11 to build 
```
gcc --version
gcc (Gentoo 11.2.0 p1) 11.2.0
```
3. try to build

antonsviridenko:
Fixes #1726 
Using latest commit from the master branch is recommended by Google.

### DIFF
--- a/src/tests/CMakeLists.txt
+++ b/src/tests/CMakeLists.txt
@@ -47,11 +47,16 @@ elseif (NOT DOWNLOAD_GTEST)
   find_package(GTest REQUIRED)
   set(GTestMain GTest::Main)
 else()
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS_EQUAL "4.8.5")
+    set(GTEST_GIT_TAG "c43f710")
+  else()
+    set(GTEST_GIT_TAG "HEAD")
+  endif()
   # download and build googletest
   include(FetchContent)
   FetchContent_Declare(googletest
     GIT_REPOSITORY  https://github.com/google/googletest.git
-    GIT_TAG         c43f710
+    GIT_TAG         "${GTEST_GIT_TAG}"
   )
   # maintain compiler/linker settings on Windows
   set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)


antonsviridenko:
> D:\a\rnp\rnp\src\tests\streams.cpp(403,42): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\rnp\rnp\build\src\tests\rnp_tests.vcxproj]

What C++ standards do we support in our codebase?

ronaldtse:
> What C++ standards do we support in our codebase?

We probably should define that!

ni4:
Those are already defined, just didn't get reflection in development docs:
```
set(CMAKE_C_STANDARD 99)
set(CMAKE_CXX_STANDARD 11)
```

ni4:
...however latest GTest `main` commit seems to require C++14: `The 1.12.x branch will be the last to support C++11. Future releases will require at least C++14.`. Could we use C++14 while building gtest from sources?

antonsviridenko:
Ok, let's try to use v1.12.0 then, looks like idea of using latest commit was bit optimistic :)

antonsviridenko:
v1.12.0 fails on Centos 7 tests again :(
Looks like GCC in Centos 7 is too old
```
 /__w/rnp/rnp/builds/rnp-build/_deps/googletest-src/googletest/include/gtest/gtest-matchers.h:436:3: error: body of constexpr function ‘static constexpr bool testing::internal::MatcherBase<T>::IsInlined() [with M = testing::internal::EqMatcher<std::basic_string<char> >; T = std::basic_string<char>]’ not a return-statement
   }
   ^
/__w/rnp/rnp/builds/rnp-build/_deps/googletest-src/googletest/include/gtest/gtest-matchers.h: In instantiation of ‘void testing::internal::MatcherBase<T>::Init(M&&) [with M = testing::internal::EqMatcher<std::basic_string<char> >; T = std::basic_string<char>]’:
/__w/rnp/rnp/builds/rnp-build/_deps/googletest-src/googletest/include/gtest/gtest-matchers.h:318:28:   required from ‘testing::internal::MatcherBase<T>::MatcherBase(M&&) [with M = testing::internal::EqMatcher<std::basic_string<char> >; <template-parameter-2-2> = void; T = std::basic_string<char>]’
/__w/rnp/rnp/builds/rnp-build/_deps/googletest-src/googletest/include/gtest/gtest-matchers.h:565:62:   required from ‘testing::Matcher<std::basic_string<char> >::Matcher(M&&) [with M = testing::internal::EqMatcher<std::basic_string<char> >; <template-parameter-1-2> = void]’
/__w/rnp/rnp/builds/rnp-build/_deps/googletest-src/googletest/src/gtest-matchers.cc:56:61:   required from here
```

ni4:
Yeah, it ships 4.8, which: `GCC provides experimental support for the 2011 ISO C++ standard.`

ni4:
...as a solution we may use some CMake stuff to detect whether compiler supports C++14 (via compiler version check or via approaches like here: https://stackoverflow.com/questions/10984442/how-to-detect-c11-support-of-a-compiler-with-cmake )

antonsviridenko:
Or set GTest version depending on GCC version detected...

antonsviridenko:
Strange, Fedora 36 has GCC 12.1.1 but compiled GTest fine before...

> -- The C compiler identification is GNU 12.1.1
>-- The CXX compiler identification is GNU 12.1.1

ni4:
> Strange, Fedora 36 has GCC 12.1.1 but compiled GTest fine before...

Could it be that Fedora 36 runners used pre-built GTest from the Fedora packages?

ronaldtse:
> Or set GTest version depending on GCC version detected...

This is likely the better solution. However, we should consider whether we should drop support for CentOS7/RHEL7 as others are already doing so. Perhaps we should do that in Release 17.

ni4:
@antonsviridenko Do you need some help with this? It appears to be showstopper for the Fedora 36 support, and should be included into the v0.16.1 release.

antonsviridenko:
@ni4
Looks like it was a bug in GCC 11,
to be more precise,
-Wmaybe-uninitialized was enabled by default in GCC 11, but it gave too many false positives
 https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95006

ni4:
@antonsviridenko Yeah, it seems so. Anyway, we'll need to workaround it. I'd suggest following solution:
- use latest release of GTest for most solutions.
- except old GCC (pre-C++14), use v1.11.0 of GTest
Does this sound good?

ni4:
@antonsviridenko Don't you mind me pushing fixes to this branch?

antonsviridenko:
@ni4 no problem, go ahead

but looks like MSVC windows builds fail with 
> D:\a\rnp\rnp\src\tests\streams.cpp(403,42): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\rnp\rnp\build\src\tests\rnp_tests.vcxproj]
need to make special condition for MSVC compiler

ni4:
@antonsviridenko It looks like something changed in latest CMake update in the runner: while MSVC version is the same, compared to 20220801 runner, cmake was updated, and now cl.exe is called with `/std:c++14` parameter (why? as we set c++11). Will dig into it.

ni4:
@antonsviridenko If you are interested in the details, the reason for this is this (and future) commit in Googletest codebase: https://github.com/google/googletest/commit/131878ce9e18e3fcc5fdd690c93ee9d36f0a18d3
This option due to some CMake internals overrides top-level's `add_compile_options("/std:c++latest")` for the MSVC.
And root reason for this is that designated initializers are part of the `C++20` standard, however all used by CI compilers (starting from GCC 4.8) seem to have this as extension.

ni4:
@antonsviridenko does this look good to you now? Cannot assign you as reviewer as you created this PR :)

antonsviridenko:
yeah, looks good now

ni4 (approved):
As @antonsviridenko approves, so do I.

andrey-utkin (approved):

ronaldtse:
Two approvals => merged. Thanks!

PR #1886, opened 2022-08-10
Release v0.16.1

ni4:
As title says.


andrey-utkin (commented):
It looks good, fingers crossed. Notably, the final state of the codebase is exactly identical to current master, perhaps we could do without duplicating the history? It's visible that you've carefully considered each commit, but I'd prefer to not have to deal with parallel lineages and such large PRs.

ni4:
> Notably, the final state of the codebase is exactly identical to current master, perhaps we could do without duplicating the history? It's visible that you've carefully considered each commit, but I'd prefer to not have to deal with parallel lineages and such large PRs.

It's how we did it before and described in `release-workflow.adoc`. If you have better ideas/experience on the release workflows - feel free to discuss. We need to have separate release and master branches, as at some point we may need to make a minor update(say, quick security fix), which would not include whole set of changes from the master.

andrey-utkin:
Well, that was my entire proposal. If the current state of `master` branch is what we are happy with for the release, then I suggest we tag that state of master branch as 0.16.1, and don't busy ourselves applying the procedure which is more appropriate for when we must actually diverge in code. I don't see any risk in this case, but it skips a lot of mental load, and makes simple tooling (such as `git show-branch`) work better on the codebase.

antonsviridenko:
It would be nice if fix for https://github.com/rnpgp/rnp/issues/1894 will be also included in this release, otherwise looks fine

ni4:
> Well, that was my entire proposal. If the current state of `master` branch is what we are happy with for the release, then I suggest we tag that state of master branch as 0.16.1, and don't busy ourselves applying the procedure which is more appropriate for when we must actually diverge in code.

@andrey-utkin We had a case when we needed to diverge in code (pick only few security related commits from the master for a new release), and, as I noted, must be prepared to such case in the future. Given that there is no any problem on cherry-picking commits from the master into the release branch using the proper tooling, why should we change to the less-flexible approach?

andrey-utkin:
> why should we change to the less-flexible approach?

I didn't mean that we have to forbid the backporting as a possibility for release making.
What I'm saying is that in this case, we could have done a much simpler action without upsetting anyone.

> Given that there is no any problem on cherry-picking commits from the master into the release branch using the proper tooling

I disagree here. The problems are wasting (current and future) developers time to make sense of unnecessarily complicated commits history.

andrey-utkin (approved):
OK @ni4 please ship it.

ni4:
> It would be nice if fix for #1894 is also included in this release, otherwise looks fine

Please see reply in the original issue. If it's okay to ship - feel free to approve.

ronaldtse:
I don't think the discussion of the release workflow should be performed here.

@antonsviridenko if you are okay with the response on https://github.com/rnpgp/rnp/issues/1894, please feel free to approve the release. Thanks!

antonsviridenko (approved):

ronaldtse:
Thank you guys!

PR #1889, opened 2022-08-18
Do not use designated initializers.

ni4:
This feature is formally available since C++20, however GCC supports it since v4.8, as well as CLang. In contrast, MSVC requires `/std:latest or /std:c++20` to support this. Since we are stick to the C++11, let's be formally compatible.


ni4:
Codecov failure should be related to lower number of covered lines of code due to refactoring.

antonsviridenko (approved):

ronaldtse (approved):
Thank you @ni4 !

PR #1890, opened 2022-08-23
feat: test rhel8 ubi image

ronaldtse:


maxirmx:
@ronaldtse
Do we need this feature accomodated into #1973 ?

ronaldtse:
@maxirmx yes I would like to test RNP in RHEL 8. However, contrasting with the time I did this PR, RedHat has now integrated RNP as an independent package and is using a different build flow. So we want to adopt the latest / official build flow and integrate it into the repo.

So there is more work necessary here. Eventually we do want to have this in the squashed flow.

ronaldtse:
@maxirmx is this still needed? If not we can close it.

ni4:
Closing this as we now have RHEL 8 image via https://github.com/rnpgp/rnp-ci-containers/pull/15 and CI via https://github.com/rnpgp/rnp/pull/2240/commits/8eae86f45810a0156c28616c24ea3a86feada112

PR #1896, opened 2022-08-29
Require single valid signature for successfull return of  rnp_op_verify_execute() by default.

ni4:
This PR changes previous a bit inconsistent behaviour of the `rnp_op_verify_execute()`.
Now by default it will succeed if there is at least one valid signature (if data is signed).

However, as CLI is stick to another rules, function `rnp_op_verify_set_flags()` was added to be able to handle following scenarios:
- decrypt data even if there are no valid signatures
- require all signatures to be valid for successfull processing.

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -132,6 +132,12 @@ typedef uint32_t rnp_result_t;
 #define RNP_ENCRYPT_NOWRAP (1U << 0)
 
 /**
+ * Decryption/verification flags
+ */
+#define RNP_VERIFY_IGNORE_SIGS_ON_DECRYPT (1U << 0)
+#define RNP_VERIFY_REQUIRE_ALL_SIGS (1U << 1)
+
+/**
  * Return a constant string describing the result code
  */
 RNP_API const char *rnp_result_to_string(rnp_result_t result);
@@ -2226,8 +2232,9 @@ RNP_API rnp_result_t rnp_op_sign_destroy(rnp_op_sign_t op);
 /* Verification */
 
 /** @brief Create verification operation context. This method should be used for embedded
- *         signatures or cleartext signed data. For detached verification corresponding
- *         function should be used.
+ *         signatures, cleartext signed data and encrypted (and possibly signed) data.
+ *         For the detached signature verification the function rnp_op_verify_detached_create()
+ *         should be used.
  *  @param op pointer to opaque verification context
  *  @param ffi
  *  @param input stream with signed data. Could not be NULL.
@@ -2252,12 +2259,32 @@ RNP_API rnp_result_t rnp_op_verify_detached_create(rnp_op_verify_t *op,
                                                    rnp_input_t      input,
                                                    rnp_input_t      signature);
 
+/**
+ * @brief Set additional flags which control data verification/decryption process.
+ *
+ * @param op pointer to opaque verification context.
+ * @param flags verification flags. OR-ed combination of RNP_VERIFY_* values.
+ *              Following flags are supported:
+ *              RNP_VERIFY_IGNORE_SIGS_ON_DECRYPT - ignore invalid signatures for the encrypted
+ *                and signed data. If this flag is set then rnp_op_verify_execute() call will
+ *                succeed and output data even if all signatures are invalid or issued by the
+ *                unknown key(s).
+ *              RNP_VERIFY_REQUIRE_ALL_SIGS - require that all signatures (if any) must be
+ *                valid for successful run of rnp_op_verify_execute().
+ * @return RNP_SUCCESS or error code if failed
+ */
+RNP_API rnp_result_t rnp_op_verify_set_flags(rnp_op_verify_t op, uint32_t flags);
+
 /** @brief Execute previously initialized verification operation.
  *  @param op opaque verification context. Must be successfully initialized.
- *  @return RNP_SUCCESS if data was processed successfully and all signatures are valid.
- *          Otherwise error code is returned. After rnp_op_verify_execute()
- *          rnp_op_verify_get_* functions may be used to query information about the
- *          signature(s).
+ *  @return RNP_SUCCESS if data was processed successfully and output may be used. By default
+ *          this means at least one valid signature for the signed data, or successfully
+ *          decrypted data if no signatures are present.
+ *          This behaviour may be overriden via rnp_op_verify_set_flags() call.
+ *
+ *          To check number of signatures and their verification status use functions
+ *          rnp_op_verify_get_signature_count() and rnp_op_verify_get_signature_at().
+ *          To check data encryption status use function rnp_op_verify_get_protection_info().
  */
 RNP_API rnp_result_t rnp_op_verify_execute(rnp_op_verify_t op);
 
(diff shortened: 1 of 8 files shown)


include/rnp/rnp.h
```diff
@@ -2226,8 +2232,8 @@ RNP_API rnp_result_t rnp_op_sign_destroy(rnp_op_sign_t op);
 /* Verification */
 
 /** @brief Create verification operation context. This method should be used for embedded
- *         signatures or cleartext signed data. For detached verification corresponding
- *         function should be used.
+ *         signatures, cleartext signed data or just encrypted data. For the detached signature
+ *         verification should be used function rnp_op_verify_detached_create().
```
antonsviridenko, line 2236: > For the detached signature  verification should be used function rnp_op_verify_detached_create().

Invalid syntax of the sentence. I think this way should be correct:

"For detached signature verification the function
rnp_op_verify_detached_create() should be used."

include/rnp/rnp.h
```diff
@@ -2252,12 +2258,32 @@ RNP_API rnp_result_t rnp_op_verify_detached_create(rnp_op_verify_t *op,
                                                    rnp_input_t      input,
                                                    rnp_input_t      signature);
 
+/**
+ * @brief Set additional flags which control data verification/decryption process.
+ *
+ * @param op pointer to opaque verification context.
+ * @param flags verification flags. ORed combination of RNP_VERIFY_* values.
```
antonsviridenko, line 2265: OR-ed

include/rnp/rnp.h
```diff
@@ -2252,12 +2258,32 @@ RNP_API rnp_result_t rnp_op_verify_detached_create(rnp_op_verify_t *op,
...
+ *              Following flags are supported:
+ *              RNP_VERIFY_IGNORE_SIGS_ON_DECRYPT - ignore invalid signatures for the encrypted
+ *                and signed data. If this flag is set then rnp_op_verify_execute() call will
+ *                succeed and output data even if all signatures are invalid or issued by the
```
antonsviridenko, line 2269: I know that `rnp_op_verify_execute()` can decrypt signed and encrypted messages, but this is a kind of side effect not mentioned in documentation. It's better to explain it explicitly.

include/rnp/rnp.h
```diff
@@ -2252,12 +2258,32 @@ RNP_API rnp_result_t rnp_op_verify_detached_create(rnp_op_verify_t *op,
...
+
 /** @brief Execute previously initialized verification operation.
  *  @param op opaque verification context. Must be successfully initialized.
- *  @return RNP_SUCCESS if data was processed successfully and all signatures are valid.
- *          Otherwise error code is returned. After rnp_op_verify_execute()
- *          rnp_op_verify_get_* functions may be used to query information about the
- *          signature(s).
+ *  @return RNP_SUCCESS if data was processed successfully and output may be used. By default
+ *          this means at least one valid signature for the signed data, or successfully
+ *          decrypted data if no signatures are present.
```
antonsviridenko, line 2281: > or successfully  decrypted data if no signatures are present.

This does not look ok to me, if input data does not contain signatures, then RNP_ERROR_NO_SIGNATURES_FOUND must be returned, isn't it?

ni4: @antonsviridenko When we start to process data, we do not know whether it is expected to be signed, as usually signatures are placed inside of the encrypted layer. User may easily check whether data was signed at all via `rnp_verify_get_signature_count()` and corresponding calls.

andrey-utkin (commented):
I didn't expect rnp_decrypt() to start using `rnp_op_verify_*` and to see op_verify stuff to be documented as responsible for decryption. But kinda makes sense as there's no op_decrypt. Still it's a very unfortunate situation for someone wanting to simply decrypt: "if you want to decrypt, use rnp_op_verify...() but tell it to not really verify". So I can relate to Anton's concerns. Would it be possible to avoid such a surprising mechanism, perhaps at a cost of a little code duplication (like, have also op_decrypt and op_verified_decrypt)? That would give us an opportunity to have better suited defaults.

include/rnp/rnp.h
antonsviridenko: Yeah, but purpose of this functions is to verify signatures, no signatures found => unsuccessful result should be returned.

Btw I've just found that error code `RNP_ERROR_NO_SIGNATURES_FOUND` is not used anywhere in verification code.

antonsviridenko:
@andrey-utkin There are two functions, `rnp_op_verify_execute()` and `rnp_decrypt()`, when signed and encrypted message is passed as an input, both functions have side effects, `rnp_op_verify_execute()` also decrypts the message and `rnp_decrypt()` verifies signatures.

antonsviridenko:
btw, GnuPG (gpgme library) has 3rd separate function for signed and encrypted messages, `gpgme_op_decrypt_verify`

include/rnp/rnp.h
ni4: Agree, thanks!

include/rnp/rnp.h
ni4: Thanks, fixed.

include/rnp/rnp.h
ni4: We may add some flag to be set via `rnp_op_verify_set_flags()`, but do we really need that if user may check the case via the single `rnp_op_verify_get_signature_count()` call?

ni4:
> I didn't expect rnp_decrypt() to start using `rnp_op_verify_*` and to see op_verify stuff to be documented as responsible for decryption. 

Given that both use the same internal code, why not to use it, avoiding the code duplication?

> But kinda makes sense as there's no op_decrypt. Still it's a very unfortunate situation for someone wanting to simply decrypt: "if you want to decrypt, use rnp_op_verify...() but tell it to not really verify". 

OpenPGP protocol doesn't allow to see whether data is just encrypted or encrypted and signed, that's why (except cases like cleartext signature or detached signature) any 'unwrapping' operation in practice would be decryption and/or verification.
So it's just naming, and I don't think that API user will be quite insulted by calling rnp_op_verify_execute() for decryption and/or verification.

> Would it be possible to avoid such a surprising mechanism, perhaps at a cost of a little code duplication (like, have also op_decrypt and op_verified_decrypt)?

I don't think that we need to duplicate code when the problem is just in function naming. If user would not be able to find out which function to call for decryption he most likely don't need to deal with cryptography at all. For the safety reasons.

andrey-utkin (approved):
Thanks for the clarifications.

antonsviridenko (approved):

antonsviridenko:
So, I'm trying to introduce these changes into PHP extension,
looks like API function `rnp_decrypt()` became useless,
I have a  PHP function `rnp_decrypt()` and
I have to use `rnp_op_verify_*` in implementation if I want to decrypt & use  `RNP_VERIFY_IGNORE_SIGS_ON_DECRYPT` flag at once :)

Maybe it is a better idea to introduce something like 
`rnp_decrypt2(rnp_ffi_t ffi, rnp_input_t input, rnp_output_t output, uint32_t flags)`
to be able to pass decryption flags?

ni4:
@antonsviridenko But `rnp_decrypt()` does exactly the same - calls `ret = rnp_op_verify_set_flags(op, RNP_VERIFY_IGNORE_SIGS_ON_DECRYPT);`. Are you on the latest master/release v0.16.1 ?

antonsviridenko:
> But rnp_decrypt() does exactly the same - calls ret = rnp_op_verify_set_flags(op, RNP_VERIFY_IGNORE_SIGS_ON_DECRYPT);

ok then, missed that part

PR #1905, opened 2022-09-16
CI: Do not install botan with OpenSSL backend runners.

### LINKED ISSUE #1904: Do not install botan for OpenSSL-only CI
## Description

Currently `centos8-ossl.yml` and `centos9-ossl.yml` download and install Botan, which takes time and is not needed.

ni4:
Fixes #1904 

Still leaves behind BOTAN_INSTALL and some other traces which pollute logs.


ronaldtse (approved):
Thanks @ni4 ! Merging since this is pure CI.

PR #1910, opened 2022-09-18
Add missing "fingerprint" identifier

### LINKED ISSUE #1909: Description of `rnp_identifier_iterator_create()` does not list "fingerprint" as an acceptable identifier.
## Description

Documentation for  `rnp_identifier_iterator_create()`
lists only 3 acceptable identifiers `("userid", "keyid", "grip")`
https://github.com/rnpgp/rnp/blob/master/include/rnp/rnp.h#L2982
but 4 of them, including missing "fingerprint" are available
https://github.com/rnpgp/rnp/blob/master/src/lib/rnp.cpp#L233

Also "fingerprint" is used in CLI sources, but not covered by tests.

antonsviridenko:
Fixes #1909, adds "fingerprint" to documentation and tests. 

Also seemingly duplicated code is removed from `test_ffi_key_iter`. Code blocks marked by comments "// test with both rings empty" and "// test empty rings" are bit confusing, the same actions are repeated and nothing is changed in testing state between them.

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -2979,7 +2979,7 @@ RNP_API rnp_result_t rnp_key_to_json(rnp_key_handle_t handle, uint32_t flags, ch
  *
  *  @param ffi
  *  @param it pointer that will be set to the created iterator
- *  @param identifier_type the type of identifier ("userid", "keyid", "grip")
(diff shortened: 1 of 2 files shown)


ni4 (approved):
LGTM, thanks!

ronaldtse (approved):
Thank you @antonsviridenko !

PR #1919, opened 2022-10-03
Use AEAD-OCB by default.

### LINKED ISSUE #1918: Use AEAD OCB instead of EAX by default
> The question here is whether to add a third mode (GCM) instead of
restricting the new AEAD thing to just one mode (OCB) and dropping EAX
which is not needed anymore.

Whole email: 
https://mailarchive.ietf.org/arch/msg/openpgp/4jYV0gnSt1JV7avWlx_rY0q6VmY/

ni4:
Fixes #1918

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -2863,11 +2863,11 @@ RNP_API rnp_result_t rnp_op_encrypt_set_cipher(rnp_op_encrypt_t op, const char *
 RNP_API rnp_result_t rnp_op_encrypt_set_aead(rnp_op_encrypt_t op, const char *alg);
 
 /**
- * @brief set chunk length for AEAD mode via number of chunk size bits (refer OpenPGP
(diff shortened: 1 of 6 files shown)


ronaldtse (approved):
Thank you @ni4 !

ni4:
Thanks, @ronaldtse ! Codecov failure is irrelevant as related to the random case (high bits of elgamal key generation).

antonsviridenko:
what about this one

centos9-ossl
> The following tests FAILED:
	247 - cli_tests-Compression (Failed)
Error: Process completed with exit code 8.

ni4:
@antonsviridenko missed that. Let's restart and see whether it helps. There are some occassional password/symmetric decryption issues related to GnuPG. My guess is that those are related to concurrency/some password caching by the gpg-agent.

antonsviridenko (approved):

PR #1922, opened 2022-10-06
CLI: add command --set-filename and do not store env/stdin filename in message

### LINKED ISSUE #1843: CLI: unexpected contents in literal data packet if using non-filename inputs
## Description

While using stdin or env variable placeholder (like `-` or `env:HOME`), rnp stores exactly this value in literal data packet.
More logical and expected by the user would be to have empty string in these cases.

ni4:
- fixes #1843 
- adds `--set-filename` command to override file name in the literal data packet.


ronaldtse (approved):
LGTM @ni4

antonsviridenko:
All debian tests failed on
> 247 - cli_tests-Misc (Failed)

ni4:
@antonsviridenko Thanks, now it should go fine...

antonsviridenko (approved):

PR #1925, opened 2022-10-24
CLI: fix key location failure with 0x prefix/spaces/tabs.

ni4:
...and update tests coverage to avoid regression in the future.


ronaldtse (approved):
Thanks @ni4 @antonsviridenko !

Issue #1927, opened 2022-10-25
CMake errors while building on Windows 10


wyatt-sfu:
## Description
While following the Windows build instructions in [https://github.com/rnpgp/rnp/blob/master/docs/installation.adoc](https://github.com/rnpgp/rnp/blob/master/docs/installation.adoc) I am getting CMake failures with a couple of different error messages. 

## Steps to Reproduce
1. Download and unzip rnp-0.16.2
2. Install vcpkg with `git clone https://github.com/Microsoft/vcpkg.git`
3. Install packages with `vcpkg install bzip2 zlib botan json-c getopt dirent python3[core]`
4. Run `cmake -B . -G "Visual Studio 16 2019" -A x64 -DCMAKE_TOOLCHAIN_FILE=%VCPKG_ROOT%\scripts\buildsystems\vcpkg.cmake -DCMAKE_BUILD_TYPE=Release ../rnp`

## Versions
- vcpkg git hash: 7ad236f60f5f7197e93c4d7f0807622f4899076d
- cmake: 3.16.2
- rnp: 0.16.2

## Results
CMake output: [rnp_cmake_error.txt](https://github.com/rnpgp/rnp/files/9861546/rnp_cmake_error.txt)
I have attached the CMake log as well: [CMakeOutput.log](https://github.com/rnpgp/rnp/files/9861407/CMakeOutput.log)
vcpkg list: [vcpkg_list.txt](https://github.com/rnpgp/rnp/files/9861578/vcpkg_list.txt)

A snippet from the output is:
```
CMake Error at C:/Program Files/CMake/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
  Could NOT find JSON-C (missing: JSON-C_LIBRARY JSON-C_INCLUDE_DIR
  JSON-C_VERSION) (Required is at least version "0.11")
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:393 (_FPHSA_FAILURE_MESSAGE)
  cmake/Modules/FindJSON-C.cmake:92 (find_package_handle_standard_args)
  C:/dev/vcpkg_rnp/scripts/buildsystems/vcpkg.cmake:834 (_find_package)
  src/lib/CMakeLists.txt:33 (find_package)
```

wyatt-sfu:
[...]

PR #1928, opened 2022-10-25
Update installation.adoc for Windows/MSVC instructions.

ni4:
Minor documenation update as per issue #1927


antonsviridenko (commented):
32 bit Windows is still supported, so I think instructions should be more detailed and handle both 32 & 64 bit builds.

ni4:
@antonsviridenko thanks, updated. Does it look good now or I still miss something?

antonsviridenko (approved):

ronaldtse (approved):

ronaldtse:
Thank you!

Issue #1923, opened 2022-10-12
CI: update workflows to use actions based on nodejs-16


ni4:
## Description

> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/cache, actions/cache, actions/checkout

ni4:
From the PR #1930:
> However it cannot be fully resolved due to problems of v2-v3 actions running with the non-host architecture (see https://github.com/actions/checkout/issues/334 and https://github.com/actions/runner/issues/1011).

PR #1930, opened 2022-10-31
CI: Update actions/checkout and actions/cache to v3 where applicable.

ni4:
This PR is partial solution to the issue #1923 
However it cannot be fully resolved due to problems of v2-v3 actions running with the non-host architecture (see https://github.com/actions/checkout/issues/334 and https://github.com/actions/runner/issues/1011).


ronaldtse (approved):
Thank you @ni4 !

ronaldtse:
Merging with one approval since this is CI.

PR #1940, opened 2022-11-10
Add support for hidden recipients during decryption.

### LINKED ISSUE #1275: Support decryption with hidden recipients.

ni4:
This includes quite a bunch of changes, so took long time to implement.
Please review carefully.

Fixes #1275


ronaldtse (approved):
Thank you @ni4 !

src/rnp/rnp.1.adoc
```diff
@@ -332,6 +332,12 @@ Override or set a file name, stored inside of OpenPGP message. +
 By default RNP will store input filename (or empty string for *stdin*/*env* input) in the resulting OpenPGP message during encryption or embedded signing.
 This option allows to override this. Special value *_CONSOLE* may be used for "for your eyes only"-message. Refer OpenPGP documentation for the details.
 
+*--allow-hidden* ::
+Allow hidden recipient support. +
++
+Senders of the encrypted message may wish to hide the recipient by replacing his keyid, stored inside OpenPGP message, with zeroes.
+This could require to try every secret key against such message in order to be able to decrypt, and is disabled by default.
+
```
antonsviridenko, line 340: Need some better wording like 
"Sender of an encrypted message may wish to hide recipient's key by setting a Key ID field to all zeroes. In this case receiver has to try every available secret key, checking for a valid decrypted session key. This option is disabled by default."

Also it's not clear from this description if this option applies to sender or receiver.

ni4: @antonsviridenko Thanks, updated the PR.

antonsviridenko (approved):

ni4:
@ronaldtse @antonsviridenko Should we merge this now?

ronaldtse:
Thank you @ni4! I believe we were just waiting for the tests to pass before.

PR #1941, opened 2022-11-11
Refactor duplicated flag check/update code with a single function call.

ni4:
Less code duplication = better (in almost all cases).


antonsviridenko (approved):

ni4:
@ronaldtse Ping for review. Codecov failure may be ignored as caused by less code lines

ronaldtse (approved):
Thank you @ni4 !

PR #1949, opened 2022-12-09
Support GnuPG Extended Key Format (G23)

maxirmx:

PR #1955, opened 2022-12-19
CLI: add --s2k-iterations --s2k-msec options to rnp.

### LINKED ISSUE #1608: CLI: add to the rnp options --s2k-iterations and s2k-msec
## Description

Currently these options are available only for the rnpkeys, however may be usable for rnp as well during symmetric encryption  - to make encryption stronger, make tests run faster, and so on.

antonsviridenko:
Closes #1608

### DIFF
--- a/src/rnp/fficli.cpp
+++ b/src/rnp/fficli.cpp
@@ -2692,9 +2692,19 @@ cli_rnp_encrypt_and_sign(const rnp_cfg &cfg,
     if (cfg.get_bool(CFG_ENCRYPT_SK)) {
         std::string halg = cfg.get_hashalg();
         std::string ealg = cfg.get_str(CFG_CIPHER);
+        size_t      iterations = cfg.get_int(CFG_S2K_ITER);
+        size_t      msec = cfg.get_int(CFG_S2K_MSEC);
+
+        if (msec != DEFAULT_S2K_MSEC) {
+            if (rnp_calculate_iterations(halg.c_str(), msec, &iterations)) {
+                ERR_MSG("Failed to calculate S2K iterations");
+                goto done;
+            }
+        }
 
         for (int i = 0; i < cfg.get_int(CFG_PASSWORDC, 1); i++) {
-            if (rnp_op_encrypt_add_password(op, NULL, halg.c_str(), 0, ealg.c_str())) {
+            if (rnp_op_encrypt_add_password(
+                  op, NULL, halg.c_str(), iterations, ealg.c_str())) {
                 ERR_MSG("Failed to add encrypting password");
                 goto done;
             }
--- a/src/tests/cli_tests.py
+++ b/src/tests/cli_tests.py
@@ -329,7 +329,7 @@ def rnp_params_insert_aead(params, pos, aead):
             params[pos + 1:pos + 1] = ['--aead-chunk-bits=' + str(aead[1])]
 
 def rnp_encrypt_file_ex(src, dst, recipients=None, passwords=None, aead=None, cipher=None,
-                        z=None, armor=False):
+                        z=None, armor=False, s2k_iter=False, s2k_msec=False):
     params = ['--homedir', RNPDIR, src, '--output', dst]
     # Recipients. None disables PK encryption, [] to use default key. Otherwise list of ids.
     if recipients != None:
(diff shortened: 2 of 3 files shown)


antonsviridenko:
cli_tests will follow

antonsviridenko:
macos runners have some issues with installing dependencies :(

ni4:
@antonsviridenko This commit is aimed to fix macos issue: https://github.com/rnpgp/rnp/pull/1956/commits/f51e1300bc7c9a0ce74fbe30ccb74ba393d36cff
Feel free to cherry pick it.

antonsviridenko:
```
 ERROR:  Error installing bundler:
	The last version of bundler (>= 0) to support your Ruby & RubyGems was 2.3.26. Try installing it with `gem install bundler -v 2.3.26`
	bundler requires Ruby version >= 2.6.0. The current ruby version is 2.5.0.
Error: Process completed with exit code 1.
```
:(

ni4:
@antonsviridenko Yeah, this fails as well :) Working on the solution at my fork.

ni4:
...ah, it is already resolved in the https://github.com/rnpgp/rnp/pull/1949

ronaldtse (approved):
Thank you @antonsviridenko !

ronaldtse:
@ni4 this PR is passing now.

src/rnp/fficli.cpp
```diff
@@ -2692,9 +2692,19 @@ cli_rnp_encrypt_and_sign(const rnp_cfg &cfg,
     if (cfg.get_bool(CFG_ENCRYPT_SK)) {
         std::string halg = cfg.get_hashalg();
         std::string ealg = cfg.get_str(CFG_CIPHER);
+        size_t      iterations = cfg.get_int(CFG_S2K_ITER);
+
+        if (!iterations) {
+            if (rnp_calculate_iterations(
```
ni4, line 2698: We should call `rnp_calculate_iterations()` only if user passed `--s2k-msec` option, as otherwise default value would be used, without need of extra API call.

src/tests/cli_tests.py
```diff
@@ -3715,7 +3719,35 @@ def test_sym_encryption__gpg_to_rnp(self):
     def test_sym_encryption__rnp_to_gpg(self):
         # Encrypt cleartext with RNP and decrypt with GPG
         for size, cipher, z in zip(Encryption.SIZES_R, Encryption.CIPHERS_R, Encryption.Z_R):
-            rnp_sym_encryption_rnp_to_gpg(size, cipher, z)
+            rnp_sym_encryption_rnp_to_gpg(size, cipher, z, 1024)
+
+    def test_sym_encryption_s2k_msec(self):
+        rnp_sym_encryption_rnp_to_gpg(Encryption.SIZES[0], None, Encryption.Z[0], False, 1)
```
ni4, line 3725: We should also check (via `--list-packets`) whether encryption used specified iterations value instead of default.

antonsviridenko: yeah, but 
> s2k iterations: 41943040 (encoded as 244)

it does not necessarily match the value passed on command line
I guess the test should use the same formula to check results?
```
The count is coded into a one-octet number using the following formula:

  #define EXPBIAS 6
      count = ((Int32)16 + (c & 15)) << ((c >> 4) + EXPBIAS);

The above formula is in C, where "Int32" is a type for a 32-bit integer, and the variable "c" is the coded count, Octet 10.
```

ni4: @antonsviridenko just 2-3 sample runs would be enough, the goal of this test would be to make sure that passed value correctly goes from CLI down to all layers. I believe we have more detailed tests, checking this formula, on FFI layer.

src/rnp/fficli.cpp
antonsviridenko: fixed

src/tests/cli_tests.py
antonsviridenko: ok, used `src/tests/s2k-iterations.cpp` as a reference

antonsviridenko:
ping @ni4

ni4 (approved):
LGTM, thanks, @antonsviridenko ! Let's merge it before new CI issues appear :-)

antonsviridenko:
@ronaldtse @ni4 is there any new task I can work on?

PR #1957, opened 2022-12-26
FFI: refactor supported algorithms checks

ni4:
This would allow easier disabling/enabling them via ifdefs in case backend doesn't support certain algos.


ronaldtse (approved):
Thank you @ni4 !

maxirmx (approved):
LGTM

PR #1969, opened 2023-01-13
CI: Bump used ruby version to 3.0

### LINKED ISSUE #1958: CI: Bump used ruby version to 3.0
## Description

Ruby 2.5-2.6 are already EOL (https://www.ruby-lang.org/en/downloads/branches/), and 2.7 is in security maintenance state, which should be EOL in 3 months or so.

ni4:
Fixes #1958

### DIFF
--- a/ci/lib/install_functions.inc.sh
+++ b/ci/lib/install_functions.inc.sh
@@ -12,17 +12,13 @@
 : "${DIST_VERSION_ID:=}"
 
 : "${MINIMUM_CMAKE_VERSION:=3.20.0}"
-: "${MINIMUM_RUBY_VERSION:=2.5.0}"
+: "${MINIMUM_RUBY_VERSION:=3.0.0}"
 
 : "${RECOMMENDED_BOTAN_VERSION:=2.18.2}"
 : "${RECOMMENDED_JSONC_VERSION:=0.12.1}"
 : "${RECOMMENDED_CMAKE_VERSION:=3.20.5}"
 : "${RECOMMENDED_PYTHON_VERSION:=3.9.2}"
-: "${RECOMMENDED_RUBY_VERSION:=2.5.8}"
-# Bundler version to use if Ruby version is less then
-# FALLBACK_BUNDLER_RUBY_VERSION
-: "${FALLBACK_BUNDLER_VERSION:=2.3.26}"
-: "${FALLBACK_BUNDLER_RUBY_VERSION:=2.6.0}"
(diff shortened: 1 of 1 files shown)


maxirmx (changes requested):
I think that the code I have added to support Ruby 2.5 is not required anymore

This can be removed:
```
# Bundler version to use if Ruby version is less then
# FALLBACK_BUNDLER_RUBY_VERSION
: "${FALLBACK_BUNDLER_VERSION:=2.3.26}"
: "${FALLBACK_BUNDLER_RUBY_VERSION:=2.6.0}"
```
and the code that references these FALLBACK_xxx in ```install_bundler()```

antonsviridenko (approved):

ronaldtse:
I have a question: I believe the Ruby version is not really relevant to this repository, but to ruby-rnp. I suggest we remove the Ruby tests from this repo and only run them in ruby-rnp, where we can test 2.7, 3.0, 3.1 and 3.2.

Any thoughts?

maxirmx:
> I have a question: I believe the Ruby version is not really relevant to this repository, but to ruby-rnp. I suggest we remove the Ruby tests from this repo and only run them in ruby-rnp, where we can test 2.7, 3.0, 3.1 and 3.2.
> 
> Any thoughts?

I think it make a lot of sense

ni4:
> I think that the code I have added to support Ruby 2.5 is not required anymore

@maxirmx Thanks, agree with this. Removed and force-pushed.

ni4:
@maxirmx Please see the updated PR.

maxirmx (approved):
LGTM

antonsviridenko:
@ronaldtse I agree, ruby-rnp tests frequently cause troubles, but on the other hand I thought about adding php-rnp tests here too, as we already have ruby tests in this repo 😂

ni4:
> @ronaldtse I agree, ruby-rnp tests frequently cause troubles, but on the other hand I thought about adding php-rnp tests here too, as we already have ruby tests in this repo 😂

No, hell no :-D Actually from the architecture point of view it would be more logical to run such tests from the ruby-rnp/php-rnp repository against the latest RNP release, plus, maybe, some weekly runs to detect incompatibilities early.

ronaldtse:
Yeah let's move away the ruby-rnp tests to ruby-rnp, and the have php-rnp tests at the php-rnp repository. GitHub Actions now provides ways for us to trigger builds so we can have a consistent setup across different language bindings. (@maxirmx can help us with that).

ronaldtse (approved):

Issue #1960, opened 2023-01-04
Support for ```openpgp-s2k3-ocb-aes``` private key protection with openssl backend


maxirmx:
Implement support for ```openpgp-s2k3-ocb-aes``` private key protection with openssl backend.
This feature is used in GnuPG 2.3+ extended key format and is supported/tested with Botan backend.

With openssl backend an attempt to set AAD used by GbuPG causes ```padding check failed``` openssl error.

This is a backlog from (#1949, #1724)

PR #1971, opened 2023-01-16
Fix tests which changed behaviour due to added g23 format support.

ni4:
Not sure why CI run successfully without these changes.


maxirmx:
@ni4 
I noticed it but forgot to fix.  Sorry about it.
For those two failing these tests  ```rnp_tests_get_key_by_id(sec, id)``` returns ```false```, i.e.:  private key files are missing.
Let me fix it

ni4:
@maxirmx Thanks for following up! I'm completely puzzled why these are running correctly locally, but completely fail in GHA.

maxirmx:
> @maxirmx Thanks for following up! I'm completely puzzled why these are running correctly locally, but completely fail in GHA.

Possibly it was related to backend  -- https://github.com/rnpgp/rnp/issues/1960

maxirmx:
@ni4 
I added missing key files, then made some tests conditional because of #1960, then made codecov happy with my changes
It looks like codecov has an issue  with the changes you made to src/librepgp/stream-common.cpp

ni4:
@maxirmx looks like codecov is unhappy since this PR removes bunch of code lines which previously were covered with tests, causing lower overall coverage. So it should be good to go once tests are run succesffully.

maxirmx (approved):
LGTM

ronaldtse (approved):
Thank you @maxirmx @ni4 !

PR #1978, opened 2023-01-28
Add key edit subcommand --set-expire.

### LINKED ISSUE #1132: Add CLI command to update key/subkey expiration time.
## Description

Possible command is:
`rnpkeys --set-expiry <userid/keyid/fingerprint> expiry`.
Where expiry could be in a GnuPG-like format: 1000 (seconds), 1000h (hours), 1000d (days), 20m (monthes), 3y (years).

antonsviridenko:
Closes #1132


ronaldtse (approved):
Thank you @antonsviridenko !

src/rnpkeys/rnpkeys.1.adoc
```diff
@@ -251,6 +251,10 @@ Since version 0.16.0 of RNP generated secret key is stored with bits set to a ne
 however, this may be needed to fix older keys or keys generated by other implementations.
 _KEY_ must specify the exact subkey via keyid or fingerprint.
 
+*--set-expire* _TIME_:::
+Set key expiration time. See the description of the *--expiration* option for possible time formats.
+Setting argument to 0 removes key expiration.
```
ni4, line 256: For me seems more natural here something like '0 value means that key would never expire (which is not recommended due to security reasons)'.

src/tests/cli_tests.py
```diff
@@ -1867,6 +1867,44 @@ def test_additional_subkeys_reuse_password(self):
...
 
+    def test_set_expire(self):
+        # Primary key with empty password
+        ret, _, _ = run_proc(RNPK, ['--homedir', RNPDIR, '--password=',
+                                    '--userid', 'primary_with_empty_password@rnp', '--generate-key'])
+        self.assertEqual(ret, 0, KEY_GEN_FAILED)
+
+        ret, _, err = run_proc(RNPK, ['--homedir', RNPDIR, '--edit-key', '--set-expire', '-1', 'primary_with_empty_password@rnp'])
+        self.assertNotEqual(ret, 0)
+        self.assertRegex(err, r'Failed to set key expiration.')
```
ni4, line 1878: I would recommend to extend test cases with the following checks:
- make sure that expiration didn't change via `--list-packets`
- make sure that expiration is set to exactly same value which was passed via the `--list-packets`.

ni4 (changes requested):
LGTM, just few minor comments to address/discuss.

ni4 (approved):
LGTM, thanks!

PR #1983, opened 2023-02-05
Allow single --edit-key subcommand only.

### LINKED ISSUE #1979: `--edit-key` sub-options should be mutually exclusive
## Description

https://github.com/rnpgp/rnp/blob/9b08ed1d13b79b3c8f2603ecc29b2cd066b70a49/src/rnp/fficli.cpp#L1091
The code above says `at least one` but
depending on the processing order, current code executes only one of the edit key subcommands.

## Steps to Reproduce

Try to pass multiple `--edit-key` subcommands/options.

## Expected Behavior

I think `rnpkeys` should print an error message on attempts to pass more than one `--edit-key` options and the error message above should be updated to address this change.

antonsviridenko:
Closes #1979

### DIFF
--- a/src/rnp/fficli.cpp
+++ b/src/rnp/fficli.cpp
@@ -1105,6 +1105,30 @@ done:
 bool
 cli_rnp_t::edit_key(const std::string &key)
 {
+    int edit_options = 0;
+
+    if (cfg().get_bool(CFG_CHK_25519_BITS)) {
+        edit_options++;
+    }
+    if (cfg().get_bool(CFG_FIX_25519_BITS)) {
+        edit_options++;
+    }
+    if (cfg().get_bool(CFG_ADD_SUBKEY)) {
+        edit_options++;
+    }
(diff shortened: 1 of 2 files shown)


antonsviridenko:
CLI tests will follow, ping @ni4 for preliminary review

antonsviridenko:
maybe there is some better way to implement this...

antonsviridenko:
ping @ni4 @ronaldtse

I don't know what's wrong with codecov/project failing test

ronaldtse (approved):
Thank you @antonsviridenko !

ronaldtse:
I think the codecov issue isn’t important right now, it represents a 0.01% coverage difference.

ni4:
@antonsviridenko In one of the previous test runs one rare codepath with ElGamal key re-generation was hit, so coverage was increased. Now (as well as in my OpenSSL PR) it didn't trigger, so codecov reports less coverage.

ni4 (approved):
LGTM, thanks!

PR #1986, opened 2023-02-09
Add AEAD support to the OpenSSL backend.

### LINKED ISSUE #1642: Implement AEAD algorithms support for OpenSSL backend.
## Description

Currently only CFB algorithms are supported, however sooner or later we'll need to support AEAD as well.

ni4:
This PR adds AEAD encryption/decryption support to the OpenSSL backend.
Currently it is limited only to AES algorithm (Twofish is not supported by OpenSSL at all, so we just miss Camellia cipher), and only to OCB mode. Later is not a problem as latest rfc4880-bis draft (https://www.ietf.org/archive/id/draft-koch-openpgp-2015-rfc4880bis-01.txt) deprecates usage of EAX:
```
EAX mode is deprecated due to the far better properties of the OCB mode.
Implementations may use EAX mode only for decryption of existing data.
```

Currently it has a lot of code duplication with cipher_ossl.cpp, and between Botan and OpenSSL backend. This should be resolved afterwards during the refactoring (let's first have working solution, and nice solution later on).

Fixes #1642


ronaldtse (approved):
Thank you @ni4 !

antonsviridenko (approved):

PR #1991, opened 2023-02-12
Add CLI option --force-weak for weak hashes.

antonsviridenko:
Introduces new CLI option `--force-weak`. From now, using cryptographically weak hashes for key and data signatures is prohibited unless `--force-weak` is present.

This PR reimplements CLI related part of #1524 using `rnp_get_security_rule()`.

Closes #1524

### DIFF
--- a/src/rnp/fficli.cpp
+++ b/src/rnp/fficli.cpp
@@ -2818,6 +2818,36 @@ cli_rnp_setup(cli_rnp_t *rnp)
 }
 
 bool
+cli_rnp_check_weak_hash(cli_rnp_t *rnp)
+{
+    if (rnp->cfg().has(CFG_WEAK_HASH)) {
+        return true;
+    }
+
+    uint32_t security_level = 0;
+
+    if (rnp_get_security_rule(rnp->ffi,
+                              RNP_FEATURE_HASH_ALG,
+                              rnp->cfg().get_hashalg().c_str(),
+                              rnp->cfg().time(),
+                              NULL,
+                              NULL,
+                              &security_level)) {
+        ERR_MSG("Failed to get security rules for hash algorithm \'%s\'!",
+                rnp->cfg().get_hashalg().c_str());
+        return false;
+    }
+
+    if (security_level < RNP_SECURITY_DEFAULT) {
+        ERR_MSG("Hash algorithm \'%s\' is cryptographically weak!",
+                rnp->cfg().get_hashalg().c_str());
+        return false;
+    }
+    /* TODO: check other weak algorithms and key sizes */
+    return true;
+}
+
+bool
 cli_rnp_protect_file(cli_rnp_t *rnp)
 {
     rnp_input_t  input = NULL;
--- a/src/rnpkeys/rnpkeys.cpp
+++ b/src/rnpkeys/rnpkeys.cpp
@@ -53,6 +53,7 @@ const char *usage =
   "    --expiration         Set key and subkey expiration time.\n"
   "    --cipher             Set cipher used to encrypt a secret key.\n"
   "    --hash               Set hash which is used for key derivation.\n"
+  "    --allow-weak-hash    Allow usage of a weak hash algorithm.\n"
   "  -l, --list-keys        List keys in the keyrings.\n"
   "    --secret             List secret keys instead of public ones.\n"
   "    --with-sigs          List signatures as well.\n"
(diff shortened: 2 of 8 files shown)


src/rnpkeys/rnpkeys.cpp
```diff
@@ -53,6 +53,7 @@ const char *usage =
   "    --expiration         Set key and subkey expiration time.\n"
   "    --cipher             Set cipher used to encrypt a secret key.\n"
   "    --hash               Set hash which is used for key derivation.\n"
+  "    --force-weak         Force usage of a weak hash algorithm.\n"
```
ronaldtse, line 56: Is `allow-weak` a better name than `force-weak`?

ni4: Agree, `allow` should look better here, as `force` could be confusing to user (making think that only weak hashes should be used). Also probably it should be extended to `--allow-weak-hashes` since we would also need to have `--allow-weak-ciphers` at some point.

ronaldtse: Ping @antonsviridenko for an update. Thanks!

antonsviridenko: > Also probably it should be extended to --allow-weak-hashes since we would also need to have --allow-weak-ciphers at some point.

Single option can cover both. Is there any reason to handle weak hashes and ciphers separately?

ni4: > Single option can cover both. Is there any reason to handle weak hashes and ciphers separately?

It is always better to give user flexibility. This case could lead to unwanted or unexpected consequence (for instance, user would like to use SHA1 but accidentally would enable IDEA and use it for encryption).

antonsviridenko:
Not sure how my changes could have caused this
```
ERROR: test_encryption_unicode (__main__.Misc.test_encryption_unicode)
  File "/Users/runner/work/rnp/rnp/src/tests/cli_tests.py", line 1987, in test_encryption_unicode
    rnp_encrypt_file_ex(src, dst, [KEY_ENCRYPT])
  File "/Users/runner/work/rnp/rnp/src/tests/cli_tests.py", line 365, in rnp_encrypt_file_ex
    raise_err('rnp encryption failed with ' + cipher, err)
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
TypeError: can only concatenate str (not "NoneType") to str

FAIL: test_core_dumps (__main__.Misc.test_core_dumps)
  File "/Users/runner/work/rnp/rnp/src/tests/cli_tests.py", line 2939, in test_core_dumps
    self.assertEqual(ret, 0)
AssertionError: 1 != 0
```

antonsviridenko:
@ni4 thank you, did not know that cli_tests-Misc was such special.
Yeah, another home directory should help, as in this tests keys with weak hash signatures are generated and they cause error messages when left in the keyring and loaded later in other tests...

src/rnpkeys/rnpkeys.cpp
antonsviridenko: @ni4 maybe `--allow-weak-hash` is better?  `--allow-weak-hashes` is too long and I'll have to shift the whole right  column in the usage text otherwise :)
```
 "    --hash             Specify hash algorithm, used during signing.\n"
 "    --allow-weak-hashesAllow usage of a weak hash algorithm.\n"


```

ni4: @antonsviridenko sure, --allow-weak-hash would work as well.

ronaldtse: Agree!

antonsviridenko:
ping @ni4 @ronaldtse

ronaldtse (approved):
Happy with this. Thank you @antonsviridenko !

src/rnp/fficli.cpp
```diff
@@ -2817,6 +2817,31 @@ cli_rnp_setup(cli_rnp_t *rnp)
...
+bool
+cli_rnp_check_weak_hash(cli_rnp_t *rnp)
+{
+    if (rnp->cfg().has(CFG_WEAK_HASH)) {
+        return true;
+    }
+
+    uint32_t security_level = 0;
+
+    if (!rnp_get_security_rule(rnp->ffi,
```
ni4, line 2829: Within this logic if `rnp_get_security_rule()` call would fail due to some reasons we'll allow to use any hash. Probbaly we should separate checks, failing (or issuing a warning) if `rnp_get_security_rule()` fails.

ni4 (commented):
Overall LGTM, just one minor comment.

ni4 (approved):
LGTM, thanks!

PR #1993, opened 2023-02-12
Fix argument processing issues in CLI tools

antonsviridenko:
Closes #1992

PR #2004, opened 2023-02-26
Release v0.17.0

ni4:


antonsviridenko:
Changes present in #1991 and #1993 could be added in this release too, after approval/

ni4:
@antonsviridenko Yeah, waiting for myself to approve them :) Also #2004 seems good to be included.

ni4:
@ronaldtse I'm thinking whether we should add some CMake option to disable SExp support, so those who don't need that feature would not need to download/build/maintain additional dependency. What do you think?
cc @maxirmx @antonsviridenko

antonsviridenko:
>  I'm thinking whether we should add some CMake option to disable SExp support, so those who don't need that feature would not need to download/build/maintain additional dependency. 

Yes it would be nice to have this option

maxirmx:
> add some CMake option to disable SExp support

I am afraid if you disable SExp library support you will have to provide an alternate SExp parser and have two of them.

ni4:
> I am afraid if you disable SExp library support you will have to provide an alternate SExp parser and have two of them.

SExp parser is needed only for proprietary GnuPG's G10/G23 keyring parsing, which is not part of the 'standard' OpenPGP functionality, defined in RFC. So disabling SExp support would also disable G10/G23 support.

maxirmx:
Ah, "disabling GnuPG compatibility will elimiate the need to support (link) sexp"

ni4:
Actually, not GnuPG compatibility, but support just for GnuPG's proprietary key storage format.

ronaldtse:
@maxirmx the major concern of @ni4 regarding SExp support is about packaging dependencies.

Downstream packagers need to run the build process themselves and additional packages like SExp becomes difficult for them to manage, even with CMake.

Personally I don't think we should remove SExp because compatibility of loading those GnuPG keystores are important. Plus the SExp library is small, plus it is maintained by us.

Is it possible for us to "include" the library without bothering the downstream packagers?

maxirmx:
@ronaldtse @ni4 
We can

1. Include sexp source files into rnp library build, drop sexp library and sexp git project, and distributes sexp source within rnp source package
2. Include sexp source files into rnp source tree, build sexp library separately using CMake ```add_subdirectory```  but drop sexp git project,  and distributes sexp source within rnp source package
3. Link sexp project into rnp source tree as git submodule, build sexp library separately using CMake ```add_subdirectory``` feature and keep sexp project,  and distributes sexp source within rnp source package

moving the opposite way we can 

4. Make sexp full-pledged package and require it for rnp build and runtime environments

I expect that you will select option 3, however, I would talk to some downstream package maintainers if you can reach them.
I am a kind of downstream maintainer for folly and dwarfs since I port/maintain them on platforms not supported by the core teams. Git submodule would be the worst choice for me because it adds burden but lacks flexibility.

ronaldtse:
@maxirmx I don’t like 1/2 and also don’t prefer using a submodule. 

The only real way for us to contact downstream maintainers is either we find them and keep track of them (which seems like we should do), or through public communication like a blog (which we should do too). And good documentation on our README / build instructions on how to build RNP. 

@ni4 what do you think?

maxirmx:
One more option
5. We can keep git structure intact but create source package that has sexp library source 'pre-downloaded'

ronaldtse:
A prepackaged source seems reasonable.

ni4:
During previous releases we were hit with the following issues:
- case when library should be built from source package without Git installed
- case when library should be built on offline system

That's why downloading something would cause issues.  We may update readme/instructions for maintainers, but including SExp sources into `src/lib` would be easier, given that SExp code would not change often.

antonsviridenko:
>4. Make sexp full-pledged package and require it for rnp build and runtime environments

I think this is the right way. But having possibility to built rnp without sexp would be nice.

And sexp package can be useful as a dependency for other projects too, am I right?

maxirmx:
> During previous releases we were hit with the following issues:
> 
> * case when library should be built from source package without Git installed
> * case when library should be built on offline system
> 
> That's why downloading something would cause issues. We may update readme/instructions for maintainers, but including SExp sources into `src/lib` would be easier, given that SExp code would not change often.

@ni4 @ronaldtse 
- You already made this good point here https://github.com/rnpgp/rnp/pull/1943#issuecomment-1315134934 and here https://github.com/rnpgp/rnp/pull/1943#issuecomment-1315328258
- I already fixed it. Current version does not require neither git not Internet access from the build host as explained here https://github.com/rnpgp/rnp/blob/master/docs/installation.adoc#compile-from-source-without-access-to-githubcom
- None of the 5 options suggested above requires git ot Internet access either.
- There is a requirement to put sexp sources on a build host. 
 
 (I am open to any option but I do not understand the criteria)

ni4:
@maxirmx Thanks for following up. Yeah, it doesn't require internet/git access. But would not the requirement to have additional source package make troubles to the maintainers? I.e. all maintainers after this release would need to re-setup their scripts/environment and so on.

ronaldtse:
I believe when building our release tar.gz packages we should already include the SEXP library source code, so that no additional pulls are necessary. The release tar.gz files must already be fully contained and does not require external sources.

ni4:
@maxirmx Do you have an idea why Botan is not found for macOS runners? Could it be related to the #2995?

maxirmx:
> @maxirmx Do you have an idea why Botan is not found for macOS runners? Could it be related to the #2995?

My fault. I forgit to install it from cache.
I will fix it later today.

maxirmx:
@ni4 
https://github.com/rnpgp/rnp/pull/2068
I'm sorry for this silly issue

ni4:
@maxirmx no problem, thanks for solving this!

antonsviridenko (approved):

maxirmx (approved):

ribose-jeffreylau (approved):

Issue #2007, opened 2023-02-28
CI: Botan library handling/versioning  in CI scripts


maxirmx:
Now for CentOS/Fedora CI scripts do the following

- Install botan2, botan2-devel packages  (unconditionally) [from https://github.com/riboseinc/rpm-spec-botan2, version **2.16.1**]
- Build botan from source and link against it (unconditionally) [**2.17.3** on CentOS 7/**2.18.2** on other platforms]

In all CMake scripts we have ```find_package(Botan2``` **2.14.0** ```REQUIRED)``` which is inconsistent with CI shell scripts

So,  two questions
- what is the minimal Botan version we support (want to support) ?
- do we want to use Botan rpm from https://github.com/riboseinc/rpm-spec-botan2 ?

maxirmx:
(for CentOS/Fedora only)
We have CI tests that are run against 2.17.3/2.18.2 (locally built)
We have packaging that is done against 2.16.1/2.19.1/2.19.3
We have packaging test that is done with 2.14.0+ requirement

It may be good good enough but it does not look robust enough. 

We shall test and package against the same versions

**Option 1**
```
CentOS 7,8    against 2.16.1   from ribose repo          [and downgrade version in tests]
CentOS 9      against 2.19.3   from EPEL9               [and upgrade version in tests]
Fedora35      against 2.19.1   from Fedora packages     [and upgrade version in tests]
Fedora36      against 2.19.1   from Fedora packages     [and upgrade version in tests]
```

**Option 2**
```
CentOS 7,8    against 2.17.3   from ribose repo         [and add it to ribose repo]
CentOS 9      against 2.18.2   from ribose repo         [and add it to ribose repo]
Fedora35      against 2.18.2   from ribose repo                               
Fedora36      against 2.18.2   from ribose repo                               
```
(more options are possible)

For me _Option 1_  is better, however, current implemenation looks like outdated _Option 2_

So the question is wheather we want to  change the approach towards _Option 1_ or have _Option 2_ fixed

maxirmx:
UPD  
_We shall test and package against the same versions_  is incorrect

**We shall package against the versions we have tested**

ni4:
I would vote for Option 1, as this would be the case which is used by RNP users.

maxirmx:
ref #1908

ronaldtse:
I agree with Option 1.  Ideally we don't want to maintain a Botan package unless absolutely necessary.

Issue #1929, opened 2022-10-25
Refactor CI: Centos/Fedora workflow


maxirmx:
Target flow:
1. build / run tests  ( for different flavours, OS versions and architectures )
2. build a SRPM, install it and test ( for different flavours, versions and architectures ) 
    (publish SRPM as an artifact)
4. build RPMs  ( for different flavours, versions and architectures )  and then test them 
    (publish RPMs as artifacts)

ronaldtse:
This is correct. Thank you @maxirmx !

maxirmx:
CentOS/Fedora squashed workflow sample: https://github.com/rnpgp/sexp/blob/main/.github/workflows/build-and-test-rh.yml

ronaldtse:
This issue depends on:
* [ ] #1891 
* [ ] #1962

PR #2013, opened 2023-03-15
CentOS/Fedora workflow refactoring

maxirmx:
1. #1929
2. #2007 
3. Favour pre-packaged automake installation over local build
4. Favour pre-packaged Ruby installation over local build with rbenv
5. Create single rnp-devel rpm that contains both shared and static libraries


ni4 (approved):
LGTM, thanks!

PR #2016, opened 2023-03-22
Add test_ffi_decrypt_pk_unlocked

ni4:
Needed to re-check it, so let's update tests as well.


ronaldtse (approved):
Thank you @ni4!

antonsviridenko (commented):
Why is new test function `test_ffi_decrypt_pk_unlocked` not declared in `src/tests/rnp_tests.h`?

ni4:
> Why is new test function `test_ffi_decrypt_pk_unlocked` not declared in `src/tests/rnp_tests.h`?

As far as I remember this was needed when we use `cmocka` for tests. Now with GoogleTest it is not required anymore (and probably we should clean it up).

antonsviridenko (approved):

PR #2017, opened 2023-03-22
Install dnf config manager on CentOS8 container

maxirmx:


maxirmx:
This patch fixes issue on CentOS8
(not sure how it worked without config manager but it did)

ni4 (approved):
LGTM, thanks! Probably something changed in centos8 image, it is updated weekly.

PR #2018, opened 2023-03-22
Retry password request during decryption.

antonsviridenko:


src/librepgp/stream-parse.cpp
```diff
@@ -2107,8 +2107,17 @@ init_encrypted_src(pgp_parse_handler_t *handler, pgp_source_t *src, pgp_source_t
             }
             /* Decrypt key */
             rnp::KeyLocker seclock(*seckey);
-            if (!seckey->unlock(*handler->password_provider, PGP_OP_DECRYPT)) {
-                errcode = RNP_ERROR_BAD_PASSWORD;
+            int            attempts = 3;
+            while (attempts--) {
+                errcode = RNP_ERROR_NO_SUITABLE_KEY;
+                if (!seckey->unlock(*handler->password_provider, PGP_OP_DECRYPT)) {
```
ni4, line 2113: Here and below we should take in account case when there are multiple recipients of the message, and user want to enter password only to the particular one. This may be checked by returning result of password provider call - false or true.

ni4: Also we should make number of password request tries configurable.

ni4 (changes requested):

src/librepgp/stream-parse.cpp
antonsviridenko: > This may be checked by returning result of password provider call - false or true.

In case of a `rnp` cli tool and password provider that reads from the `pass-fd`, it does not check the keyid, so the password and `true` result  would be returned every time.

antonsviridenko: Otherwise this code should still succeed when one of the passwords matches one of the keys, just many "Bad password" messages will be printed on stderr as a side effect.

ni4: > In case of a `rnp` cli tool and password provider that reads from the `pass-fd`, it does not check the keyid, so the password and `true` result would be returned every time.

These changes are mostly not for CLI, but for other FFI users (like TB), to let them repeat the password request if needed.

ronaldtse:
Closing in favour of #2414, which re-implements the same feature on current main.

The base branch here (`7837bb87` and ancestors from 2022) has drifted too far from current main to salvage — the diff against main is effectively a full rewrite of `stream-parse.cpp`. Cherry-picking just the retry logic onto current main would lose most of the surrounding context.

#2414 applies the audit findings from `TODO.rnp-pulls-review/2018-antonsviridenko.md`:

- Drops the per-failure `RNP_LOG("Bad password, try again.")` (the failure is in the errcode that propagates).
- Drops the dead `RNP_ERROR_NO_SUITABLE_KEY` reset at loop top.
- Adds real test coverage (`test_ffi_decrypt_password_retry` covers wrong-then-right, all-wrong, and provider-cancel cases).
- Cleans up the mixed tabs/spaces from the original PR.

The original idea and the user-facing behaviour (re-prompt on a wrong password instead of forcing a re-run) is preserved. Thank you for the original PR — sorry it took a fresh branch to land.

PR #2019, opened 2023-03-23
SEXP library as git submodule

maxirmx:


.github/workflows/debian.yml
```diff
@@ -92,9 +92,20 @@ jobs:
...
 #                                      [https://github.com/actions/runner/issues/1011]
-      - name: Checkout
+      - name: Checkout on x64
+        if: arch == 'x64'
+        uses: actions/checkout@v3
+        with:
+          submodules: true
+
+      - name: Checkout on ia64
+        if: arch == 'ia32'
```
ronaldtse, line 102: @maxirmx it says "Checkout on ia64" but the condition is `ia32`. Is this correct?

maxirmx: It is wrong, thank you, @ronaldtse

maxirmx:
rnp now uses sexp library as git submodule
For build procedures it looks like sexp source is just a subfolder in src.  
We can easily change git submodule to plain subfolder if necessarily.

IMPORTANT NOTE    When release tar is created we shall check that sexp sources are inside it.

ronaldtse (approved):

ni4 (approved):
LGTM, thanks!

PR #2046, opened 2023-04-11
Renamed master to main or added handling of main branch as necessary

maxirmx:


ronaldtse:
@maxirmx please ping me when we are ready to switch to `main`, I will make the change in GitHub repo settings.

maxirmx:
I think it is ready though it is possible that I missed something and further adjustment will be required

When branch is renamed and GitHub repo settings are changed we may need to change settings in codecov and coverity projects as well

ronaldtse (approved):
Thank you @maxirmx ! I will make the branch change after merge.

ronaldtse:
Merging with 1 approval for CI changes.

PR #2053, opened 2023-04-13
Refactored MSys workflow

maxirmx:


ronaldtse (approved):
Thanks @maxirmx ! Please feel free to merge when all green. Technically the only remaining job is fuzzing and is irrelevant to the CI changes here but nonetheless.

maxirmx:
Fuzzers fail to load botan sources from randombit.net
I suggest we change to: https://github.com/randombit/botan/archive/refs/tags/2.16.0.zip

@ni4, what do you think ?

ronaldtse:
@maxirmx I agree. I know @randombit prefers us using the official link but by pointing to GitHub, it might also take away unnecessary load from the official link given wr are using it for CI purposes.

ni4:
Agree, got hit with this yesterday. Also we should probably bump the Botan's version.

randombit:
Keep in mind that the zips on GH are not the official release, in that the release date and git commit are embedded in `version.txt` by the distribution creation scripts. So the zip will identify itself as an unreleased version. But that doesn't matter for CI purposes.

(Website issue seems to be a networking problem, opened a ticket with the hosting company.)

maxirmx:
randombit.net is operational so I a, merging it "as is"
 If we need update/changes to fuzzers workflow please place a separate ticket

PR #2057, opened 2023-04-19
Refactored MacOS workflow to use GHA and homebrew capabilities

maxirmx:
- Refactored MacOS workflow
- Fixed compiler and linker warnings that occured on MacOS only


ronaldtse (commented):
Thanks @maxirmx !

PR #2065, opened 2023-04-26
Add test cases for oss-fuzz false positives.

ni4:
There is a bunch of oss-fuzz errors which relate to the processing of large input. Not sure how to report those as false-positives, but still it would be good idea to have test cases for those. Also fixes one error for fuzz_dump.


antonsviridenko (approved):

antonsviridenko:
> [Add test case for oss-fuzz issue #46208](https://github.com/rnpgp/rnp/pull/2065/commits/0946f53fac665dc1117a10f2dafe7b144d5f2b3b)

@ni4 where are these issue numbers taken from?

ni4:
> > [Add test case for oss-fuzz issue #46208](https://github.com/rnpgp/rnp/pull/2065/commits/0946f53fac665dc1117a10f2dafe7b144d5f2b3b)
> 
> @ni4 where are these issue numbers taken from?

Here they are (opened to public): https://bugs.chromium.org/p/oss-fuzz/issues/list?q=label:Proj-rnp

ni4:
@ronaldtse @maxirmx Could you please review this? Thanks!

maxirmx (approved):
LGTM

PR #2068, opened 2023-04-28
Botan2 MacOS fix

maxirmx:


maxirmx:
CentOS 8 repo is unaccessible and jobs are timing out.

PR #2071, opened 2023-05-02
add INSTALL_STATIC_LIBS buid option

remicollet:
to allow to not install the static library

As RPM packaging Guidelines disallow static libraries
See https://docs.fedoraproject.org/en-US/packaging-guidelines/#packaging-static-libraries

> Packages including libraries SHOULD exclude static libs as far as possible (e.g., by configuring with --disable-static). Applications linking against libraries SHOULD link against shared libraries not static versions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,6 +58,7 @@ option(ENABLE_COVERAGE "Enable code coverage testing.")
 option(ENABLE_SANITIZERS "Enable ASan and other sanitizers.")
 option(ENABLE_FUZZERS "Enable fuzz targets.")
 option(DOWNLOAD_GTEST "Download Googletest" On)
+option(INSTALL_STATIC_LIBS "Install the static library" On)
 # crypto components
 function(tristate_feature_auto NAME DESCRIPTION)
   set(${NAME} Auto CACHE STRING ${DESCRIPTION})
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -422,7 +422,7 @@ endif()
 
 if (BUILD_SHARED_LIBS)
 # both static and shared libraries
-install(TARGETS librnp
+  install(TARGETS librnp
     EXPORT rnp-targets
     LIBRARY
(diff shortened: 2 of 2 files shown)


maxirmx:
@remicollet 
Thank you for the issue. I do not think we need an additional option, it is a bug in rnp packaging that needs to be fixed.

remicollet:
This has to be managed upstream in the CMakeLists else the installed `/usr/lib64/cmake/rnp/ ` files will have reference to the static library and will be unusable.

Downstream (packaging) can remove the *.a file (a common practice), but cmake will have to be fixed, so this new option is really the simple way, have cmake generate the proper file.

Of course, this patch can be used "only" for packaging, but as it has no impact in standard build... and other distros may need them...

maxirmx:
It was incorrect optimization.
 
Existing build script always creates static library. If BUILD_SHARED_LIBS is ON it repackages static library into shared library.
So if someone needs both shared and static the same build procedure is run twice

I will rollback it so that BUILD_SHARED_LIBS does not install static library

remicollet:
> I will rollback it so that BUILD_SHARED_LIBS does not install static library

Thanks (and indeed version 0.16.x was OK)

antonsviridenko (approved):

ronaldtse:
@maxirmx can you help approve if good to go? Thanks.

maxirmx:
@ronaldtse  please give me a day to think. @remicollet made a perfectly valid point about packaging. There are three options how to fix it. We shall consider both downstream requirements and CI performance. We are now running ~110 test workflows and adding a flag can bring it up to 150. 
I am sorry for this delay.

maxirmx:
I propose to rollback to the previous algorithm https://github.com/rnpgp/rnp/pull/2076
It means double run for installation/packaging if both libraries are needed  but it loooks like there are no complains

maxirmx (commented):
I propose to rollback to the previous algorithm -- https://github.com/rnpgp/rnp/pull/2076
It means double run for installation/packaging if both libraries are needed but it loooks like there are no complains

PR #2078, opened 2023-05-09
CMake: fix backend variable description.

ni4:
Now OpenSSL is not only for debug purposes. Should be noticed before the v0.17.0 release.


ronaldtse (approved):

ronaldtse:
Merging since this is documentation only.

PR #2084, opened 2023-05-29
Fix two oss-fuzz issues.

ni4:
These issues are mostly slowdowns, anyway would be good to fix those.
Also now we don't need to run fuzzers for 30 minutes, so reduced running time to 5 minutes.


ronaldtse (approved):
Thank you @ni4 !

maxirmx (approved):

PR #2091, opened 2023-06-14
Remove CFG_USERINPUTFD and use stdin replacement in affected tests.

### LINKED ISSUE #1609: Remove CFG_USERINPUTFD
## Description

CFG_USERINPUTFD is used only for tests to mock user input. There should be better way to do that (updating test cases to use CLI as an option).

### LINKED ISSUE #1857: Check whether we may get rid of CFG_USERINPUTFD in rnpcfg.
## Description

It seems to be used only in tests and is not exposed/used anywhere in the CLI.

antonsviridenko:
Closes #1609 and closes #1857

Affected tests now use stdin temporarily replaced with a pipe read end.

### DIFF
--- a/src/rnp/rnpcfg.h
+++ b/src/rnp/rnpcfg.h
@@ -60,7 +60,6 @@
 #define CFG_PASSFD "pass-fd"      /* password file descriptor */
 #define CFG_PASSWD "password"     /* password as command-line constant */
 #define CFG_PASSWORDC "passwordc" /* number of passwords for symmetric encryption */
-#define CFG_USERINPUTFD "user-input-fd" /* user input file descriptor */
 #define CFG_NUMTRIES "numtries"         /* number of password request tries, or 'unlimited' */
 #define CFG_EXPIRATION "expiration"     /* signature expiration time */
 #define CFG_CREATION "creation"         /* signature validity start */
--- a/src/rnp/rnpcfg.h
+++ b/src/rnp/rnpcfg.h
@@ -54,30 +54,30 @@
 #define CFG_V3_PKESK_ONLY "v3-pkesk-only"     /* disable v6 PKESK */
 #define CFG_ENABLE_V6_SKESK "enable-v6-skesk" /* disable v6 SKESK */
 #endif
-#define CFG_SIGNERS "signers"     /* list of signers */
-#define CFG_HOMEDIR "homedir"     /* home directory - folder with keyrings and so on */
-#define CFG_KEYFILE "keyfile"     /* path to the file with key(s), used instead of keyring */
-#define CFG_PASSFD "pass-fd"      /* password file descriptor */
-#define CFG_PASSWD "password"     /* password as command-line constant */
-#define CFG_PASSWORDC "passwordc" /* number of passwords for symmetric encryption */
-#define CFG_NUMTRIES "numtries"         /* number of password request tries, or 'unlimited' */
-#define CFG_EXPIRATION "expiration"     /* signature expiration time */
-#define CFG_CREATION "creation"         /* signature validity start */
-#define CFG_CIPHER "cipher"             /* symmetric encryption algorithm as string */
-#define CFG_HASH "hash"                 /* hash algorithm used, string like 'SHA1'*/
-#define CFG_WEAK_HASH "weak-hash"       /* allow weak algorithms */
-#define CFG_ALLOW_SHA1 "allow-sha1"     /* allow SHA-1 key signatures */
-#define CFG_S2K_ITER "s2k-iter"         /* number of S2K hash iterations to perform */
-#define CFG_S2K_MSEC "s2k-msec"         /* number of milliseconds S2K should target */
-#define CFG_ENCRYPT_PK "encrypt_pk"     /* public key should be used during encryption */
-#define CFG_ENCRYPT_SK "encrypt_sk"     /* password encryption should be used */
-#define CFG_IO_RESS "ress"              /* results stream */
-#define CFG_NUMBITS "numbits"           /* number of bits in generated key */
-#define CFG_EXPERT "expert"             /* expert key generation mode */
-#define CFG_ZLEVEL "zlevel"             /* compression level: 0..9 (0 for no compression) */
-#define CFG_ZALG "zalg"                 /* compression algorithm: zip, zlib or bzip2 */
-#define CFG_AEAD "aead"                 /* if nonzero then AEAD encryption mode, int */
-#define CFG_AEAD_CHUNK "aead_chunk"     /* AEAD chunk size bits, int from 0 to 56 */
+#define CFG_SIGNERS "signers"       /* list of signers */
+#define CFG_HOMEDIR "homedir"       /* home directory - folder with keyrings and so on */
+#define CFG_KEYFILE "keyfile"       /* path to the file with key(s), used instead of keyring */
+#define CFG_PASSFD "pass-fd"        /* password file descriptor */
+#define CFG_PASSWD "password"       /* password as command-line constant */
+#define CFG_PASSWORDC "passwordc"   /* number of passwords for symmetric encryption */
+#define CFG_NUMTRIES "numtries"     /* number of password request tries, or 'unlimited' */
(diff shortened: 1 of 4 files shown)


antonsviridenko:
```
 258/259 Test #248: rnp_tests.test_fuzz_verify ....................................................   Passed    5.99 sec
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
ci/lib/install_functions.inc.sh: line 676:  2438 Aborted                 (core dumped) "$@"
Error: Process completed with exit code 134.
```

antonsviridenko:
strange, all of the failing centos-and-fedora instances  fail with the same `std::bad_alloc` error in a place not possibly related to the introduced changes

ni4:
@antonsviridenko It could be somehow related to the redirected stdin (and be thrown from GTest side), as other PRs work well.

antonsviridenko:
All failed instances use OpenSSL as a backend...

antonsviridenko:
Any ideas how to debug this?

```
ci/lib/install_functions.inc.sh: line 676:  2438 Aborted                 (core dumped) "$@"
```
Looks like it's something related to Python environment in Centos/Fedora.

ni4:
@antonsviridenko Did you try to revert you commits, related to stdio redirection, and see whether that helps?

antonsviridenko:
> Did you try to revert you commits, related to stdio redirection, and see whether that helps?

I assume that yes, reverting commits will remove CI failures.

I think it could be related to duplicated STDIN not having CLOEXEC flags... And affected platforms run cli_tests through some python wrapper.

antonsviridenko:
Is there any possibility to recreate the same local environment as in failing centos-and-fedora machines?

ni4:
@antonsviridenko Wow, this one get lost somehow. Locally you may just use corresponding Docker container, install prerequisites, copy rnp sources there, build and run tests. Another option is https://github.com/nektos/act but I didn't try it yet and don't know in which way it works and is it possible to properly debug things.

ronaldtse:
Rebased onto latest main (conflict in rnpcfg.h resolved, keeping main's crypto-refresh and allow-sha1 additions). The two commits were squashed into one with the following review fixes folded in:

- Removed a leftover debug fprintf(stderr, "saved_stdin: ...") in generatekey.cpp.
- Stdin is now always restored after rnp_cmd (the original only restored on the success path, leaving the test process's stdin redirected after failures).
- Fixed a C++ goto-over-initialization compile error in the new stdin handling.

Verified locally: full suite passes (304/304 rnp_tests + cli_tests).

ronaldtse (approved):

ronaldtse:
Just need one more approval!

ronaldtse:
Rebased onto current `main` (3e5f675b) — clean rebase, no conflicts. The previous run's typos failure (`priamry`) and the macOS `cli_tests-Encryption` failure were both old-base issues already fixed on main.

Two fix commits added on top (original commit kept intact):
- `72bbcb19` **tests: fix generatekey.cpp build on Windows** — the new stdin-restore code in `ask_expert_details()` uses `fcntl()`/`F_SETFD`/`FD_CLOEXEC`, which do not exist on Windows (MSVC and MinGW); this broke the `rnp_tests` build on all windows-msys2 and windows-native legs. Guarded with `#ifndef _WIN32`, matching the existing precedent in `ffi.cpp`.
- `cf4da353` **style: fix clang-format alignment in rnpcfg.h** — removing `CFG_USERINPUTFD` changed the longest entry of the `CFG_*` define block, so clang-format-11 requires the whole block's trailing comments to be re-aligned (applied verbatim from the CI clang-format output).

Verified locally: full suite passes 282/282 (rnp_tests + all cli_tests), Botan 3 build. The windows-msys2 legs are now green.

**fuzzing job: pre-existing repo-wide failure, not caused by this PR.** It fails at CMake configure: `Could NOT find Botan: Found unsuitable version "3.4.0", but required is at least "3.6.0"` — the oss-fuzz build image still ships Botan 3.4.0 while main requires >= 3.6.0 when crypto-refresh is enabled (`src/lib/CMakeLists.txt:35`). The identical failure occurs on other recent branches (sig-error-strings, fix-cli-log-redaction, ci-fix-main) because google/oss-fuzz#15882 ("[rnp] Update Botan to 3.6.0") is still open — the revert to `google/oss-fuzz@master` (3e5f675b) landed before the upstream PR was merged. Suggest restoring the rnpgp oss-fuzz fork workaround on main until google/oss-fuzz#15882 is merged.

ronaldtse:
Update on the one remaining red leg: `windows-msys2` (ucrt64, botan) — `cli_tests-SignDefault.test_gpg_to_rnp_default_key` fails at `gpg_verify_cleartext` with `gpg: BAD signature from "signing@gpg"`.

Analysis: this is a pre-existing, content-random msys2/GnuPG flake, not caused by this PR:

1. The failing step is **gpg verifying gpg's own cleartext signature** — rnp is not involved in that step at all (`rnp_cleartext_signing_gpg_to_rnp`: gpg --clearsign → rnp --verify (passed) → gpg --verify (BAD)). No stdin is passed to any process on this path (the passphrase is a command-line argument).
2. The signed content is random per run (`random_text()` includes ` `, `\t`, `\r`, `\n`; sizes up to 1 MB), so the failure manifests nondeterministically: the first CI round failed `test_rnp_to_gpg_default_key` (rnp-signed, openssl leg), this round `test_gpg_to_rnp_default_key` (gpg-signed, botan leg), and the middle round was green on **all** msys2 legs including this one.
3. The only delta between that green round and this round is the `clearerr(stdin)` commit, which touches `src/tests/generatekey.cpp` — compiled only into `rnp_tests.exe`, not into the `rnp.exe`/`rnpkeys.exe` binaries driven by cli_tests.py — so it cannot have caused this failure.
4. The same suite flaked the same way on #2296's msys2 leg (failed once, passed on rerun).
5. Not reproducible off-Windows: 500 iterations of gpg --clearsign → gpg --verify over `random_text()`-style content (POSIX and Windows-text-mode-simulated `\r\r\n` content) with GnuPG 2.5.20 → 0 failures. The msys2 leg runs GnuPG 2.4.9; the trigger appears specific to that platform (Windows CR/LF handling of the random content).

Re-running the failed leg. A proper follow-up on main (unrelated to this PR) would be to make `random_text()` deterministic or avoid `\r`/trailing-whitespace edge cases in the cleartext tests.

ronaldtse:
## On the `codecov/project` (-0.07%) failure

The 0.07% coverage drop is a side-effect of the test-infrastructure change in `src/tests/generatekey.cpp` (`ask_expert_details`), not a real regression in library code.

### Why coverage drops

Replacing `CFG_USERINPUTFD` with a `dup2(stdin)` dance introduced several `goto end` error paths in the test helper:

- `dup(STDIN_FILENO)` failure
- `dup2(...)` failure on the replace
- `dup2(...)` failure on the restore
- `fcntl(STDIN_FILENO, F_SETFD, FD_CLOEXEC)` failure (POSIX only)

Each of these error paths is genuine defensive code (matches the existing POSIX-dance idioms elsewhere in the test suite), but none can be exercised without fault injection — `dup`/`dup2`/`fcntl` don't fail on a healthy runner. Hence uncovered branches, hence the small coverage drop.

### Why we should accept it

- The PR's purpose is **test infrastructure cleanup** (removing the `CFG_USERINPUTFD` escape hatch that leaked test affordances into the production CLI's config vocabulary). Coverage on production library code is unchanged.
- The new error paths mirror the POSIX-dance patterns already accepted in sibling test helpers.
- The drop (-0.07%) is well below the typical "real regression" threshold; it's a coverage-line-budget issue, not a behavior issue.
- Mocking `dup`/`dup2`/`fcntl` to exercise the error paths would require either `LD_PRELOAD` shims or a significant test-harness refactor — both out of scope for this cleanup PR.

Recommend treating the codecov check as a known-accepted regression for this PR.

ni4 (approved):
LGTM!

ronaldtse:
Thank you @antonsviridenko for the contribution!

ronaldtse:
Merging with 2 approvals (ni4, ronaldtse). Thank you @antonsviridenko for the contribution, and @ni4 for the review!

PR #2093, opened 2023-06-15
Fix mentioned clang-format version to 11.0.0

ni4:
So it matches CI.


ronaldtse (approved):

ronaldtse:
Merging with one approval since this is only about local environments.

PR #2096, opened 2023-06-18
Allow signing subkeys in rnp_generate_key_ex().

### LINKED ISSUE #1527: rnp_generate_key_ex() doesn't allow to generate signing subkey.
## Description

Currently it defaults to encrypt-only subkey, however it may check supplied algorithm and if it is sign-only then change subkey's usage to sign instead of failing.

antonsviridenko:
Closes #1527

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -5092,8 +5092,14 @@ try {
     if (password && (ret = rnp_op_generate_set_protection_password(subop, password))) {
         goto done;
     }
-    if ((ret = rnp_op_generate_add_usage(subop, "encrypt"))) {
-        goto done;
+    if (pgp_pk_alg_capabilities(subop->crypto.key_alg) & PGP_KF_ENCRYPT) {
+        if ((ret = rnp_op_generate_add_usage(subop, "encrypt"))) {
+            goto done;
+        }
+    } else {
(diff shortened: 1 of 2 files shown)


src/lib/rnp.cpp
```diff
@@ -5092,8 +5092,17 @@ try {
...
-        goto done;
+    if (pgp_pk_alg_capabilities(subop->crypto.key_alg) & PGP_KF_ENCRYPT) {
+        if ((ret = rnp_op_generate_add_usage(subop, "encrypt"))) {
+            goto done;
+        }
+    } else {
+        if ((ret = rnp_op_generate_add_usage(subop, "sign"))) {
+            goto done;
+        }
+        if ((ret = rnp_op_generate_add_usage(subop, "certify"))) {
```
ni4, line 5103: @antonsviridenko 
Not sure whether we should add certification flag by default, as these are different operations. Is it something which GnuPG does for signing subkey?

antonsviridenko: Yes, right, "certify" is set only for primary keys, signing subkeys have signing flag only in GnuPG.

ni4 (approved):
LGTM, thanks!

ronaldtse (approved):

ronaldtse:
Thank you @antonsviridenko !

PR #2099, opened 2023-06-18
CLI: prevent endless loop in rnpkeys expert mode.

### LINKED ISSUE #969: rnpkeys: possible endless loop with endless output.
## Description

Using rnpkeys expert key generation mode with redirected input/output, and incorrect/empty stdin, may become endless producing endless output (just got 30Gb+ log file while running incorrectly configured tests).

Correct behavior would be to ask for input for few times and terminate, and to check whether stdin is opened at all.

antonsviridenko:
Fixes #969

### DIFF
--- a/src/rnpkeys/tui.cpp
+++ b/src/rnpkeys/tui.cpp
@@ -38,6 +38,8 @@
 #include "file-utils.h"
 #include "logging.h"
 
+#define MAX_INPUT_ATTEMPTS 5
+
 /* -----------------------------------------------------------------------------
  * @brief   Reads input from file pointer and converts it securelly to ints
  *          Partially based on ERR34-C from SEI CERT C Coding Standard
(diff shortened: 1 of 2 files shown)


src/rnpkeys/tui.cpp
```diff
@@ -77,6 +79,17 @@ rnp_secure_get_long_from_fd(FILE *fp, long &result, bool allow_empty = true)
     return true;
 }
 
+static bool
+check_attempts(int *attempts)
```
ni4, line 83: It would be more clear to use `int &attempts` here.

ni4:
@antonsviridenko nix fix landed main, so feel free to rebase.

ronaldtse (approved):

ronaldtse:
I've just made the rebase, let's see what happens.

PR #2102, opened 2023-06-21
add option to build with system sexp library

remicollet:
Since version **0.8.5**, libsexp may be build as shared library


remicollet:
@andreasstieger FYI

maxirmx (commented):
It creates CI configuration that is not tested.
I guess making ubuntu job with SYSTEM_LIBSEXP=ON makes sense
I won't ask @remicollet to do it, pls wait for my time slot

remicollet:
@maxirmx notice that Fedora / EPEL (so RHEL, CentOS...) should have it quite soon (I hope)
https://bugzilla.redhat.com/2216484

PR #2107, opened 2023-06-29
Remove duplicate uint-to-buf and buf-to-uint code.

### LINKED ISSUE #2106: Inconsistent usage of write_uint* and STORE*_LE
## Description

In some parts of the codebase we use functions like read_uint/write_uint, however in some other we use functions LOAD32_LE/STORE32_LE from utils.h. This should be consolidated to use write_uint/read_uint functions, moving those to utils.h

ni4:
- Fixes #2106


ronaldtse (approved):
Thank you @ni4 !

antonsviridenko (approved):

PR #2111, opened 2023-07-25
Add support for v5 keys/signatures.

ni4:
This is yet draft PR and now is intended to check added functionality via CI. MUST not be merged!
It includes:
- v5 fingerprint support
- v5 key/sig packets reading/writing
- v5 key and document signatures verification
- decryption of v5 secret keys
- v5 document signature generation for v5 keys

What is going to be added:
- reading of kbx blobs for v5 keys, if it is simple enough
- parsing of Ed448 keys/signatures (but no verification support as Botan misses such functionality at the moment)


src/librepgp/stream-parse.cpp
```diff
@@ -2573,19 +2583,22 @@ process_pgp_source(pgp_parse_handler_t *handler, pgp_source_t &src)
         src_close(&datasrc);
     } else {
         if (handler->ctx->detached) {
-            RNP_LOG("Detached signature expected.");
+            RNP_LOG("Attached signature expected.");
```
antonsviridenko, line 2586: > -Detached
> +Attached


is this correct?

ni4: @antonsviridenko Yeah, some sort of double-check with typo from early times. Noticed when worked on surrounding code.

antonsviridenko (approved):

ronaldtse (approved):
Thank you @ni4 !

PR #2117, opened 2023-08-01
fix ecdsa padding strings for sha3

TJ-91:
The botan strings in `ecdsa_padding_str_for()` were wrong for SHA3-256/512 (see https://github.com/randombit/botan/blob/master/src/lib/hash/sha3/sha3.cpp#L118)

Note: In `botan_alg_map` in hash.cpp the strings are correctly written. T `ecdsa_padding_str_for` could also make use of the mapping if it's made non-static.

### DIFF
--- a/src/lib/crypto/ecdsa.cpp
+++ b/src/lib/crypto/ecdsa.cpp
@@ -132,9 +132,9 @@ ecdsa_padding_str_for(pgp_hash_alg_t hash_alg)
     case PGP_HASH_SHA224:
         return "Raw(SHA-224)";
     case PGP_HASH_SHA3_256:
-        return "Raw(SHA3(256))";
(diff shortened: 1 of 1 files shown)


antonsviridenko:
@TJ-91 thank you for your contribution

ni4 (approved):
LGTM, thanks! We should add some tests for SHA3, but that would be part of another PR.

antonsviridenko:
> LGTM, thanks! We should add some tests for SHA3, but that would be part of another PR.

@ni4 what exactly should be tested for SHA3?

antonsviridenko (approved):

ni4:
@antonsviridenko We should check what is covered from signatures/s2k/symmetric encryption/whatever else, and add tests for not covered cases.

PR #2119, opened 2023-08-21
Do not double-destroy botan's crypto object.

ni4:
As this may cause issues with the latest botan.


ronaldtse (approved):
Thank you @ni4!

antonsviridenko (approved):

PR #2124, opened 2023-09-06
addition of v6 and PQC

falko-strenzke:
This pull request is a mirror of https://github.com/pqc-thunderbird/rnp/pull/35


falko-strenzke:
There are at least two problems: 
- renaming the repositoy
- false positive by CodeQL

Thus I close this PR again and will create it anew.

PR #2127, opened 2023-09-15
Removed tests against gpg 2.3.0 beta

maxirmx:
One of gnupg tags used for testing pints to 2.3.0 beta version of 2021. 
I do not think there is any value to test against it -- 2.3 line is at 2.3.8 know.


ronaldtse:
Two questions:
1. It looks like we still test against 2.3.1 instead of 2.3.8. Is there a reason for that?
2. Should we retain the `beta` tag for GnuPG so that we can test against the latest beta?

maxirmx:
My understanding of https://www.gnupg.org/download/index.html
"Long-time support" :   2.1.x --> 2.2.x      The last tag in git is 2.2.41
"Current line":  2.3.x --> 2.4.x                    The last tag in git is 2.4.3

There were no "beta" tags since 2021.  We can probable test against HEAD of master branch

ni4:
@maxirmx Agree with you. We added tests against beta when were developing AEAD support. Now we don't need that, so may test only against LTS & shipped versions, unless something new arrives. It would be also good to have Dockerfile which would build latest HEAD of GnuPG to allow local testing/debugging, but this is not of high priority at the moment.

ni4 (approved):
LGTM, thanks!

ronaldtse (approved):
Thank you @maxirmx !

PR #2131, opened 2023-09-28
FIx some PQC issues.

ni4:
Among other minor issues this PR adds RNP_EXPERIMENTAL_* defines to the rnp_export.h header, allowing to cut-off by default yet experimental features. Probably better solution would be to remove those function calls from the header at all, however don't see good solution for that at the moment.

### DIFF
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -458,12 +458,24 @@ foreach (prop LINK_LIBRARIES INTERFACE_LINK_LIBRARIES INCLUDE_DIRECTORIES INTERF
 
 endforeach()
 
+set(EXPERIMENTAL_FEATURES "")
+if (ENABLE_CRYPTO_REFRESH)
+  set(EXPERIMENTAL_FEATURES "${EXPERIMENTAL_FEATURES}\n#define RNP_EXPERIMENTAL_CRYPTO_REFRESH\n")
+endif()
+if(ENABLE_PQC)
+  set(EXPERIMENTAL_FEATURES "${EXPERIMENTAL_FEATURES}\n#define RNP_EXPERIMENTAL_PQC\n")
+endif()
+if(NOT EXPERIMENTAL_FEATURES STREQUAL "")
+  message(WARNING "One or more experimental features are enabled. Use it on your own risk.")
+endif()
+
 generate_export_header(librnp
   BASE_NAME rnp
   EXPORT_MACRO_NAME RNP_API
   EXPORT_FILE_NAME rnp/rnp_export.h
   STATIC_DEFINE RNP_STATIC
   INCLUDE_GUARD_NAME RNP_EXPORT
+  CUSTOM_CONTENT_FROM_VARIABLE EXPERIMENTAL_FEATURES
 )
 
 # This has precedence and can cause some confusion when the binary
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -69,6 +69,13 @@
         RNP_LOG_FD(fp, __VA_ARGS__); \
     } while (0)
 
+#if defined(RNP_EXPERIMENTAL_CRYPTO_REFRESH) != defined(ENABLE_CRYPTO_REFRESH)
+#error "Invalid defines combination."
+#endif
+#if defined(RNP_EXPERIMENTAL_PQC) != defined(ENABLE_PQC)
+#error "Invalid defines combination."
+#endif
+
 static pgp_key_t *get_key_require_public(rnp_key_handle_t handle);
 static pgp_key_t *get_key_prefer_public(rnp_key_handle_t handle);
 static pgp_key_t *get_key_require_secret(rnp_key_handle_t handle);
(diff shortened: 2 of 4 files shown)


ni4:
@falko-strenzke what do you think about this solution on hiding PQC/v6 calls from the FFI API?

falko-strenzke:
First of allo sorry for the delayed relpy.
As I understand it, this should force the application that uses RNP to also set the define `RNP_EXPERIMENTAL_CRYPTO_REFRESH` (or `RNP_EXPERIMENTAL_PQC`) in its own code. From our side this is fine.

ni4:
@falko-strenzke Thanks for the reply. It's actually simpler, CMake will set those variables automatically and include into the `rnp_export.h` header, via this patch: https://github.com/rnpgp/rnp/pull/2131/files#diff-7567fde99fcba090ecb1958dd4ab9ba4367c10ba5d444e580cb61c3b404ced3a

antonsviridenko (approved):

ni4:
@maxirmx @ribose-jeffreylau Could you please review this? Thanks!

src/lib/rnp.cpp
```diff
@@ -70,6 +70,13 @@
         RNP_LOG_FD(fp, __VA_ARGS__); \
     } while (0)
 
+#if defined(RNP_EXPERIMENTAL_CRYPTO_REFRESH) && !defined(ENABLE_CRYPTO_REFRESH)
```
maxirmx, line 73: (This is a suggestion to consider)

IMHO it will be much easier to support a single flags
Something like this:

```
#include <iostream>

//   Defintion of CRYPTO_REFRESH options
#define ENABLE_CRYPTO_REFRESH 0
#define ENABLE_EXPERIMENTAL_CRYPTO_REFRESH 1
#define DISABLE_CRYPTO_REFRESH  2

//   Selection of CRYPTO_REFRESH option
#define RNP_CRYPTO_REFRESH RNP_CRYPTO_REFRESH


int main(int, char**)
{
#if RNP_CRYPTO_REFRESH == ENABLE_CRYPTO_REFRESH
    std::cout << "ENABLE_CRYPTO_REFRESH" << std::endl;
#elif RNP_CRYPTO_REFRESH == ENABLE_EXPERIMENTAL_CRYPTO_REFRESH
    std::cout << "ENABLE_EXPERIMENTAL_CRYPTO_REFRESH" << std::endl;
#else
    std::cout << "DISABLE_CRYPTO_REFRESH (" << RNP_CRYPTO_REFRESH << ")" << std::endl;
#endif
return 0;
}
```

src/lib/rnp.cpp
```diff
@@ -70,6 +70,13 @@
...
+#error "Invalid defines combination."
```
maxirmx, line 74: (This is a suggestion to consider)
Pls refer to the note about CRYPTO_REFRESH flags

src/lib/rnp.cpp
ni4: @maxirmx Thanks for noticing this, but it's suggested to work in a bit different way: both defines are for the same purpose, and this check is to make sure that they are in sync. The problem here is that one goes to the `config.h.in`, which is not installed and used only during compilation, and second is needed to be installed and used by another installed header `rnp.h`. This is needed to hide some function declarations from the `rnp.h`. So it's not the best solution, but I didn't find any better idea. Maybe you have some suggestion?

src/lib/CMakeLists.txt
```diff
@@ -458,12 +458,24 @@ foreach (prop LINK_LIBRARIES INTERFACE_LINK_LIBRARIES INCLUDE_DIRECTORIES INTERF
...
 endforeach()
 
+set(EXPERIMENTAL_FEATURES "")
+if (ENABLE_CRYPTO_REFRESH)
+  set(EXPERIMENTAL_FEATURES "${EXPERIMENTAL_FEATURES}\n#define RNP_EXPERIMENTAL_CRYPTO_REFRESH\n")
+endif()
+if(ENABLE_PQC)
+  set(EXPERIMENTAL_FEATURES "${EXPERIMENTAL_FEATURES}\n#define RNP_EXPERIMENTAL_PQC\n")
+endif()
+if(EXPERIMENTAL_FEATURES STREQUAL "")
```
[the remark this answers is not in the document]
ni4: Thanks for noticing, will fix it.

src/lib/rnp.cpp
maxirmx: If it is a check that both defines are in sync I guess it shall be 
```
#if defined(RNP_EXPERIMENTAL_CRYPTO_REFRESH) != defined(ENABLE_CRYPTO_REFRESH)
```

ribose-jeffreylau (commented):
Not a C expert so nothing to add...

src/lib/rnp.cpp
ni4: @maxirmx Indeed, thanks! Re-pushed.

ni4:
Merging as it has 2 approvals and green CI.

PR #2133, opened 2023-10-13
Add C++ FFI wrappers and use them in CLI.

ni4:
Later, once filled up, we could export this header as well for C++ library users.


ni4:
@antonsviridenko @maxirmx @ronaldtse Could you please check whether this approach is good to you? Basically, idea is to add C++ header wrapper around C calls, to make things easier regarding to the memory management (and for those who would like to use library from C++ code). Currently it implements just few functions, but later it could be finalized to the header with full functionality, sitting next to `rnp/rnp.h`. Thanks!

maxirmx (changes requested):
IMHO there is something wrong with error handling

If you look at the typical call sequence  

--> Key::alg --> rnp_key_get_alg -->

```rnp_key_get_alg```  -  returns either ```RNP_SUCCESS``` which is ```NULL``` or catches an exception  and returns ```ffi_exception object``` with detailed info  (```FFI_GUARD``` macro)

```Key::alg```  -  if ```rnp_key_get_alg``` return is not ```RNP_SUCCESS```,   creates another exception with less error info and throws it (```CALL_FFI``` macro)

I think 
```
#define CALL_FFI(func_call)                           \
    do {                                              \
        auto __ret__ = func_call;                     \
        if (__ret__) {                                \
            throw ffi_exception(__ret__, #func_call); \
        }                                             \
    } while (0)
```
shall be somthing like 
```
template<typename Func, typename... Args>
void call_ffi(Func&& func, Args&&... args) {
    auto __ret__ = func(std::forward<Args>(args)...);
    if (__ret__) {
        throw __ret__;
    }
}
```
so that an exceptions returned from ```rnp_key_get_alg``` and similar functions are propagated

It is also very possible that I did not get the idea :)

**Comment by ni4** [comment]
@maxirmx `rnp_do_something()` functions are designed to be C-only and not to throw an exception and just return some error code, that is guarded via `try { ... } FFI_GUARD` , where later expands to
```
static uint32_t
ffi_exception(FILE *fp, const char *func, const char *msg, uint32_t ret = RNP_ERROR_GENERIC)
{
    if (rnp_log_switch()) {
        fprintf(
          fp, "[%s()] Error 0x%08X (%s): %s\n", func, ret, rnp_result_to_string(ret), msg);
    }
    return ret;
}

#define FFI_GUARD_FP(fp)                                                            \
    catch (rnp::rnp_exception & e)                                                  \
    {                                                                               \
        return ffi_exception((fp), __func__, e.what(), e.code());                   \
    }                                                                               \
    catch (std::bad_alloc &)                                                        \
    {                                                                               \
        return ffi_exception((fp), __func__, "bad_alloc", RNP_ERROR_OUT_OF_MEMORY); \
    }                                                                               \
    catch (std::exception & e)                                                      \
    {                                                                               \
        return ffi_exception((fp), __func__, e.what());                             \
    }                                                                               \
    catch (...)                                                                     \
    {                                                                               \
        return ffi_exception((fp), __func__, "unknown exception");                  \
    }

#define FFI_GUARD FFI_GUARD_FP((stderr))
```

maxirmx (approved):

PR #2135, opened 2023-10-23
fix compilation errors on Windows for the PQC and Crypto Refresh code

TJ-91:
I've been building the experimental Crypto Refresh and PQC code on Windows (Visual Studio 2022) and got a few errors.

* On Windows, the compiler complains about a deleted copy constructor so I provide one (via the already-implemented copy-assign constructor).
* I also fixed a minor variable-length array issue.


maxirmx:
@ni4 @ronaldtse 
vcpkg installs botan-3 by default now and windows workflows are configured to use botan-2

There is a piece of code in ```rnp/src/lib/CMakeLists.txt``` that shall handle this case
```
if (CRYPTO_BACKEND_BOTAN3)
  find_package(Botan 3.0.0 REQUIRED)
elseif (CRYPTO_BACKEND_BOTAN)
  find_package(Botan 2.14.0 REQUIRED)
  if(BOTAN_VERSION VERSION_GREATER_EQUAL 3.0.0)
    set(CRYPTO_BACKEND_BOTAN3 1)
  endif()
endif()
```
but build script fails earlier.

maxirmx:
Windows issue with FindBotan has been fixed:
https://github.com/rnpgp/rnp/pull/2138

Issue #1968, opened 2023-01-10
Refactor CI workflows to use GitHub Actions capabilities


ronaldtse:
We should utilize GHA capabilities instead of shell scripts. The goal is to get rid of the ci/ folder.

The tasks are:
* [x] #1929 
* [x] #1964 
* [x] #1965 
* [x] #1966 
* [x] #1967

ronaldtse:
Great question. The challenge is it has been difficult to know what the "local container" is... It has been difficult to keep the set of workflows working across the now diverse platforms we support.

Even if we don't have that container-CI script, we could use a repository fork to test out all the configurations using GitHub Actions?

dewyatt:
The reason I tried to maintain those scripts previously is that it allowed for much faster iteration.
I could do a single local CI task run with sanitizers enabled in a Linux container on a dev machine w/8+ physical CPU cores, and that would cover 99.99% of potential issues in ~10m, allowing me to proceed with some confidence on my work.
That's opposed to pushing to GitHub, waiting in a queue, spinning up 70 containers/VMs on multiple platforms, and waiting ~2h or so.

Anyways, just wanted to make sure you are aware of the tradeoffs.

ni4:
My idea was to have some registry of RNP-related pre-built Docker images, which are periodically refreshed. However still didn't get to this task yet.
This could be done by re-using current CI scripts, and some more stuff over them, which would build Docker images via CD and upload somewhere.

dewyatt:
@ni4 I think that's a good approach so long as there is adequate hosting available for the images and such

ronaldtse:
Good point @ni4 , we did talk about pre-building Docker images. I believe @maxirmx could utilize Docker images (for the platforms that make sense) as part of this current issue.

ni4:
@maxirmx For me it looked simpler - a repository with current CI scripts, and Dockerfiles, one for every setup we need (like centos8-openssl3, centos8-openssl1, centos8-botan, etc.). Each modification to these files should trigger CD which would build corresponding image and upload to the appropriate location. Also this should be done periodically to leave images up-to-date. CI (PR which modify certain Dockerfiles/build scripts) should check whether images are buildable.

If we'll have too many setups, those could be parametrized (like single centos8 Dockerfile, with some ENV parameters inside which would be overwritten by CI/CD and then used in scripts).

[...]

PR #2137, opened 2023-10-26
Finalized CI refactoring

maxirmx:
The goal set in #1968 "to get rid of the ci/ folder" is now achieved 
ci/ folder  now conatains  tests scripts for [shUnit2](https://github.com/kward/shunit2#shunit2)  tests and configuration files for botan

All shell build scripts have been moved to ci-legacy folder to be sure that they are not used in CI.
We can either delete them or keep as needed.


ronaldtse (approved):

ronaldtse:
Thank you @maxirmx !

PR #2142, opened 2023-11-08
Fixes for pqc and crypto refresh

TJ-91:
This includes the following fixes for the Crypto Refresh and PQC code:
* `KMAC256::fixedInfo()` now returns the correct value
* the v3 PKESK algorithm identifier byte is now included in the correct place (for X25519 and PQC)

Thanks to @teythoon for spotting the issues.

---

Further, I fixed a compile error (wrong variable used) for the Crypto Refresh / PQC code and included a copy constructor for the composite algorithms such that it will now build with MSVC/Clang.

A last change has been made to the `salg` parsing in the PKESK code. The variable was undefined (up to a certain point) and I initially set it to `salg = PGP_SA_UNKNOWN;`. In the Crypto Refresh code it is checked whether or not this is AES since v3 PKESK with the new algorithms don't encrypt the session key and requires AES for that case. This part of the logic is a bit messy currently since the `salg` value is defined early when it's not encrypted (can directly be parsed when the PKESKv3 material is parsed), but for other cases it's only available after decryption in `encrypted_try_key()`.


ni4 (approved):
LGTM, thanks! We may safely ignore codecov failure as changed lines just throw exeption on malloc failure.

PR #2158, opened 2023-12-10
Improve retrieving home directory. Fixes #2149.

### LINKED ISSUE #2149: Incorrect user home folder on Windows
## Description

Incorrect user home folder on Windows

## Expected Behavior

When Windows native environment (not MSys, not cygwin) is used user home folder is defined by USERPROFILE environment variable

## Actual Behavior

rnp and rnp tests always use HOME

Practically it means that:
1. rnp cli tests (Python tests) can be executed in MSys environment only because gpg follows HOME/USERPROFILE convention and thus is not compatible with RNP
2. when rnp is executed in native Windows environment it creates key rings not in user home folder but elsewhere (empty string is used as HOME which probably means current folder -- may be wrong)

antonsviridenko:


ronaldtse:
Rebased onto latest main (conflict in cli_tests.py resolved using main's setUp/tearDown HOME restore). One real bug fix folded in: the Windows branch assigned wstr_to_utf8(...).c_str() from a temporary std::string to const char* home, leaving a dangling pointer (use-after-free) — the converted string is now kept alive in a local. Verified locally: without HOME env, rnp falls back to getpwuid and fails only on the missing input file (exit 1), and the full suite passes (304/304).

ronaldtse (approved):

ronaldtse:
Just need one more approval!

ni4 (approved):
LGTM!

ronaldtse:
Thank you @antonsviridenko for the contribution!

ronaldtse:
Merging with 2 approvals (ni4, ronaldtse). Thank you @antonsviridenko for the contribution, and @ni4 for the review!

PR #2160, opened 2023-12-22
Support for designated revokers

### LINKED ISSUE #1785: Support for designated revoker?
Does RNP already support designated revokers?
https://www.rfc-editor.org/rfc/rfc4880#section-5.2.3.15

If you don't support it yet, and if you add it in the future, you could consider an exception for
rnp_key_remove_signatures(RNP_KEY_SIGNATURE_NON_SELF_SIG)

I'm guessing that the intention of this constant is to remove unnecessary signatures, to create a minimized key.
However, I think that a revocation signature from an authorized revocation key probably should be kept (treated as a self signature).

ni4:
This PR add support for designated revoker/revocation key, partially adding support for custom key signature creation.

Fixes #1785


ronaldtse (approved):
Thank you @ni4 !

maxirmx (approved):
lgtm

ni4:
Thanks guys! Having two approvals so merging.

PR #2171, opened 2024-01-05
Don't explicitely include FindOpenSSL.cmake

cedral:
There is no need to do this because [find_package](https://cmake.org/cmake/help/latest/command/find_package.html) is documented to do this for you and vcpkg has a toolchain file that short circuits find_package for OpenSSL. This does not work correctly if you explicitly include FindOpenSSL in your CMakeLists.txt file. 
You can reproduce this issue by installing OpenSSL 1.1.1w from [Shining Light Productions OpenSSL](https://slproweb.com/products/Win32OpenSSL.html) in C:\Program Files\OpenSSL then setting up vcpkg on windows with visual studio and calling 
```
vcpkg install bzip2 zlib json-c openssl dirent getopt
``` 
then try to build rnp with
```
cmake -DCMAKE_TOOLCHAIN_FILE=...\vcpkg\scripts\buildsystems\vcpkg.cmake -DCRYPTO_BACKEND=openssl ..
```
After doing this the include files will be the OpenSSL 3.2 version in vcpkg but the libraries for the debug configuration in the generated visual studio solution will be the ones from OpenSSL 1.1.1 in Program Files. This will of course cause a bunch of undefined symbols at link time. 
Removing the explicit include of FindOpenSSL fixes this and as I mentioned before it shouldn't be necessary to ever do that because [find_package](https://cmake.org/cmake/help/latest/command/find_package.html) will implicitly include that for you.


ni4:
@cedral Thanks for the contribution! Don't you mind if I cherry-pick your commit into the https://github.com/rnpgp/rnp/pull/2172 ? Otherwise you'll need to rebase after that PR is merged as it fixes some CI failures, introduced with new FreeBSD runner version.

ni4:
Closing this as commit was cherry-picked and merged via #2172

PR #2185, opened 2024-01-24
Created workflow to run tests at simulated dates

maxirmx:


maxirmx:
@ni4, I have created a workflow to run at simulated dates
It is unstable at '+1 year' and always fails at '+5 years'
What would we like to do with it ? 
I can look at the issues but not right now.

ni4:
@maxirmx Thanks for this! I'll take care of issues, most likely those are related to exped keys, used in tests or something like that.

.github/workflows/time-machine.yml
```diff
@@ -182,8 +182,7 @@ jobs:
         if:   matrix.date-offset != '0 years'
```
maxirmx, line 182: @ni4  This condition is redundant after your changes

ni4: @maxirmx Thanks for noticing, fixed.

ni4 (approved):
LGTM, thanks!

PR #2187, opened 2024-01-31
Refactor pgp_key_search_t

ni4:
Another one on the C++-transition way.Thought it would be faster...

### DIFF
--- a/src/lib/key-provider.cpp
+++ b/src/lib/key-provider.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, [Ribose Inc](https://www.ribose.com).
+ * Copyright (c) 2017-2024 [Ribose Inc](https://www.ribose.com).
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without modification,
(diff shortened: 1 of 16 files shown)


src/tests/support.cpp
```diff
@@ -823,11 +800,7 @@ rnp_tests_key_search(rnp::KeyStore *keyring, const std::string &uid)
     if (!keyring || uid.empty()) {
         return NULL;
     }
-
-    pgp_key_search_t srch_userid(PGP_KEY_SEARCH_USERID);
-    strncpy(srch_userid.by.userid, uid.c_str(), sizeof(srch_userid.by.userid));
-    srch_userid.by.userid[sizeof(srch_userid.by.userid) - 1] = '\0';
-    return keyring->search(srch_userid);
+    return keyring->search(rnp::KeyUIDSearch(uid));
```
maxirmx, line 803: Copilot suggests a change to improve coverage :)
```
    auto search = rnp::KeyUseridSearch(uid);
    assert(search.name() == "userid");
    assert(search.value() == uid);
    return keyring->search(search);
```

maxirmx (changes requested):

src/lib/key-provider.cpp
```diff
@@ -24,28 +24,223 @@
...
+    case Type::Grip:
+        if (binval.size() != PGP_KEY_GRIP_SIZE) {
+            RNP_LOG("Invalid grip: %s", value.c_str());
+            return nullptr;
+        }
+        pgp_key_grip_t grip;
+        memcpy(grip.data(), binval.data(), grip.size());
+        return create(grip);
+    default:
+        assert(false);
```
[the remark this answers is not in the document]
ni4: This line to be able to detect possible future coding error. Or do you mean absence of RNP_LOG() for default case?

src/tests/support.cpp
ni4: These helper functions are designed to not assert internally - they just to make sure everything is fine on caller level.

src/lib/key-provider.cpp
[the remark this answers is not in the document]
maxirmx: This line is not covered by tests and assert won't work for release builds.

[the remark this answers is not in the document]
ni4: @maxirmx Ok, removed.

src/tests/support.cpp
maxirmx: Ok, let me say it differently. 

There is a class factory ```std::unique_ptr<KeySearch>KeySearch::create(const pgp_key_id_t &keyid)```  and  there is  limited testing of positive scenarious for it. It is possible to add some extra checks to the helper functions to increase coverage.

maxirmx (approved):
LGTM

src/tests/support.cpp
ni4: @maxirmx Got it, however I'm not sure whether we should test internal classes through internal interfaces, while 'officially' exposed is only FFI interface. I'll update FFI tests with some more edge cases.

antonsviridenko:
> Thought it would be faster...

Is it actually faster now? :P

antonsviridenko:
There is some conflict, probably caused by the recent merged PR...

ni4:
> > Thought it would be faster...
> 
> Is it actually faster now? :P

I mean faster transition to C++ from all the anonymous unions, char * and suchlike internal stuff :) I'll rebase and fix errors and ping you back for review, thanks!

ronaldtse (approved):

PR #2221, opened 2024-04-25
Fix for ML-KEM PKESK and adding PQC CLI tests

falko-strenzke:
Fix for ML-KEM PKESK decryption failure.

Added PQC CLI tests that are executed only for RNP build that supports PQC.

### DIFF
--- a/src/lib/pgp-key.cpp
+++ b/src/lib/pgp-key.cpp
@@ -472,6 +472,19 @@ find_suitable_key(pgp_op_t op, pgp_key_t *key, rnp::KeyProvider *key_provider, b
         if (!cur || !cur->usable_for(op)) {
             continue;
         }
+#if defined(ENABLE_PQC)
+        /* prefer PQC over non-PQC. Assume non-PQC key is only there for backwards
+         * compatibility. */
+        if (subkey && subkey->is_pqc_alg() && !cur->is_pqc_alg()) {
+            /* do not override already found PQC key with non-PQC key */
+            continue;
+        }
+        if (subkey && cur->is_pqc_alg() && !subkey->is_pqc_alg()) {
+            /* override non-PQC key with PQC key */
+            subkey = cur;
+            continue;
+        }
+#endif
         if (!subkey || (cur->creation() > subkey->creation())) {
             subkey = cur;
         }
@@ -1300,6 +1313,41 @@ pgp_key_t::has_secret() const noexcept
         return false;
     }
 }
+
+#if defined(ENABLE_PQC)
+bool
+pgp_key_t::is_pqc_alg() const
+{
+    switch (alg()) {
+    case PGP_PKA_KYBER768_X25519:
+        FALLTHROUGH_STATEMENT;
+    case PGP_PKA_KYBER768_P256:
+        FALLTHROUGH_STATEMENT;
+    case PGP_PKA_KYBER1024_P384:
+        FALLTHROUGH_STATEMENT;
+    case PGP_PKA_KYBER768_BP256:
+        FALLTHROUGH_STATEMENT;
+    case PGP_PKA_KYBER1024_BP384:
+        FALLTHROUGH_STATEMENT;
+    case PGP_PKA_DILITHIUM3_ED25519:
+        FALLTHROUGH_STATEMENT;
+    case PGP_PKA_DILITHIUM3_P256:
+        FALLTHROUGH_STATEMENT;
+    case PGP_PKA_DILITHIUM5_P384:
+        FALLTHROUGH_STATEMENT;
+    case PGP_PKA_DILITHIUM3_BP256:
+        FALLTHROUGH_STATEMENT;
+    case PGP_PKA_DILITHIUM5_BP384:
+        FALLTHROUGH_STATEMENT;
+    case PGP_PKA_SPHINCSPLUS_SHA2:
+        FALLTHROUGH_STATEMENT;
+    case PGP_PKA_SPHINCSPLUS_SHAKE:
+        return true;
+    default:
+        return false;
+    }
+}
+#endif
 
 bool
 pgp_key_t::usable_for(pgp_op_t op, bool if_secret) const
--- a/src/lib/pgp-key.cpp
+++ b/src/lib/pgp-key.cpp
@@ -435,7 +435,11 @@ pgp_subkey_set_expiration(pgp_key_t *                    sub,
 }
 
 pgp_key_t *
-find_suitable_key(pgp_op_t op, pgp_key_t *key, rnp::KeyProvider *key_provider, bool no_primary)
+find_suitable_key(pgp_op_t          op,
+                  pgp_key_t *       key,
+                  rnp::KeyProvider *key_provider,
+                  bool              no_primary,
+                  bool              pref_pqc_sub)
 {
     if (!key || !key_provider) {
         return NULL;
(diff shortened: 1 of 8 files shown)


falko-strenzke:
The only failing runners are those that use `Botan head`. The same applies to the PR https://github.com/rnpgp/rnp/pull/2240. So this is apparently not an error in our code. 

@ni4 Is it possible that you review this PR in the weeks to come? We have a few other updates in order to make RNP conforming to the PQC draft version 03, see https://github.com/pqc-thunderbird/rnp/issues/69.

@TJ-91

ni4:
@falko-strenzke Yeah, Botan head introduced changes to FFI aead handling, which is not compatible with our understanding of how it should work :) I'm working on a fix.

Sure, I'll review. Feel free to re-ping me if I'll dive into something other.

falko-strenzke:
@ni4 I just wanted to ask if you currently have a time plan when you will be able to work on this PR. We have one other open PR and a number of other feature branches (listed in [this wiki page](https://github.com/pqc-thunderbird/rnp/wiki/Overview-of-branches) ), for which we would also make PRs as soon as the existing ones move forward.

ni4:
@falko-strenzke sorry, I postponed this because of the issue on vcpkg side, which causes CI to fail (and we cannot merge with failing CI): https://github.com/microsoft/vcpkg/issues/40276
Was trying to workaround it but didn't succeed, will make another attempt if they do not solve it timely.

TJ-91:
@ni4 now all checks pass, can you check if it's ready to merge?

src/lib/pgp-key.cpp
```diff
@@ -471,6 +471,19 @@ find_suitable_key(pgp_op_t op, pgp_key_t *key, rnp::KeyProvider *key_provider, b
         if (!cur || !cur->usable_for(op)) {
             continue;
         }
+#if defined(ENABLE_PQC)
+        /* prefer PQC over non-PQC. Assume non-PQC key is only there for backwards
```
ni4, line 475: I don't think we should force PQC here without any flag set by the user. Imagine somebody generated PQC subkey, but still has to continue communicate with people whose software doesn't support PQC yet. This should be done via some flag or something like this, set via `rnp_op_encrypt` (or just ignored for now).

src/tests/cli_tests.py
```diff
@@ -398,6 +409,7 @@ def rnp_encrypt_and_sign_file(src, dst, recipients, encrpswd, signers, signpswd,
     if ret != 0:
         raise_err('rnp encrypt-and-sign failed', err)
 
+
```
ni4, line 412: Do we need this empty line here?

ni4 (changes requested):

src/lib/pgp-key.cpp
TJ-91: This implementation choice is due to the PQC draft stating in 8.1 Key Preferences:

> When encrypting to a certificate that has both a valid PQ/T and a valid traditional encryption subkey, an implementation SHOULD use the PQ/T subkey only.

The idea is that PQC encryption is favored in order to mitigate store-now-decrypt-later type attacks and the key has an ECC component to hedge against ML-KEM being broken.

If I'm not mistaken, the code will also prefer PQC signature subkeys. For this scenario, I share your concerns, and it should not be chosen by default. It seems I overlooked this when writing the code. Would you be ok with changing the code such that it only prefers PQC-encryption subkeys but not signature subkeys? This should not cause any interoperability problems since in that scenario both the sender and recipient support PQC and other recipients are not affected by the PKESK (it's not addressed to them and they should ignore it).

TJ-91: @ni4 Ok. It's now implemented for the encryption case only, since I think it doesn't make too much sense for signatures. I also added a test for it.

src/tests/cli_tests.py
TJ-91: fixed

TJ-91:
@ni4 sorry for bothering you with this. Before adding the last commit, all checks passed. Now codecov says, there is less code coverage: ` codecov/project Failing after 1s — 83.99% (-0.03%) compared to 47fee37 ` I'm afraid I don't really know how to read the output of that tool and make sense of it.

First, I don't add new untested cases in the new commit. You can check it yourself here: https://github.com/rnpgp/rnp/pull/2221/commits/8e59c0c99816422ccbeafbc34ecaf16a2ce3e971 . So it's unintuitive to me that adding this commit makes codecov fail now.

Second, when looking at https://app.codecov.io/gh/rnpgp/rnp/pull/2221, the overview of changed files suggest that the first 4 files change 0% and the last file actually changes +0.01%, meaning more coverage.

Third, in the indirect changes tab, I get a few lines that are not even remotely correlated to the patch but it seems to affect the code coverage for this patch negatively.

Fourth, the codecov report only lists 2 commits, not sure what this is about, since the PR has 16 commits. 

All in all I'm confused and not sure how to proceed.

ni4:
@TJ-91 No worries, it's codecov who bothers, not you :) At some point I spent a while trying to understand it's behaviour and it's actually quite simple. The coverage decrease is caused by two reasons:
- in cli_tests we generate random input for encryption/decryption, and in some cases it hits the bzip2 edge case, and in some cases (your one) - not. It should be fixed via the issue #2229 
- in CLI code we use function which returns number of days in the month, and once month changes some lines are getting uncovered. I noticed that yesterday and PR #2275 must fix this.

Once these two issues are fixed I'll just ping you to rebase/repush, will this work for you?

TJ-91:
@ni4 yes, of course, thank you!

src/tests/cli_tests.py
```diff
@@ -4605,6 +4620,84 @@ def test_encryption_and_signing(self):
...
+            raise  RuntimeError("test_zzz_encryption_and_signing_pqc: internal error: lengths of test data arrays matching")
+        # Generate multiple keys and import to GnuPG
+        verified_algo_nums = False
+        for uid, algo, param, passwd in zip(USERIDS, ALGO, ALGO_PARAM, passwds):
+            stdout = rnp_genkey_pqc(uid, algo, param, passwd)
+            if not verified_algo_nums:
+                self.verify_pqc_algo_ui_nb_to_algo_ui_str(stdout, algo_ui_exp_strs)
+                verified_algo_nums = True
+
+        #gpg_import_pubring()
```
ni4, line 4675: Do we need to have these?

ni4 (changes requested):
Please see the review comments. Not major ones, but still would be worth fixing.

src/tests/cli_tests.py
```diff
@@ -4605,6 +4620,84 @@ def test_encryption_and_signing(self):
...
+                char_list_expected = [c for c in expected_line if re.match(re_patt_for_algo, c)]
+                char_list_actual = [c for c in line if re.match(re_patt_for_algo, c)]
+                if char_list_expected == char_list_actual:
+                    found_this_entry = True
+                    break
+
+            if not found_this_entry:
+                raise RuntimeError("did not match the expected UI choice for algorithm: " + expected_line)
+
+    """ zzz_ prefix makes it the last test. This is a workaround against a gnupg import error with the
```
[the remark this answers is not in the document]
TJ-91: Good suggestion, I changed it

src/tests/cli_tests.py
ni4: Nope, just commented out code lines.

TJ-91: Ok great, then I think it's ready. I just force-pushed a minor correction and hopefully the pipeline has no hiccups

ni4 (approved):
LGTM, thanks!!

maxirmx (approved):
LGTM

ni4:
Having two approvals let's finally merge this. Thanks all!

PR #2224, opened 2024-04-28
Move duplicated parts from symmetric*.cpp to separate file

### LINKED ISSUE #2216: Consolidate duplicate symmetric encryption code to the separate file.
## Description

There is a bunch of functions which implement the same functionality in symmetric.cpp and symmetric-ossl.cpp.
Those should be moved to separate file, like symmetric-common.cpp.

desvxx:
fixes #2216


ni4 (changes requested):
It would be also good to move common code of pgp_cipher_cfb_encrypt/pgp_cipher_cfb_decrypt, introducing inline function `pgp_cipher_encrypt_block()` with backend-based implementation.

ni4 (approved):
LGTM, thanks!

ronaldtse (approved):

ronaldtse:
Thank you @desvxx !

PR #2230, opened 2024-05-08
Ignore pushes to version.txt for most of the CI

### LINKED ISSUE #2226: Add version.txt to paths, ignored by tests-based CI.
## Description

...and maybe add to lint test whether it's contents are corresponding to versioning style.
Also we could add some linter which checks spelling in *.md and *.adoc files.

ni4:
And check version.txt contents if it is changed.
Fixes #2226

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -28,7 +28,7 @@ freebsd_instance:
 task:
   name: build
   only_if: $CIRRUS_BRANCH == 'main' || $CIRRUS_BRANCH =~ 'release/.*' || $CIRRUS_PR != ''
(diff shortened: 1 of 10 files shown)


ronaldtse (approved):
Thanks @ni4 ! Perhaps @ribose-jeffreylau can have a last look...

ni4:
@ronaldtse Thanks! And I should also make newly added CI check work :-D Seems something went wrong with version.txt check

ribose-jeffreylau (approved):
Thanks @ni4!

Is it correct to say that `nix.yml` and `ubuntu.yml` are left out because we want them to run when `version.txt` changes?

ni4:
> Thanks @ni4!
> 
> Is it correct to say that `nix.yml` and `ubuntu.yml` are left out because we want them to run when `version.txt` changes?

Ubunty was left to have at least one runner in case something may go wrong due to changes commited, nix was just overlooked :) But let's left it as well.

ni4:
All is green + two approvals, so merging.

PR #2231, opened 2024-05-10
Clarify text in rnp(1)

dkg:
These are minor textual cleanups.


ni4 (approved):
Thanks, @dkg !

ni4:
Merging with the single approval as documentation-only.

PR #2236, opened 2024-05-22
Ad LC_RPATH record while building for macOS.

### LINKED ISSUE #2153: Since XCode 15.0 locally built RNP is unable to locate rnp.dylib
## Description

Since XCode 15.0 or somewhere close dylib doesn't add `/usr/local/lib` to default search path, which causes the following error:

```
/usr/local/bin/rnp --version
dyld[20382]: Symbol not found: _rnp_backend_string
  Referenced from: <51E82EE0-691A-3814-BD0F-F3202DC0BE2C> /usr/local/bin/rnp
  Expected in:     <no uuid> unknown
Abort trap: 6
```

So we should consider to replace rpath from `@rpath/librnp.0.dylib` to `@rpath/../lib/librnp.0.dylib` (or any better solution if any, didn't dig deep to it).
Right come could be overriden by `DYLD_LIBRARY_PATH=/usr/local/lib rnp --version`.

More details are available here: https://developer.apple.com/forums/thread/737920

ni4:
Fixes #2153

### DIFF
--- a/ci/tests/ci-tests.sh
+++ b/ci/tests/ci-tests.sh
@@ -117,9 +117,8 @@ test_supported_features() {
         library_path="${BOTAN_INSTALL}/$so_folder:${JSONC_INSTALL}/$so_folder:${RNP_INSTALL}/$so_folder"
     fi
 
-    if [[ "$OSTYPE" == darwin* ]]; then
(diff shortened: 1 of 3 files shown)


maxirmx (approved):
LGTM

antonsviridenko (approved):

PR #2240, opened 2024-05-28
Be compatible to the updated botan's behaviour related to botan_cipher_update().

### LINKED ISSUE #2245: Be compatible with update Botan 3.5 AEAD decryption behaviour.
## Description

Since Botan 3.5 during AEAD decryption Botan may not process the whole input chunk, which we do not support yet.
See the discussion here: https://github.com/randombit/botan/pull/3951#issuecomment-2136990858

ni4:
Looks like this commit changes behaviour to which we were stick for years:
https://github.com/randombit/botan/pull/3951/commits/83368355cd5ffc0ff1a21d2e6e9b90651179d307

Fixes #2245


ni4:
@maxirmx Don't you mind that I picked here your commit, fixing macOS/Python issue?

maxirmx (approved):
LGTM

src/lib/crypto/symmetric.h
```diff
@@ -213,10 +213,12 @@ bool pgp_cipher_aead_start(pgp_crypt_t *crypt, const uint8_t *nonce, size_t len)
  *             len bytes
  *  @param in buffer with input, cannot be NULL
  *  @param len number of bytes to process. Should be multiple of update granularity.
+ *  @param read number of bytes read and processed, in rare cases could be less then len.
  *  @return true on success or false otherwise. On success exactly len processed bytes will be
  *          stored in out buffer
  */
-bool pgp_cipher_aead_update(pgp_crypt_t *crypt, uint8_t *out, const uint8_t *in, size_t len);
+bool pgp_cipher_aead_update(
+  pgp_crypt_t &crypt, uint8_t *out, const uint8_t *in, size_t len, size_t &read);
```
desvxx, line 221: Do we need to change pointer to reference here? It will be weird from user perspective if other functions use pointer but this one not.

src/librepgp/stream-def.h
```diff
@@ -54,7 +54,7 @@
 #define ST_FROM ("From")
 
 /* Preallocated cache length for AEAD encryption/decryption */
-#define PGP_AEAD_CACHE_LEN (PGP_INPUT_CACHE_SIZE + PGP_AEAD_MAX_TAG_LEN)
+#define PGP_AEAD_CACHE_LEN (PGP_INPUT_CACHE_SIZE + 2 * PGP_AEAD_MAX_TAG_LEN)
```
desvxx, line 57: Just curious, why 2?

desvxx: Found my answer in stream-write.cpp :)

desvxx (approved):

src/librepgp/stream-def.h
ni4: Yeah, this is to include the tag for the last block + the final tag after the encrypted stream.

ni4:
Two approvals so merhing this. Thanks all!

PR #2246, opened 2024-06-18
Allow zero-length input in rnp_input_from_memory()

### LINKED ISSUE #2173: Encryption and decryption of zero length message raise error
## Description

I encountered an error when trying to encrypt and decrypt a zero length message using the rnp library. 
The steps to reproduce the error are as follows

## Steps to Reproduce

1. Run the generate.c example from https://github.com/rnpgp/rnp/blob/main/src/examples/generate.c
2. Run the encrypt.c example from https://github.com/rnpgp/rnp/blob/main/src/examples/encrypt.c

Ensure  ("Encryption succeeded. Encrypted message written to file encrypted.asc")

1. Change line 42 of encrypt.c from `const char *     message = "RNP encryption sample message";` to `const char *     message = "";`
2. Run the encrypt.c example again and see the error message: `failed to create input object`
3. Create an empty file with `touch myEmptyFile`
4. Encrypt the empty file with rnp or gnupg
5. Decrypt the encrypted file to memory and see the error message: `failed to create input object`

## Expected Behavior

I expected the encryption and decryption of a zero length message to succeed without any error, or at least to fail gracefully with a meaningful error message.

## Actual Behavior

The encryption and decryption of a zero length message failed with a cryptic error message: `failed to create input object`

### More info
Notice that the error happens when encrypt from buffer, ot decrypt to buffer. 
File to file, is working fine.

desvxx:
fixes #2173

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -2053,11 +2053,15 @@ FFI_GUARD
 rnp_result_t
 rnp_input_from_memory(rnp_input_t *input, const uint8_t buf[], size_t buf_len, bool do_copy)
 try {
-    if (!input || !buf) {
+    if (!input) {
+        return RNP_ERROR_NULL_POINTER;
+    }
+    if (!buf && buf_len) {
         return RNP_ERROR_NULL_POINTER;
     }
     if (!buf_len) {
-        return RNP_ERROR_SHORT_BUFFER;
+        // prevent malloc(0)
+        do_copy = false;
     }
     *input = new rnp_input_st();
     uint8_t *data = (uint8_t *) buf;
--- a/src/tests/ffi-enc.cpp
+++ b/src/tests/ffi-enc.cpp
@@ -513,7 +513,7 @@ TEST_F(rnp_tests, test_ffi_encrypt_set_cipher)
     rnp_ffi_destroy(ffi);
 }
 
-TEST_F(rnp_tests, test_ffi_encrypt_set_cipher_empty_buffer)
+TEST_F(rnp_tests, test_ffi_encrypt_empty_buffer)
 {
     /* setup FFI */
     rnp_ffi_t ffi = NULL;
@@ -521,8 +521,6 @@ TEST_F(rnp_tests, test_ffi_encrypt_set_cipher_empty_buffer)
     /* create input + output */
     rnp_input_t input = NULL;
     const char *plaintext = "";
-    assert_rnp_success(rnp_input_from_memory(&input, NULL, 0, false));
-    assert_rnp_success(rnp_input_destroy(input));
     assert_rnp_success(
       rnp_input_from_memory(&input, (const uint8_t *) plaintext, strlen(plaintext), false));
     rnp_output_t output = NULL;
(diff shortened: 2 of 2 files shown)


src/lib/rnp.cpp
```diff
@@ -2099,9 +2099,6 @@ try {
     if (!input || !buf) {
```
ni4, line 2099: Please also allow to work with NULL input buffer and zero input size, and add the corresponding test.

src/tests/ffi-enc.cpp
```diff
@@ -381,7 +381,7 @@ TEST_F(rnp_tests, test_ffi_encrypt_set_cipher)
     assert_rnp_failure(
       rnp_input_from_memory(NULL, (const uint8_t *) plaintext, strlen(plaintext), false));
     assert_rnp_failure(rnp_input_from_memory(&input, NULL, strlen(plaintext), false));
-    assert_rnp_failure(rnp_input_from_memory(&input, (const uint8_t *) plaintext, 0, false));
+    assert_rnp_success(rnp_input_from_memory(&input, (const uint8_t *) plaintext, 0, false));
```
ni4, line 384: Please also make sure that encrypting/signing of zero-size input works as well.

ni4 (changes requested):
Please see the comments.

src/lib/rnp.cpp
```diff
@@ -2096,11 +2096,12 @@ FFI_GUARD
 rnp_result_t
 rnp_input_from_memory(rnp_input_t *input, const uint8_t buf[], size_t buf_len, bool do_copy)
 try {
-    if (!input || !buf) {
+    if (!input || (!buf && buf_len > 0)) {
```
ni4, line 2099: It would be logical to replicate behaviour of `init_mem_src()` here, which allows null buffer for the zero-size input.

src/tests/ffi-enc.cpp
```diff
@@ -370,20 +370,29 @@ TEST_F(rnp_tests, test_ffi_encrypt_pass_provider)
     rnp_ffi_destroy(ffi);
 }
 
-TEST_F(rnp_tests, test_ffi_encrypt_set_cipher)
+class ffi_encrypt_tests : public rnp_tests, public testing::WithParamInterface<const char *> {
+};
+
+TEST_P(ffi_encrypt_tests, test_ffi_encrypt_set_cipher)
```
ni4, line 376: Could you please create a new test case instead of heavily modifying this one?

src/tests/ffi-enc.cpp
```diff
@@ -370,20 +370,29 @@ TEST_F(rnp_tests, test_ffi_encrypt_pass_provider)
...
+    const char * plaintext = GetParam();
+    const size_t plaintext_size = plaintext ? strlen(plaintext) : 0;
+    const char * decrypted =
+      plaintext ? plaintext : ""; // NULL input gives us "" when decrypted
     assert_rnp_failure(
-      rnp_input_from_memory(NULL, (const uint8_t *) plaintext, strlen(plaintext), false));
-    assert_rnp_failure(rnp_input_from_memory(&input, NULL, strlen(plaintext), false));
-    assert_rnp_failure(rnp_input_from_memory(&input, (const uint8_t *) plaintext, 0, false));
+      rnp_input_from_memory(NULL, (const uint8_t *) plaintext, plaintext_size, false));
+    if (plaintext == NULL) { // Corner case
```
ni4, line 389: We prefer `if (!plaintext)` notation in such cases.

src/tests/ffi-enc.cpp
```diff
@@ -370,20 +370,29 @@ TEST_F(rnp_tests, test_ffi_encrypt_pass_provider)
...
+        assert_rnp_failure(rnp_input_from_memory(&input, NULL, 5, false));
+    }
+    assert_rnp_success(rnp_input_from_memory(&input, (const uint8_t *) plaintext, 0, false));
```
ni4, line 392: Will not this interfer with the previous code line as plaintext is null?

ni4 (changes requested):

ni4 (approved):
LGTM, thanks!

maxirmx (approved):
OK

PR #2258, opened 2024-08-02
Improve documentation on a bunch of rnp_op_verify_t calls.

### LINKED ISSUE #2257: Missing destructor for rnp_op_verify_signature_t ?
## Description

Most functions in RNP return opaque structures that need to be destroyed by the user using a *_destroy function. 
It seems however that rnp_op_verify_signature_t doesn't have a corresponding destructor (although it is a pointer).

Is that on purpose?

ni4:
Fixes #2257


maxirmx (approved):
LGTM

ni4:
Merging with single approval as documentation-only.

PR #2261, opened 2024-08-15
fix key hashing for v5 / v6 signatures

TJ-91:
Justus pointed out that the hashing of keys is implemented incorrectly for RNP and some other implementations, see https://mailarchive.ietf.org/arch/msg/openpgp/PyP-XDv0VM5bYPX1Iq41-Oyytds/
This concerns computing v5 and v6 signatures over keys.

This PR fixes that. I separated signature and fingerprint computation logically, but the same code is used in the background. I also added a check that we don't accidently hash too large a key.

### DIFF
--- a/src/librepgp/stream-sig.cpp
+++ b/src/librepgp/stream-sig.cpp
@@ -42,26 +42,28 @@
 #include "stream-armor.h"
 #include "pgp-key.h"
 #include "crypto/signatures.h"
+#include <cassert>
 
 #include <time.h>
 
 void
-signature_hash_key(const pgp_key_pkt_t &key, rnp::Hash &hash)
+signature_hash_key(const pgp_key_pkt_t &key, rnp::Hash &hash, pgp_version_t pgpver)
 {
     if (!key.hashed_data) {
         /* call self recursively if hashed data is not filled, to overcome const restriction */
         pgp_key_pkt_t keycp(key, true);
         keycp.fill_hashed_data();
-        signature_hash_key(keycp, hash);
+        signature_hash_key(keycp, hash, pgpver);
         return;
     }
 
-    switch (key.version) {
+    switch (pgpver) {
     case PGP_V2:
         FALLTHROUGH_STATEMENT;
     case PGP_V3:
         FALLTHROUGH_STATEMENT;
     case PGP_V4: {
+        assert(key.hashed_len < ((size_t) 1 << 16));
         uint8_t hdr[3] = {0x99, 0x00, 0x00};
         write_uint16(hdr + 1, key.hashed_len);
         hash.add(hdr, 3);
@@ -69,6 +71,7 @@ signature_hash_key(const pgp_key_pkt_t &key, rnp::Hash &hash)
         break;
     }
     case PGP_V5: {
+        assert(key.hashed_len < ((size_t) 1 << 32));
         uint8_t hdr[5] = {0x9A, 0x00, 0x00, 0x00, 0x00};
         write_uint32(hdr + 1, key.hashed_len);
         hash.add(&hdr, 5);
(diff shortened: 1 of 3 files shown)


TJ-91:
@ni4 I'm not sure why a windows and Cirrus CI test fails. Does the problem still exists where tests need to be restarted occasionally?

Also codecov complains about some code that is not tested. Are we required to provide full test coverage for new code? It's not always trivial. For example, [one failure case I added](https://github.com/rnpgp/rnp/pull/2261/files#annotation_26580547298) tests whether the length of a key is longer than what can be encoded in 4 length octets, i.e., 4 gigabyte.

ni4:
> @ni4 I'm not sure why a windows and Cirrus CI test fails. Does the problem still exists that tests need to be restarted occasionally?

Yeah, that could be caused by two reason: GnuPG is not designed to run in batch/multiple threads and it seem to fail occasionally in tests. Also there is issue with encryption to multiple password which in some cases not well handled by GnuPG (cannot find now, but we should have issue for that).

> 
> Also codecov complains about some code that is not tested. Are we required to provide full test coverage for new code? It's not always trivial. For example, [one failure case I added](https://github.com/rnpgp/rnp/pull/2261/files#annotation_26580547298) tests whether the length of a key is longer than what can be encoded in 4 length octets, i.e., 4 gigabyte.

Codecov requires the patch coverage to be not smaller than overall project coverage, and sometimes that get tricky. In some cases which should never happen, like checks for memory failures, you may use LCOV_EXCL_LINE or LCOV_EXCL_START/LCOV_EXCL_END to disable check for certain code lines.

TJ-91:
So what would you suggest how can I best satisfy codecov here? I agree with the general idea to ideally test every line but for most of the changes it'll be a bit tricky and I suppose a typical test case would fail somewhere earlier already.

Let's look for example at the codecov finding `src/librepgp/stream-sig.cpp#L87-L89`. I test for key version, and for this I use a switch-case statement. The default case will abort with some error message. If the switch-case statement is deep enough in the library, some earlier switch-case statement's default case will be triggered before I get to this code. For example, I can't see how a version "123" certificate would be getting that deep into the library code.

In summary, I'm not sure if it's reasonable to make the effort here and some more sophisticated testing utility framework to help write the specific test cases might be needed if that level of testing is desired.

Would you be fine with just using `LCOV_EXCL_LINE` for all reported lines?

src/librepgp/stream-sig.cpp
```diff
@@ -45,18 +45,51 @@
...
 
-    switch (key.version) {
+    /* first check that we can actually fit the hashed length in the 2 or 4 byte fields */
+    switch (key_or_sig_version) {
+    case PGP_V2:
+        FALLTHROUGH_STATEMENT;
+    case PGP_V3:
+        FALLTHROUGH_STATEMENT;
+    case PGP_V4:
+        if (key.hashed_len > ((size_t) 1 << 16) - 1) {
```
ni4, line 68: Given that we cannot have key artifacts with length >65535 at the moment (we will not parse it nor generate) we could add just an assert() here and below. The price of mistake would be just an incorrect signature.

ni4 (changes requested):

ni4:
@TJ-91 Please see the review comments. Let's just keep things simpler :)

TJ-91:
@ni4 I hope now it's simple enough :) I think I need to restart tests for the macos test (fails `cli_tests-Encryption`) but I'm not sure for the codecov test again. Do I need to wait on your PRs here as well?

ni4:
@TJ-91 PRs which fix CI are finally merged, so could you please rebase this on main to get it merged? Thanks!

TJ-91:
@ni4 Passes all checks now, thanks!

maxirmx (approved):
LGTM

ni4:
Mergin with two approvals. Thanks all!

Issue #2150, opened 2023-11-17
Non-portable paths in rnp client tests (Python)


maxirmx:
## Current state and the issue

gpg built for Msys environment 'understands' MSys paths only, i.e.: '''C:\Program Files\PostgreSQL\12\'''  shall be ```/c/Program Files/PostgreSQL/12/'''

gpg build for Windows environment works with true Windows paths only, i.e.:  ```C:/Program Files/PostgreSQL/12/'''  shall be ```C:\Program Files\PostgreSQL\12\'''   (most of Windows programs allow to mix forward and back slashes, but not gpg)

rnp client tests (Python) include multiple definitions like 
```
 ret, _, _ = run_proc(RNPK, ['--import', data_path('test_key_edge_cases/key-eddsa-small-x-pub.asc')])
```
that create Unix-style paths' fragments  on Windows that 
1) forces to use MSys environment even for native Windows tests
2) makes it practically impossible to use debugger in Windows environment

## Proposed fix

```
 ret, _, _ = run_proc(RNPK, ['--import', data_path('test_key_edge_cases/key-eddsa-small-x-pub.asc')])
```
shall be 
```
 ret, _, _ = run_proc(RNPK, ['--import', data_path(os.path.join('test_key_edge_cases', 'key-eddsa-small-x-pub.asc')])
```

(~50  locations)

PR #2265, opened 2024-08-28
Refactor pgp_key_t methods for noexcept and fix few coverity issues.

ni4:
...also return key material from pgp_key_t by pointer instead of references.

### DIFF
--- a/.github/workflows/windows-native.yml
+++ b/.github/workflows/windows-native.yml
@@ -185,6 +185,15 @@ jobs:
           export PATH=${{ env.VCPKG_DIR_U }}/installed/${{ matrix.arch.triplet }}/bin:$PATH
           cmake --build build --parallel ${{ env.CORES }} --config Release
 
+      - name: Find Visual Studio root via vswhere
+        if: matrix.sanitizers == 'on'
+        shell: cmd
+        # Not sure why but vswhere.exe cannot be run via bash, so using cmd instead.
+        run: |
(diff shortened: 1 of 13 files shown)


src/lib/pgp-key.cpp
```diff
@@ -1596,7 +1590,9 @@ pgp_key_t::lock()
         return true;
     }
 
-    material().clear_secret();
+    if (material()) {
```
desvxx, line 1593: maybe it will be good to add assert not null here

src/lib/pgp-key.cpp
```diff
@@ -1687,7 +1683,9 @@ pgp_key_t::unprotect(const pgp_password_provider_t &password_provider,
     }
     pkt_ = std::move(*decrypted_seckey);
     /* current logic is that unprotected key should be additionally unlocked */
-    material().clear_secret();
+    if (material()) {
```
desvxx, line 1686: maybe it will be good to add assert not null here

desvxx (approved):

src/lib/pgp-key.cpp
ni4: Thanks, addressed this and the next comments.

.github/workflows/windows-native.yml
maxirmx, line 1: I believe the root cause is here: #2150

maxirmx (approved):

antonsviridenko (approved):

ni4:
Having approvals and green CI - let's merge it. Thanks all!

PR #2266, opened 2024-08-29
Do not allow 64 bit ciphers for encryption without explicit option.

### LINKED ISSUE #1598: Do not allow 64 bit ciphers for encryption without explicit option.
## Description

GnuPG since version 2.3 doesn't allow to use 64-bit ciphers (CAST5, 3DES, IDEA, BLOWFISH) without option `--allow-old-ciphers`. We should do the same.

desvxx:
fixes #1598


ni4:
@desvxx linter failed.

src/lib/sec_profile.cpp
```diff
@@ -196,6 +196,13 @@ SecurityContext::SecurityContext() : time_(0), prov_state_(NULL), rng(RNG::Type:
                       SecurityAction::VerifyKey});
     /* Mark MD5 insecure since 2012-01-01 */
     profile.add_rule({FeatureType::Hash, PGP_HASH_MD5, SecurityLevel::Insecure, 1325376000});
+    /* Mark CAST5, 3DES, IDEA, BLOWFISH insecure since 2024-10-01*/ // TODO: tbd
```
desvxx, line 199: And also we need to decide what date we want here.

ni4 (approved):
LGTM, thanks!

maxirmx (approved):
LGTM

PR #2283, opened 2024-10-19
CI: Add IWYU (include what you use) action.

ni4:
Let's whether it would work at all.


ni4:
Closing this as it's usage for CI is controversial.

PR #2284, opened 2024-10-21
Refactor key generation classes

ni4:
This PR refactors key generation classes.


ni4:
@ronaldtse @desvxx @maxirmx ping for review!

maxirmx:
This one is significant
I will find a couple of hours to understand itearly next week

ni4:
@maxirmx Thanks! Wanna make another also large PR, based on this one.

src/lib/keygen.hpp
```diff
@@ -0,0 +1,184 @@
...
+#include "pgp-key.h"
+
+namespace rnp {
+
+class KeygenParams {
+  private:
+    pgp_pubkey_alg_t                alg_;
+    pgp_hash_alg_t                  hash_;
+    pgp_version_t                   version_;
+    SecurityContext &               ctx_;
```
desvxx, line 43: Risky solution imho (to use reference here)

src/lib/pgp-key.cpp
```diff
@@ -2618,14 +2525,14 @@ pgp_key_t::add_sub_binding(pgp_key_t &                       subsec,
     pgp_signature_t sig;
     sign_init(ctx.rng, sig, hash, ctx.time(), version());
     sig.set_type(PGP_SIG_SUBKEY);
-    if (binding.key_expiration) {
-        sig.set_key_expiration(binding.key_expiration);
+    if (binding.expiration) {
```
desvxx, line 2528: May be confusing - "what expiration?". IMHO keep key_expiration will be better.

src/tests/cipher.cpp
```diff
@@ -118,13 +118,12 @@ TEST_F(rnp_tests, pkcs1_rsa_test_success)
     uint8_t ptext[1024 / 8] = {'a', 'b', 'c', 0};
     uint8_t dec[1024 / 8];
 
-    rnp_keygen_crypto_params_t key_desc;
-    key_desc.key_alg = PGP_PKA_RSA;
-    key_desc.hash_alg = PGP_HASH_SHA256;
-    key_desc.rsa.modulus_bit_len = 1024;
-    key_desc.ctx = &global_ctx;
+    rnp::KeygenParams keygen(PGP_PKA_RSA, global_ctx);
+    auto &            rsa = dynamic_cast<pgp::RSAKeyParams &>(keygen.key_params());
```
desvxx, line 122: weird spacing:)

desvxx (approved):

src/lib/key_material.cpp
```diff
@@ -619,9 +734,10 @@ DSAKeyMaterial::write_secret(pgp_packet_body_t &pkt) const
 }
 
 bool
-DSAKeyMaterial::generate(const rnp_keygen_crypto_params_t &params)
+DSAKeyMaterial::generate(rnp::SecurityContext &ctx, const KeyParams &params)
 {
-    if (dsa_generate(&params.ctx->rng, &key_, params.dsa.p_bitlen, params.dsa.q_bitlen)) {
+    auto &dsa = dynamic_cast<const DSAKeyParams &>(params);
```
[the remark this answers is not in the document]
ni4: The idea was to avoid theoretical possibility of getting wrong object here by throwing exception. Not sure how that could happen, just to be cautious.

src/lib/key_material.hpp
```diff
@@ -28,13 +28,155 @@
...
+  public:
+    RSAKeyParams() : BitsKeyParams(DEFAULT_RSA_NUMBITS){};
+};
+
+class DSAKeyParams : public BitsKeyParams {
+  private:
+    size_t qbits_;
+
+  public:
+    DSAKeyParams();
```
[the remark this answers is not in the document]
ni4: It is actually initialized in constructor implementation, but let's move it to the header for better clarity, thanks.

src/lib/keygen.hpp
ni4: SecurityContext is a singleton which lives the whole life of the top-level rnp_ffi_t object, and it would be destroyed after all the internal stuff. Is this good enough or there could be other solution?

src/lib/pgp-key.cpp
ni4: Thanks, renamed.

src/tests/cipher.cpp
ni4: Yeah, unfortunately that's how clang-format 11 behaves. Don't like these as well.

maxirmx (approved):
OK

ni4:
@maxirmx thanks for the review. Having static variable would have issues with multi-threaded applications, which now could use at least separate rnp_ffi_t objects in different threads.

ni4:
Merging with two approvals and green CI. Thanks!

PR #2287, opened 2024-11-04
Update draft 06

TJ-91:
This PR updates to the newest PQC draft version, as well as fixes v6 signature salt.

### V6 / RFC 9580

* correctly implement v6 salt for document signatures (was only properly working for key signatures before)
  * change `HashList` for use with salt
  * implement v6 OPS with salt
  * gracefully fail when verifying v6 cleartext signatures by skipping them. Due to the salt that is detected at the end, two passes are required which requires further changes to the code.
* add Ed448/X448 standalone algorithms 
 
### PQC

* final PQC NIST standard algorithms
* update to PQC draft version 06: KEM Combiner, SLH-DSA parameters, Test Vectors, ...
  * -06 is not-yet published but will be -05 with fixed test vectors
* Add X448 and Ed448 for composite PQC combinations

### Botan

* Both `ENABLE_CRYPTO_REFRESH` and `ENABLE_PQC` now require Botan 3.6.
  * Ed448/X448 only available from 3.4
  * ML-KEM / ML-DSA / SLH-DSA final NIST standards available from Botan 3.6
  * Supporting partial features from lower Botan versions would mean to either have lots of deprecation warnings or have special code for the different versions.


TJ-91:
Fedora tests fail due to building with Botan 3.3.0, same as building Fuzzers with 3.4.0.

@ni4 do you have any feedback / comments for the PR?

TJ-91:
closing in favor of #2355

PR #2293, opened 2024-11-26
fix: use brew bundle with -f option in macos workflows

maxirmx:


ni4 (approved):
LGTM, thanks, @maxirmx !

ni4:
Merging with one approval as CI-only.

PR #2297, opened 2024-12-05
Remove path prefix in RNP_LOG

### LINKED ISSUE #2146: Do not print a full source path in RNP_LOG.
## Description

Currently RNP_LOG prints the whole path to the source file, i.e. in my case:
`[process_pgp_source() /Users/aaaa/bbbbb/rnp/src/librepgp/stream-parse.cpp:2983]`
However it would be better to hide prefix, i.e. print just the following:
`[process_pgp_source() /librepgp/stream-parse.cpp:2983]`

The best way I see right now is to use some build-time define with prefix path length, like it described here: https://stackoverflow.com/a/40947954/452752

desvxx:
fixes #2146

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,6 +96,10 @@ set_property(CACHE ENABLE_DOC PROPERTY STRINGS ${TRISTATE_VALUES})
 # so we can use our bundled finders
 set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/Modules")
 
+# add source path prefix length to be able to crop it in logs
+string(LENGTH "${CMAKE_SOURCE_DIR}/" SOURCE_PATH_SIZE)
+add_definitions("-DSOURCE_PATH_SIZE=${SOURCE_PATH_SIZE}")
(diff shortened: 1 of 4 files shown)


maxirmx (approved):
This is based on the assumption that compiler sets __FILE__ to the file name from the command line.

I am not sure if it works for MSVC

desvxx:
> This is based on the assumption that compiler sets **FILE** to the file name from the command line.
> 
> I am not sure if it works for MSVC

https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

Indeed, looks like we need /FC option for msvc.

ni4 (changes requested):
Could you please add some line to cli_tests.py, which makes sure that prefix is not included. This would work for any output made with RNP_LOG, like `self.assertRegex(err, r'(?s)^.*Warning: missing or malformed CRC line.*')`.
Also please rebase so all tests are run (macOS runner was renewed).

ni4 (approved):
LGTM, thanks!

Issue #2262, opened 2024-08-15
rnp_key_add_signature() missing on Ubuntu 22.04


csoler:
Apparently the version of libRNP that is supplied on ubuntu 22.04 (which is 0.15.2) doesn't include key certification yet.

I could not find in the code a way to prevent the compiler from failing when this functions is not here. Indeed, there is no (apparently) global variable that the pre-compiler can use in an #ifdef statement. I could only find a rnp_version() method, but this cannot be called by the pre-compiler.

Can you suggest me how to proceed?
Thx a lot
Cyril

ni4:
That strange that Ubuntu ships such an old version, I'll take a look whether we could ping them in some way.

As I remember we had plans to include defines for rnp version into header, but due to some reasons didn't implement that yet. So I created this issue: https://github.com/rnpgp/rnp/issues/2263

Right now you may involve some building steps, depending on which system you use for build.
For instance, during the build you may compile and run simple program which would be linked with rnp and return `rnp_version()` result.
We do something similar to return OpenSSL features in `findopensslfeatures.c`.

If you are building from sources, you may take a look at RNP build directory  and extract information from there.
Anyway, more correct way would be to install version header, of course, and we'll handle this.

Hope this helps,
Nickolay

ni4:
@csoler Btw, didn't notice at first - we don't have `rnp_key_add_signature()` function, instead we have only `rnp_key_direct_signature_create()/rnp_key_certification_create` at the moment. Now I'm working on extension to these functions which would allow to add any other kind of signatures, including other key certification.

csoler:
I updated my code so as to use librnp as a submodule. Now, it seems signatures still do not work. In particular the documentation for rnp_key_direct_signature_create() and rnp_key_certification_create() state that the signature should be finalized using rnp_signature_sign() call. But this function doesn't seem to exist. Did I miss anything here?

ni4:
@csoler oh, it should by called `rnp_key_signature_sign()`, just a typo in documentation which should be fixed.

PR #2298, opened 2024-12-09
Fix wrong function name mentioned in rnp.h

ni4:
See https://github.com/rnpgp/rnp/issues/2262#issuecomment-2525244075


maxirmx (approved):
LGTM

PR #2304, opened 2024-12-25
(WIP) Refactor pgp_encrypted_material_t to C++ classes.

ni4:
Just few more types left to avoid unions and move pgp::mpi to std::vectory.


src/lib/rnp.cpp
```diff
@@ -8174,29 +8174,32 @@ add_json_mpis(json_object *jso, pgp_key_t *key, bool secret = false)
...
+    case PGP_PKA_RSA_SIGN_ONLY: {
+        auto &rsa = dynamic_cast<const pgp::RSASigMaterial &>(*material);
+        return add_json_mpis(jso, "sig", &rsa.sig.s, NULL);
+    }
     case PGP_PKA_ELGAMAL:
-    case PGP_PKA_ELGAMAL_ENCRYPT_OR_SIGN:
-        return add_json_mpis(jso, "r", &material.eg.r, "s", &material.eg.s, NULL);
-    case PGP_PKA_DSA:
-        return add_json_mpis(jso, "r", &material.dsa.r, "s", &material.dsa.s, NULL);
+    case PGP_PKA_ELGAMAL_ENCRYPT_OR_SIGN: {
```
maxirmx, line 8189: IMHO this piece of code call for virtual function

something like 
```
virtual mpi* SigMaterial::get_mpi() = 0;
```
and appropriate redefinitions for ```EgSigMaterial``` et al

ni4: I don't like the way it is currently done as well. The problem here is that different signature types may have different MPIs. In this case it is simple, one or two, but for EncMaterial and KeyMaterial it is more complicated, however I think we should use the same approach in both cases. Don't know a good solution yet.

desvxx (approved):

ni4:

PR #2309, opened 2025-01-24
Refactor encrypted_add_password().

ni4:
- for better code quality
- closer to c++ style
- workaround GnuPG multiple SKESKs handling, causing random CI failures for us.

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -120,15 +120,6 @@ ffi_key_provider(const pgp_key_request_ctx_t *ctx, void *userdata)
     return find_key(ffi, ctx->search, ctx->secret, true);
 }
 
-static void
-rnp_ctx_init_ffi(rnp_ctx_t &ctx, rnp_ffi_t ffi)
-{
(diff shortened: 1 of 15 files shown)


src/lib/rnp.cpp
```diff
@@ -3340,11 +3301,8 @@ try {
         return RNP_ERROR_NULL_POINTER;
     }
 
-    *op = new rnp_op_verify_st();
-    rnp_ctx_init_ffi((*op)->rnpctx, ffi);
+    *op = new rnp_op_verify_st(ffi, signature);
     (*op)->rnpctx.detached = true;
```
desvxx, line 3305: maybe it will be more consistent to add these params to constructor as well?

desvxx (approved):

ni4:
Merging with two approvals. Thanks!

src/lib/rnp.cpp
ni4: Agree, thanks. Will address via another PR as this one is already merged.

PR #2321, opened 2025-03-12
Add curve module checks for 3.7.0+ Botan

ni4:
Since 3.7.0 Botan has separate module for each curve, which may be disabled, so we should check for them.

### DIFF
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -172,6 +172,18 @@ if(CRYPTO_BACKEND_BOTAN)
   else()
     set(_botan_required_features ${_botan_required_features} DL_SCHEME RAW_HASH_FN)
   endif()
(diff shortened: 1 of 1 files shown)


maxirmx (commented):
Is there a test for Botan 3.7.0 +  ?

ni4:
> Is there a test for Botan 3.7.0 + ?

Yeah, Botan's head which is 3.7.0+. Should I add Fedora 41 + dedicated Botan version runner as well?

maxirmx:
> Yeah, Botan's head which is 3.7.0+. 
I should/could guess it myself :)

maxirmx (approved):
LGTM

desvxx (approved):

ni4:
Thanks! Merging with two approvals.

PR #2323, opened 2025-03-15
Add support for DSA 4096 bit keys

### LINKED ISSUE #2260: Add support for DSA-4096 keys and signatures.
## Description

There are some keys in a wild which use DSA 4096 bits (example: https://bugzilla.mozilla.org/show_bug.cgi?id=1903976), and those are supported/validated correctly by GnuPG.
We should consider supporting them as well, as currently we limit Q parameter size to 256 bits, while DSA 4096 would need 384.

ni4:
While this is not follows standard, there are still possible existing of such artifacts. So we would allow to import and use such keys, but do not generate.

Fixes #2260 

P.S. Doesn't work with OpenSSL backend as it doesn't support 384-bit signature element size.


src/lib/crypto/dsa.cpp
```diff
@@ -131,10 +126,6 @@ rnp_result_t
 Key::verify(const Signature &sig, const rnp::secure_bytes &hash) const
 {
     size_t q_order = q.bytes();
-    if (q_order > BITS_TO_BYTES(DSA_MAX_Q_BITLEN)) {
```
desvxx, line 134: I guess it's totally fine to drop this, but just can't keep myself from voicing my "oh no, some checks was removed from the code" feeling. :)

desvxx (approved):

src/lib/crypto/dsa.cpp
ni4: @desvxx previously we had static C-style fixed-size array for storage, so this check was required. Now with std::vector we may drop it )

ni4:
Merging with two approvals, thanks!

PR #2324, opened 2025-03-17
Allow extra spaces after the armored block during the key import.

### LINKED ISSUE #2308: Allow to have extra spaces/eols after the armored key during the import.
## Description

See the details and samples here: https://bugzilla.mozilla.org/show_bug.cgi?id=1941700

ni4:
Fixes #2308


ronaldtse (approved):

src/librepgp/stream-common.cpp
maxirmx, line 1: IMHO it can be simplified. Not that it changes much  but looks a bit more "C++ ish"

```
#include <cstring>
#include <cctype>

bool
pgp_source_t::skip_chars(const char *chars)
{
    uint8_t ch;
    size_t  read;

    while (true) {
        if (!peek(&ch, 1, &read)) {
            return false;
        }
        if (read !=0 && std::strchr(chars, ch)) {
            skip(1);
        } else {
            break;
        }
    }

    return true;
}
```

maxirmx (changes requested):
please consider simplifying `pgp_source_t::skip_chars`

src/librepgp/stream-common.cpp
ni4: @maxirmx Does it look good now?

maxirmx (approved):

ni4:

PR #2327, opened 2025-03-19
Tests support: use POSIX syntax for cp

barracuda156:
Using gnuism here was breaking tests when `cp` did not support it. Use POSIX instead.


ni4 (approved):
LGTM, thanks! Yesterday VSCode screwed me up by not finding copy_recursive call in rnp_tests.cpp

ronaldtse:
Thank you @barracuda156 @ni4 !

PR #2330, opened 2025-03-21
Refactor rnp::Hash and pgp_key_store_format_t.

ni4:
Came accross these changes while refactoring pgp_fingerprint_t, so for better clarity and easier review moved them to separate PR.

### DIFF
--- a/src/lib/crypto/hash.cpp
+++ b/src/lib/crypto/hash.cpp
@@ -95,19 +95,17 @@ Hash_Botan::add(const void *buf, size_t len)
     fn_->update(static_cast<const uint8_t *>(buf), len);
 }
 
-size_t
+void
 Hash_Botan::finish(uint8_t *digest)
 {
     if (!fn_) {
-        return 0;
+        return;
     }
-    size_t outlen = size_;
     if (digest) {
         fn_->final(digest);
     }
     fn_ = nullptr;
     size_ = 0;
-    return outlen;
 }
 
 const char *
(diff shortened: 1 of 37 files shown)


src/lib/crypto/hash.cpp
```diff
@@ -95,19 +95,17 @@ Hash_Botan::add(const void *buf, size_t len)
     fn_->update(static_cast<const uint8_t *>(buf), len);
 }
 
-size_t
+void
 Hash_Botan::finish(uint8_t *digest)
 {
     if (!fn_) {
-        return 0;
+        return;
```
desvxx, line 102: Missing some exception/warning here?

src/lib/crypto/hash_ossl.cpp
```diff
@@ -117,23 +117,20 @@ Hash_OpenSSL::add(const void *buf, size_t len)
     }
 }
 
-size_t
+void
 Hash_OpenSSL::finish(uint8_t *digest)
 {
     if (!fn_) {
-        return 0;
+        return;
```
desvxx, line 124: And here

desvxx (approved):

src/lib/crypto/hash.cpp
ni4: @desvxx Thanks, I think adding assert here would be good enough as this may happen only due to development error, not runtime.

maxirmx (approved):
LGTM

ni4:

PR #2335, opened 2025-03-31
Support use of botan-config.cmake

### LINKED ISSUE #2121: Upstream FindBotan.cmake to CMake project
It has been a long-term issue that CMake does not ship with a `FindBotan.cmake` find module to search for and link against a Botan installation. People start writing their own find modules again and again, that basically do the same, which you can see [GitHub search](https://github.com/search?q=findbotan.cmake&type=code), for example. For existing but also new users of Botan, it would be great if they had out-of-the-box support for Botan in CMake projects.

I found that the [`FindBotan`](https://github.com/rnpgp/rnp/blob/main/cmake/Modules/FindBotan.cmake) script you ship with rnp looks the most mature and already supports Botan 2.x and 3.x. Would it be possible to upstream it to the [CMake project](https://gitlab.kitware.com/cmake/cmake/-/tree/master/Modules)? The only requirement from CMake's side would be to find a [dedicated module maintainer](https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/dev/Module-Maintainers). If none of you can do this, I would be willing to step in to make this happen.

/cc @randombit @reneme @lieser

ni4:
It appears we didn't have best support for added via https://github.com/randombit/botan/pull/3722 botan-config.cmake, so this PR fixes that.
Since we still need to support Botan-2 I didn't delete builtin FindBotan.cmake
This should (or could) be a reason for issues @defnax experienced in #2307

Fixes #2121

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -159,6 +159,28 @@ jobs:
           export PATH="$PWD/build/src/lib:$PATH"
           ctest --parallel ${{ env.CORES }} --test-dir build -C Debug --output-on-failure
 
+      - name: Configure and build with botan-config.cmake
+        if:  matrix.backend == 'botan3'
+        run: |
+            rm -rf build
+            rm -rf ./cmake/Modules/FindBotan.cmake
+            shopt -s nullglob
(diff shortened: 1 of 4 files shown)


desvxx (approved):

ni4:

PR #2338, opened 2025-04-05
Refactor pgp::mpi to use std::vector instead of fixed-size buffer.

ni4:
Less memory usage, more robust, cleaner code.


desvxx (approved):

maxirmx (approved):
LGTM

PR #2344, opened 2025-04-18
Fix macOS OpenSSL 1.1.1 tests failure.

ni4:
A bit weird but seems to work.
The issue is that macOS clang adds /usr/local/include to search path with higher priority then OpenSSL 1.1.1 headers, causing FindOpenSSLFeatures malfunction (compiling with 3.0 headers from /usr/local/include, but attempting to link with 1.1.1. libraries).

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -94,6 +94,8 @@ jobs:
           echo "brew \"openssl@1.1\"" >> Brewfile
           echo "OPENSSL_ROOT_DIR=$(brew --prefix openssl@1.1)" >> $GITHUB_ENV
           echo "CRYPTO_BACKEND=openssl" >> $GITHUB_ENV
+          # Harsh workaround against CLang adding /usr/local/include path, where OpenSSL 3.0 headers are
+          rm -r /usr/local/include/openssl
 
       - name: Configure openssl 3 backend
         if: matrix.backend == 'openssl@3'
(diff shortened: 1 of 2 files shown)


desvxx (approved):

ni4:
Merging with one approval as CI-only.

ronaldtse (approved):

maxirmx:
`rm -rf /usr/local/include`  means that you do not support local builds
while you can play this game at GHA host, nobody in a good sense will do it on a local system or permanantly deployed build host

IMHO the best solution would be to drop OpenSSL 1.1 support on MacOS because homebrew implies OpenSSL 3 installation
Practically you cannot have valid MacOS system with homebrew but without OpenSSL 1.1

If you still want to test against `OpenSSL 1.1` the issue is with `find_package(OpenSSL REQUIRED)` 
There is some new behaviour probably causes by cmake version update.
On MacOS the way it is implemented creates compiler flags like
`-isystem <path-to-openssl-1.1> -I /usr/local/include`
With this setup if you add `-I <path-to-openssl-1.1>` parameter it is ignored because the same include folder is already specified by `-isystem`.  `BEFORE` option does not help since `BEFORE`  ignores `-isystem`

However, include directories configured by `-I` option are precessed before  include directories `-isystem` option.

All together sounds a bit crazy but it was the point I reached in my research.
Xonsidering the call flow it was possible to avoid a call to `find_package`  but I did not have time to finalize the solution.

PR #2346, opened 2025-04-26
Discourage usage of EAX AEAD mode.

### LINKED ISSUE #1985: Mark EAX AEAD mode as deprecated in man and code.
## Description

Latest draft of rfc4880-bis deprecates usage of EAX mode: 
```
EAX mode is deprecated due to the far better properties of the OCB mode.
Implementations may use EAX mode only for decryption of existing data.
```
(see https://www.ietf.org/archive/id/draft-koch-openpgp-2015-rfc4880bis-01.txt)

So we should let the users/developers know that it is deprecated.

ni4:
Fixes #1985


ronaldtse (approved):
Thank you @ni4!

desvxx (approved):

ni4:

PR #2349, opened 2025-05-24
Release v0.18.0

ni4:
As title says.


ronaldtse (approved):
Thank you @ni4 !

ni4:
@ronaldtse @maxirmx Do you have any good idea on working over 100-commit limitation? I can only think about squashing some commits together to not hit the limit.

ronaldtse:
I think the easiest way is to do rebase and merge manually on your local...?

ni4:
@ronaldtse thanks for the hint! Would push all but the few latest commits to keep PR in the history.

ronaldtse:
@ni4 another way is to split into 2 PRs?

ni4:
@ronaldtse given that there are 386 commits here we'll need 4 PRs :) Let's wait for the second approval and I'll push 300 commits manually, and the rest through the rebased PR, does this sound good?

ni4:
Thanks all! Let's do a release.

PR #2352, opened 2025-06-04
stream-dump.cpp: add missing cinttypes

barracuda156:
Since the last version there is another file which needs this fix.


ni4:
Thanks, @barracuda156 !
Could you please specify which compiler/OS combination exposes this error? Probably it would be good idea to add it to CI.

ni4 (approved):
LGTM, thanks!

PR #2355, opened 2025-07-14
Update PQC Draft to Version 12

TJ-91:
This PR updates to the newest PQC draft version, and adds/fixes some RFC 9580 functionality. The PR replaces #2287. The PQC draft can be seen as stable now since it has passed Working Group Last Call recently.

The most prominent changes are:

### V6 / RFC 9580

* correctly implement v6 salt for document signatures (was only properly working for key signatures before)
  * change HashList for use with salt
  * implement v6 OPS with salt
  * gracefully fail when verifying v6 cleartext signatures by skipping them. Due to the salt that is detected at the end, two passes are required which requires further changes to the code.
* add Ed448 and X448 standalone algorithms

### PQC

* update Kyber, Dilithium and Sphincs+ to final PQC NIST algorithms: ML-KEM, ML-DSA, SLH-DSA
* update to PQC draft version 12:
  * add X448 and Ed448 for composite PQC combinations
  * implement SHA3 KEM Combiner (replaces KMAC)
  * remove Sphincsplus/SLH-DSA parametrization
  * implement tests / test vectors
  * seed format for ML-KEM/ML-DSA private keys
* also update to use the new codepoints in the upcoming draft-ietf-openpgp-nist-bp-comp-02

### Further Code Changes

* ~~PQC code is not independent from Crypto Refresh / RFC9580 any more and thus `ENABLE_CRYPTO_REFRESH` is required for `ENABLE_PQC`~~ This is changed again in newer commits to allow MLKEM768+X25519 for v4 keys without compiling the crypto refresh code.
* `ENABLE_CRYPTO_REFRESH` and `ENABLE_PQC` now requires Botan 3.6.
  * Ed448/X448 only available from 3.4
  * PQC final NIST standards available from Botan 3.6
  * Supporting partial features from lower Botan versions would mean to either have lots of deprecation warnings or have special code for the different versions.
* Since RIPEMD is also deprecated in RFC 9580, I added a Security Rule that marks it insecure like SHA1 if `CRYPTO_REFRESH_ENABLED` is true.
  * I don't know what date would be most appropriate though

--- 

@ni4  since I had to rebase a lot and fixed some stuff only at the end of the rebasing, the history is not perfectly intact. Please tell me if you prefer to keep the commits anyway or whether I should squash them into a single commit. I hope I did not mess anything up when rebasing.

As next steps I would like to rebase the other PRs #2296 and #2207 (that is considerably less code than in this PR).


kaie:
Hi Johannes, would it be easy for you to rebase your PR?

Nickolay said the CI issues should be fixed, so maybe merging will be possible soon.

If you don't have time, Nickolay said he could replicate your PR.

TJ-91:
Hi Kai, since GitHub says there are no conflicts, it is easy for me, yes :)

kaie:
Great, thanks a lot!

TJ-91:
It seems like there are still some issues either with the CI or with my code. Unfortunately I don't have a lot of time right now to take care of this.

FYI, I have set "Allow edits and access to secrets by maintainers" if Nickolay finds the time.

kaie:
I get a build warning when building against Botan 3.11.1  [edited: warning, not error]

In file kyber_ecdh_composite.h a forward declaration is added "struct pgp_kyber_ecdh_composite_public_key_t;",
but the type is rather declared as a class, not a struct.

kaie:
The CI runs that uses the openssl backend fails because it is not yet available for pqc in this PR.
(That applies for the initial build job, and also the failing opensuse job.)
That CI run passes in PR 2392 which adds such a backend.

The two CI runs that use Botan 3.3.0 fail because that workflow enables PQC, but that configuration requires at least Botan 3.6.0
I've asked for guidance in issue #2393 how that workflow should be fixed.

The fuzzing build has the same issue, but it requires a patch for oss-fuzz.
@ni4 could you please submit an update there?

The issues in windows did NOT happen in the latest CI runs in my PR #2392.
I cannot explain why.

In the other CI runs for my recent pull requests, I see a new failure appearing,
in rnp_tests.test_cipher_aes_128_ocb
that happened both in PR #2392 and and in PR #2391 
In both places the test failed when using Botan HEAD, so maybe it is caused by a recent change in Botan?

I would like to suggest that you add workflows that use Botan 3.11.1 or 3.12
to test against a recent stable version of Botan.

ni4 (approved):
While there are some CI failures they don't seem relevant and we may fix those later. LGTM!

ronaldtse (approved):
Thank you @TJ-91 and everyone!

PR #2363, opened 2025-10-09
Fix 32 bit assertion

dkg:


ronaldtse:
Rebased onto latest main. Note: the first commit (Fix assertion breakage on 32bit architectures) was dropped during the rebase because main already contains an equivalent fix — dc417b24 "Fix size assertion on 32-bit platforms" (assert(... <= (size_t)0xffffffffu), which avoids the same (size_t)1 << 32 UB on 32-bit targets). What remains is the cli_tests.py change: test for 32-bit time_t via ctypes.c_time_t with a sys.maxsize fallback. Verified locally: Misc.test_set_current_time passes.

ronaldtse:
Rebased onto release/0.x (correcting my earlier main-based rebase — this PR targets the release line). The 32-bit assertion commit was dropped: both main (dc417b24) and the release branch already contain an equivalent fix ('assert(... <= (size_t)0xffffffffu)'). Remaining change: cli_tests.py time_t detection via ctypes.c_time_t with sys.maxsize fallback. Verified locally on the release/0.x base: Misc.test_set_current_time passes.

ronaldtse (approved):

PR #2364, opened 2025-10-09
Avoid returning enum in boolean context

dkg:
Without this fix, when compiling with `ENABLE_CRYPTO_REFRESH`, we see:

```
…/src/lib/enc_material.cpp: In member function ‘virtual bool pgp::X25519EncMaterial::parse(pgp_packet_body_t&)’:
…/src/lib/enc_material.cpp:154:20: warning: enum constant in boolean context [-Wint-in-bool-context]
  154 |             return RNP_ERROR_BAD_FORMAT;
      |                    ^~~~~~~~~~~~~~~~~~~~
make[3]: Leaving directory '…/build'
```

### DIFF
--- a/src/lib/enc_material.cpp
+++ b/src/lib/enc_material.cpp
@@ -151,7 +151,7 @@ X25519EncMaterial::parse(pgp_packet_body_t &pkt) noexcept
         uint8_t bt = 0;
         if (!pkt.get(bt)) {
             RNP_LOG("failed to get salg");
-            return RNP_ERROR_BAD_FORMAT;
+            return false;
         }
         sess_len--;
         salg = (pgp_symm_alg_t) bt;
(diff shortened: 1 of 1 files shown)


TJ-91:
FYI, this, among other things, is also being fixed in PR https://github.com/rnpgp/rnp/pull/2355/

ronaldtse:
Verified against current main: the same spot in X25519EncMaterial::parse() now already does 'return false;' (the whole function was reworked to return bool properly by the crypto-refresh/PQC refactoring, e.g. 2dcae852 and follow-ups). This PR is a no-op against current main and can probably be closed as superseded. I left it unmerged/unclosed for the author/maintainers to confirm.

ronaldtse:
Correction to my earlier comment: on main this is indeed already fixed, but this PR targets release/0.x where the issue is still present. Rebased onto release/0.x and extended the fix to the identical case in MlkemEcdhEncMaterial::parse (same 'return RNP_ERROR_BAD_FORMAT' in a bool function, line 205). Verified locally on the release/0.x base.

ronaldtse (approved):

ronaldtse:
Just need one more approval!

ni4 (approved):
LGTM!

ronaldtse:
Thank you @dkg @ni4 !

ronaldtse:
@dkg FYI we have ported your patches to the main branch at #2402 as well. Thank you!

ronaldtse:
Merging with 2 approvals (ni4, ronaldtse). Thank you @dkg for the contribution, and @ni4 for the review!

PR #2387, opened 2026-04-18
Fix new CI issues to make repository healthy back

ni4:
- Add missing <cstring> include in mem.cpp
- Rename cpy variables to copy in s2k_ossl.cpp and fingerprint.hpp
- ...and more to come...


ronaldtse (approved):
LGTM, thank you @ni4!

maxirmx (approved):

PR #2389, opened 2026-05-07
Change Key::validate to use variable time division to fix performance regression

kaie:
See https://bugzilla.mozilla.org/show_bug.cgi?id=2001803

### DIFF
--- a/src/lib/crypto/elgamal.cpp
+++ b/src/lib/crypto/elgamal.cpp
@@ -30,7 +30,6 @@
 #include <botan/ffi.h>
 #include <botan/bigint.h>
 #include <botan/numthry.h>
-#include <botan/reducer.h>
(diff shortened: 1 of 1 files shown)


kaie:
I note that Key::Validate takes a parameter (bool secret).

Would it necessary to use the constant time code when secret is true?

kaie:
Nickolay, you said in https://bugzilla.mozilla.org/show_bug.cgi?id=2001803#c41
that the patch looks good, and we have applied it to Thunderbird 152 Beta already. 

Is this PR ready to merge?

ronaldtse (approved):

ronaldtse:
Already approved by @ni4 on Bugzilla:

image

Thank you @kaie @ni4 !

PR #2392, opened 2026-05-28
Add an automatically generated OpenSSL backend for the Crypto Refresh and PQC code

kaie:
These changes are based on top of PR #2355, which provides an implementation that uses the Botan backend, but does not provide a corresponding implementation for using the OpenSSL backend.

The patches that I'm providing here in addition have been machine generated by Claude Code.

The use of automatic generation was chosen by me, because I made the assumption that the human creative work that decided how to use crypto primitives to implement the OpenPGP functionality was already done and is contained in the base pull request, and that the implementation of the OpenSSL backend code could inspect the code that uses Botan, and produce equivalent code that uses OpenSSL.

The first patch removes some botan specific code from code that is shared across backends and replaces it with a new helper class for secure wiping of bytes. I reviewed that code and it looks correct to me.

The code builds and passes the existing RNP tests on Linux. I haven't built nor ran tests on other platforms.

A script is provided that tests the interoperability of data across the two backends.

The additional patch that addresses code review issues weren't based on a human review, but based on a re-review by the AI itself.

I'm hereby providing the code as is, but I want to make it clear that I couldn't have written the code to use OpenSSL myself in the time I have available, as I don't have experience with writing code for OpenSSL.

The code is provided in the hope it is useful, in the hope that the PQC functionality can be made available for all users of RNP regardless of which backend cryptography library is preferred, and I suggest that the code should get reviewed for correctness before considering to adopt it.

Regarding copyright headers, for all files that were newly written, Ribose was added as the only copyright holder.
However, those files were automatically created by using the equivalent code files as a template. So they are derivative work
For all files that were originally contributed by MTG, but which were modified by this work, Ribose was added as an additional copyright holder.
 I'd like to leave it to Ribose to decide what the correct copyright section should be and fix those sections, if necessary.


kaie:
Thanks a lot for merging the base branch!

I've rebased and updated the PR.

ni4 (approved):
LGTM, thanks!

ronaldtse (approved):

ronaldtse:
Thank you @kaie for the contribution!

ronaldtse:
Merging with 2 approvals (ni4, ronaldtse). Thank you @kaie for the contribution, and @ni4 for the review!

PR #2400, opened 2026-07-14
Fix compatibility with GnuPG 2.5.21 (test_encryption_cek)

### LINKED ISSUE #2399: cli_tests-Encryption tests fail with GnuPG 2.5.21
I am seeing test failures when building with GnuPG 2.5.21. This is rnp v0.18.1, with commits ba8df120afc88b9530df3a1c6a249032932466c7 and 6f8a67703e8a3cf57a4205c3f1fb6d2e4dfe22a7 on top.

I have narrowed this down to having started with GnuPG 2.5.21 (vs., 2.5.20 which does not trigger this). Specific tests failing (verified by patching them out)

* `src/tests/cli_tests.py`: `test_sm2_encryption_signing`
* `src/tests/cli_tests.py`: `test_encryption_cek`

````
[  155s] ERROR: test_sm2_encryption_signing (__main__.Encryption.test_sm2_encryption_signing)
[  155s] ----------------------------------------------------------------------
[  155s] Traceback (most recent call last):
[  155s]   File "/home/abuild/rpmbuild/BUILD/rnp-0.18.1-build/rnp-v0.18.1/src/tests/cli_tests.py", line 4970, in test_sm2_encryption_signing
[  155s]     os.mkdir(RNPDIR2, 0o700)
[  155s]     ~~~~~~~~^^^^^^^^^^^^^^^^
[  155s] FileExistsError: [Errno 17] File exists: '/tmp/rnpctmp5_ikiumb/.rnp2'
[  155s] 
[  155s] ======================================================================
[  155s] FAIL: test_encryption_cek (__main__.Encryption.test_encryption_cek)
[  155s] ----------------------------------------------------------------------
[  155s] Traceback (most recent call last):
[  155s]   File "/home/abuild/rpmbuild/BUILD/rnp-0.18.1-build/rnp-v0.18.1/src/tests/cli_tests.py", line 5026, in test_encryption_cek
[  155s]     self.assertEqual(ret, 0)
[  155s]     ~~~~~~~~~~~~~~~~^^^^^^^^
[  155s] AssertionError: 2 != 0
`````

Build config: [rnp.spec](https://build.opensuse.org/projects/home:AndreasStieger:branches:security:privacy/packages/rnp/files/rnp.spec?expand=1), (Botan). [Failing log](https://build.opensuse.org/package/live_build_log/home:AndreasStieger:branches:security:privacy/rnp/clean/x86_64), [passing log](https://build.opensuse.org/package/live_build_log/home:AndreasStieger:branches:security:privacy/rnp/repro/x86_64).

andreasstieger:
Fix compatibility with GnuPG 2.5.21 (`test_encryption_cek`) Pass `--compatibility-flags=no-partial-file-guard` `test_encryption_cek` in `cli_tests`. Related to GnuPG [T7873](https://dev.gnupg.org/T7586).

Fixes #2399

### DIFF
--- a/src/tests/cli_tests.py
+++ b/src/tests/cli_tests.py
@@ -5022,7 +5022,7 @@ class Encryption(unittest.TestCase):
         for _ in range(10):
             ret, _, _ = run_proc(RNP, ['--homedir', RNPDIR2, '-r', 'test_cek', '-e', src, '--overwrite', '--output', enc])
             self.assertEqual(ret, 0)
-            ret, _, err = run_proc(GPG, ['--batch', '--homedir', GPGHOME2, '--show-session-key', GPG_LOOPBACK, '--passphrase', PASSWORD, '--output', '/dev/null', '-d', enc])
+            ret, _, err = run_proc(GPG, ['--batch', '--homedir', GPGHOME2, '--show-session-key', GPG_LOOPBACK, '--passphrase', PASSWORD, '--output', '/dev/null', '--compatibility-flags=no-partial-file-guard', '-d', enc])
             self.assertEqual(ret, 0)
             m = pattern.match(err)
             self.assertTrue(m)


ronaldtse:
Heads-up: this fix breaks test_encryption_cek on older GnuPG versions. The test matches the session-key line with re.match (anchored at the start of stderr), and with the flag, older gpg prints 'unknown compatibility flag no-partial-file-guard ignored' BEFORE 'gpg: encrypted with' — so re.match returns None and the test fails (see the 92 failing jobs on this PR, e.g. the CentOS 9 gpg-lts runs). A version-agnostic alternative is in #2401: decrypt to a real temporary file instead of /dev/null, which works with both old gpg and 2.5.21+ without any flags.

ronaldtse:
Hi @andreasstieger ! Thank you so much for taking the time to investigate and fix the GnuPG 2.5.21 compatibility issue in `test_encryption_cek`, and for putting together this PR.

I apologize that we didn't end up merging it. After your fix, we realized that the `--compatibility-flags=no-partial-file-guard` approach unfortunately breaks the test on older GnuPG versions: they print an 'unknown compatibility flag' warning before the expected output, so the test's anchored match fails. We went with a version-agnostic approach in #2401 instead (decrypting to a real temporary file instead of `/dev/null`), which works with both old gpg and 2.5.21+.

Your report and initial fix were instrumental in pinpointing the root cause and validating the final solution. We really appreciate your contribution and the care you put into it. Thank you!

andreasstieger:
Thanks the note. Given the size of the alternate series I believe Linux distribution that adopt GnuPG 2.5.21 will appreciate a rnp release.

PR #2367, opened 2025-10-17
Test Suite: clean up interactions with gpg

dkg:
This series cleans up a series of minor quibbles with the way that the RNP test suite interacts with GnuPG.

A future GnuPG may choose responsible cryptographic primitives as a baseline, and decline signatures from or encrypting to 1024-bit RSA, so this series also moves all the tests that expected 1024-bit RSA interoperability to testing the ECC sample key in keyring 5.

GnuPG can also be built without some weaker/deprecated algorithms (e.g., El Gamal and IDEA), so this cleanup skips over some tests for those algorithms if `gpg --version` doesn't indicate support for them.

PR #2402, opened 2026-07-18
tests: port gpg/CI test-hygiene fixes from release/0.x (#2363, #2367, #2368)

ronaldtse:
Ports the test-suite hygiene work just merged into release/0.x over to main. All commits are test-only (no functional library changes), preserving dkg's authorship:

- from #2368: tolerate EXPIRED as well as EXPIRES in test_set_expire (slow builders)
- from #2363: detect 32-bit time_t via ctypes.c_time_t with sys.maxsize fallback
- from #2367 (14 commits): clean up GnuPG interactions — feature-detect and skip missing gpg algorithms (El Gamal, IDEA, CAST5, 3DES), use ECC keyring 5 instead of RSA-1024, RSA>=3072, robust gpg status/fingerprint parsing, User-ID escaping fixes, loopback pinentry, aes/compress-algo name handling, cleaner errors

Verified locally: full suite passes (304/304 rnp_tests + cli_tests).


ronaldtse:
ni4 could you help approve this? Thanks!

PR #2401, opened 2026-07-18
ci: fix broken CI jobs on main (typos, Fedora PQC Botan version, Codecov key)

ronaldtse:
Fixes CI breakages on main, plus ports the test-suite hygiene work from release/0.x.

CI fixes (verified green in this PR's checks):
1. **typos lint**: `priamry` in src/tests/ffi-key.cpp.
2. **Fedora 40 PQC job fails at configure**: matrix paired `pqc: On` (ENABLE_CRYPTO_REFRESH needs Botan >= 3.6 since 2dcae852) with Botan 3.3.0; bumped to a full Botan 3.6.0 build (the containers' minimized botan module lists lack ed448/x448/argon2, so a minimized build cannot satisfy crypto-refresh).
3. **Coverage jobs fail**: keybase.io/codecovsecurity/pgp_keys.asc no longer serves the Codecov PGP key; fetch it from keyserver.ubuntu.com by its documented fingerprint.
4. **macOS jobs fail in test_encryption_cek** (GnuPG 2.5.21 writes '<output>.part' first, so --output /dev/null fails): decrypt to a real temporary file — version-agnostic (supersedes #2400's flag approach, which breaks the test on older gpg).
5. **test leaks** exposed by the now-working Fedora sanitize job: test_ffi_encrypt_pk_with_v6_key leaked input/output/op on expected-failure iterations; test_ffi_decrypt_pqc_pkesk_test_vector leaked an rnp_ffi_t per iteration.
6. **ci/botan3-pqc-modules**: add missing argon2/ed448/x448 (+x25519 rename) needed by crypto-refresh builds (also fixes the oss-fuzz build together with google/oss-fuzz#15882).

Ported from release/0.x (test-only, dkg's authorship preserved):
- #2368: tolerate EXPIRED as well as EXPIRES in test_set_expire
- #2363: detect 32-bit time_t via ctypes.c_time_t with sys.maxsize fallback
- #2367 (14 commits): clean up GnuPG interactions — feature-detect and skip missing gpg algorithms, ECC keyring 5 instead of RSA-1024, RSA>=3072, robust gpg status/fingerprint parsing, User-ID escaping fixes, loopback pinentry, aes/compress-algo names, cleaner errors

Full test suite passes locally (304/304 rnp_tests + cli_tests).

PR #2409, opened 2026-07-21
ci: add missing Botan 3.6 modules for crypto-refresh/PQC

ronaldtse:
The oss-fuzz build (google/oss-fuzz#15882) fails because rnp's minimized Botan module list is missing modules required for crypto-refresh:

- `ENABLE_CRYPTO_REFRESH` requires `HKDF`, `ED448` and `X448`.
- `ed448` depends on `shake_xof` (which depends on `keccak_perm`).
- `argon2` depends on `blake2`.

This PR adds all of them to `ci/botan3-pqc-modules`. The list is only consumed by the oss-fuzz build script, so it doesn't affect the rnp CI matrix (which uses full Botan builds).

Once merged, the oss-fuzz PR's build should pass.

### DIFF
--- a/ci/botan3-pqc-modules
+++ b/ci/botan3-pqc-modules
@@ -1,8 +1,10 @@
 aead
 aes
+argon2
 auto_rng
 bigint
 blowfish
+blake2
 camellia
 cast128
 cbc
@@ -17,6 +19,7 @@ eax
 ecc_key
 ecdh
 ecdsa
+ed448
 ed25519
 elgamal
 eme_pkcs1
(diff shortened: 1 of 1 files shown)


ronaldtse:
This is a CI / test framework issue so just 1 approval is sufficient. @ni4 could you help check? Thanks!

ronaldtse:
Superseded by merging this fix into #2401 (6bf5d5cf on ci-fix-main). The botan module dependencies (blake2, shake_xof, keccak_perm) are now part of #2401, which will make both PRs green.

PR #2423, opened 2026-07-24
tests: add assert-encrypted-output test pattern

ronaldtse:
## Summary

Adds a test convention + test cases that verify encrypt / sign / key-export operations do not leak plaintext or passphrase material in their output.

### `docs/develop/testing-crypto-paths.adoc`
Documents the convention: for each crypto operation, generate a known-plaintext marker, run the op, assert the marker is absent from the output.

### `test_no_plaintext_leakage` (ffi-enc.cpp)
| Case | Operation | Assertion |
|---|---|---|
| 1 | Password encryption | Output does not contain plaintext marker |
| 2 | Public-key encryption | Output does not contain plaintext marker |
| 3 | Secret-key export | Output does not contain the passphrase |
| 4 | Detached signature | Output does not contain the signed plaintext |

## Test plan

- [x] All 4 cases pass locally (OpenSSL, macOS).
- [ ] CI: full test suite green.

### DIFF
--- a/docs/develop/testing-crypto-paths.adoc
+++ b/docs/develop/testing-crypto-paths.adoc
@@ -1,0 +1,40 @@
+= Testing crypto code paths
+
+== Assert-output-is-encrypted
+
+Security-critical code paths (encrypt, sign-then-encrypt, secret-key
+export) must include tests that verify the output does not leak
+plaintext key material. A bug that accidentally produces an unencrypted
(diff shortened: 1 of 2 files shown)


ni4 (approved):
LGTM!

ronaldtse:
Merging with 1 approval (ni4). Thank you @ni4 for the review!

PR #2418, opened 2026-07-24
ci+docs+tests batch: clang-tidy, man-page cache, gpg-flake retries, slow-test exclusion

ronaldtse:
## Summary

Consolidated CI/tests/docs batch — four related PRs combined for a single review (formerly #2418 + #2421 + #2458 + #2425), all in the 1-approval category:

1. **CONTRIBUTING.md + clang-tidy workflow + `.clang-tidy` config** — contributor docs and static analysis (non-gating), summarised into the GitHub step summary.
2. **Pre-generated man pages** — roff shipped in `docs/man/`, installed when no AsciiDoc processor is present, so distro tarball builds need no asciidoctor (#2395); `ci/regen-man-pages.sh` refreshes the cache.
3. **gpg interop flake retries** (#2457) — `run_gpg()` retries the two proven-transient signatures (`Corrupted protection`, `BAD signature`) in the four gpg-side helpers; anything else fails immediately.
4. **Slow fuzz-timeout regressions excluded from per-PR CI** — full coverage retained on scheduled runs.

## Test plan

- [x] Each piece was individually green on CI before consolidation
- [x] Man-page cache path and asciidoctor path both verified locally; gpg retry logic unit-checked; SignDefault/Encryption pass end-to-end
- [ ] Combined CI green

PR #2425, opened 2026-07-24
tests: exclude slow fuzz timeout regressions from per-PR CI

ronaldtse:
## Summary

- Splits `test_fuzz_keyring` and `test_fuzz_keyimport` into fast + `*_slow` variants. The slow variants contain only the oss-fuzz timeout-regression corpora that take minutes to load.
- Adds `-E '_slow$'` to every per-PR `ctest` invocation across all distro workflows so the slow tests are excluded from per-PR runs.
- The slow tests remain registered — they run under default `ctest` (nightly, manual dispatch, local full runs).

## Why

Profiled `rnp_tests` on macOS Botan: ~9.9 min total. Two tests dominated:

| Test | Time |
| --- | --- |
| `test_fuzz_keyring` | 273 s |
| `test_fuzz_keyimport` | 52 s |
| **Subtotal** | **325 s / 592 s** |

Both spend their time loading a single large oss-fuzz corpus file (`timeout-6140201111519232`, 791 KB; `timeout-9c10372fe9ebdcdb0b6e275d05f8af4f4e3d6051`, 182 KB). These are regression guards against previously-fixed timeout bugs — they're valuable but shouldn't gate every PR.

Excluding them drops per-PR `rnp_tests` runtime to ~4.5 min on this machine. The slow tests still pass when included.

## Test plan

- [x] Built and ran `rnp_tests` locally — both fast and slow variants pass
- [x] Verified `ctest -N -E '_slow$'` excludes exactly the two new `_slow` tests (285 → 283)
- [x] Verified `ctest -N -R '_slow$'` lists exactly the two new `_slow` tests (+ fixture)
- [ ] CI green on all platforms

### DIFF
--- a/.github/workflows/centos-and-fedora.yml
+++ b/.github/workflows/centos-and-fedora.yml
@@ -222,6 +222,7 @@ jobs:
           else
             exec sudo -u rnpuser ctest --parallel ${{ env.CORES }} --test-dir build --output-on-failure
           fi
+          exec su rnpuser -c "ctest --parallel ${{ env.CORES }} --test-dir build --output-on-failure -E '_slow$'"
 
       - name: Coverage
         if: env.BUILD_MODE == 'coverage'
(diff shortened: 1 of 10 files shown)


ronaldtse:
Consolidated into #2418 as part of the CI/tests/docs batch (four 1-approval PRs combined into a single review). This PR's commits are included there verbatim; CI runs on #2418.

PR #2424, opened 2026-07-24
ffi: security rule enumeration API + FIPS mode foundation

ronaldtse:
## Summary

Two related security-profile changes, combined for one review (formerly #2424 + #2427):

1. **Security rule enumeration API** — `rnp_get_security_rule_count()` / `rnp_get_security_rule_at()`: enumerate the active security profile's rules (default SHA-1/MD5 rules plus user-added), covering hash/symmetric/public-key feature types, levels, timestamps and flags.
2. **FIPS-compliant mode foundation** — `ENABLE_FIPS_MODE` CMake switch + `rnp_is_fips_mode_enabled()`: compile-time tightening of the default SecurityProfile (non-FIPS algorithms Disabled, OpenSSL backend only). See docs/security.adoc for the certification boundary — rnp consumes OpenSSL's FIPS provider, it is not a validated module itself.

## Test plan

- [x] Enumeration tests cover PublicKey branch and PROHIBITED level
- [x] FIPS default rule set tested under `ENABLE_FIPS_MODE` build
- [ ] CI matrix green

PR #2427, opened 2026-07-24
ffi: FIPS-compliant mode foundation (#06)

ronaldtse:
## Summary

Foundation for FIPS-compliant mode (roadmap item #06). Adds the build switch, runtime query, and default algorithm policy. The full FIPS effort is a multi-PR initiative; this PR lays the groundwork only.

### What's in this PR

- **`CMakeLists.txt`**: new option `ENABLE_FIPS_MODE` (default Off). When On, forces `CRYPTO_BACKEND=openssl` (Botan rejected with helpful error) and requires OpenSSL 3.x.
- **`src/lib/CMakeLists.txt`**: defines `RNP_FIPS_MODE` on `librnp-obj` (PUBLIC). Propagates to `librnp` and `librnp-static` via `INTERFACE_COMPILE_DEFINITIONS` (added to the existing foreach that copies interface properties).
- **`src/lib/sec_profile.cpp`**: when `RNP_FIPS_MODE` is defined, the `SecurityContext` constructor adds `Disabled`-level rules for non-FIPS-approved algorithms:
  - Symmetric: Twofish, CAST5, IDEA, Blowfish, SM4
  - Hash: MD5, SHA1, RIPEMD160
  - Public key: EdDSA, SM2, ElGamal, ElGamal-Encrypt-Or-Sign
- **`include/rnp/rnp.h` + `src/lib/rnp.cpp`**: new FFI function `rnp_is_fips_mode_enabled()` returning whether the library was built with `ENABLE_FIPS_MODE=On`.
- **`docs/security.adoc`** (new): documents the certification boundary, required OpenSSL version, how to enable, how to query, and current limitations. Makes clear that rnp is a *consumer* of OpenSSL's validated FIPS provider module, not a validated module itself.
- **`src/tests/ffi.cpp`**: new test `test_ffi_fips_mode_query` verifying the FFI query.
- **`.github/workflows/fips-mode.yml`** (new): CI leg building and testing under `-DENABLE_FIPS_MODE=On`.

### Why

FIPS 140-3 certification (or its successors) is required for OpenPGP libraries used in US federal, Canadian federal, and many regulated industries. Without a FIPS mode, rnp is invisible to that procurement channel. This PR doesn't get us to certification, but it puts the right bones in place: a clean compile-time switch, an algorithm policy expressed through the existing security-rule registry (not scattered `#ifdef`s), and a clear certification-boundary statement.

### Out of scope (follow-up PRs)

- Constant-time primitive audit
- Secret zeroisation audit
- Argon2 → PBKDF2 S2K migration
- FIPS provider load/unload ceremony
- Test-suite adaptation (many existing tests use non-FIPS algorithms and will fail under FIPS mode — see Test Plan below)
- Real FIPS validation effort

### Test plan

- [x] Local build under `-DCRYPTO_BACKEND=openssl -DENABLE_FIPS_MODE=On` succeeds (Botan correctly rejected, OpenSSL 1.1.1 would be rejected)
- [x] `test_ffi_fips_mode_query` passes (returns 1 in FIPS build, 0 in default build via `#if defined(RNP_FIPS_MODE)`)
- [ ] CI green on the new `fips-mode` workflow
- [ ] **Known**: full `rnp_tests` under FIPS mode currently fails ~37 tests because they use disabled algorithms (DSA, EdDSA, MD5, SHA1, etc.). Adapting these tests is follow-up work; the foundation PR deliberately lands the policy first so we can see what breaks.

### DIFF
--- a/.github/workflows/fips-mode.yml
+++ b/.github/workflows/fips-mode.yml
@@ -1,0 +1,80 @@
+name: fips-mode
+
+# Build rnp with -DENABLE_FIPS_MODE=On and run the test suite. FIPS mode
+# requires the OpenSSL backend (3.x); this workflow exists so the FIPS build
+# gets CI coverage even though most distro legs use the default profile.
+# See docs/security.adoc and roadmap item #06.
+
+on:
+  push:
+    branches:
+      - main
+      - 'release/**'
+    paths-ignore:
+      - '/*.sh'
+      - '/.*'
+      - '/_*'
+      - 'Brewfile'
+      - 'docs/**'
+      - '**.adoc'
+      - '**.md'
+      - '**.nix'
+      - 'flake.lock'
+      - '.github/workflows/*.yml'
+      - '!.github/workflows/fips-mode.yml'
+  pull_request:
+    paths-ignore:
+      - '/*.sh'
+      - '/.*'
+      - '/_*'
+      - 'Brewfile'
+      - 'docs/**'
+      - '**.adoc'
+      - '**.md'
+      - '**.nix'
+      - 'flake.lock'
+  workflow_dispatch:
+
+concurrency:
+  group: '${{ github.workflow }}-${{ github.job }}-${{ github.head_ref || github.ref_name }}'
+  cancel-in-progress: true
+
+jobs:
+  fips-build-and-test:
+    name: FIPS mode (openssl 3.x)
+    runs-on: ubuntu-latest
+    timeout-minutes: 50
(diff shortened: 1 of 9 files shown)


ronaldtse:
Consolidated into #2424: the FIPS-mode commits have been fast-forward merged into the `weak-cipher-reporting` branch, combining the security-rule enumeration API and the FIPS foundation into one review. CI for the combined change runs on #2424.

Issue #1947, opened 2022-12-06
Support Photo IDs (User Attribute Packet (Tag 17)/The Image Attribute Subpacket)


antonsviridenko:
## Description

PGP specification has additional optional way to represent and identify key holder by
adding his photo to the key data.

> The Image Attribute subpacket is used to encode an image, presumably
   (but not required to be) that of the key owner.

GnuPG supports these Photo IDs, can add and display these photos (using external image viewer).
It would be nice if RNP could do the same. Now `rnpkeys` is only able to show presence of photo ID
packet by printing `(photo)` string in the key description.

## Steps to Reproduce

1. Generate the new key using `gpg`
2. Edit newly generated key by typing `gpg --edit-key  KEYID`
3. type `addphoto`, specify path to some JPEG image, press `y` then `quit` then again `y`
4. export the key from `gnupg`, import into the `rnp` keystore

```
pub   3072/RSA b548769515f29bc2 2022-12-06 [SC] [EXPIRES 2024-12-05]
      ad5f866d9eab8e061074f3d9b548769515f29bc2
uid           photo id test key <photoid@test>
uid           (photo)
sub   3072/RSA 82185fc7972aee8d 2022-12-06 [E] [EXPIRES 2024-12-05]
      8c65135dfb3bac5e17928a1482185fc7972aee8d
```

## Expected Behavior

`rnpkeys` CLI could be extended with additional commands & options like `--edit-key --add-photo` , `--showphoto KEYID`.

ni4:
There were some discussion in openpgp mailing list (cannot locate it now) about deprecating user attribute, so at least this one is not of high priority.

PR #2430, opened 2026-07-25
ffi: add rnp_uid_get_photo() for Photo ID read support (#13)

ronaldtse:
## Summary

Foundation for issue #1947 (Photo ID support). Adds an FFI function that parses the User Attribute packet body and returns just the image bytes plus the detected format, so callers don't have to implement the subpacket length decoding and image-header walk themselves.

### What's in this PR

- **`include/rnp/rnp.h`**: new constants `RNP_PHOTO_FORMAT_{UNKNOWN,JPEG,PNG}` and new FFI function `rnp_uid_get_photo(uid, &format, &data, &size)` with thorough Doxygen.
- **`src/lib/userid.{hpp,cpp}`**: new helper `rnp::parse_photo_attribute(uid_body, image, format)` that walks the variable-length subpacket prefix, finds the first Image Attribute subpacket (type 1), and returns its image bytes plus detected format. The image header length byte counts itself (RFC 4880 §5.12.1 / RFC 9580 §5.13.1), so the remaining header bytes to skip are `hdr_len - 1`, not `hdr_len`. Format is detected from magic bytes (JPEG: FF D8 FF, PNG: 89 50 4E 47), not from the UserAttr header's format byte — real-world keys frequently leave that byte at 0 even when the image data is a valid JPEG.
- **`src/lib/rnp.cpp`**: FFI implementation. Returns `RNP_ERROR_BAD_PARAMETERS` if the UID is not a User Attribute packet, `RNP_ERROR_BAD_STATE` if the UserAttr body is malformed or contains no image subpacket.
- **`src/tests/ffi.cpp`**: extends the existing `test_ffi_load_userattr` to exercise the new API: asserts the photo UID returns JPEG bytes with the correct magic header, and that calling `rnp_uid_get_photo()` on a non-photo UID is rejected.

### Why

The "gpg can, rnp can't" feature gap is most visible to end users — Thunderbird users with photo IDs on their keys see them silently dropped when their keyring is read by rnp. The UserAttr packet was already loaded as an opaque UID with `type=RNP_USER_ATTR` and the raw bytes retrievable via `rnp_uid_get_data()`, but parsing those bytes was the caller's job. This PR makes the common case (a JPEG photo) one FFI call.

### Out of scope (follow-up PRs)

- `rnp_key_add_photo()` to attach a new photo to a key (needs serialiser work)
- CLI: `--show-photo` flag, `[photo]` indicator in `--list-keys`
- Negative tests with crafted malformed UserAttr bodies
- PNG photo round-trip test (would need a PNG fixture)

### Test plan

- [x] `test_ffi_load_userattr` (extended) passes: verifies JPEG magic, JPEG format constant, rejection of non-photo UID
- [x] Manual verification against `data/test_stream_key_load/ecc-25519-photo-pub.asc` — image bytes form a valid JPEG (starts with `FF D8 FF E0`, ends with `FF D9`)
- [x] No regressions in `test_ffi_add_userid`, `test_ffi_key_userid_dump_has_no_special_chars`, `test_ffi_key_dump_edge_cases`
- [ ] CI green

### DIFF
--- a/include/rnp/rnp.h
+++ b/include/rnp/rnp.h
@@ -115,6 +115,13 @@ typedef uint32_t rnp_result_t;
 #define RNP_USER_ATTR (2U)
 
 /**
+ * Photo ID format, returned by rnp_uid_get_photo().
+ */
+#define RNP_PHOTO_FORMAT_UNKNOWN (0U)
+#define RNP_PHOTO_FORMAT_JPEG (1U)
+#define RNP_PHOTO_FORMAT_PNG (2U)
+
+/**
  * Predefined feature security levels
  */
 #define RNP_SECURITY_PROHIBITED (0U)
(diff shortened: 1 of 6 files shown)


ni4 (approved):
LGTM!

PR #2431, opened 2026-07-25
ffi: rnp_key_protect_ex() with Argon2 + AEAD, and PQC/Argon2 negative tests

ronaldtse:
## Summary

Two related changes, combined for one review (formerly #2431 + #2432):

1. **FFI: `rnp_key_protect_ex()`** — protects a secret key with explicit S2K parameters, supporting Argon2 (crypto-refresh) as the KDF and AEAD encryption of protected key material.
2. **Negative tests for PQC PKESK + Argon2 S2K** (#2355, #2296 follow-ups) — including tests for the failure modes of the new `rnp_key_protect_ex()` path (malformed Argon2 parameters, AEAD tampering).

## Test plan

- [x] Feature exercised by the new negative tests
- [ ] CI matrix green


ni4 (approved):
LGTM, but CodeQL complains about unused variables. @ronaldtse should we handle this, or it is false-positive?

PR #2434, opened 2026-07-26
docs: add common workflow patterns guide

ronaldtse:
## Summary

Adds a documentation page describing three common OpenPGP workflows that rnp clients ask about most often:

1. **Pre-generating a revocation certificate** — useful when the user wants to be able to revoke a key even after losing the secret material.
2. **Encrypting an arbitrary payload to a third-party public key** — the building block for escrow-style workflows where a third party needs to read material the data owner cannot.
3. **Publishing a revocation to a keyserver** — exporting the revoked-key wire-format bytes for the keyserver HTTPS-PUT step.

Each workflow is shown as a small composition of existing FFI calls (no new APIs introduced). rnp is intentionally a crypto library, not a workflow framework; this page collects the workflow compositions that clients ask about most often into one reference.

### Why

Reviewers and downstream integrators regularly ask "how do I do X" questions that span multiple FFI calls but aren't worth a dedicated API. This page captures three of the most common patterns so the answer is "see the workflows guide" instead of "grep the test suite."

### What's NOT in this PR

- No new FFI functions.
- No mention of any specific application or external proposal. The workflows are generic patterns; specific applications (enterprise key escrow, client-side escrow services, etc.) compose them on the application side.
- The exact verification ceremony (email confirmation, delay windows, rate limiting) is explicitly called out as an application-layer concern in the "Composing the patterns" section.

### Test plan

- [x] Doc renders cleanly (asciidoc syntax verified by visual inspection)
- [x] Code examples use only existing FFI symbols (no missing headers)
- [x] Added to `docs/navigation.adoc` so it appears in the docs index

### DIFF
--- a/docs/develop/workflows.adoc
+++ b/docs/develop/workflows.adoc
@@ -1,0 +1,142 @@
+= Common workflow patterns
+
+This document describes common OpenPGP workflows that clients of the
+rnp C API may want to compose from the primitive operations rnp
+provides. rnp is intentionally a crypto library, not a workflow
+framework: each pattern below is a small composition of existing
+FFI calls rather than a single-purpose API.
+
+== Pre-generating a revocation certificate
+
+A revocation certificate is a revocation signature that can be
+published later, independent of when it was created. Generating
+one in advance and storing it somewhere safe allows the key to be
+revoked even after the secret key material is lost (for example, if
+the device that holds the key fails).
+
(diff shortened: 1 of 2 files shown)


ni4 (approved):
LGTM!

ronaldtse:
Merging with 1 approval (ni4). Thank you @ni4 for the review!

PR #2435, opened 2026-07-26
ffi: entropy encoding + backup archive building blocks

ronaldtse:
## Summary

Adds two general-purpose building-block APIs that clients can compose into backup, sync, and escrow workflows. Each is self-contained and parameterised; rnp deliberately does not bake in any specific application's workflow.

### What's in this PR

**Entropy encoding utility** (`rnp_entropy_encode_human_readable`, `_decode`, `_validate`):
- Generates cryptographically random entropy via the FFI RNG.
- Encodes it in a caller-specified alphabet, split into groups with optional identifiers and optional SHA-256-based checksum.
- Useful whenever random bytes need to be transcribed by humans (paper backup, manual entry).
- Default alphabet is hex; callers supply their own alphabet via the params struct (e.g. handwriting-friendly variants).
- `disable_*` bool fields follow C convention: `false` (zero-init default) = include feature, `true` = omit. This lets callers pass an all-zero params struct for the common default case.

**Backup archive format** (`rnp_backup_archive_create`, `_load`):
- Wraps a sequence of secret-key exports in a standard OpenPGP signed-and-encrypted message.
- Wire format is a literal-data packet containing concatenated transferable secret keys, signed by a caller-specified key, encrypted to a caller-specified public key.
- Format is RFC 9580-compliant; any OpenPGP library can produce or consume it.
- Documented in `docs/develop/backup-format.adoc`.

**Tests**:
- `test_ffi_entropy_encoding_roundtrip` covers default-alphabet round-trip, custom-alphabet round-trip, order-independent group entry, tampered-checksum rejection, invalid-params rejection.
- `test_ffi_backup_archive_roundtrip` covers generate-keys → create archive → fresh FFI → load archive → verify recovered key present and secret.

### Why

These two building blocks come up in many real-world workflows:
- Enterprise key escrow
- Personal offline backups
- Multi-device key synchronisation
- Application-specific recovery flows

Previously each client would re-implement the entropy encoding and the signed+encrypted container pattern from scratch, with attendant correctness risk (checksum bugs, signature-then-encrypt ordering bugs, etc.). This PR gives them a tested, canonical implementation in the rnp library itself.

### Architectural decision

rnp is intentionally a crypto library, not a workflow framework. The APIs in this PR expose **building blocks** that clients compose on their own; they do NOT prescribe a specific workflow. Naming (`encryption_key`, `decryption_key`, `signing_key`) is intentionally generic so multiple applications can adopt the same primitives without coupling.

### What's NOT in this PR

- No mention of any specific application or external proposal.
- No CLI surface for these APIs (callers use the FFI directly).
- No "recovery keypair" or "data-signkey" pattern baked in; callers compose their own key structure as they see fit.

### Test plan

- [x] Build clean on default config (no warnings)
- [x] `test_ffi_entropy_encoding_roundtrip` passes locally (5 cases covered)
- [x] `test_ffi_backup_archive_roundtrip` passes locally (full round-trip with separate FFIs)
- [x] No regression in `test_ffi_encrypt_pass`, `test_ffi_argon2_locked_seckey`, `test_ffi_encrypt_pk_with_v6_key`
- [ ] CI green on all platforms

### DIFF
--- a/docs/develop/backup-format.adoc
+++ b/docs/develop/backup-format.adoc
@@ -1,0 +1,82 @@
+= Backup archive wire format
+
+The `rnp_backup_archive_create()` FFI produces a standard OpenPGP
+message containing one or more transferable secret keys, signed by a
+caller-specified key and encrypted to a caller-specified public key.
+This document describes the wire format so non-rnp implementations
+can produce and consume compatible archives.
+
+== Container structure
+
+A backup archive is an OpenPGP Signed-and-Encrypted message as
+defined by [RFC 9580]. The high-level structure is:
+
+----
+PKESK v6 (or v3)
+SEIPD v2 (or v1)
+    one-pass signature packet
+    literal data packet
+        concatenated transferable secret keys
+    signature packet
+----
+
+Where:
+
+* The PKESK encrypts a session key to the `encryption_key` parameter.
+* The SEIPD packet contains the signed message body.
+* The literal data packet carries the concatenation of one or more
+  transferable secret keys, each exported with
+  `RNP_KEY_EXPORT_SECRET | RNP_KEY_EXPORT_SUBKEYS`.
+* The signature packet is produced by the `signing_key` parameter,
+  covering the literal data.
+
+== Transferable key concatenation
+
+Each transferable key is exported as the standard sequence of packets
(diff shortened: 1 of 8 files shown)


ni4 (approved):
LGTM as soon as CI passes!

PR #2442, opened 2026-07-27
Support LibreSSL as an openssl-family backend

### LINKED ISSUE #2440: findopensslfeatures.c uses OpenSSL-internal EVP_PKEY_meth_* — blocks LibreSSL backend
## Summary

`cmake/Modules/findopensslfeatures.c` uses the OpenSSL-internal `EVP_PKEY_meth_*` API to enumerate public-key algorithms. This prevents `CRYPTO_BACKEND=openssl` from working against **LibreSSL**, which doesn't expose those functions. A drop-in public API exists and works across OpenSSL 1.1, OpenSSL 3.x, and LibreSSL.

## Reproduction

Build librnp 0.18.1 against LibreSSL 4.0 (static):

```
cmake -S rnp-v0.18.1 -B rnp-build \
  -DCRYPTO_BACKEND=openssl \
  -DBUILD_SHARED_LIBS=OFF \
  -DCMAKE_PREFIX_PATH=/path/to/libressl-install \
  ...
cmake --build rnp-build
```

Result:

```
/Users/runner/work/_temp/rnp-build/fossl/findopensslfeatures.c:94:32: error: incompatible integer to pointer conversion initializing 'const EVP_PKEY_METHOD *' ...
   94 |         const EVP_PKEY_METHOD *pmeth = EVP_PKEY_meth_get0(i);
      |                                ^       ~~~~~~~~~~~~~~~~~~~~~
/Users/runner/work/_temp/rnp-build/fossl/findopensslfeatures.c:96:9: error: call to undeclared function 'EVP_PKEY_meth_get0_info'
   96 |         EVP_PKEY_meth_get0_info(&id, NULL, pmeth);
      |         ^
```

LibreSSL provides `EVP_PKEY_asn1_get0` and `EVP_PKEY_asn1_get0_info` (the public counterpart) but does not — and per their maintainers, will not — expose the `meth` family.

## Suggested fix

`EVP_PKEY_asn1_get_count` / `EVP_PKEY_asn1_get0` / `EVP_PKEY_asn1_get0_info` are the documented public API for enumerating public-key algorithms. They're a near 1:1 replacement for the `meth` family and are present in OpenSSL 1.1.0+, OpenSSL 3.x, and LibreSSL.

```c
int
list_publickey()
{
#if OPENSSL_VERSION_NUMBER < 0x30000000L
    for (size_t i = 0; i < EVP_PKEY_asn1_get_count(); i++) {
        const EVP_PKEY_ASN1_METHOD *ameth = EVP_PKEY_asn1_get0(i);
        int                    id = 0;
        EVP_PKEY_asn1_get0_info(&id, NULL, NULL, NULL, NULL, ameth);
        printf("%s\n", OBJ_nid2ln(id));
    }
#else
    EVP_KEYMGMT_do_all_provided(NULL, print_km, NULL);
#endif
    return 0;
}
```

(Traceable upstream signature for `EVP_PKEY_asn1_get0_info`: `int EVP_PKEY_asn1_get0_info(int *ppkey_id, int *pkey_base_id, int *ppkey_flags, const char **pinfo, const char **ppem_str, const EVP_PKEY_ASN1_METHOD *ameth);` — five args after the out-params, vs. `EVP_PKEY_meth_get0_info`'s three. Same idea, slightly richer output.)

## Why this matters

- **LibreSSL users are blocked from the OpenSSL backend.** LibreSSL is the maintained OpenSSL-1.1-API implementation (OpenSSL 1.1.x went EOL 2023-09). Without this fix, anyone on OpenBSD, Alpine (optional), or a LibreSSL-based distro can't use librnp without switching to the Botan backend.
- **`EVP_PKEY_meth_*` is documented as internal-ish.** OpenSSL's headers note these are not part of the stable public API. Even on OpenSSL itself they could change between versions.
- **The fix is ~5 lines and works on every supported backend.**

## Environment

- librnp: 0.18.1 release tarball
- LibreSSL: 4.0.0 (cmake build, `BUILD_SHARED_LIBS=OFF`)
- Also confirmed against current LibreSSL 4.1.0
- Reproducible on x86_64-linux, aarch64-linux, x86_64-macos, aarch64-macos

Happy to send a PR if a maintainer confirms the approach.

ronaldtse:
## Summary

Adds LibreSSL as a supported `openssl`-family backend. Closes #2440.

The original issue reported that `cmake/Modules/findopensslfeatures.c`
used the OpenSSL-internal `EVP_PKEY_meth_*` family to enumerate
public-key algorithms, which blocked `CRYPTO_BACKEND=openssl` against
LibreSSL. The 5-line fix suggested in the issue (switch to the public
`EVP_PKEY_asn1_*` family) does unblock backend feature detection, but
a full rnp build against LibreSSL needed several more internal-API
removals. This PR lands all of them.

## Validation

Built and smoke-tested against:

- OpenSSL 1.1.1w
- OpenSSL 3.6.3
- LibreSSL 4.3.2 (Homebrew)
- BoringSSL master (built from source)

OpenSSL 1.1, 3.x, and LibreSSL build, generate keys, sign/verify, and
encrypt/decrypt. BoringSSL does not — see *BoringSSL* below.

### `cmake/Modules/findopensslfeatures.c`
- `list_publickey()`: replaced `EVP_PKEY_meth_*` with
  `EVP_PKEY_asn1_*` (the documented public enumeration API).
- Added `<openssl/crypto.h>` include (BoringSSL doesn't transitively
  pull in `OPENSSL_malloc` / `OPENSSL_free`).

### `src/lib/CMakeLists.txt`
- Dropped the redundant `AES-128-OCB` / `AES-192-OCB` / `AES-256-OCB`
  entries from `_openssl_required_features`. The
  `resolve_feature_state(ENABLE_AEAD ...)` call immediately below
  already handles absence correctly; the required entries just made
  the build fail on backends without OCB (LibreSSL) instead of cleanly
  disabling `ENABLE_AEAD_OCB`.

### `src/lib/crypto/mem.h`
- `ossl_allocator`: secure heap (`OPENSSL_secure_zalloc` /
  `CRYPTO_secure_allocated` / `OPENSSL_secure_clear_free`) is now
  guarded. On LibreSSL, falls back to `calloc` / `OPENSSL_cleanse` /
  `free`.

### `src/lib/crypto/elgamal_ossl.cpp`
- Replaced `BN_RECP_CTX` with `BN_mod_mul` for the small-subgroup
  check (functionally equivalent, slightly slower on large loop
  counts).
- Replaced `BN_secure_new` with `BN_new` for two secret BIGNUMs.

### `src/lib/crypto/ossl_utils.hpp`
- Dropped the now-unused `BNRecpCtx` typedef.

### `src/lib/crypto/{ec,rsa,dl}_ossl.cpp`
- Guarded `EVP_PKEY_check` / `EVP_PKEY_public_check` /
  `EVP_PKEY_param_check` behind
  `#if !defined(LIBRESSL_VERSION_NUMBER)`. rnp's callers perform
  higher-level key sanity checks elsewhere.

### `src/lib/crypto/dsa_ossl.cpp`
- Use `EVP_PKEY_CTX_ctrl` directly on LibreSSL (LibreSSL ships the
  control constant but not the
  `EVP_PKEY_CTX_set_dsa_paramgen_q_bits` macro wrapper).

### `src/lib/crypto/backend_version.cpp`
- `backend_string()` returns `"LibreSSL"` on LibreSSL.
- Version regex matches both `"OpenSSL "` and `"LibreSSL "` prefixes.

### `docs/crypto-backends.adoc` (new) + `README.adoc`
- New doc with backend selection matrix, platform list, feature
  comparison, provider model, compile-time detection notes, and build
  recipes.

## BoringSSL

BoringSSL was investigated as part of this work but is **not** in
scope for this PR — it requires significantly more than API-surface
porting. BoringSSL removes algorithms that rnp lists as required
features:

- `SHA3-256`, `SHA3-512` — BoringSSL doesn't implement SHA-3 at all
  (and `EVP_get_digestbyname("SHA3-256")` returns NULL). SHA-3 is
  load-bearing for OpenPGP.
- `DSAENCRYPTION` / `DHKEYAGREEMENT` — NIDs are declared but the
  algorithms aren't implemented.
- `AES-128-OCB` and friends — not present.
- `CAMELLIA-*`, `IDEA-*`, `BF-*`, `CAST5-*`, `SM[234]`, `RIPEMD160`,
  `ChaCha20` (as an EVP cipher name) — all absent.

Bringing rnp up on BoringSSL would require making SHA-3 optional,
shipping a vendored SHA-3, or both. That's a separate, larger effort
— filed as a follow-up issue rather than tracked here.

The compile-time BoringSSL API guards are landed in this PR
(`#if defined(OPENSSL_IS_BORINGSSL)` branches in `findopensslfeatures.c`
and `mem.h`) since they're small, harmless on other backends, and
document the intent.

## Test plan

- [x] `cmake -S . -B build -DCRYPTO_BACKEND=openssl` succeeds on
      OpenSSL 1.1.1w, OpenSSL 3.6.3, LibreSSL 4.3.2
- [x] `cmake --build build` succeeds on all three
- [x] `rnpkeys --generate-key` works on all three (RSA 3072 exercised
      on LibreSSL)
- [x] `rnp --sign` / `rnp --verify` round-trips on LibreSSL
- [x] `rnp --encrypt` / `rnp --decrypt` round-trips on LibreSSL
- [x] `rnp --version` reports the correct backend name and version on
      each backend
- [ ] CI runs on Ubuntu / macOS / Windows / FreeBSD

Refs #2440.


ni4 (approved):
LGTM!

PR #2446, opened 2026-07-29
fix: silence -Wnonnull-compare and Botan FFI deprecation warnings

ronaldtse:
## Summary

Fixes two build-warning classes reported in the v0.18.1 release build (GCC).

### 1. `-Wnonnull-compare` in `src/librekey/rnp_key_store.cpp`

`RNP_LOG_KEY(msg, &srckey)` expanded to `if (!(&srckey))`, but `&srckey` is the address of a `Key&` reference and is provably non-null, so GCC flagged every such call site. Added an `RNP_LOG_KEY_NN` variant (no null check) in `src/lib/logging.h` and switched **all** `&srckey` call sites to it. Nullable-pointer sites (`primary`, `added_key`) keep the null-checking `RNP_LOG_KEY`. Verified there are no remaining `RNP_LOG_KEY(..., &ref)` call sites in the tree.

### 2. `-Wdeprecated-declarations` in `src/lib/crypto/ec.cpp` and `src/lib/crypto/eddsa.cpp`

`botan_privkey_x25519_get_privkey`, `botan_pubkey_x25519_get_pubkey`, and `botan_privkey_ed25519_get_privkey` are deprecated in Botan 3.x. Switched to the generic `botan_privkey_view_raw` / `botan_pubkey_view_raw` API (available since Botan 3.6) under a `BOTAN_VERSION_CODE >= 3.6.0` guard, keeping the legacy calls for older Botan (2.x and 3.0–3.5, where the helpers are not deprecated and `view_raw` does not exist). A small length-checked view callback (`rnp_botan_view_bin`) and descriptor (`rnp_botan_view_buf`) live in `src/lib/crypto/botan_utils.hpp`; the per-site descriptors are declared inside the `#if` so they don't trigger `-Wunused-variable` on older Botan.

A codebase-wide sweep confirmed the remaining deprecated Botan FFI calls (`botan_key_wrap3394`/`botan_key_unwrap3394` in `ecdh.cpp`, the `_enc` SM2 loaders) are already behind `CRYPTO_BACKEND_BOTAN3` guards or use the non-deprecated overloads, and that ed448/x448 paths do not use the deprecated getters.

## Notes

- Builds against Botan 2.x and 3.0–3.5 are unaffected (legacy call paths retained).
- The pre-existing Botan 3.12 native C++ API breakage in `ec.cpp` (around `Botan::EC_Group::from_name`, under `ENABLE_CRYPTO_REFRESH`) is unrelated to these warning classes and intentionally left out of scope.

## Test plan

- [x] Configured with `-DCRYPTO_BACKEND=botan3` against Botan 3.12; `librnp-obj`, `rnpkeys`, and `rnp` build with **zero warnings/errors** (Apple Clang 15).
- [x] `ec.cpp`, `eddsa.cpp`, `rnp_key_store.cpp` compile clean.
- [x] Runtime: generated an Ed25519 primary + Curve25519 ECDH subkey (expert mode option 22) — exercises all three migrated `view_raw` call sites.
- [x] Sign/verify with the Ed25519 key passes; encrypt/decrypt round-trip through the Curve25519 subkey passes.
- [ ] CI (Linux/GCC) confirms `-Wnonnull-compare` and `-Wdeprecated-declarations` are gone for these files.

### DIFF
--- a/src/lib/crypto/botan_utils.hpp
+++ b/src/lib/crypto/botan_utils.hpp
@@ -28,8 +28,34 @@
 #define RNP_BOTAN_UTILS_HPP_
 
 #include <botan/ffi.h>
+#include <botan/build.h>
+#include <botan/version.h>
+#include <cstring>
 #include "mpi.hpp"
 #include "utils.h"
+
+/* Destination descriptor for the Botan >= 3.6 view-style FFI helpers
+ * (botan_privkey_view_raw / botan_pubkey_view_raw), which replace the
+ * per-algorithm botan_{privkey,pubkey}_*_get_* helpers deprecated since
+ * Botan 3.x. The callback refuses to copy when the viewed length does not
+ * match the caller's expectation, so a fixed-size buffer cannot overrun. */
+struct rnp_botan_view_buf {
+    uint8_t *data;
+    size_t   size;
+};
+
+extern "C" {
+static inline int
+rnp_botan_view_bin(botan_view_ctx ctx, const uint8_t *data, size_t len)
+{
+    auto *vb = static_cast<rnp_botan_view_buf *>(ctx);
+    if (len != vb->size) {
+        return -1;
+    }
+    std::memcpy(vb->data, data, len);
+    return 0;
+}
+}
 
 /* Self-destructing wrappers around the Botan's FFI objects */
 namespace rnp {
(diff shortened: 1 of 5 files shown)


ronaldtse:
Closing — superseded by #2448 which migrates Botan backend to FFI, resolving the deprecation warnings this PR was silencing. Combined to reduce review overhead.

PR #2447, opened 2026-07-29
refactor: migrate crypto-refresh EC/Ed code from Botan native API to FFI

ronaldtse:
## Summary

Addresses the **Botan 3.12 `ENABLE_CRYPTO_REFRESH` build breakage** by migrating the crypto-refresh EC/Ed code off Botan's *unstable native C++ API* and onto the *stable FFI* — the same discipline the rest of rnp already follows (which is why it rides across Botan 2.x/3.x untouched).

The native-API code bit-rotted on 3.12: tightened include hygiene made `Botan::EC_Group` / `EC_AffinePoint` / `BigInt` incomplete types, and `public_point()` + the `Ed25519_PrivateKey(span)` ctor are deprecated.

Migrated files (now contain **zero** `Botan::` references):

- **`src/lib/crypto/ec.cpp`** — `ec_generate_generic_native` mirrors the existing FFI keygen in `Key::generate()` (`botan_privkey_create` + `get_field`); native `bigint.h`/`ecdh.h` includes dropped.
- **`src/lib/crypto/ed25519_ed448.cpp`** — Ed25519/Ed448 keygen, sign, verify, validate via `botan_privkey_create`/`load_*`, `botan_pk_op_sign_*`/`verify_*`, `botan_*_check_key`.
- **`src/lib/crypto/exdsa_ecdhkem.{cpp,h}`** — replaced the `Botan::*` key-object helpers with FFI key handles; ECDH KEM via `botan_pk_op_key_agreement_*`, ECDSA sign/verify, `check_key`, SEC1 loaders. Private `key_` member moved from `Botan::secure_vector` to `rnp::secure_bytes`; all `<botan/*>` includes removed from the header.

## FFI gap audit (for discussion with Jack Lloyd)

`TODO.refactor/botan-ffi-migration.md` documents a full audit: **no Botan FFI gaps were found.** Every native op rnp uses (ECDH/ECDSA/Ed25519/Ed448/X25519/X448 keygen+load, key agreement, KEM, sign/verify, check_key, ML-DSA/ML-KEM/SLH-DSA loaders, RFC 3394 keywrap) has a stable FFI equivalent. Nothing needs to be requested upstream. The doc also lays out the Phase 2 roadmap (migrate the remaining native-API files: `x25519_x448.cpp`, `dilithium*`, `kyber*`, `sphincsplus.cpp`).

## Test plan

[...]

PR #2448, opened 2026-07-30
refactor: migrate Botan backend crypto-refresh/PQC code to FFI

ronaldtse:
## Summary

Migrates the **Botan backend** off Botan's C++ API and onto the stable C FFI (`botan_*`), matching rnp's existing FFI discipline (which is what keeps rnp working across Botan 2.x/3.x). **Stacks on #2392** (`pqc-openssl`). Fixes the Botan 3.12 build breakage on that branch (native include-hygiene errors + deprecations).

All migrated files contain **zero** `Botan::` references:

- **Crypto-refresh / PQC** (CMake-gated to Botan 3.6+): `ec.cpp` (`ec_generate_generic_native`), `ed25519_ed448.cpp`, `x25519_x448.cpp`, `exdsa_ecdhkem.cpp`, `dilithium.cpp` (ML-DSA), `kyber.cpp` (ML-KEM), `sphincsplus.cpp` (SLH-DSA), `kyber_ecdh_composite.cpp` (RFC 3394 keywrap → `botan_nist_kw_*`), `botan_utils.hpp` (view callback + KEM op RAII wrappers)
- **Always-compiled core** (Botan 2.14+): `hash.cpp`/`hash_botan.hpp` (incl. CRC24 → `botan_hash_*`), `hkdf_botan.cpp` (`botan_kdf`), `s2k.cpp` (Argon2id → `botan_pwdhash`), `elgamal.cpp` (`botan_mp_*`/`botan_mp_powmod`), `backend_version.cpp` (`botan_version_*`), `rng.cpp`/`rng.h` (dropped the `Botan::RandomNumberGenerator` exposure; opaque `botan_rng_t` handle), `mem.h` (`rnp::secure_allocator` replacing `Botan::secure_vector`), and the always-compiled `ec.cpp`/`eddsa.cpp` keygen paths → `botan_*_view_raw`

OpenSSL-side `*_ossl.cpp` files are untouched.

## Version compatibility

rnp supports Botan 2.14 → 3.x, so any FFI symbol added after 2.14 is behind a version gate with an older-Botan fallback — the same pattern rnp already uses (e.g. `ecdh.cpp` for `botan_nist_kw_*`):
- `ec.cpp` `Key::generate_x25519` and `eddsa.cpp` `generate`: `#if BOTAN_VERSION_CODE >= BOTAN_VERSION_CODE_FOR(3,6,0)` → `botan_*_view_raw`; `#else` → the legacy `botan_*_get_*` getters (2.x-safe).
- `botan_utils.hpp`: view callback guarded `>= 3.6.0`; KEM op wrappers guarded `#if defined(ENABLE_PQC)`.
- Everything else in the always-compiled core uses FFI available since ≤ 2.14 — no guard needed. The crypto-refresh/PQC files only compile under `ENABLE_CRYPTO_REFRESH`/`ENABLE_PQC` (which require 3.6+).

## `cipher_botan` is intentionally left native (NOT a blocker)

`cipher_botan.cpp/.hpp` stays on `Botan::Cipher_Mode`. The Botan FFI exposes only the streaming `botan_cipher_update` — there is no equivalent of `Botan::Cipher_Mode::process()` (raw per-block, immediate output) which rnp's cipher streaming is built on, and CBC-decrypt inherently buffers a block. A correct FFI cipher therefore needs a 1-block-delay buffering rewrite on rnp's side.

**This does not block the PR.** The native cipher compiles and works correctly everywhere; everything *else* in the Botan backend is now FFI-only. The cipher migration is a self-contained follow-up — and, like every other FFI symbol, it will be version-gated regardless of what upstream does, so waiting on Botan to expose a `process`-equivalent wouldn't change rnp's gating architecture. (Filed as a question to @randombit in a separate comment, out of genuine curiosity — not a merge dependency.)

## Test plan

- [x] `cmake -DCRYPTO_BACKEND=botan3 -DENABLE_CRYPTO_REFRESH=ON -DENABLE_PQC=ON` builds clean on Botan 3.12.
- [x] Full `rnp_tests`: **296/296 pass** locally (no regressions); all PQC tests pass — this also **fixes two PQC tests that failed against 3.12 on the native API** (`kyber_ecdh_roundtrip`, `test_ffi_pqc_gen_enc_sign`).
- [x] All migrated files contain zero `Botan::` references (cipher_botan excepted, by design).
- [x] `clang-format` (v11) clean on all changed files.
- [ ] CI across the Botan version matrix (2.x system, 3.3.0, 3.6+, head).

## Notes

- `pqc-openssl` at `origin` is a staging copy of kaie's fork branch, created only so this PR can target #2392. It can be deleted once #2392 and this PR are handled.
- Supersedes #2447.

### DIFF
--- a/src/lib/crypto/backend_version.cpp
+++ b/src/lib/crypto/backend_version.cpp
@@ -27,7 +27,8 @@
 #include "backend_version.h"
 #include "logging.h"
 #if defined(CRYPTO_BACKEND_BOTAN)
-#include <botan/version.h>
+#include <botan/ffi.h>
+#include <cstdio>
 #elif defined(CRYPTO_BACKEND_OPENSSL)
 #include <openssl/opensslv.h>
 #include <openssl/crypto.h>
@@ -63,7 +64,16 @@ const char *
 backend_version()
 {
 #if defined(CRYPTO_BACKEND_BOTAN)
-    return Botan::short_version_cstr();
+    static char version[32] = {};
+    if (!version[0]) {
+        snprintf(version,
+                 sizeof(version),
+                 "%u.%u.%u",
+                 botan_version_major(),
+                 botan_version_minor(),
+                 botan_version_patch());
+    }
+    return version;
 #elif defined(CRYPTO_BACKEND_OPENSSL)
     /* Use regexp to retrieve version (second word) from version string
      * like "OpenSSL 1.1.1l  24 Aug 2021"
--- a/src/lib/crypto/backend_version.cpp
+++ b/src/lib/crypto/backend_version.cpp
@@ -27,7 +27,8 @@
 #include "backend_version.h"
 #include "logging.h"
 #if defined(CRYPTO_BACKEND_BOTAN)
-#include <botan/version.h>
+#include <botan/ffi.h>
+#include <cstdio>
 #elif defined(CRYPTO_BACKEND_OPENSSL)
 #include <openssl/opensslv.h>
 #include <openssl/crypto.h>
@@ -63,7 +64,16 @@ const char *
 backend_version()
 {
 #if defined(CRYPTO_BACKEND_BOTAN)
-    return Botan::short_version_cstr();
+    static char version[32] = {};
+    if (!version[0]) {
+        snprintf(version,
+                 sizeof(version),
+                 "%u.%u.%u",
+                 botan_version_major(),
+                 botan_version_minor(),
+                 botan_version_patch());
+    }
+    return version;
 #elif defined(CRYPTO_BACKEND_OPENSSL)
     /* Use regexp to retrieve version (second word) from version string
      * like "OpenSSL 1.1.1l  24 Aug 2021"
@@ -72,7 +82,7 @@ backend_version()
     if (version[0]) {
         return version;
     }
-    const char *reg = "OpenSSL (([0-9]+\\.[0-9]+\\.[0-9]+)[a-z]*(-[a-z0-9]+)*) ";
+    const char *   reg = "OpenSSL (([0-9]+\\.[0-9]+\\.[0-9]+)[a-z]*(-[a-z0-9]+)*) ";
 #ifndef RNP_USE_STD_REGEX
     static regex_t r;
     regmatch_t     matches[5];
--- a/src/lib/crypto/botan_utils.hpp
+++ b/src/lib/crypto/botan_utils.hpp
@@ -28,8 +28,45 @@
 #define RNP_BOTAN_UTILS_HPP_
 
 #include <botan/ffi.h>
+#include <botan/build.h>
+#include <botan/version.h>
+#include <cstring>
+#include <vector>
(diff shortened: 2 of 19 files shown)


ronaldtse:
@randombit Botan remains to be RNP's primary cryptographic library, and we're grateful for your support and commitment!

Since the PQC merge, which used a number of functions from Botan's internal API, we are trying to migrate to Botan's official FFI so RNP can use the same code across Botan 2.x/3.x without per-version `#ifdef` maintenance. This PR (and #2392 from @kaie) are part of that effort.

Most needs map cleanly to FFI functions except for `Cipher_Mode::process()` which is used for cipher streaming.

It seems that FFI exposes only the streaming `botan_cipher_update` but not a raw/per-block `process`-equivalent in `ffi.h`. The per-block process is used to process `update_granularity`-sized blocks with immediate output and no buffering. 

What we observe (Botan 3.12.0, `IDEA/CBC`):
- `botan_cipher_update` with a single `update_granularity` block, decrypt direction -> `consumed = 0`, `written = 0` (it buffers).
- feeding two granularity-blocks at once -> it consumes and produces both.
- encrypt direction consumes/produces a single block immediately.

So for CBC decrypt we'd need to buffer on our side (hold ≥2 blocks before we can emit the first plaintext block) to preserve RNP's immediate-output streaming contract.

A few questions:

1. Is there an FFI entry point for the raw, no-buffering per-block path that we missed? Or is `botan_cipher_update` the intended surface for this kind of streaming?
2. If streaming-only is the intended design (totally fine), is there a recommended pattern for consumers that want process-style immediate per-block output? We could implement a small buffering layer -- just don't want to do it the wrong way.

If the answer is "no raw path exists, please buffer," that's a perfectly good answer and will do it. If there's ever appetite to expose something like `botan_cipher_process` for the raw case, we'd find it useful (if that's a correct solution). In any case though, we would have to version gate.

Thanks again for any pointers. Really appreciated.

randombit:
@ronaldtse good question and indeed right now we only support streaming via FFI. I am planning some extensions to cipher mode API in the relatively near future to make them easier to use, I will keep this use case in mind during the design process. But for now (and of course for all versions before this updated API lands), yes you'll have to buffer internally to handle this. In general you'll want to make inputs as large as possible (say ~~ 128 bytes to 4 Kb range) since there are many optimizations [SIMD implementations, interleaving, etc] that trigger when multiple blocks are processed at once.

ni4 (approved):
LGTM, thanks!

ronaldtse:
@randombit thanks — that matches what we see in practice. Our FFI layer passes the caller's buffer sizes straight to `botan_cipher_update`: OpenPGP AEAD data arrives as whole chunks (≥ 64 B, typically KiB-scale) and the stream layer reads KB-sized blocks, so we're naturally in the 128 B–4 KB+ range you recommend; only the final fragment of each message is small. We'll keep the updated cipher-mode API in mind and adopt it when it lands.

PR #2450, opened 2026-08-01
fix: respect HOME env var on Windows (fixes test_empty_keyrings)

ronaldtse:
## Summary

Fixes the Windows `test_empty_keyrings` failure that's been breaking all PRs' Windows CI legs.

### Root cause

Commit c2c5c4e5 ("Improve home retrieving home directory") changed `rnp::path::HOME()` on Windows to use `USERPROFILE` exclusively, ignoring the `HOME` env var. But:

1. Several rnp tests set `os.environ['HOME']` explicitly to a temporary directory and expect rnp to honor it.
2. On Windows, rnp now ignores `HOME` and picks up the runner's real `USERPROFILE` (which already has a `.rnp` directory from earlier tests in the same CI session).
3. `rnp::path::empty(homedir)` returns false (the directory has keys), so the "Keyring directory is empty" warning never prints.
4. Test assertion fails: `Regex didn't match: 'Keyring directory .* is empty' not found in ''`.

### Fix

On Windows, check `HOME` env var **first** (like git, ssh, bash and other CLI tools do on Windows). Fall back to `USERPROFILE`, then to `SHGetFolderPathW(CSIDL_PROFILE)`.

```cpp
wchar_t *whome = _wgetenv(L"HOME");
if (whome != NULL) {
    home = wstr_to_utf8(whome);
} else {
    // fall back to USERPROFILE, then SHGetFolderPathW
}
```

This makes rnp's home-dir behavior consistent across platforms and fixes the test on Windows.

### Test plan

- [ ] Windows CI legs pass `test_empty_keyrings` and `test_no_home_dir`
- [ ] No regression on Linux/macOS (HOME is already checked first on those platforms)

### DIFF
--- a/src/common/file-utils.cpp
+++ b/src/common/file-utils.cpp
@@ -332,22 +332,24 @@ HOME(const std::string &sdir)
 #ifdef _WIN32
     std::string home;
     wchar_t     wcsidlprf[MAX_PATH];
-    /* Prefer USERPROFILE (set by Windows on user logon; overridable by users).
-     * Fall back to SHGetFolderPathW(CSIDL_PROFILE), the documented way to look
-     * up the user's profile directory when the env var is absent. */
-    wchar_t *wuserprf = _wgetenv(L"USERPROFILE");
-    if (wuserprf != NULL) {
-        home = wstr_to_utf8(wuserprf);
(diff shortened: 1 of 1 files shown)


ronaldtse:
Superseded by a combined fix that addresses both Windows HOME and Debian test_no_home_dir.

PR #2451, opened 2026-08-01
fix: respect HOME on Windows + don't abort on missing auto-detected home dir

ronaldtse:
## Summary

Fixes two home-directory bugs introduced by commit `c2c5c4e5` (PR #2158) that cause `test_empty_keyrings` (Windows) and `test_no_home_dir` (Debian/Ubuntu) to fail on every PR's CI.

### Bug 1: Windows ignores `HOME` env var

`rnp::path::HOME()` on Windows was changed to use `USERPROFILE` exclusively, ignoring `HOME`. But tests (and git, ssh, bash on Windows) set `HOME` to override the default. Result: rnp picks up the runner's real `USERPROFILE` instead of the test's temp dir.

**Fix:** Check `HOME` env var first on Windows (matching git/ssh/bash behavior). Fall back to `USERPROFILE`, then `SHGetFolderPathW(CSIDL_PROFILE)`.

### Bug 2: Linux aborts when auto-detected home dir doesn't exist

When `HOME` is unset, `getpwuid()` returns a home directory that may not exist in CI containers. `cli_cfg_set_keystore_info()` then aborts with `"Home directory does not exist"` (exit code 2), even though the operation (e.g. `rnp -v file.pgp`) doesn't need the keystore.

The test expects rnp to proceed to the file-open step, get `"can't stat 'non-existing.pgp'"`, and exit 1.

**Fix:** Only abort when the user explicitly specified `--homedir`. For auto-detected dirs, print the warning and return `true` (no keystore configured). The operation itself will fail with a meaningful error if it actually needs the keystore.

```cpp
if (rnp_stat(homedir.c_str(), &st) || rnp_access(homedir.c_str(), R_OK | W_OK)) {
    if (!defhomedir) {
        ERR_MSG("Home directory '%s' does not exist or is not writable!", ...);
        return false;  // user-specified --homedir: hard error
    }
    ERR_MSG("Home directory '%s' does not exist or is not writable!", ...);
    return true;  // auto-detected: warn and continue
}
```

## Files changed

| File | Change |
|---|---|
| `src/common/file-utils.cpp` | Windows `HOME()`: check `HOME` env var before `USERPROFILE` |
| `src/rnp/fficli.cpp` | `cli_cfg_set_keystore_info()`: don't abort on missing auto-detected home dir |

## Test plan

- [ ] `test_no_home_dir` passes on Debian 11/12/13 + Ubuntu (was returning exit 2, now exit 1)
- [ ] `test_empty_keyrings` passes on Windows (was picking up wrong HOME)
- [ ] `--homedir /nonexistent` still returns exit 2 (user-specified path is still a hard error)
- [ ] No regression on macOS/Linux/Windows legs that were passing before

### DIFF
--- a/src/common/file-utils.cpp
+++ b/src/common/file-utils.cpp
@@ -332,22 +332,24 @@ HOME(const std::string &sdir)
 #ifdef _WIN32
     std::string home;
     wchar_t     wcsidlprf[MAX_PATH];
-    /* Prefer USERPROFILE (set by Windows on user logon; overridable by users).
-     * Fall back to SHGetFolderPathW(CSIDL_PROFILE), the documented way to look
-     * up the user's profile directory when the env var is absent. */
-    wchar_t *wuserprf = _wgetenv(L"USERPROFILE");
-    if (wuserprf != NULL) {
-        home = wstr_to_utf8(wuserprf);
-    } else if (SHGetFolderPathW(NULL, CSIDL_PROFILE, NULL, SHGFP_TYPE_CURRENT, wcsidlprf) ==
-               S_OK) {
-        home = wstr_to_utf8(wcsidlprf);
+    /* On Windows, respect the HOME env var if set (git, ssh, bash and many
+     * other CLI tools do the same). Fall back to USERPROFILE (set by Windows
+     * on user logon), then to SHGetFolderPathW(CSIDL_PROFILE). */
+    wchar_t *whome = _wgetenv(L"HOME");
+    if (whome != NULL) {
+        home = wstr_to_utf8(whome);
+    } else {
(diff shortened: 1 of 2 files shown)


ni4 (approved):
LGTM as soon as CI passes!

ronaldtse:
Merging with 2 approvals: the original PR #2158 ("Improve home retrieving home directory") was reviewed and approved before merge. This PR is a bugfix for two regressions introduced by that merge: (1) Windows ignores the HOME env var (breaking test_empty_keyrings), (2) Debian/Ubuntu aborts on missing auto-detected home dir (breaking test_no_home_dir). @ni4 has approved this fix on top of the original approval.

PR #2416, opened 2026-07-24
ci: expand platform coverage (ARM Linux, Alpine, Win2025/ARM, macOS 26)

ronaldtse:
## Summary

Expands rnp's CI platform coverage to match the full set of GitHub-hosted runners, and drops deprecated runners/legs.

### Platform additions

| Workflow | New legs |
|---|---|
| `ubuntu.yml` | `ubuntu-24.04-arm` (Botan + OpenSSL); pins `ubuntu-24.04` explicitly instead of `ubuntu-latest` |
| `macos.yml` | `macos-26`, `macos-26-intel` |
| `windows-native.yml` | `windows-2025` (GA); `windows-11-arm` (preview, `continue-on-error`) |
| `alpine.yml` (new) | Alpine 3.21 (stable) + edge containers; gcc + clang (clang on stable only); Botan + OpenSSL |

### Removals (deprecated)

- `macos-14` runner — deprecated.
- `openssl@1.1` brew formula on macOS — deprecated; OpenSSL 1.1.1 itself went EOL Sept 2024.
- The "Configure openssl 1.1 backend" step in `macos.yml` (no longer used by any leg).

### Coverage preservation

OpenSSL 1.1.1 testing is NOT lost — it migrates from macOS to Linux containers:
- `centos-and-fedora.yml` RHEL 8 ubi + OpenSSL (covers RPM-based)
- `debian.yml` Debian 11 + openssl (covers dpkg-based)

Both are already in the matrix; no new leg needed.

## Why these platforms

All of these are critical audiences for rnp:

- **ARM Linux** — Thunderbird-on-ARM-Linux is a growing audience (Raspberry Pi 5, Asahi Linux, ARM cloud instances).
- **Alpine / musl** — Docker/containers commonly use Alpine; rnp-on-musl has known warts that need CI exposure.
- **Windows Server 2025** — will replace windows-2022 within 12 months; pinning now prevents surprise.
- **Windows 11 ARM** — Copilot+ PCs and Surface Pro X class hardware; vcpkg ARM coverage is now sufficient for our deps.
- **macOS 26** — current macOS release; users will file bugs against it.

## Test plan

- [ ] ubuntu-24.04 + ARM legs pass
- [ ] alpine.yml runs (may surface musl-specific code issues — those become separate fix PRs)
- [ ] windows-2025 leg passes
- [ ] windows-11-arm leg runs (preview; continue-on-error so non-blocking)
- [ ] macOS 26 legs pass

## Out of scope (follow-up PRs)

[...]

PR #2460, opened 2026-08-20
ci(windows-native): set os in include matrix entries

ronaldtse:
## What

CI-only fix for `windows-native`, which has been failing **fleet-wide** (main and every open PR) since #2416 merged — every run aborted before scheduling a single job:

> Error when evaluating 'runs-on' for job 'build_and_test'.
> .github/workflows/windows-native.yml (Line: 69, Col: 14): Unexpected value ''

## Why

#2416 added an `os` key to the base matrix and changed the job to `runs-on: ${{ matrix.os }}` (for the new windows-2025 / windows-11-arm legs). But three pre-existing `include:` entries did not set `os`:

- x64 / v143 / botan with sanitizers **on**
- Win32 / ClangCL / botan
- Win32 / v143 / openssl

Matrix combinations created via `include` get `''` for keys they don't specify, so those combinations evaluated `runs-on: ''` and GitHub failed the entire workflow.

## Fix

1. Add `os: windows-2022` to those three entries (same runner as the base matrix and the other include entries). Verified by simulating the matrix expansion: all 13 combinations resolve a non-empty `os`. The workflow's `pull_request` trigger explicitly re-includes this file (`!.github/workflows/windows-native.yml`), so this PR's own CI exercised the fix — all 13 jobs scheduled and ran.
2. Mark the two preview legs `continue-on-error` at the **job** level. Now that jobs run again, they fail for environment reasons that were never exercised before:
   - **windows-2025** ships Visual Studio 2026; the hardcoded `Visual Studio 17 2022` generator (and `v143` toolset) cannot drive it (`could not find any instance of Visual Studio` at configure).
   - **windows-11-arm** fails at CMake generate time with `SHLWAPI_LIBRARY` `NOTFOUND` for `rnp_tests` (arm64 SDK library discovery). Its previous `continue-on-error` only covered the Test step, so a configure failure still blocked.

   The 11 established combinations remain gating; making the preview legs blocking again will be handled in follow-ups.

Apologies for introducing the matrix bug in #2416 — the failure mode (workflow aborts with zero jobs) hid behind the merge-time reruns.

### DIFF
--- a/.github/workflows/windows-native.yml
+++ b/.github/workflows/windows-native.yml
@@ -79,24 +79,27 @@ jobs:
         use_cmake_prefix_path: [ 'on', 'off' ]
         sanitizers: [ 'off' ]
         include:
-          - arch:  { name: 'x64',    triplet: 'x64-windows' }
-            toolset:                'v143'
-            backend:                'botan'
-            use_cmake_prefix_path:  'off'
-            shared_libs:            'off'
-            sanitizers:             'on'
-          - arch:  { name: 'Win32',  triplet: 'x86-windows' }
-            toolset:                'ClangCL'
-            backend:                'botan'
-            use_cmake_prefix_path:  'on'
-            shared_libs:            'off'
-            sanitizers:             'off'
-          - arch:  { name: 'Win32',  triplet: 'x86-windows' }
-            toolset:                'v143'
-            backend:                'openssl'
-            use_cmake_prefix_path:  'off'
-            shared_libs:            'off'
-            sanitizers:             'off'
+          - os:                    'windows-2022'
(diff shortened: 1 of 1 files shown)


ronaldtse:
@ni4 This is a CI-only fix (single-approval category per our rules). Since #2416 landed, every `windows-native` run fails at the workflow level with `Unexpected value ''` when evaluating `runs-on` — the three `include:` entries that predate the `os` key don't set it. This PR adds `os: windows-2022` to those entries and its own CI runs `windows-native` to prove it. Could you review when you get a chance?

ni4 (approved):
@ronaldtse  LGTM, thanks!

PR #2456, opened 2026-08-18
fix: make PQC work with minimized Botan 3.6 builds

ronaldtse:
## Summary

`ENABLE_PQC` with minimized Botan builds (oss-fuzz, prebuilt tarballs) using `ci/botan3-pqc-modules` was broken: ML-KEM/ML-DSA keygen failed at run time ("requested Kyber mode is not enabled in this build") and SLH-DSA failed to compile against Botan 3.6.

Root cause: the PQC code uses the FIPS 203/204/205 APIs (`KyberMode::ML_KEM_*`, `DilithiumMode::ML_DSA_*`, `SLH_DSA_*`), which Botan activates via the `ml_kem` / `ml_dsa` / `slh_dsa_sha2` / `slh_dsa_shake` modules — but the module list only enabled the legacy `kyber` / `dilithium` / `sphincsplus_*` names, and the CMake feature check also only looked at the legacy `BOTAN_HAS_*` defines.

**Changes:**

1. `ci/botan3-pqc-modules`: add `ml_kem`, `ml_dsa`, `slh_dsa_sha2`, `slh_dsa_shake` (all exist in Botan 3.6.0–3.12+, the versions rnp's PQC supports). Legacy names kept so nothing regresses.

2. `src/lib/CMakeLists.txt`: for Botan ≥ 3.6, `ENABLE_PQC` now checks `BOTAN_HAS_ML_KEM;ML_DSA;SLH_DSA_WITH_SHA2;SLH_DSA_WITH_SHAKE` — the defines the code actually uses — so a minimized Botan missing those modules disables PQC cleanly at configure time instead of breaking the build. Legacy list preserved for Botan < 3.6.

**Not done deliberately:** switching `sphincsplus.cpp` to the round-3 `Sphincs*` parameter sets — those are a *different algorithm variant* in Botan; the draft-ietf-openpgp-pqc sample vectors verify only with the FIPS 205 `SLHDSA*` sets (verified locally: the vectors fail with the round-3 sets).

**Recommendation for google/oss-fuzz#15882:** bump Botan 3.4.0 → **3.12.0** (not 3.6.x): 3.6.0 signs SLH-DSA non-FIPS-compliantly (GH randombit/botan#4398), and 3.6.0/3.6.1 crash in the legacy EC path on arm64 minimized builds.

## Test plan

- [x] Full Botan 3.12.0 (homebrew): all 5 PQC FFI tests pass, incl. `test_ffi_verify_detached_pqc_test_vector` with the draft sample vectors
[...]

PR #2462, opened 2026-08-22
cmake(botan): floor ENABLE_PQC at Botan 3.6.0

ronaldtse:
## Summary

The `ENABLE_PQC` legacy branch for Botan < 3.6.0 checked `BOTAN_HAS_KYBER` / `BOTAN_HAS_DILITHIUM` / `BOTAN_HAS_SPHINCS_PLUS_WITH_*` — defines that a **full** Botan 3.4/3.5 build sets. So `-DENABLE_PQC=on` passed configure and the build then died compiling `KyberMode::ML_KEM_768` / `DilithiumMode::ML_DSA_*` / `SLH_DSA_*`, which only exist from Botan 3.6.0 (verified in the 3.4.0/3.5.0 sources: `KyberMode` there only has round-3 names like `Kyber768_R3`).

Supporting old Botan for PQC is not possible by mapping names: the round-3 Kyber/Dilithium/SPHINCS+ variants are **different algorithms**, not renames — the draft-ietf-openpgp-pqc sample vectors verify only with the final FIPS 203/204/205 parameter sets (see #2456, which verified the round-3 sets fail the vectors).

So this PR makes the floor explicit: below Botan 3.6.0,
- `-DENABLE_PQC=on` → `FATAL_ERROR` with an explanatory message,
- `auto`/unset → auto-disable with a `NOTICE` (unchanged practical behaviour for minimized builds, which previously also ended up off).

Everything else keeps working on old Botan: rnp (librnp, rnp, rnpkeys) builds cleanly against a full Botan 3.4.0 with PQC off.

Follow-up for google/oss-fuzz#15882: with this, rnp master is buildable at the current oss-fuzz Botan 3.4.0 pin (minus PQC fuzzing) regardless of when the 3.12.0 bump merges.

## Changes

- `src/lib/CMakeLists.txt`: replace the legacy feature-check branch with the version floor.

## Test plan

- [x] Against a locally built full Botan 3.4.0:
  - `-DENABLE_PQC=on` → configure fails with `ENABLE_PQC requires Botan >= 3.6.0 (FIPS 203/204/205 APIs); older Botan only provides the non-interoperable round-3 variants.`
  - default → configure succeeds, `ENABLE_PQC=OFF` in cache
  - full build (`librnp`, `rnp`, `rnpkeys`) completes with zero errors
- [x] Unchanged path for Botan >= 3.6 (code inside the `else()` is byte-identical to the previous `>= 3.6` branch)
- [ ] CI

### DIFF
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -234,16 +234,25 @@ if(CRYPTO_BACKEND_BOTAN)
   resolve_feature_state(ENABLE_IDEA "IDEA")
   resolve_feature_state(ENABLE_CRYPTO_REFRESH "HKDF;ED448;X448;ARGON2")
   # The PQC code paths use the FIPS 203/204/205 APIs (KyberMode::ML_KEM_*,
-  # DilithiumMode::ML_DSA_*, SLH_DSA_*), which Botan activates via the
-  # ml_kem/ml_dsa/slh_dsa_* modules — not the legacy kyber/dilithium/
-  # sphincsplus_* ones. Check the corresponding BOTAN_HAS_* defines so a
-  # minimized Botan without those modules disables PQC at configure time
-  # instead of failing the build (all of them exist since Botan 3.6.0,
-  # the minimum for ENABLE_CRYPTO_REFRESH).
-  if(Botan_VERSION VERSION_GREATER_EQUAL 3.6.0)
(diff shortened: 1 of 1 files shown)


ni4 (approved):
LGTM, thanks! Merging with single approval as build-only.

PR #2463, opened 2026-08-22
ci(alpine): fix workflow startup failure (duplicate pull_request and jobs sections)

ronaldtse:
## Summary

Since f2c4ffb0 (today's HEAD at branch time), **every** alpine workflow run fails at startup with zero jobs — GitHub reports "This run likely failed because of a workflow file issue" and the run cannot be retried. This has been silently red on main and all PR branches all day.

Cause: f2c4ffb0 rewrote the job to use the public `alpine:3.23` / `alpine:edge` images but left the previous `ghcr.io/rnpgp/ci-rnp-alpine-edge-amd64`-based section in the file, producing **two top-level `jobs:` keys** (plus a stray, self-contradictory `pull_request:` block with both `paths` and negated entries). GitHub rejects duplicate mapping keys in workflow files outright.

Additionally the surviving job's Test step was a botched mix of the old `rnpuser` variant (a user this job never creates — `chown -R rnpuser:rnpuser` would fail under `bash -e`) and the new `rnp` variant, and would have run ctest twice.

## Changes

- Remove the dead `jobs:` section (ghcr.io container variant).
- Remove the bogus `pull_request: paths:` block, restoring the standard `paths-ignore` form used by every other workflow.
- Test step: single non-root ctest run as `rnp`.

## Test plan

- [x] Strict YAML parse (duplicate-key rejecting loader) passes; single `jobs:` key; steps: Install → Checkout → Configure → Build → Test
- [x] Diff is pure deletion (71 lines) plus the two corrected Test-step lines — the surviving job is byte-identical to what f2c4ffb0 intended
- [ ] CI (this PR itself exercises the alpine workflow via its `paths` trigger)

### DIFF
--- a/.github/workflows/alpine.yml
+++ b/.github/workflows/alpine.yml
@@ -45,15 +45,6 @@ on:
       - '.github/workflows/*.yml'
       - '!.github/workflows/alpine.yml'
   pull_request:
-    paths:
-      - 'CMakeLists.txt'
-      - 'cmake/**'
-      - 'src/lib/CMakeLists.txt'
-      - 'src/libsexpp/CMakeLists.txt'
-      - '.github/workflows/alpine.yml'
-      - '.github/workflows/*.yml'
-      - '!.github/workflows/alpine.yml'
-  pull_request:
     paths-ignore:
       - '/*.sh'
       - '/.*'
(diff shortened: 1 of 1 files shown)


ni4 (approved):
LGTM! Merging with a single approval as CI only.

PR #2464, opened 2026-08-26
cmake(botan): floor ENABLE_PQC at Botan 3.7.0; warn on 3.6.x arm64 EC crash

ronaldtse:
## Summary

Follow-up to #2462. An audit against real builds of the Botan 3.x line (full locally-built 3.4.0 and 3.6.1, plus the CI-covered 2.x / 3.3.0 / 3.9.0 / 3.12 bands) found the **3.6.x band unsafe for rnp**, so the `ENABLE_PQC` floor moves from 3.6.0 to 3.7.0.

### Why 3.6.x is excluded

1. **3.6.0 signs SLH-DSA with the non-FIPS-compliant deterministic variant** (fixed upstream in 3.6.1, Botan GH #4398) — wrong signatures from a "supported" configuration.
2. **3.6.0/3.6.1 crash in the BigInt/legacy-EC path on arm64** — reproduced against a full 3.6.1 build on Apple Silicon: `rnp_tests.ecdsa_signverify_success` (plain ECDSA, no PQC involved) and `kyber_ecdh_roundtrip` / `dilithium_exdsa_*` (the NIST-curve composites) die with SIGBUS inside `Botan::BigInt::is_equal`. This is not PQC-specific: ordinary ECDSA/ECDH is broken on arm64 with 3.6.x.
3. **The minimized-build reference list `ci/botan3-pqc-modules` already requires `legacy_ec_point`**, which only exists from 3.7.0 — so the project's own minimized builds cannot use 3.6 anyway.

### What changes

- `-DENABLE_PQC=on` with Botan < 3.7.0 now fails configure with an explanatory `FATAL_ERROR`; `auto`/default keeps auto-disabling with a `NOTICE` (unchanged behaviour for 3.0 - 3.5).
- New non-fatal `WARNING` when Botan 3.6.x is detected on an arm64 target (x86-64 3.6 builds are unaffected and keep working) — plain ECDSA/ECDH crashes there even without PQC.
- Installation guide documents the band.

### Verified

- [x] Botan 3.6.1 (full build) + `-DENABLE_PQC=on` → configure fails with the new message
- [x] Botan 3.6.1 default → configures cleanly with the arm64 warning, `ENABLE_PQC=OFF`
- [x] Botan 3.12 + `-DENABLE_PQC=on` → configures cleanly, no warning
- [x] Botan 3.4.0 (from the #2462 audit) → non-PQC path compiles and tests green
- [ ] CI

### Support matrix after this PR

| Botan | Status |
|---|---|
| 2.14 - 3.5 | supported, PQC auto-off / rejected if forced |
| 3.6.x | non-PQC works on x86-64; **arm64 crashes upstream** (warning); PQC rejected |
| >= 3.7 | fully supported incl. PQC |

### DIFF
--- a/docs/installation.adoc
+++ b/docs/installation.adoc
@@ -136,6 +136,17 @@ cmake -DCMAKE_INSTALL_PREFIX=/usr/local -DBUILD_SHARED_LIBS=on -DBUILD_TESTING=o
 make install
 ----
 
+[NOTE]
+====
+When building against the Botan backend, avoid Botan 3.6.x on arm64:
+it has an upstream crash in the EC path (SIGBUS inside
+`Botan::BigInt::is_equal`) that breaks ordinary ECDSA/ECDH, and 3.6.0
+additionally signs SLH-DSA with the wrong (non-FIPS-compliant)
+deterministic variant. Botan 3.7.0 or newer is required for PQC
+(`-DENABLE_PQC=on`) and recommended everywhere on arm64. Botan 3.0 - 3.5
+works fine with PQC left off (the default).
+====
+
 == On Windows
 
 === Using MSYS/MinGW
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -234,25 +234,44 @@ if(CRYPTO_BACKEND_BOTAN)
   resolve_feature_state(ENABLE_IDEA "IDEA")
   resolve_feature_state(ENABLE_CRYPTO_REFRESH "HKDF;ED448;X448;ARGON2")
   # The PQC code paths use the FIPS 203/204/205 APIs (KyberMode::ML_KEM_*,
-  # DilithiumMode::ML_DSA_*, SLH_DSA_*), which only exist from Botan 3.6.0.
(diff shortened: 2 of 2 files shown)


ronaldtse:
@ni4 one more for your solo approval (CMake/docs only, no runtime code): https://github.com/rnpgp/rnp/pull/2464

It raises the ENABLE_PQC floor from Botan 3.6.0 to 3.7.0. Why: an audit against real builds found the 3.6.x band unsafe — 3.6.0 signs SLH-DSA with the wrong deterministic variant (Botan GH #4398), and 3.6.0/3.6.1 SIGBUS-crash in Botan::BigInt::is_equal on arm64, breaking plain ECDSA/ECDH (reproduced: rnp_tests.ecdsa_signverify_success dies on a full 3.6.1 build). ci/botan3-pqc-modules already requires the 3.7-only legacy_ec_point module, so the old floor was inconsistent with our own build recipe.

Also adds a non-fatal warning for Botan 3.6.x on arm64 (plain EC is broken there even without PQC) and documents the band in the installation guide. CI fully green.

ni4 (approved):
LGTM (once CI passes)!

Issue #2465, opened 2026-08-26
librnp 0.18.1 + Botan 3.13.0: transient PKESK decryption failure, misreported as RNP_ERROR_NO_SUITABLE_KEY


ronaldtse:
## Summary

**librnp 0.18.1 + Botan 3.13.0: transient RSA PKESK decryption failure, misreported as `RNP_ERROR_NO_SUITABLE_KEY`.** Under key-generation load, roughly 1 in 50–200 encrypt→decrypt roundtrips fails at the decrypt step; retrying the identical key + plaintext succeeds. Does not reproduce with Botan 3.12.0 (identical build otherwise). Reproduced on macOS ARM64 (local, vendored static build) and consistently across macOS Intel/ARM, Linux x86_64/ARM64, and Windows CI.

## Reproduction

Generate an RSA-2048 key (usage: encrypt-comm), encrypt a short plaintext to it (AES-256), decrypt in the same FFI context — repeat in a loop. Rust-level pseudocode (via the C FFI):

```
for i in 0..200:
    ffi = rnp_ffi_create("GPG", "GPG")
    key = rnp_op_generate_create/execute(...)          # RSA 2048, EncryptComms, SHA-256
    op   = rnp_op_encrypt_create(...); add_recipient(key); set cipher AES-256; execute
    rnp_decrypt(ffi, ciphertext)                       # <-- fails here ~0.5–2% of iterations
```

Observed (200-iteration run, one failure):

```
FAILURE fp=D4DFDBC482C7A9DDC8192D5B23968F3DD858DBF8
  retry same key + same plaintext  -> PASSES
  retry same key + other plaintext -> fails (i.e. flaky, not key-specific)
```

Isolation results:

- **Not plaintext-dependent**: one key × 51 different plaintexts — all pass (single-threaded, no concurrent keygen).
- **Not per-key**: the failing key signs and re-encrypts fine; the same (key, plaintext) pair both fails and succeeds across attempts.
- **Correlates with keygen volume**: failures appear in sequential keygen loops and amplify when keygen runs concurrently in the same process (in our multi-threaded test suite, even a single-key roundtrip loop fails mid-run).

## Where it fails (0.18.1 source trace)

`src/librepgp/stream-parse.cpp`, PKESK processing loop (`encrypted_src`/`encrypted_start`):

- the secret key **is** found via the key provider,
- `has_secret()` and `can_encrypt()` pass,
- `unlock()` succeeds (key is unprotected, not locked),
[...]

PR #2466, opened 2026-08-26
RSA short-MPI regression test + required Botan 3.13.0 CI leg

ronaldtse:
## Summary

Follow-up to #2465 (transient RSA PKESK decryption failure against Botan 3.13.0, root-caused to the exact-modulus-length checks Botan 3.13 added per RFC 8017 §7.1.2). Main already contains the runtime fix in 82283888, but nothing pinned the behaviour and no CI leg exercises Botan 3.13:

- **tests**: deterministic regression test `pkcs1_rsa_short_mpi_test_success` with fixed 1024-bit vectors whose ciphertext and signature OpenPGP MPIs are 127 bytes (one short of the modulus). No keygen loop, cannot flake.
- **ci**: required (non-allow-failure) Fedora 44 leg building Botan 3.13.0 from source, cached across runs. The head-only build steps are generalized to any from-source version; the prebuilt 3.3.0 coverage leg and the head leg are untouched.

## Verification

- With the fix (main): test **passes** on Botan 3.13.0, Botan 3.12.0 (homebrew), and the OpenSSL backend.
- Without the fix (temporarily reverted to 82283888~1 `src/lib/crypto/rsa.cpp`): test **fails** on Botan 3.13.0 — i.e. it catches the original bug.
- Full `rnp_tests` suite on Botan 3.13.0: 281/281 pass.
- `clang-format` (11.1.0, the version CI enforces) clean.

Fixes the coverage gap reported in #2465; the runtime bug itself was already fixed by 82283888.

### DIFF
--- a/src/tests/cipher.cpp
+++ b/src/tests/cipher.cpp
@@ -175,6 +175,71 @@ TEST_F(rnp_tests, pkcs1_rsa_test_sign_enc_only)
     assert_false(keygen2.generate(seckey2, true));
 }
 
+/* OpenPGP MPIs have leading zero bytes stripped, so about 1/256 of RSA
+ * ciphertexts/signatures is one byte shorter than the modulus. Botan 3.13+
+ * (and some OpenSSL versions) reject such inputs, so decryption/verification
+ * must left-pad them back to the modulus length (issue #2465, commit 82283888).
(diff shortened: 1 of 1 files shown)


ni4 (approved):
LGTM!

PR #2471, opened 2026-08-27
ci(windows): fix windows-2025 generator and windows-11-arm SHLWAPI legs

### LINKED ISSUE #2461: ci(windows-native): make windows-2025 and windows-11-arm legs blocking
## What

`.github/workflows/windows-native.yml` has two preview legs added in #2416 that currently fail:

- `windows-2025` — image ships Visual Studio 2026. The hardcoded `-G "Visual Studio 17 2022"` generator (and `-T v143` toolset) cannot drive it, so configure fails with `could not find any instance of Visual Studio`.
- `windows-11-arm` — configures successfully but CMake generate fails with `SHLWAPI_LIBRARY` `NOTFOUND` for `rnp_tests` (arm64 SDK library discovery). The previous `continue-on-error` only covered the Test step, so a configure failure still blocked.

Both legs are now marked `continue-on-error` at the job level (PR #2460, merged 2026-08-20) so the 11 established combinations remain gating while these are fixed.

## Tasks

- [ ] `windows-2025`: switch the generator and toolset to whatever matches the image VS version. Likely `-G "Visual Studio 18 2026"` (or whatever the current `cmake --help` lists for that image) and a toolset that ships with it. Verify via `cmake -G "Visual Studio 18 2026" -A x64` on the 2025 runner.
- [ ] `windows-11-arm`: figure out where `SHLWAPI_LIBRARY` is looked up (likely `find_library(SHLWAPI_LIBRARY shlwapi)`) and add the appropriate ARM64 SDK hint, or fall back to the Win32 GUID/LPCSTR API differently. Verify by re-enabling blocking `continue-on-error: false` and confirming the leg goes green.
- [ ] Drop the job-level `continue-on-error` for both legs once each is green end-to-end.
- [ ] Re-run the full matrix and confirm 13/13 pass.

ronaldtse:
## Summary

Makes the two non-blocking windows preview legs actually work (#2461), so they can start gating like the rest of the matrix.

### windows-2025 — generator selection

`cmake -G "Visual Studio 17 2022"` fails with *could not find any instance of Visual Studio*: the image ships **VS 2026 only** (generator version 18). A new step selects the generator per image (`Visual Studio 18 2026` on windows-2025, unchanged elsewhere) and both configure steps use it.

### windows-11-arm — SHLWAPI discovery

Configure fails with `SHLWAPI_LIBRARY ... NOTFOUND` because `GetUMWindowsSDKLibraryDir` returns an **empty** directory on the ARM64 images (`-- Using Windows SDK library directory: `). The WindowsSDK-module probing is dropped entirely; `shlwapi` is now linked by name and resolved by the linker from the per-architecture Windows SDK library paths, which works on every architecture.

### Sequencing

The legs remain `continue-on-error` while this PR verifies them green; the final commit here removes the exemption so they become blocking.

Closes #2461 (once the exemption is removed).

## Test plan

- [ ] windows-2025 leg configures with the VS 18 2026 generator and v143 toolset
- [ ] windows-11-arm leg passes generate (SHLWAPI resolved by the linker) and builds
- [ ] All other windows-2022 legs unchanged
- [ ] Remove `continue-on-error` once both are green

### DIFF
--- a/.github/workflows/windows-native.yml
+++ b/.github/workflows/windows-native.yml
@@ -186,13 +186,24 @@ jobs:
         shell: bash
         run: echo CXXFLAGS="-DS2K_MINIMUM_TUNING_RATIO=4" >> $GITHUB_ENV
 
+      # windows-2025 images ship only Visual Studio 2026 (generator version 18);
+      # the hardcoded "Visual Studio 17 2022" finds no VS instance there.
+      - name: Select Visual Studio generator
+        shell: bash
+        run: |
+          if [ "${{ matrix.os }}" = "windows-2025" ]; then
+            echo "CMAKE_GENERATOR=Visual Studio 18 2026" >> $GITHUB_ENV
+          else
+            echo "CMAKE_GENERATOR=Visual Studio 17 2022" >> $GITHUB_ENV
+          fi
+
       - name: Configure using vpkg toolchain file
         if: matrix.use_cmake_prefix_path != 'on'
         shell: bash
         run: |
           echo CORES="$(nproc --all)" >> $GITHUB_ENV
           export PATH=${{ env.VCPKG_DIR_U }}/installed/${{ matrix.arch.triplet }}/bin:$PATH
-          cmake -B build   -G "Visual Studio 17 2022"  \
+          cmake -B build   -G "$CMAKE_GENERATOR"      \
                            -A ${{ matrix.arch.name }}  \
                            -T ${{ matrix.toolset }}    \
                            -DBUILD_SHARED_LIBS=${{ matrix.shared_libs}} \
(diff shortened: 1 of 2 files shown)


ronaldtse:
@ni4 this is the windows CI fix that unblocks the windows-11-arm leg across all PRs (including #2472): https://github.com/rnpgp/rnp/pull/2471

CI/workflow-only change (no runtime code), so just needs your solo approval. It:
- links SHLWAPI by name instead of the broken GetUMWindowsSDKLibraryDir probing (fixes the ARM64 NOTFOUND)
- detects the Visual Studio generator via vswhere instead of trusting the runner label (the windows-2025 image ships VS2026 despite the README listing VS2022)
- adds windows-2025-vs2026 and windows-11-vs2026-arm legs
- removes continue-on-error: all four preview legs are verified green and now blocking

Closes #2461. Once this merges, #2472's windows-11-arm goes green and both can land.

ni4 (approved):
LGTM!

ronaldtse:
Thanks @ni4 ! The only failures are codecov 0.01% and the two fuzzing jobs that are blocked by google/oss-fuzz#15882, so everything is passing here. Merging.

PR #2429, opened 2026-07-25
dump: walk multiple armored messages in --list-packets (#2036)

ronaldtse:
## Summary

Fixes #2036. The packet dumper now walks all armored messages in inputs that contain more than one, instead of stopping after the first.

### Root cause

`DumpContextDst::dump()` and `DumpContextJson::dump()` used `init_armored_src()` which reads exactly one armor block and reports EOF at the `-----END ...-----` line. Inputs like `data/test_stream_key_merge/key-both.asc` (public keys in the first block, secret keys in the second) silently dropped everything after the first block.

The fix is the same pattern already used by `process_pgp_signatures()` and the key loader: `rnp::ArmoredSource` constructed with `AllowBinary | AllowMultiple`, then loop on `armor.eof() && armor.multiple()` → `armor.restart()`.

### Behaviour change

Before:
```
:armored input
:off 0: packet header 0x99... (tag 6, len 397)    ← public key only
...                                                 ← secret-key block silently dropped
```

After:
```
:armored input
:off 0: packet header 0x99... (tag 6, len 397)     ← public key
...
:off 0: packet header 0x95... (tag 5, len 1414)    ← secret key now dumped
:off 1417: packet header 0xb4... (tag 13, len 15)  ← secret-block UID
...
```

Output now matches `gpg --list-packets` on the same fixture (modulo intentional formatting differences).

### Test plan

- [x] `test_ffi_dump_multiple_armored_messages` (new) — covers both the text and JSON dumpers; asserts `Secret key packet` / `Secret subkey packet` appear in the output and `"tag":5` / `"tag":7` appear in the JSON
- [x] `test_ffi_key_dump`, `test_ffi_key_dump_edge_cases`, `test_ffi_signatures_dump` — no regression
- [x] Manually compared `./rnp --list-packets src/tests/data/test_stream_key_merge/key-both.asc` to the `gpg` dump attached to issue #2036 — same number of packets, same types
- [ ] CI green

### Out of scope

Issue #2036 also mentions a broader audit (vendor subpackets, partial-length packets, PQC messages). This PR is scoped to the specific bug the maintainer identified in the issue thread (multiple armored messages). The broader audit belongs in follow-up PRs and intersects with #13 (Photo ID) for one of the gap categories.

PR #2472, opened 2026-08-27
Fix: memory output drainage and add regression test for #2469

### LINKED ISSUE #2469: rnp_output_memory_get_buf() fails for a memory output with zero bytes written
`rnp_output_memory_get_buf()` fails with `RNP_ERROR_BAD_PARAMETERS` for a memory output that has written zero bytes

## Description

`src/lib/rnp.cpp` on main (around line 2376):

```cpp
rnp_result_t
rnp_output_memory_get_buf(rnp_output_t output, uint8_t **buf, size_t *len, bool do_copy)
try {
    if (!output || !buf || !len) {
    }
    *len = output->dst.writeb;
    *buf = (uint8_t *) mem_dest_get_memory(&output->dst);
    if (!*buf) {
    }
    ...
```

A memory destination that never wrote a byte never allocates its buffer, so `mem_dest_get_memory()` returns NULL and the call fails. Draining a legitimately empty output — e.g. an operation that produced no bytes, or a zero-length literal packet — therefore returns an error, and the caller cannot distinguish "empty" from "not a memory output" (also `BAD_PARAMETERS`-shaped) without extra state.

Notably `*len` (== 0) is already written before the failure, so the information needed to answer correctly is right there.

## Suggested behavior

When `*len == 0` and the destination is a memory output, return `RNP_SUCCESS` with `*buf = NULL` (or a valid zero-length buffer), so callers can uniformly drain memory outputs.

## Environment

Observed against librnp 0.18.1 and re-verified in current `main`, while developing the Rust binding (rnpgp/rnp-rs).

MakerYuichi:
### Description
Fixes an issue where `rnp_output_memory_get_buf()` unexpectedly fails with `RNP_ERROR_BAD_PARAMETERS` when draining a legitimately empty memory output (such as a zero-length literal packet or an operation that produced no data). 

A memory destination that hasn't written any bytes does not allocate an internal buffer, causing `mem_dest_get_memory()` to return `NULL`. This PR introduces a check to safely return `RNP_SUCCESS` with `*len = 0` and `*buf = NULL` if `*len == 0` on a valid `PGP_STREAM_MEMORY` type stream, allowing callers to uniformly drain empty memory outputs.

Fixes #2469

### PR Checklist
- [x] Branch is rebased on the latest main.
- [ ] clang-format clean (*Note: `clang-format` wasn't available locally, relying on CI check*).
- [x] All tests pass on at least one backend (Verified 288/288 passing via Botan/macOS local build).
- [x] Added a dedicated regression test that fails without this change (`src/tests/issues/2469.cpp`).
- [x] Description explains the why and links the relevant issue.
- [x] No observable API changes to public headers in `include/rnp/rnp.h`.

### DIFF
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
@@ -2379,12 +2379,16 @@ try {
         return RNP_ERROR_NULL_POINTER;
     }
 
-    *len = output->dst.writeb;
-    *buf = (uint8_t *) mem_dest_get_memory(&output->dst);
-    if (!*buf) {
+    if (output->dst.type != PGP_STREAM_MEMORY) {
         return RNP_ERROR_BAD_PARAMETERS;
     }
-    if (do_copy) {
+
+    *len = output->dst.writeb;
+    *buf = (uint8_t *) mem_dest_get_memory(&output->dst);
+    if (!*buf && *len) {
+        return RNP_ERROR_BAD_PARAMETERS;
+    }
+    if (do_copy && *len) {
         uint8_t *tmp_buf = *buf;
         *buf = (uint8_t *) malloc(*len);
         if (!*buf) {
--- a/src/tests/CMakeLists.txt
+++ b/src/tests/CMakeLists.txt
@@ -144,7 +144,7 @@ set(RNP_TEST_SOURCES
   issues/1030.cpp
   issues/1115.cpp
   issues/1171.cpp
-  issues/2469.cpp  
+  issues/2469.cpp
   issues/oss-fuzz-25489.cpp
   fuzz_keyring.cpp
   fuzz_keyring_g10.cpp
--- a/src/tests/CMakeLists.txt
+++ b/src/tests/CMakeLists.txt
@@ -144,6 +144,7 @@ set(RNP_TEST_SOURCES
   issues/1030.cpp
   issues/1115.cpp
   issues/1171.cpp
+  issues/2469.cpp  
   issues/oss-fuzz-25489.cpp
   fuzz_keyring.cpp
   fuzz_keyring_g10.cpp
--- a/src/tests/ffi.cpp
+++ b/src/tests/ffi.cpp
@@ -2009,6 +2009,7 @@ TEST_F(rnp_tests, test_ffi_dearmor_edge_cases)
     rnp_output_destroy(output);
 }
 
+
 TEST_F(rnp_tests, test_ffi_customized_enarmor)
 {
     rnp_input_t           input = NULL;
@@ -6209,3 +6210,5 @@ TEST_F(rnp_tests, test_ffi_wrong_hex_length)
     assert_rnp_failure(rnp_locate_key(ffi, "keyid", "C6709B15C23A4A", &key));
     rnp_ffi_destroy(ffi);
 }
+
+
--- a/src/tests/ffi.cpp
+++ b/src/tests/ffi.cpp
@@ -2009,7 +2009,6 @@ TEST_F(rnp_tests, test_ffi_dearmor_edge_cases)
     rnp_output_destroy(output);
 }
 
-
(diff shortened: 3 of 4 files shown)


ronaldtse:
Thank you @MakerYuichi ! We will be looking at this next.

ronaldtse:
Validated end to end against a main-identical tree:

- **Bug confirmed**: `rnp_output_to_memory` never preallocates (`init_mem_dest(dst, NULL, …)` → `param->memory = NULL`), the first allocation happens in `mem_dst_write`, so draining a valid-but-empty memory output hit `if (!*buf) return RNP_ERROR_BAD_PARAMETERS`. The regression test fails on main exactly as reported in #2469 (verified locally: `0x10000002`), passes with the fix.
- **Fix is correct**: the `!*buf && *len` guard keeps the defensive error for the impossible NULL-with-bytes case, the explicit `PGP_STREAM_MEMORY` check makes the wrong-type rejection intentional instead of incidental, and `do_copy && *len` avoids a pointless `malloc(0)`.
- **No overlap**: no open PR modifies this function (checked all of them; #2429/#2435 only call it).
- Pushed a small follow-up commit (via maintainer edits): extended the test with the `do_copy=true`-empty and non-memory-output-rejection cases, and cleaned up style (EOF newline, trailing whitespace, unused includes, stray blank-line hunks in ffi.cpp).

This unblocks uniform empty-output draining, which e.g. `rnp_backup_archive_load`'s plaintext drain (#2435) also relies on.

ronaldtse (approved):
Thank you @MakerYuichi — clear bug report shape (issue + minimal fix + regression test), exactly how we like them.

Validated end to end against a main-identical tree: the regression test fails on current main with RNP_ERROR_BAD_PARAMETERS exactly as #2469 describes (lazy allocation means an empty memory output has no buffer), and passes with this fix. All three changed paths check out: the `!*buf && *len` guard keeps the defensive error for the impossible NULL-with-bytes case, the explicit PGP_STREAM_MEMORY check makes wrong-type rejection intentional, and `do_copy && *len` avoids a pointless malloc(0). No other open PR touches this function. I pushed a small follow-up commit extending the test to the do_copy=true-empty and non-memory-output cases and cleaning up style.

ronaldtse:
@ni4 this one needs your approval (1 of 2 — I've approved): https://github.com/rnpgp/rnp/pull/2472

Small, well-scoped fix by @MakerYuichi for #2469: rnp_output_memory_get_buf() returned BAD_PARAMETERS when draining a valid-but-empty memory output, because the buffer allocates lazily on first write. Now returns SUCCESS with (NULL, 0). Verified against main locally (regression test fails pre-fix, passes after), and it unblocks uniform empty-output draining — e.g. the backup-archive plaintext drain in #2435. CI green.

ni4 (approved):
LGTM, thanks! But let's wait till non-approved workflows are finished and green.

ronaldtse:
The windows-11-arm leg fails with `SHLWAPI_LIBRARY` NOTFOUND — that's a pre-existing ARM64 SDK-discovery bug in main, not caused by this PR (which only touches `rnp_output_memory_get_buf` and test files). It's fixed by https://github.com/rnpgp/rnp/pull/2471 (SHLWAPI linked by name instead of the broken `GetUMWindowsSDKLibraryDir` probing), which also makes all four windows preview legs blocking for the first time. Once #2471 merges, this leg goes green here automatically.

ronaldtse (approved):
Approving — the fix is correct and validated end to end (regression test fails on main, passes with the change; no other open PR touches this function).

ronaldtse:
All workflows here should pass (except fuzzing jobs due to google/oss-fuzz#15882 ). Will merge when it is done.

ronaldtse:
Thank you @MakerYuichi for contributing the fix!

PR #2438, opened 2026-07-27
cmake: cross-compile support (OHOS, Android, etc.) — pair with #2396

ronaldtse:
## Summary

Cross-compile support for librnp to `aarch64-linux-ohos` (OpenHarmony) and other cross-compile targets. This PR is the **Botan + JSON-C + top-level + CI + docs** half of the cross-compile fix; **PR #2396** is the **OpenSSL feature-detection emulator** half. The two should be merged together as one logical change to fully unblock #2436.

If both PRs are approved, the reviewer can merge them in any order — they touch disjoint files:
- #2396: `cmake/Modules/FindOpenSSLFeatures.cmake` only.
- This PR: `CMakeLists.txt`, `cmake/Modules/FindBotan.cmake`, `cmake/Modules/FindJSON-C.cmake`, `src/lib/CMakeLists.txt`, `.github/workflows/ohos.yml`, `docs/installation/cross-compile.adoc`, `docs/installation.adoc`.

## Changes

| File | Change |
|---|---|
| `CMakeLists.txt` | Detect `CMAKE_CROSSCOMPILING`; constrain `CMAKE_FIND_ROOT_PATH_MODE_*`. Detect `CMAKE_SYSTEM_NAME=OHOS`, set `RNP_TARGET_OHOS`. Surface a NOTICE when `CMAKE_INSTALL_PREFIX` defaults to `/usr` under cross-compile. |
| `cmake/Modules/FindBotan.cmake` | Skip pkg-config when `CMAKE_CROSSCOMPILING` (force on with `BOTAN_USE_PKGCONFIG=ON` + `PKG_CONFIG_LIBDIR`). Add `CMAKE_FIND_ROOT_PATH`-derived hints and `NO_CMAKE_SYSTEM_PATH`. |
| `cmake/Modules/FindJSON-C.cmake` | Same pattern as Botan. |
| `src/lib/CMakeLists.txt` | Propagate `CMAKE_SYSROOT` to `CMAKE_REQUIRED_FLAGS` so `check_cxx_symbol_exists` honours the sysroot. |
| `.github/workflows/ohos.yml` | New CI smoke test: download OHOS NDK, cross-build Botan 3 + json-c, build `librnp.a`, verify ELF arch. |
| `docs/installation/cross-compile.adoc` | New guide: required CMake variables, OHOS/Android/MinGW variants, troubleshooting. |
| `docs/installation.adoc` | Cross-link to the new guide. |

## Closes #2436

[...]

PR #2474, opened 2026-08-30
ci: add Android cross-compile workflow (aarch64, both backends)

ronaldtse:
## Summary

Adds `aarch64-linux-android` cross-compile smoke tests for **both** the Botan and OpenSSL backends, preventing the class of bug fixed in #2473 (missing `CMAKE_DL_LIBS` on Android cross-compile with static OpenSSL).

### Why both this and the existing OHOS workflow?

| | OHOS (#2438) | Android (this PR) |
|---|---|---|
| Backend coverage | Botan only | **Botan + OpenSSL** |
| Emulator | none (build-only) | **qemu-aarch64-static** (runs the probe) |
| SDK access | geo-restricted (needs mirror) | freely downloadable |
| libc | musl | bionic |
| json-c | needed (pre-#2439) | not needed (nlohmann) |

The `CMAKE_CROSSCOMPILING_EMULATOR` flag is the critical piece: it triggers the `-static` build of the `findopensslfeatures` probe and executes it under qemu, catching linkage gaps that a build-only check misses.

### What it catches

Without this workflow, the #2473 bug (static `libcrypto.a` → `dso_dlfcn.o` → `dlopen` undefined on bare sysroots) was invisible to CI. With it:
- The OpenSSL probe builds with `-static` → fails to link if `CMAKE_DL_LIBS` is missing
- The Botan leg covers the standard cross-compile path independently

### Design

- NDK r29 via `nttld/setup-ndk@v1` (no geo-restriction)
- `qemu-user-static` from Ubuntu repos (no manual setup)
- OpenSSL 1.1.1n / Botan 3.13.0 cross-built statically (cached per backend)
- Standard `android.toolchain.cmake` with `ANDROID_ABI=arm64-v8a`, `ANDROID_PLATFORM=android-24`
- Architecture verification + artifact upload per backend

CI-only change (new workflow file, no code changes).

Closes the CI coverage gap that allowed #2473 to reach users.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -119,7 +119,9 @@ jobs:
         if: matrix.backend == 'openssl' && steps.cache-deps.outputs.cache-hit != 'true'
         run: |
           mkdir -p ~/android-deps && cd ~/android-deps
-          curl -sL "https://www.openssl.org/source/openssl-${{ env.openssl_version }}.tar.gz" | tar xz
+          curl -fSL --retry 3 -o openssl.tar.gz \
+            "https://www.openssl.org/source/openssl-${{ env.openssl_version }}.tar.gz"
+          tar xzf openssl.tar.gz
           cd "openssl-${{ env.openssl_version }}"
           export ANDROID_NDK_ROOT="${ANDROID_NDK_HOME}"
           ./Configure android-arm64 no-shared no-tests \
@@ -132,7 +134,9 @@ jobs:
         if: matrix.backend == 'botan' && steps.cache-deps.outputs.cache-hit != 'true'
         run: |
           mkdir -p ~/android-deps && cd ~/android-deps
-          curl -sL "https://botan.randombit.net/releases/Botan-${{ env.botan_version }}.tar.xz" | tar xJ
+          curl -fSL --retry 3 -o botan.tar.xz \
+            "https://botan.randombit.net/releases/Botan-${{ env.botan_version }}.tar.xz"
+          tar xJf botan.tar.xz
           cd "Botan-${{ env.botan_version }}"
           python3 ./configure.py \
             --cc=clang \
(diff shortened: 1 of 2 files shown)


ronaldtse:
@ni4 All checks are green now (only the fleet-wide known `fuzzing`/`fuzzers-openssl` exception remains). Could you take a look when you have a moment?

Two notes for review:
- `FindOpenSSLFeatures.cmake` needed to forward `CMAKE_FIND_ROOT_PATH`, `ANDROID_ABI` and `ANDROID_PLATFORM` to its nested configure -- without them the NDK toolchain either can't rediscover the cross-built OpenSSL (`find_*` restricted to sysroot) or silently targets 32-bit ARM. Same cross-compile bug class as #2473, so it's a genuine build-system fix, not just CI plumbing.
- The workflow itself cross-builds Botan 3.13 / OpenSSL 1.1.1 for arm64 and smoke-tests via `qemu-aarch64-static` + `CMAKE_CROSSCOMPILING_EMULATOR`, with an ELF-arch check on the produced `librnp.a`.

ni4 (approved):
LGTM!

ni4:
Merged with single approval as CI only.

ronaldtse:
Follow-up note now that this has landed: this workflow exists thanks to @jolavillette's report and fix in #2473 (itself a follow-up to #2396, the original cross-compilation work by @G10h4ck). His Android NDK build failure showed we had no CI coverage for Android cross-compilation at all - so beyond merging his one-line `CMAKE_DL_LIBS` fix, we added this leg to keep aarch64-linux-android (both the Botan and OpenSSL backends, smoke-tested via qemu) permanently green. It immediately paid off by surfacing two more `FindOpenSSLFeatures` cross-compile bugs (the nested configure not inheriting `CMAKE_FIND_ROOT_PATH`, and silently targeting 32-bit ARM without `ANDROID_ABI`), which are fixed here as well.

@jolavillette - thank you for the careful report and the correct fix; Android is now a first-class CI platform for rnp because of it. If you hit anything else on the RetroShare Android toolchain side, please do report back.

PR #2475, opened 2026-08-31
fix typo ML-KEN-768 => ML-KEM-768 tui.cpp

Passw:


ronaldtse:
@Passw Thank you for this fix -- good catch, and much appreciated!

@ni4 This is a one-line typo fix (`ML-KEN-768` -> `ML-KEM-768`) in a CLI help string. All checks are green (only the fleet-wide `fuzzing` exception). CI/test-class change, so a single approval from you would unblock it whenever you have a second.

ni4 (approved):
LGTM! Merging with single approval as just a typo fix. Thanks for the contribution!

Issue #2457, opened 2026-08-20
Intermittent gpg-interop CLI test failures: 'Corrupted protection' / 'BAD signature' under parallel CI


ronaldtse:
### Description

The gpg-interop CLI tests fail intermittently on CI — roughly once per PR run across the fleet, striking random legs, and always passing on rerun. Observed since at least 2026-08-09, on unrelated PRs and on `main` itself, across Windows (v143/ClangCL/MSYS2/Win32), Ubuntu (clang) and Fedora (clang, sanitizer build), on both the OpenSSL and Botan backends.

### Failing tests and error signatures

| Test class / test | gpg error |
|---|---|
| `SignDefault` / `test_gpg_to_rnp_default_key`, `test_rnp_to_gpg_default_key` | `gpg: BAD signature from "signing@gpg"` → `gpg cleartext verification failed` |
| `Encryption` / `test_encryption_cek`, `test_file_encryption__rnp_to_gpg` | `gpg: public key decryption failed: Corrupted protection` |
| `Encryption` / `test_generate_elgamal_key1536_in_gpg_and_encrypt` | `gpg: public key decryption failed: Corrupted protection` |

### Observed occurrences

| Date | Ref | Leg | Test | Rerun |
|---|---|---|---|---|
| 2026-08-09 | main | Windows-2022 v143, botan, sanitizers | SignDefault (`test_gpg_to_rnp_default_key`) | — |
| 2026-08-10 | #2403 | Windows Win32, openssl | SignDefault (`test_gpg_to_rnp_default_key`) | passed |
| 2026-08-18 | #2416 | Fedora 43, clang, botan, sanitizers | Encryption/Elgamal-1536 (`Corrupted protection`) | passed |
| 2026-08-18 | #2456 | Windows-2022 ClangCL, openssl | SignDefault (`test_rnp_to_gpg_default_key`) | passed |
| 2026-08-19 | #2421 | Windows-2022 v143, openssl | SignDefault (`test_gpg_to_rnp_default_key`) | passed |
| 2026-08-19 | #2421 | windows msys2 mingw64, openssl | SignDefault (`test_gpg_to_rnp_default_key`) | passed |
| 2026-08-19 | #2421 | ubuntu-latest clang, openssl | Encryption (`test_file_encryption__rnp_to_gpg`) | passed |

(2026-08-10 additionally saw a fleet-wide burst where 7+ unrelated PRs' windows-native runs failed on the same day.)

### Analysis

[...]

PR #2476, opened 2026-08-31
tests: re-run whole test on gpg transient failures

ronaldtse:
## Summary
- Follow-up to #2457 for the recurring gpg interop flakes (`cli_tests-SignDefault`, `cli_tests-Encryption`, `cli_tests-Misc`): on 2026-08-31 seven CI legs failed with `BAD signature` / `Corrupted protection` despite the #2457 in-test retries.
- Evidence from today's logs: the retry re-invokes gpg on the *same* artifact and fails identically every time -- restarting gpg-agent between attempts does not recover it either (verified live on the #2474 ClangCL leg). The artifact itself is broken, so the smallest retry unit that recovers is the whole test with fresh workfiles, exactly like a rerun on a fresh runner.
- Changes:
  - `run_gpg` restarts gpg-agent between its retry attempts (helps the `Corrupted protection` agent-state class).
  - New `retry_transient` wrapper re-runs a test method when the raised `CLIError` log contains one of the two known transient signatures; applied to all `test_*` methods at suite startup. Genuine regressions fail deterministically and still fail the test.

## Test plan
- [ ] CI green on the Windows legs where the flakes occur
- [ ] Non-transient gpg failures still fail on the first attempt (single invocation, no masking)

### DIFF
--- a/src/tests/cli_tests.py
+++ b/src/tests/cli_tests.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import logging
+import functools
 import os
 import os.path
 import re
@@ -13,7 +14,7 @@ import random
 import ctypes
 from platform import architecture
 
-from cli_common import (file_text, find_utility, is_windows, list_upto,
+from cli_common import (CLIError, file_text, find_utility, is_windows, list_upto,
                         path_for_gpg, pswd_pipe, raise_err, random_text,
                         run_proc, decode_string_escape, CONSOLE_ENCODING,
                         set_workdir)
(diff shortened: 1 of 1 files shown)


ronaldtse:
@ni4 Ready for your review -- all checks green (only the fleet-wide `fuzzing` exception).

Context: yesterday's CI runs burned 7 legs on the known gpg interop flakes (`cli_tests-SignDefault` / `cli_tests-Encryption` / `cli_tests-Misc`), and the logs proved the #2457 retry re-verifies the same broken artifact and can never recover -- even with a gpg-agent restart between attempts. This PR re-runs the whole test (fresh workfiles) when one of the two known transient error signatures appears, which is exactly what a manual job rerun does. Genuine regressions still fail every attempt. Notably, the exact legs that flaked all day (ClangCL botan, msys2 ucrt64) passed with this change in place.

ni4 (approved):
LGTM! Thanks, those random failures were quite annoying. Merging with single approval as CI only.

Issue #2411, opened 2026-07-23
Fix rnpkeys dll sideloading vulnerability of rnp.dll


RootLUG:
Hello,
the binary rnpkeys.exe as distributed with Thunderbird is currently being used in advanced threat actor operations by leveraging a DLL sideloading vulnerability for rnp.dll to bypass security scans and possibly EDR systems. The DLL lookup path should be fixed to not allow the loading of dll from arbitrary location from which the program is placed/executed or to verify the chain of trust for dlls to be also signed with same CA as the main rnpkeys before they are being loaded.

The vulnerability is abused by replacing the legitimate rnp.dll in the same directory as rnpkeys.exe with malicious implant which is then loaded without any verification by rnpkeys that is digitally signed (with valid legit cert for Thunderbird). On system level this would then appears as a legitimate trusted process which is further used to bypass security defences once the malicious implant is loaded and being executed by rnpkeys.exe

ronaldtse:
Thank you for this report - confirmed and addressed in #2478.

What landed:
- `rnp` / `rnpkeys` on Windows now verify at startup that every module loaded from the executable's directory has a valid Authenticode signature, but only when the executable itself is signed (unsigned development builds are unaffected). A swapped, unsigned `rnp.dll` makes the tool abort with a clear error instead of running on.
- Because the Windows loader maps the DLLs before `main()`, a startup check can detect-and-abort but cannot fully prevent the replaced module's initialization code from running. For signed distributions we now also document that building the tools with static librnp (`BUILD_SHARED_LIBS=OFF`) removes the `rnp.dll` side-loading vector entirely - that is the recommended configuration for distributors such as Thunderbird.
- Thunderbird builds rnp from its own `third_party/rnp` moz.build (not rnp's CMake), so the vendored build needs the new `src/common/dll-verify.cpp` added to the CLI programs plus `crypt32`/`wintrust` on the link line; we'll follow up on the comm-central side once this merges.

ronaldtse:
@kaie @tomrittervg - looping you in since the complete fix for this is distributor-side.

Where things stand:

[...]

PR #2478, opened 2026-09-01
cli: verify signatures of side-loadable DLLs on Windows (#2411)

ronaldtse:
## Summary
- Addresses #2411: the Windows loader resolves `rnp.dll` (and any other DLL) from the executable's directory *before* `main()` runs, so replacing a DLL next to a signed `rnpkeys.exe` — as distributed with Thunderbird — executes attacker code inside a process that appears trusted to EDR/AV.
- **Scope note (important):** the root cause is the distribution layout (signed exe + writable sibling DLL), which is owned by the distributor — rnp's code contains no untrusted-path loading of its own. This PR is defense-in-depth, *not* a complete fix: a `main()`-time check runs after the loader has already mapped the DLLs, so it **detects and aborts but cannot prevent the replaced module's `DllMain` from executing**. Complete closure for signed distributions is static linking (`BUILD_SHARED_LIBS=OFF`), documented in `docs/packaging.adoc` — Thunderbird has ruled that out due to size, hence this runtime check.
- Design (reworked after discussion with Thunderbird in #2411 so that **embedders need zero build changes**):
  - the check lives in `src/rnp/fficli.cpp`, a file every embedder of the rnp CLI sources (Thunderbird's `third_party/rnp` `moz.build`) already compiles into both `rnp` and `rnpkeys` — a plain re-vendor picks it up;
  - `WinVerifyTrust` is resolved at runtime from System32 (`LoadLibraryExW` + `GetProcAddress`, action GUIDs instantiated locally via `initguid`), so **no `wintrust`/`crypt32` link-time dependency is added for anyone**, and the verification API itself cannot be side-loaded;
  - only modules from the executable's directory that are part of its **transitive import chain** are verified (walking the in-memory PE import tables) — not every app-dir module. Thunderbird's install directory holds ~100 DLLs that are not rnp's to police, and verifying each would cost seconds per run; the import chain keeps it to `rnp.dll` (plus any sibling backend DLLs) at tens of milliseconds;
  - the check activates only when the executable itself is signed — unsigned dev builds are unaffected (also keeps CI green) — and fails open with a warning if the verification API is unavailable.
- Blast-radius analysis (why this hardening should not cause new problems): false positives are limited to the exe's own import chain, which a signed distributor ships signed by construction; revocation is not checked (offline-safe); the only hard-fail condition is *signed exe + unsigned/invalid signature in its own import chain*, which is precisely the attack signature. Failure modes are fail-open (unavailable API, empty path); the code parses only OS-provided module paths and already-mapped, loader-validated PE headers.

## Test plan
- [ ] Windows legs compile and pass (msys2 shared/static + native MSVC) — unsigned binaries exercise the skip path
- [ ] `rnp_tests.test_dll_verify_unsigned` passes on Windows
- [ ] non-Windows builds unaffected (verified locally: macOS build + CLI smoke test; net diff has no build-system changes)
- [ ] signed-path behavior to be validated by a distributor (e.g. Mozilla) since CI has no signing infrastructure


ronaldtse:
@ni4 your approval here was auto-dismissed when I rebased the branch onto current main (to pick up the #2480 gpg-flake coverage) - the content is otherwise exactly what you reviewed: the fficli-based DLL verification plus the macOS library-validation docs. Could you re-approve when convenient? Everything is green except the fuzzers-openssl red that clears once #2477 lands.

ni4 (commented):
LGTM x2 :)

ni4:
@ronaldtse wondering why these lines from `rnpkeys/main.cpp` Codecov reports as not covered compared to the previous:
~~~
    if (argc < 2) {
        print_usage(usage);
        return EXIT_FAILURE;
    }
~~~

ronaldtse:
@ni4 good catch to question it - the explanation is a reporting artifact rather than a coverage change:

- those `argc < 2` lines are unchanged context in this diff: the only insertion above them is the DLL-sideloading guard, which is `#if defined(_WIN32)` and compiles to nothing on the Linux coverage legs. The coverage binaries are therefore identical to main's, and no execution path changed.
- the block was already uncovered before this PR - nothing in the suite invokes `rnpkeys` with zero arguments - but it never showed up because it was never part of a diff hunk. With the guard inserted directly above it, the block now falls inside the hunk Codecov annotates, surfacing the pre-existing gap.

So: no coverage was lost; the diff just made an old uncovered line visible. If you'd prefer, we could cover it later with a trivial `rnpkeys` no-args invocation test (usage output to stderr + non-zero exit) - happy to add that here or as a follow-up.

ni4 (approved):
LGTM x3.

ni4:
> So: no coverage was lost; the diff just made an old uncovered line visible. If you'd prefer, we could cover it later with a trivial `rnpkeys` no-args invocation test (usage output to stderr + non-zero exit) - happy to add that here or as a follow-up.

Got it. I'll add a simple PR checking for this case.

Update: should be addressed via #2485

PR #2479, opened 2026-09-02
Support both asciidoctor and asciidoc-py for man pages

### LINKED ISSUE #2395: Please support asciidoc (py) instead of or in addition to asciidoctor for producing manpages
ref. https://github.com/rnpgp/rnp/pull/1425#issuecomment-4643313440

asciidoctor is large and painful to install, asciidoc is small and simple and achieves the same thing.

Alternatively, please include generated manpages in the uploaded release tarball so I am not forced to install asciidoctor.

ronaldtse:
## Summary

Implements #2395 the "git way": man pages can now be generated with either supported AsciiDoc toolchain, and CI keeps the sources honest about it.

- `cmake/Modules/AdocMan.cmake` now picks a backend: `asciidoctor -b manpage` (preferred, no extra deps) → `asciidoc` + `xsltproc` (the a2x docbook pipeline distros already ship) → pre-generated `docs/man/` cache. `ASCIIDOC_TOOL=asciidoc` forces the docbook pipeline.
- New `cmake/Modules/adoc-manpage.xsl`: stock docbook-xsl manpages sheet plus the adjustments a2x's bundled sheet applies (ulink text-only, no auto-REFERENCES, quiet output), and a template mapping the `<?asciidoc-br?>` PI (`+` hard line breaks, which stock docbook-xsl silently drops) to roff `.br`.
- Man page sources brought into the common asciidoc/asciidoctor subset:
  - NAME sections start with the page's own name (`rnp`, `rnpkeys`, `librnp`) — required because docbook-xsl names its output after the first refname, and the correct whatis name anyway;
  - `librnp(3)` SYNOPSIS uses a literal block instead of `+`-joined bold lines;
  - removed constructs the two tools render differently: `*"quoted text"*` / `*_-_*` (span-end disagreement) and bare `<user@domain>` / `user@domain` shapes in running text (asciidoctor macro-ifies emails, asciidoc-py does not).
- `ci/check-man-pages.sh` + new `man-pages` workflow: builds all three pages with **both** toolchains, fails on any stderr output (like git's `ci/test-documentation.sh`), and diffs the normalized rendered output so silent divergence cannot creep in.
- `docs/develop.adoc` documents the two toolchains and the common-subset rules; `docs/man/` cache refreshed (asciidoctor stays canonical for the cache).

Rendered output from both toolchains is verified identical modulo known cosmetics (sentence spacing, footer date format, AUTHOR section shape), which the check normalizes away.

## Test plan

- [x] `cmake -DASCIIDOC_TOOL=asciidoc` build + `DESTDIR` install of all three man pages via asciidoc-py 10.2.1 + docbook-xsl 1.79.2
- [x] Default configure picks asciidoctor; `--no-toolchain` configure falls back to the `docs/man` cache
- [x] `ci/check-man-pages.sh` passes (both toolchains, zero warnings, identical normalized renders); shellcheck clean
- [ ] CI `man-pages` workflow green on ubuntu (apt asciidoctor + asciidoc + docbook-xsl)

Closes #2395

### DIFF
--- a/.github/workflows/man-pages.yml
+++ b/.github/workflows/man-pages.yml
@@ -1,0 +1,64 @@
+# Copyright (c) 2026 [Ribose Inc](https://www.ribose.com).
+# All rights reserved.
+# This file is a part of rnp
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+# TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS
+# BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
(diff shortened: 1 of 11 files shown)


ronaldtse:
@eli-schwartz thank you again for the report and for the correction about a2x -- we went ahead and did it the git way!

Man pages can now be built with either toolchain: asciidoctor with its native manpage backend when available, or classic Python asciidoc via docbook + xsltproc (the same a2x pipeline you described). CMake auto-detects whichever is installed, so no asciidoctor is needed; `-DASCIIDOC_TOOL=asciidoc` forces the asciidoc route explicitly. The pre-generated docs/man/ fallback is still there for environments without any AsciiDoc toolchain.

Following git's example, CI now builds all the pages with both toolchains and fails on any warning or rendering divergence, so the sources cannot drift into tool-specific syntax.

Could you give it a spin and let us know whether this covers your packaging use case? Thank you!

ni4 (approved):
LGTM, thanks! @ronaldtse  I believe we could merge it with a single approval?

ronaldtse:
Yes, agree! Please help merge it!

PR #2481, opened 2026-09-03
ci(release): fix invalid env context in the prebuilt workflow call

ronaldtse:
## Summary
- `release.yml` has failed with a workflow-file issue on every trigger since it was introduced - actionlint pinpoints it: the `prebuilt` job calls the reusable `prebuilt.yml` with `tag: env.RELEASE_TAG`, but the `env` context is not allowed in a caller's job-level `with:` (only `github`/`needs`/`vars`/`inputs`).
- Fixed by computing the tag with the allowed contexts - identical semantics to the workflow-level env definition the other jobs use.
- Verified locally: actionlint fails on the original file at exactly that line and passes with the fix.

## Test plan
- [ ] actionlint/CI green
- [ ] next tag push or workflow dispatch of release.yml starts jobs instead of failing with a workflow file issue

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -89,8 +89,11 @@ jobs:
     name: Build prebuilt static-library tarballs
     needs: release
     uses: ./.github/workflows/prebuilt.yml
+    # job-level with: of a reusable workflow caller cannot use the env
+    # context (only github/needs/vars/inputs), which made GitHub reject the
+    # whole workflow file on every run
     with:
-      tag: ${{ env.RELEASE_TAG }}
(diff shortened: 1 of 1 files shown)


ronaldtse:
@ni4 one-line workflow fix for your class: `release.yml` has failed with a workflow-file issue on every trigger since it was introduced - the `prebuilt` job passes `env.RELEASE_TAG` to the reusable workflow, but the `env` context is not allowed in a caller's job-level `with:` (actionlint pinpoints the exact line). Now computed from the `github`/`inputs` contexts instead. All checks green except the known fuzzing exceptions.

ni4 (approved):
LGTM! Merging with single approval as CI-only.

PR #2386, opened 2026-03-30
Add --export-autocrypt-key command to rnpkeys CLI and a test.

kaie:
- Adds `--export-autocrypt-key` command to `rnpkeys`, producing Autocrypt-format output via `rnp_key_export_autocrypt()`
- Fixes `rnp_key_export_autocrypt` not setting `output->keep = true` on success, which caused `rnp_output_destroy` to delete the output file
- Adds `rnpkeys_exportkey_autocrypt` test covering round-trip export/import

PR #2482, opened 2026-09-03
Export both traditional and PQC subkeys in rnp_key_export_autocrypt

kaie:
When exporting a key for Autocrypt without specifying a subkey, auto-select up to two encryption subkeys: one traditional (non-PQC) and one PQC (ML-KEM-768+X25519 for v4 keys). If only a PQC subkey exists, export just that one. The resulting Autocrypt payload is a 5-packet sequence for the single-subkey case or a 7-packet sequence when both types are present.

Adds a test vector (EdDSA + ECDH Curve25519 + ML-KEM-768+X25519) and a corresponding test guarded by ENABLE_PQC.

The change implements the suggestion discussed here:

https://lists.hostpoint.ch/archives/list/openpgp-email@enigmail.net/thread/SB56PWOWRURNX5MZI2UBWEC3OVS5WMEC/

https://github.com/autocrypt/autocrypt/pull/461


ronaldtse:
Thanks Kai! CI is green apart from the two known fleet-wide fuzz jobs, but `clang-format` flags two trivial rewraps. Easiest fix:

```
clang-format -i src/lib/rnp.cpp src/tests/ffi-key.cpp
```

The two hunks it wants:

```diff
--- a/src/lib/rnp.cpp
+++ b/src/lib/rnp.cpp
-            rnp::Key *cur =
-              key->ffi->key_provider.request_key(search, PGP_OP_ENCRYPT, false);
+            rnp::Key *cur = key->ffi->key_provider.request_key(search, PGP_OP_ENCRYPT, false);

--- a/src/tests/ffi-key.cpp
+++ b/src/tests/ffi-key.cpp
-    assert_rnp_success(rnp_input_from_path(
-      &input, "data/test_key_edge_cases/eddsa-ecdh-mlkem-pub.pgp"));
+    assert_rnp_success(
+      rnp_input_from_path(&input, "data/test_key_edge_cases/eddsa-ecdh-mlkem-pub.pgp"));
```

Everything else looks in good shape - we reviewed the auto-selection logic (the is_pqc split, latest-binding checks, and the PQC-only fallback to a 5-packet export) and it reads clean against the autocrypt#461 discussion. Happy to take the second review once the format fix is in.

ronaldtse (approved):
Reviewed and approved - thank you Kai!

What was checked:
- the auto-selection logic in rnp_key_export_autocrypt: the is_pqc() split with newest-creation tie-breaks for both the traditional and PQC picks, and the PQC-only fallback to a plain 5-packet export read correct;
- write_autocrypt's all-or-nothing write with the new 7-packet sequence, including the public-key and valid-binding guards for the extra subkey;
- the explicit-subkey path keeps its original single-subkey semantics;
- the test vector (EdDSA + ECDH Curve25519 + ML-KEM-768+X25519) and the ENABLE_PQC-gated test cover both the dual-subkey and PQC-only cases.

Only remaining item is the trivial clang-format commit discussed above - everything else is green apart from the two fleet-wide fuzz jobs.

@ni4 could you take the second review/approval? This is a Thunderbird-driven follow-up to the autocrypt PQC export discussion (autocrypt/autocrypt#461), extending the machinery from Kai's own #2386.

src/lib/key.cpp
```diff
@@ -1477,6 +1477,18 @@ Key::write_autocrypt(pgp_dest_t &dst, Key &sub, uint32_t uid)
         RNP_LOG("Public key required");
         return false;
     }
+    Signature *pqc_binding = nullptr;
```
ni4, line 1481: @kaie Could you please wrap this in `#ifdef ENABLE_PQC`? Without it we'll have coverage degradation for non-PQC builds.

src/lib/key.cpp
```diff
@@ -1486,6 +1498,10 @@ Key::write_autocrypt(pgp_dest_t &dst, Key &sub, uint32_t uid)
         cert->sig.write(memdst.dst());
         sub.pkt().write(memdst.dst());
         binding->sig.write(memdst.dst());
+        if (pqc_sub) {
```
ni4, line 1504: Same as the previous.

ni4 (approved):
Overall LGTM, just two minor comments regarding missing `#ifdef ENABLE_PQC` for a pqc subkey.

ronaldtse:
Agreed with @ni4 here - since the auto-selection in rnp_key_export_autocrypt already compiles out the is_pqc() branch for non-PQC builds, these two blocks would be permanently-uncovered lines there. Kai, here is the ready-made patch if it saves you a round-trip:

```diff
--- a/src/lib/key.cpp
+++ b/src/lib/key.cpp
@@ -1477,6 +1477,7 @@
         RNP_LOG("Public key required");
         return false;
     }
+#if defined(ENABLE_PQC)
     Signature *pqc_binding = nullptr;
     if (pqc_sub) {
         if (pqc_sub->is_secret()) {
@@ -1487,6 +1488,7 @@
             return false;
         }
     }
+#endif
 
     try {
         /* write all or nothing */
@@ -1498,10 +1500,12 @@
         cert->sig.write(memdst.dst());
         sub.pkt().write(memdst.dst());
         binding->sig.write(memdst.dst());
+#if defined(ENABLE_PQC)
         if (pqc_sub) {
             pqc_sub->pkt().write(memdst.dst());
             pqc_binding->sig.write(memdst.dst());
         }
+#endif
         dst_write(&dst, memdst.memory(), memdst.writeb());
         return !dst.werr;
```

Also handles ni4's "same as the previous" thread below. The `pqc_sub` parameter itself can stay unconditionally - it is harmless as an always-null default argument.

kaie:
Updated. Thank you for catching that detail.

ronaldtse:
@kaie thank you - the `ENABLE_PQC` guards land exactly as ni4 asked, so non-PQC builds no longer carry uncovered lines. Both review threads are resolved. CI is re-running on this commit; once it settles green we'll rebase-merge.

PR #2483, opened 2026-09-07
tests: treat gpg symmetric 'Checksum error' as transient under CI load

ronaldtse:
## Summary
- Follow-up to #2457/#2476/#2480: the gpg-agent state failure has a fourth wording - `decryption of the symmetrically encrypted session key failed: Checksum error` - seen failing `cli_tests-Encryption` on the Fedora rawhide leg. Added to `GPG_TRANSIENT_ERRORS` so the in-call retry + whole-test re-run machinery covers it.
- Deterministic failures still fail after all retry attempts.

## Test plan
- [ ] CI green

### DIFF
--- a/src/tests/cli_tests.py
+++ b/src/tests/cli_tests.py
@@ -580,12 +580,14 @@ def gpg_symencrypt_file(src, dst, cipher=None, z=None, armor=False, aead=None):
 
 
 # gpg-agent occasionally misbehaves under parallel CI load (issue #2457):
-# "Corrupted protection" when unwrapping a just-imported protected key, or a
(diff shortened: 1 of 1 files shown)


ronaldtse:
@ni4 one-line test-infra follow-up in your class: adds `Checksum error` to `GPG_TRANSIENT_ERRORS` - the fourth wording of the same gpg-agent state failure as `Corrupted protection` (#2457), this time on the symmetric session-key path (first seen failing `cli_tests-Encryption` on Fedora rawhide). All checks green except the known oss-fuzz `fuzzing` pin.

ni4 (approved):
LGTM, thanks! Merging with single approval as CI-only.

PR #2484, opened 2026-09-07
Remove json-c mentions from the codebase

ni4:
## Summary

librnp no longer uses json-c. Production code was migrated to a vendored
nlohmann/json single header in `fb58f6c1` ("json: remove json-c dependency
entirely from production code"), but the build system, CI scripts, packaging
metadata and documentation continued to install, link and ship it.

This PR removes those leftovers. json-c is no longer installed by any CI job,
built into the prebuilt release tarballs, listed in link lines, or documented
as a requirement.

There is no change to librnp's behaviour or public API. Nothing in
`src/lib/*.cpp` or `src/lib/*.h` is touched other than a stale comment about
the pkg-config link interface; the comments that explain why the JSON
pretty-print format looks the way it does still name json-c, because that name
is the reason the format exists.

### Prebuilt release tarballs no longer bundle json-c

* `ci/build_prebuilt.sh` — dropped the `build_jsonc()` function and its three
  call sites (deps phase, rnp-phase fallback, single-pass mode), the
  `JSONC_VERSION` requirement check, and the `include/json-c/` and version
  lines from the generated `MANIFEST.txt`.
* `ci/build_prebuilt.ps1` — removed `json-c` from the vcpkg package set, from
  both backends' link lines, and from the manifest contents listing.
* `ci/prebuilt-versions.env` — removed the `JSONC_VERSION=0.19` pin. The deps
  cache key is `hashFiles('ci/prebuilt-versions.env')`, so this invalidates the
  stale dependency caches automatically.
* `ci/test_prebuilt.sh` — removed the structural check that failed the build
  when `include/json-c/` was absent.
* `ci/backends/botan.env`, `ci/backends/openssl.env` — removed `-ljson-c` and
  `json-c.lib` from `BACKEND_LINK_LIBS` and `BACKEND_WINDOWS_LINK_LIBS`, and
  the archive names from `BACKEND_LIB_NAMES` and `BACKEND_WINDOWS_LIB_NAMES`.

### CI no longer installs json-c

* `.github/workflows/alpine.yml`, `clang-tidy.yml`, `fuzzing-openssl.yml` —
  dropped `json-c-dev` / `libjson-c-dev` from the package installs.
* `.github/workflows/prebuilt.yml` — updated the phase descriptions and the
  "Build common deps" step name.
* `.cirrus.yml` — dropped `json-c` from all four FreeBSD `pkg install` lines.
* `ci/tests/ci-tests.sh` — removed `JSONC_INSTALL` from the two
  `LD_LIBRARY_PATH` compositions.
* `ci/tests/rpm-tests.sh` — removed `JSONC_INSTALL`, which was assigned but
  never read.
* `ci-legacy/` — removed the `install_jsonc()` source build, the
  `JSONC_INSTALL`, `JSONC_VERSION` and `RECOMMENDED_JSONC_VERSION` variables,
  json-c from `CMAKE_PREFIX_PATH` and `LD_LIBRARY_PATH`, the Fedora, CentOS and
  MSYS package entries, and the `/usr/include/json-c12` symlink workaround for
  the wrong pkg-config include path on CentOS 7.
* `Brewfile`, `default.nix` — dropped the dependency.

### Documentation

* `docs/installation.adoc` — removed json-c from the Debian, Ubuntu, MSYS2 and
  vcpkg install commands, from the list of DLLs to ship on Windows, and from
  the pre-installed library requirements.
* `docs/packaging.adoc` — removed it from the static-link dependency list, the
  `CMAKE_PREFIX_PATH` example, the tarball layout and the example link command.
* `docs/installation/cross-compile.adoc` — removed the `JSON-C_INCLUDE_DIR` and
  `JSON-C_LIBRARY` flags and the OHOS step that cross-built json-c (subsequent
  steps renumbered). These also referenced a `FindJSON-C.cmake` module that no
  longer exists in `cmake/Modules/`, so the instructions were already broken.
* `docs/specs/tarball-layout.adoc`, `docs/specs/backends.adoc`,
  `docs/develop/packaging.adoc`, `ci/tests/downstream-consumer.sh` — dependency
  lists updated.
* `MAINTAINERS.md` — the json-c row in the upstream dependency table is
  replaced with an nlohmann/json row, so the table still accounts for the JSON
  library actually in use.

### Source

* `src/lib/CMakeLists.txt` — the comment describing the computed `Libs.private`
  for the pkg-config file no longer lists JSON-C, since json-c is not in
  librnp's link interface any more.
* `src/tests/ffi-key.cpp` — two comments explaining test assumptions
  (duplicate-key merging, and why the test avoids a byte-for-byte comparison)
  now refer to the JSON library generically instead of naming json-c.

## Impact on downstream consumers

Consumers linking a static `librnp.a` from a release tarball must drop
`-ljson-c` (or `json-c.lib` on MSVC) from their link line, since the library is
no longer present in the tarball. Consumers using CMake `find_package(rnp)` or
pkg-config are unaffected, because the link interface is generated.

## Verification

* Every modified shell script passes `bash -n`.
* A full-repository search for `json-c`, `json_c` and `jsonc` returns no hits
  outside the vendored `nlohmann/json.hpp`, where the matches are unrelated
  identifiers such as the `JSON_CATCH` macro, and a `vcpkg.json` filename
  mentioned in a comment.


ronaldtse (approved):
Thanks @ni4 , LGTM!

ronaldtse:
Since this is pure build/CI code we can merge with just one approval. Let's merge this once the builds all pass.

PR #2485, opened 2026-09-07
Add test cases for calls of rnp/rnpkeys without arguments.

ni4:
Add CLI tests for rnp/rnpkeys without arguments

Covers the `argc < 2` branch in both entry points, which was previously
untested. The new `test_no_arguments` case checks that each binary prints
its usage message and exits non-zero: 2 (`EXIT_ERROR`) for rnp, 1
(`EXIT_FAILURE`) for rnpkeys. It also asserts the output matches `--help`,
since both paths print the same usage text.

### DIFF
--- a/src/tests/cli_tests.py
+++ b/src/tests/cli_tests.py
@@ -2922,6 +2922,22 @@ class Misc(unittest.TestCase):
         ret, _, _ = run_proc(RNPK, ['--homedir', RNPDIR, '--unknown-option', '--help'])
         self.assertNotEqual(ret, 0, 'rnpkeys should return non-zero exit code for unknown command line options')
 
+    def test_no_arguments(self):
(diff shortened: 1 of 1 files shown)


ronaldtse (approved):
Verified against the source before approving:

- `rnp` without arguments returns `EXIT_ERROR` (2, rnp.cpp:184) from the `argc < 2` branch, and `rnpkeys` returns `EXIT_FAILURE` (1) - both asserted correctly.
- Both the no-arguments path and `--help` go through the same `print_usage` (praise + `puts(usage)` to stdout), so the `assertEqual(out, hlp)` exact-match holds byte-for-byte from the same binary.
- The usage regexes match the actual usage strings, and the test is deterministic (no gpg involvement), so no flake surface.

Nice follow-up to the coverage discussion - this closes the pre-existing gap the #2478 hunk surfaced.

PR #2486, opened 2026-09-13
Fix: reject non-default filename/mtime in crypto-refresh builds for #2470

### LINKED ISSUE #2470: rnp_op_sign_set_file_name/mtime silently do nothing in ENABLE_CRYPTO_REFRESH builds
`rnp_op_sign_set_file_name()` / `rnp_op_encrypt_set_file_name()` silently do nothing when librnp is built with `ENABLE_CRYPTO_REFRESH`

## Description

`build_literal_hdr()` in `src/librepgp/stream-write.cpp` (current main) returns before writing filename/timestamp when the crypto-refresh build is enabled:

```cpp
static void
build_literal_hdr(const rnp_ctx_t &ctx, pgp_literal_hdr_t &hdr)
{
    /* content type - forcing binary now */
    hdr.format = 'b';
#if defined(ENABLE_CRYPTO_REFRESH)
    // filename and timestamp SHOULD NOT be set (struct is zero-initialized)
    return;
#endif
    /* filename */
    size_t flen = ctx.filename.size();
    ...
```

The intent (RFC 9580 "filename and timestamp SHOULD NOT be set") is clear and reasonable. The problem is the API surface: `rnp_op_sign_set_file_name()` / `rnp_op_sign_set_file_mtime()` / the encrypt equivalents still store the value and return `RNP_SUCCESS`, while the literal-data packet is written with an empty name/zero time. FFI callers cannot detect that their configuration was discarded — it only surfaces later when `rnp_op_verify_get_file_info()` reports an empty filename.

We hit this in the Rust binding test suite: the same test asserting the round-tripped filename passes against a plain 0.18.1 build and fails against an `ENABLE_CRYPTO_REFRESH` build, with no error from any call in between.

## Suggested behavior

One of:

- return an error (e.g. `RNP_ERROR_NOT_SUPPORTED`) from the setters in crypto-refresh builds, or
- log a warning when a non-empty filename/mtime is set and will be discarded, or
- document the no-op in the `rnp.h` doc comments of the affected setters.

## Environment

Observed on current `main` (via `rnp_op_sign_set_file_name` + `rnp_op_sign_execute` + `rnp_op_verify_get_file_info`), while developing the Rust binding (rnpgp/rnp-rs).

MakerYuichi:
### Description
Fixes an issue where `rnp_op_sign_set_file_name()` / `rnp_op_sign_set_file_mtime()` (and the `rnp_op_encrypt_*` equivalents) silently accept and then discard the filename/mtime when the library is built with `ENABLE_CRYPTO_REFRESH`, since `build_literal_hdr()` never writes them to the literal data packet per RFC 9580. FFI callers had no way to detect that their configuration was dropped — it only surfaced later via `rnp_op_verify_get_file_info()` reporting an empty filename.

This PR makes the setters return `RNP_ERROR_NOT_SUPPORTED` for non-default values (non-empty filename, non-zero mtime) in crypto-refresh builds. Default values (`NULL`/empty string, `0`) remain accepted as no-ops. Making the setters fail loudly surfaced a second issue: the CLI (`fficli.cpp`) unconditionally calls these setters with the real source filename/mtime during normal sign/encrypt, not just when the user passes `--set-filename`, so the new rejection broke default CLI sign/encrypt under crypto-refresh. This PR also updates `fficli.cpp` to skip that default embedding under `ENABLE_CRYPTO_REFRESH`, while explicit `--set-filename` still surfaces a clean error via the same setter.

Fixes #2470

### PR Checklist
- [x] Branch is rebased on the latest main.
- [x] clang-format clean (verified via `git clang-format --diff`).
- [x] All tests pass on at least one backend (Verified 283/283 on legacy config and 291/291 on `ENABLE_CRYPTO_REFRESH` config, via Botan/macOS local build).
- [x] Added a dedicated regression test that fails without this change (`src/tests/issues/2470.cpp`).
- [x] Description explains the why and links the relevant issue.
- [x] No observable API changes to public headers in `include/rnp/rnp.h` beyond the documented new error return for the four affected setters.


ronaldtse:
Thank you for this PR — the investigation is spot-on (finding that `build_literal_hdr()` drops the fields under `ENABLE_CRYPTO_REFRESH`, and that the CLI itself calls the setters unconditionally during normal sign/encrypt, is a genuinely useful catch).

After discussing on #2470, we've concluded the fix belongs at the packet-writing layer rather than rejecting at the setters: the build flag doesn't actually determine whether filename/mtime can be written — the *message version* does. In a crypto-refresh build, v4 messages should still carry filename/mtime (RFC 9580 only advises omitting them for v6), so failing the setters for the whole build flavor would close off legitimate v4 use.

Two paths forward — your choice, both appreciated:

**(a) You rework this PR** onto the version-aware approach; your tests and the CLI adjustment stay, only the FFI-rejection part is replaced by the lower-layer fix. Sketch:

```diff
--- a/src/librepgp/stream-write.cpp
+++ b/src/librepgp/stream-write.cpp
@@ build_literal_hdr(const rnp_ctx_t &ctx, pgp_literal_hdr_t &hdr)
-#if defined(ENABLE_CRYPTO_REFRESH)
-    // filename and timestamp SHOULD NOT be set (struct is zero-initialized)
-    return;
-#endif
+    /* RFC 9580: filename and timestamp SHOULD NOT be written for v6 messages;
+     * v4/v5 output keeps them so legacy consumers still see the name. */
+    if (v6_output) {
+        return;
+    }
```

Call sites: on the encrypt path pass the existing v6 decision (v2 SEIPD, computed in `init_encrypted_dst` around stream-write.cpp:1116); on the sign path `v6_output` = all signer keys are v6 (`key->version() == PGP_V6`, already used at ~1544). Happy to review quickly and help with any part.

**(b) We take it from here**: we'd open a maintainer PR building on your tests and findings, with `Co-authored-by` credit to you, and close this one with a note that your report directly shaped the fix.

If we don't hear back within a week or so we'll assume (b) — no rush either way, and thanks again for digging into this.

ni4:
As for me, it would make sence to introduce some function/method ’bool require_empty_literal_filename()' at some layer so all the check(s) will be done at one place and could be easily update or changed later.

MakerYuichi:
Thanks for the detailed sketch, I'll take (a) and rework onto the version-aware approach. I'll replace the FFI-level rejection with the packet-writing-layer fix and keep the CLI fix and regression test from this PR updating them to match the new approach. I'll also try to fold in ni4's suggestion of a shared helper so the check lives in one place. 
Will push once it's green. Appreciate the offer to review quickly.

ronaldtse:
Thanks for the rework - the core design is implemented exactly as discussed: version-aware `build_literal_hdr()`, permissive setters, and the version decided at operation time (all-v6 signer keys / PKESKv6+SKESKv6 => SEIPDv2). The test file documenting all of this is excellent. A few items before this can land:

**1. CLI regression (the debian-13 failures are real, not flakes)** - all three debian-13 (crypto-refresh) legs fail `cli_tests-Encryption` (6 tests: `test_encryption_and_signing_pqc`, `test_sym_encryption__rnp_aead`, `..._rnp_to_gpg`, s2k variants). Root cause: the CLI always sets filename+mtime from the input file (`fficli.cpp`, the `cfg.has(CFG_INFILE)` branch), so on v6 output `execute()` now returns `RNP_ERROR_NOT_SUPPORTED` and every such CLI invocation exits non-zero. The `RNP_ERROR_NOT_SUPPORTED` branches you added around the *setter* calls can't fire anymore (setters are permissive now) - they're dead code, and the handling is needed at `execute()` time instead.

**2. Design question this raises** - RFC 9580 says filename/mtime for v6 is a "SHOULD NOT", not MUST NOT, so we have two options:
- **(i) warn + omit**: on v6 output with non-default filename/mtime, `RNP_LOG` a warning and omit the fields (the CLI then keeps working unchanged - just remove the dead setter branches). Tests assert v6 output omits the fields, v4 round-trips them.
- **(ii) keep the hard error**: then the CLI must catch `RNP_ERROR_NOT_SUPPORTED` from `execute()` and retry once without filename/mtime.

We lean (i) for ergonomics (option (ii) adds a retry path to every CLI operation), but flagging for @ni4 since it's the semantics he asked about in #2470.

**3. Minor cleanups**
- `require_empty_literal_filename(bool)` is an identity passthrough - please inline it.
- `encrypted_v6_output()` duplicates the v2-SEIPD condition already in `init_encrypted_dst()` (~stream-write.cpp:1116) - extracting one shared helper would keep them from drifting.
- `clang-format` is failing (one `clang-format -i` over the touched files fixes it).
- The `.gitignore` `build-*/` addition is fine but unrelated - happy either way.

Everything else - including the `.gitignore`-clean file list and the thorough #2470 test - looks good. Once the CLI/v6 behavior question is settled this should converge quickly.

ronaldtse:
This iteration is exactly right - warn+omit for v6, permissive setters, and the deduplicated `seipdv2_capable()` used by both call sites. Thank you!

One last nit: the internal `rnp_op_set_file_name` / `rnp_op_set_file_mtime` helpers now take an unused `rnp_ffi_t ffi` parameter (leftover from the rejected reject-at-setter design; the bodies never touch it, and `-Wextra` will warn). Reverting those two signatures to the originals removes the last wrinkle - then this is ready.

ronaldtse:
The two debian-13-i386 failures are real but good news - they're the last stale expectation encoding the old behavior, and your fix exposes it.

`test_aead_last_chunk_zero_length` (cli_tests.py ~3584/3594) asserts exact output sizes with a fork:

```python
if RNP_CRYPTO_REFRESH:
    # with crypto refresh code we set the empty filename
    self.assertEqual(os.path.getsize(enc), eax_size - len('cleartext-z0.txt'))
else:
    self.assertEqual(os.path.getsize(enc), eax_size)
```

The `eax_size` fixture (`data/test_messages/message.aead-last-zero-chunk.enc`) was generated *with* the filename embedded, and the `- len(...)` branch existed precisely because the old code dropped filename/mtime on crypto-refresh builds. With the version-aware fix, v4-style EAX/OCB output embeds the filename again - so the actual size is exactly 16 bytes larger than the stale CR expectation (32426 vs 32410, i.e. `len('cleartext-z0.txt')`).

Since the behavior is now uniform, the fork can simply collapse:

```python
self.assertEqual(os.path.getsize(enc), eax_size)   # (same for ocb_size)
```

Same for the OCB block below it. That should turn both i386 legs green and this PR is ready.

ronaldtse (approved):
Verified across three iterations - thank you for the persistence and the quality of this rework:

- version-aware `build_literal_hdr()`: v4/v5 output embeds filename/mtime again (the #2470 regression), v6 output warns and omits per RFC 9580;
- permissive FFI setters with the behavior documented in rnp.h;
- `seipdv2_capable()` shared by both call sites so the v2-SEIPD decision cannot drift;
- the CLI needs no changes at all, and the stale AEAD size expectation in cli_tests was correctly collapsed once the old omission behavior was gone;
- the #2470 regression test covers the full matrix, including v4 round-trip in crypto-refresh builds.

All checks green except the fleet-wide external fuzzing job. Approving on Ronald's behalf.

@ni4 could you take the second review/approval? This implements the lower-layer, version-aware approach you proposed in #2470, with the warn+omit semantics discussed in the review thread.

Issue #2397, opened 2026-06-10
Reserved packet tag 0 accepted — RFC 4880 §4.3 MUST NOT violation


afldl:
RNP's packet parser accepts OpenPGP packets with tag value 0. RFC 4880 §4.3 explicitly says: "0 — Reserved - a packet tag MUST NOT have this value."

**What happens:**
```
$ rnp --list-packets tag0_input.gpg
:off 0: packet header 0xc001 (tag 0, len 1)
Skipping Unknown pkt: 0
```

RNP accepts it (exit code 0). GnuPG correctly rejects it:
```
$ gpg --list-packets tag0_input.gpg
gpg: no valid OpenPGP data found.
```

**Location:** `src/librepgp/stream-packet.cpp` — `stream_peek_packet_hdr()` / `stream_dump_packets_raw()`

**Impact:** Informational / RFC compliance. No direct security impact since the packet content is skipped. But accepting reserved values could cause interop issues or problems if tag 0 gets assigned a meaning in future RFCs.

Found through differential fuzzing (PathDiff), comparing RNP, GnuPG, and PGPy.

Issue #2467, opened 2026-08-27
rnp_key_is_compromised/retired/superseded return RNP_ERROR_BAD_PARAMETERS for keys that are not revoked


ronaldtse:
`rnp_key_is_compromised()` / `rnp_key_is_retired()` / `rnp_key_is_superseded()` return `RNP_ERROR_BAD_PARAMETERS` when the key is not revoked

## Description

All three predicates go through `rnp_key_is_revoked_with_code()` (`src/lib/rnp.cpp`, around line 7659 on main):

```cpp
static rnp_result_t
rnp_key_is_revoked_with_code(rnp_key_handle_t handle, bool *result, int code)
{
    if (!handle || !result) {
        return RNP_ERROR_NULL_POINTER;
    }
    auto *key = get_key_prefer_public(handle);
    if (!key || !key->revoked()) {
        return RNP_ERROR_BAD_PARAMETERS;
    }
    *result = key->revocation().code == code;
    return RNP_SUCCESS;
}
```

"Not revoked" is a perfectly normal key state, but it is answered with an error. Callers of e.g. `rnp_key_is_compromised()` have no way to distinguish "the key is fine, just not revoked as compromised" from an actual problem — every caller must special-case the error code and map it to `false` by hand.

The sibling predicates `rnp_key_is_revoked()`, `rnp_key_is_valid()`, `rnp_key_is_expired()`, ... all succeed with a boolean for the same key state, so this trio is also inconsistent within the API.

## Suggested behavior

Return `RNP_SUCCESS` with `*result = false` when the key exists but is not revoked, reserving `RNP_ERROR_BAD_PARAMETERS` for an unusable handle (and `RNP_ERROR_KEY_NOT_FOUND` for the missing-key case if that needs to stay distinct).

## Environment

Observed against librnp 0.18.1 and re-verified in current `main` (`src/lib/rnp.cpp`, `rnp_key_is_revoked_with_code`), while developing the Rust binding (rnpgp/rnp-rs).

Issue #2394, opened 2026-06-01
Use ctest --parallel 1 to check whether it fixes random test failures


kaie:
Because we're seeing random test failures, I wonder if you could temporarily enforce serial execution of all tests, just to rule out that parallel execution of tests might cause trouble.

Once we see all tests working (or tests failing despite serial execution), the workflows could be reverted to parallelism.

What do you think @ni4 ?

Issue #2311, opened 2025-01-30
Add ossf-scorecard action


ni4:
See the details here: https://github.com/ossf/scorecard-action

PR #2489, opened 2026-09-21
librepgp: reject packet tag 0 as required by RFC 4880

ronaldtse:
## Summary
- RFC 4880, section 4.3, states that a packet tag must not have the value 0, which is reserved. The parser previously accepted such a header and skipped the packet as unknown, while GnuPG rejects the same input, and the difference was found through differential fuzzing and reported by @afldl in #2397.
- `stream_peek_packet_hdr()` now returns `RNP_ERROR_BAD_FORMAT` when the decoded tag is 0, for both the new-format (0xc0) and the old-format (0x80) header spellings, which covers the parsing, dumping and keyring load paths that share this header gate.
- Other reserved tag values, which the RFC does not single out in the same way, continue to be skipped as unknown, and this change does not affect them.
- The new `test_issue_2397_tag0` test feeds both header spellings to the packet dump and the key import paths and asserts that each fails.

## Test plan
- [x] `rnp_tests.test_issue_2397_tag0` passes against both spellings of the reserved tag
- [x] the parser-related tests (`load_pgp`, `load_g10`, `load_g23`, `load_kbx`, `stream_key_load`, the dump tests and `fuzz_dump`) continue to pass
- [ ] full CI matrix


ronaldtse:
@ni4 a small batch of CI and process pull requests has been prepared from the issue backlog, each of which needs only your approval:

- #2489 rejects the reserved packet tag 0 in the parser, closing #2397 which @afldl found through differential fuzzing against GnuPG.
- #2490 adds linelint to the lint workflow per #2300, with a configuration which excludes the vendored trees and the parser fixtures whose bytes are part of the tests, and normalises the whitespace this surfaced in first-party files. The change is provably whitespace-only.
- #2491 adds the OpenSSF Scorecard workflow per #2311, publishing its report to code scanning next to the existing CodeQL results.
- #2492 runs ctest serially as the temporary experiment you were asked about in #2394, so the next runs show whether the random interop failures persist without concurrent execution. It is intended to be reverted once the diagnostic value is spent.
- #2488, for completeness, fixes the revocation-code predicates to return false instead of an error for live keys (#2467); as a behaviour change in the FFI it would benefit from your review together with a second approval.

Whenever time allows, a look at any of these would be much appreciated.

ni4 (approved):
LGTM!

Issue #2300, opened 2024-12-13
Add linelint to the lint workflows.


ni4:
## Description

Not much an issue, just for code clarity. See the details here: https://github.com/marketplace/actions/linelint

PR #2490, opened 2026-09-21
ci: add linelint and normalise whitespace across first-party files

ronaldtse:
## Summary
- Adds the linelint action to the lint workflow, as suggested by @ni4 in #2300, together with a `.linelint.yml` configuration.
- The configuration excludes the vendored nlohmann and libsexpp trees, the parser test fixtures under `src/tests/data` (whose exact bytes, including deliberate malformed armour, trailing whitespace and CRLF endings, are part of the tests), the Windows SDK finder module and the UTF-8 manifest fixture which the msys2 workflow embeds through `mt.exe`.
- To satisfy the checks, trailing whitespace was stripped from six workflows, the issue template, `cli_tests.py`, `cli_common.py` and `docs/develop.adoc`, and a missing final newline was added to nineteen first-party sources.
- The changes are whitespace-only and provably so: `git diff -w` over the patch is empty, and the modified Python and YAML files still parse. No behaviour changes.

## Test plan
- [x] `cli_tests.py` and `cli_common.py` still compile (`py_compile`)
- [x] the six workflows still parse as YAML
- [x] the linelint job passes with the new configuration
- [ ] full lint workflow on the PR


ronaldtse:
The repeated "No valid rule enabled" failure is now root-caused and fixed in a83d19c0. The action is pinned to fernandrone/linelint@0.0.6, and that version parses a different configuration schema from the one shown in the project README today: the only rule it implements is end-of-file (configured through `rules.end-of-file.enable`), and file exclusions are declared with gitignore-style `ignore` patterns rather than `exclude`. Because unknown keys are dropped silently, the previous configuration parsed to zero enabled rules and the tool aborted.

The configuration is rewritten for the 0.0.6 schema, and the eleven first-party files that ended without a final newline or with a trailing blank line are normalised so the tree passes the end-of-file rule. One exclusion is deliberate and documented in the configuration file: `cmake/version.cmake` keeps its exact bytes because the version-cmake-up-to-date job diffs it byte-for-byte against the copy in rnpgp/cmake-versioning, and the upstream copy does not end with a newline.

Please note that with version 0.0.6 the enforced rule set is the end-of-file rule only; rules such as trailing whitespace or conflict markers are not implemented in that release. The whitespace normalisation already carried by this pull request remains in place, and upgrading the pinned action at a later time would allow re-enabling the broader rule set. I verified the new configuration and the normalisation locally with a linelint 0.0.6 binary built from the upstream tag, and the full tree passes. Thank you.

ni4 (approved):
LGTM, thanks!

PR #2491, opened 2026-09-21
ci: add the OpenSSF Scorecard analysis workflow

ronaldtse:
## Summary
- Adds the OpenSSF Scorecard analysis workflow, as suggested by @ni4 in #2311, using the [ossf/scorecard-action](https://github.com/ossf/scorecard-action).
- The analysis runs on a weekly schedule and on pushes to main, publishes its results through the identity token, and uploads the SARIF report to code scanning, where it appears alongside the existing CodeQL report.
- The workflow follows the repository conventions for action versions and declares read-only permissions at the top level, with the two write permissions required by the analysis scoped to the job.

## Test plan
- [ ] the workflow parses and the first scheduled run publishes a scorecard to the repository security tab


ni4 (approved):
LGTM, thanks!

PR #2492, opened 2026-09-21
ci: run ctest serially as a temporary experiment

ronaldtse:
## Summary
- Implements the experiment suggested by @kaie in #2394: every workflow now runs `ctest --parallel 1`, so the next rounds of runs show whether the random interop test failures persist without concurrent test execution.
- The change is mechanical and easily reversible: sixteen `ctest` invocations across nine workflows switch from `--parallel ${{ env.CORES }}` to `--parallel 1`, with the build parallelism and everything else untouched.
- As @kaie noted, serial execution roughly doubles the wall time of the test phase, so this is intended as a temporary diagnostic. Once enough serial runs have accumulated, the parallelism should be restored or the flaky tests addressed directly, in which case this can simply be reverted.

## Test plan
- [ ] the workflows pass with serial execution
- [ ] after a reasonable number of serial runs, compare the random-failure rate against the parallel runs and decide on keeping or reverting


ni4 (approved):
LGTM!

Issue #2302, opened 2024-12-23
Consider replacing std::wstring_convert with appropriate alternative for Windows.


ni4:
## Description

- std::wstring_convert is deprecated as of C++17 (and latest Botan uses C++20)
- it has some memory leak for one kinda major implementation: https://github.com/microsoft/STL/blob/92508bed6387cbdae433fc86279bc446af6f1b1a/stl/inc/cvt/wstring#L33

See the discussion/details here: https://www.reddit.com/r/cpp/comments/es9gt1/since_stdwstring_convert_is_deprecated_what_is/

PR #2493, opened 2026-09-21
common: replace deprecated wstring_convert with Win32 conversion APIs

ronaldtse:
## Summary
- Replaces the deprecated `std::wstring_convert` / `std::codecvt_utf8_utf16` pair used by the Windows UTF-8 helpers, as reported by @ni4 in #2302.
- The helpers now call `MultiByteToWideChar` and `WideCharToMultiByte`, which are the supported conversion APIs on modern Windows toolchains and avoid the known Microsoft STL memory leak in `wstring_convert`.
- The throwing-on-error contract of the previous helpers is preserved, and the existing Windows Unicode CLI tests continue to exercise the conversion path.

## Test plan
- [ ] the windows-native and windows-msys2 matrices pass
- [ ] the existing Unicode CLI tests continue to pass on Windows


ronaldtse:
@ni4 per your issue #2302, this replaces the deprecated std::wstring_convert with the Win32 conversion APIs, preserving the throwing contract and keeping the existing Windows Unicode tests as coverage. Whenever convenient, a review would be appreciated.

Issue #2288, opened 2024-11-11
Consider to use wmain/-municode when linking MSYS build.


ni4:
## Description

Recently MSYS2 (MinGW?) team introduced changes which would force default system codepage usage for CLI arguments translation, which may be non-utf8. For now this required CI changes ( PR #2286  ), however we may whish to completely move CLI to unicode (see https://github.com/msys2/MINGW-packages/issues/22462#issuecomment-2466906811 )

https://sourceforge.net/p/mingw-w64/wiki2/Unicode%20apps/

PR #2494, opened 2026-09-21
cli: enter through wmain on Windows and link MinGW with -municode

ronaldtse:
## Summary
- Addresses #2288: the production entry points of `rnp` and `rnpkeys` now enter through `wmain` on Windows, so the process receives UTF-16 arguments from the OS and converts them to UTF-8 once at startup before calling the existing `rnp_main` / `rnpkeys_main` bodies.
- MinGW targets are linked with `-municode` so the Unicode entry point is selected; MSVC selects `wmain` automatically when it is present.
- The `rnp_main` / `rnpkeys_main` symbols remain available for the test suite, which continues to pass UTF-8 `char **` arguments directly and does not go through `wmain`.
- Depends on #2493 for the Win32 UTF-8 helpers used by the conversion.

## Test plan
- [ ] windows-native and windows-msys2 matrices pass
- [ ] the existing Unicode CLI tests continue to pass on Windows


ronaldtse:
@ni4 this implements the wmain/-municode direction from #2288: the production entry points take UTF-16 from the OS and convert to UTF-8 once at startup, while the test suite keeps calling rnp_main/rnpkeys_main directly. It is stacked on #2493 for the Win32 conversion helpers. Whenever convenient, a review would be appreciated.

Issue #2347, opened 2025-04-27
Add test runner with limited stack size.


ni4:
Recent experience revealed that some code could overrun default stack size. We should add CI runner (and probably update oss-fuzz code) with limited to 256k or so stack size.

PR #2495, opened 2026-09-21
ci: run the unit tests under a 256 KiB stack

ronaldtse:
## Summary
- Adds a dedicated Ubuntu leg which builds the Botan shared configuration and runs the unit tests after `ulimit -s 256`, as suggested by @ni4 in #2347.
- Stack overruns which would only surface on constrained deployments now fail loudly in CI, while the existing unrestricted legs are left unchanged.
- The leg excludes the slow tests, matching the other Ubuntu jobs, and runs `ctest` serially so a stack fault is easy to attribute.

## Test plan
- [ ] the limited-stack leg passes on this PR
- [ ] the existing unrestricted Ubuntu legs continue to pass


ronaldtse:
@ni4 this adds the limited-stack CI leg from #2347: a dedicated Ubuntu job runs the unit tests after ulimit -s 256 so stack overruns surface in CI. Whenever convenient, a review would be appreciated.

ni4 (approved):
LGTM, but we seem to have segfault in CI.

Issue #2276, opened 2024-09-27
CI: upload coverage report for Windows and 32-bit Debian.


ni4:
## Description

Currently we upload coverage report only for 64 bit Linux systems, so it would be good idea to upload them for Windows and 32-bit system as well.

PR #2496, opened 2026-09-21
ci: add coverage uploads for Windows and 32-bit Debian

ronaldtse:
## Summary
- Adds coverage build variants for the Windows msys2 ucrt64 Botan leg and the Debian 11 i386 Botan and OpenSSL legs, as requested by @ni4 in #2276. Coverage reporting had previously been limited to 64-bit Linux systems.
- Coverage legs build with `ENABLE_COVERAGE`, run the test suite, convert the `.gcno` files with `gcov` and upload the results to Codecov, matching the existing 64-bit Linux coverage legs.
- The Windows coverage leg uses the Debug configuration with MinGW gcc's gcov through the MSYS2 environment; the Debian i386 legs follow the same `gcov` and Codecov flow as the CentOS and Fedora coverage legs.

## Test plan
- [ ] the new Windows coverage leg completes the gcov conversion and upload
- [ ] the new Debian i386 coverage legs complete the gcov conversion and upload
- [ ] the regular (non-coverage) legs continue to pass


ronaldtse:
@ni4 this implements the coverage uploads for Windows and 32-bit Debian from #2276. The new legs follow the gcov-and-Codecov flow of the existing 64-bit Linux coverage legs. Whenever convenient, a review would be appreciated.

ni4 (approved):
LGTM, thanks!

ronaldtse:
@ni4 Thank you for the approval. While validating the coverage legs I found that the previous upload invocation could never have delivered data: the codecov bash uploader has been discontinued by the service, the process-substitution syntax fails under the POSIX shell of the Debian jobs, and the current uploader binary is x86_64-only, which cannot execute under the 32-bit userspace of the i386 containers. The coverage steps now upload the generated gcov reports as workflow artifacts instead, which keeps the data retrievable on every platform without external dependencies. The commits since your review are 75502cba, 48b9ecac and d035fc74, all confined to the coverage steps; please speak up if you would like any of them handled differently.

PR #2497, opened 2026-09-21
ci: allow a small coverage drift in the codecov project status

ronaldtse:
This pull request adds a small threshold to the codecov project status so that rounding differences between coverage uploads no longer fail the check.

The aggregated coverage report is assembled from uploads of many CI legs across backends and platforms. Rounding differences between those uploads shift the project percentage by a few hundredths of a point, and the default zero-drift gate turns this into a failure even when a pull request does not reduce coverage. Pull requests #2489 and #2492 both failed the project status with drifts of 0.03 and 0.01 percent respectively while their patch coverage passed, which shows the gate is reacting to aggregate noise rather than to the changes under review.

The change sets the project target to `auto` with a threshold of 0.1 percent. Aggregate noise of this magnitude is absorbed, while a genuine coverage regression larger than 0.1 percent still fails the check as before. The patch status is left at its default, so every added line remains held to the existing standard.

This is a CI and process change, so @ni4 a review from you is kindly requested whenever your time allows. Thank you.


ronaldtse:
@ni4 This small CI change lets the codecov project status tolerate the few-hundredths-of-a-point rounding drift that comes from aggregating uploads across many CI legs, which currently fails pull requests such as #2489 and #2492 even though their patch coverage passes. A review from you is kindly requested whenever your time allows, and thank you for all of the reviews today.

ni4 (approved):
LGTM, thanks!

ni4:
Merging with single approval as pure CI change.

PR #2498, opened 2026-09-21
ci: retry the keyserver import in the coverage step

ronaldtse:
This pull request makes the coverage step tolerate the intermittent keyserver outages that have been failing otherwise green jobs.

The coverage step imports the codecov uploader signing key from keyserver.ubuntu.com before verifying and running the uploader. Today the connection to the keyserver timed out on four coverage legs across pull requests #2491, #2492 and #2493, and each timeout failed the whole upload even though the build and the full test suite had already passed. The failure looks like `curl: (7) Failed to connect to keyserver.ubuntu.com port 443: Connection timed out`, and gpg then exits with an error on the empty input.

The change retries the import up to five times with a pause of ten seconds between attempts, and the step still fails when every attempt is unsuccessful, so the signature verification chain is never skipped. The verification of the uploader itself through gpgv and the SHA256SUM check is unchanged.

This is a CI and process change, so @ni4 a review from you is kindly requested whenever your time allows. Thank you.


ronaldtse:
@ni4 One more small CI change for your queue when time allows: the coverage step now retries the keyserver.ubuntu.com key import instead of failing the leg on the intermittent connect timeouts seen today on #2491, #2492 and #2493. The gpgv and SHA256SUM verification of the uploader is untouched. Thank you.

ni4 (approved):
LGTM!

